PR #14, opened 2025-11-22
feat[#13]: Add integration tests to CI pipeline

### LINKED ISSUE #13: Add running integration tests on CI
# Feature request / General issue

We would like to run ROS2 integration tests on the CI (GitHub Actions), so that they are executed automatically on pull requests and on pushes to the main branch.

### Proposal

Add a new `integration-tests` job to the existing GitHub Actions workflow that runs **after** the current build/unit test job (e.g. depending on `build-and-test` or equivalent).

The job should:
- check out the repository,
- set up the ROS2 environment,
- build the workspace with `colcon build`,
- run ROS2 integration tests (e.g. using `colcon test` and launch-based tests),
- report the test results (e.g. via `colcon test-result` and CI artifacts).

### Motivation

- Catch integration regressions early on pull requests.
- Ensure multiple ROS2 nodes and packages work correctly together.
- Provide consistent, reproducible test results for all contributors.
- Increase confidence in changes and reduce manual testing effort.

### Alternatives considered

- Running integration tests only locally (not enforced, not reproducible).
- Running integration tests only on a nightly/scheduled workflow (slower feedback).
- Using an external CI system instead of GitHub Actions (more complexity, split checks).

### Additional context

- Tests are ROS2 integration tests run via `colcon`.
- In the future we might:
  - expand to a matrix of ROS2 distros / OS versions,
  - use Docker images with ROS2 preinstalled to speed up CI.

bburda:
# Pull Request

## Description

- Split CI test step into unit and integration tests
- Add python3-requests test dependency to package.xml
- Update README with instructions for running tests separately

## Type of change

- [ ] Bug fix (non-breaking change which fixes an issue)
- [x] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to not work as expected)
- [x] Documentation update

## Related issues

Closes #13

## How has this been tested?

On CI

## Checklist

- [x] My code follows the project's coding style
- [x] I ran `colcon build` and the build succeeds without warnings
- [x] I ran `colcon test` and all tests pass locally
- [ ] I added/updated tests where applicable
- [x] I updated documentation where applicable
- [x] My commit messages are clear and descriptive

### DIFF
--- a/README.md
+++ b/README.md
@@ -61,11 +61,11 @@ Run linters:
 
 ```bash
 source install/setup.bash
-colcon test --packages-select ros2_medkit_gateway --ctest-args -L linters
+colcon test --ctest-args -L linters
 colcon test-result --verbose
 ```
 
-Run only unit tests:
+Run only unit tests (everything except integration):
 
 ```bash
 source install/setup.bash
@@ -84,12 +84,13 @@ colcon test-result --verbose
 ### CI/CD
 
 All pull requests and pushes to main are automatically built and tested using GitHub Actions.
-The CI workflow runs on Ubuntu 24.04 with ROS 2 Jazzy and includes:
+The CI workflow runs on Ubuntu 24.04 with ROS 2 Jazzy, executes a single `colcon test` to cover:
 
 - Code linting and formatting checks
 - Unit tests
 - Integration tests with demo automotive nodes
-- Automatic upload of test logs and results for all runs
+
+After every run the workflow always calls `colcon test-result --verbose` and uploads the generated logs/results as artifacts for debugging.
 
 ## Contributing
 
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -20,12 +20,6 @@ jobs:
         with:
           required-ros-distributions: jazzy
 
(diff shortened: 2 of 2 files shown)


README.md
```diff
@@ -49,11 +49,25 @@ colcon build --symlink-install
...
+
 ```bash
 colcon test
 colcon test-result --verbose
 ```
 
+Run only unit tests:
+
+```bash
+colcon test --ctest-args -E test_integration
```
mfaferek93, line 72: @bburda 
maybe we should be more specific here and use the -R option. Instead of excluding the integration tests, we could just call our unit tests by name with -R "test_gateway_node". WDYT?

bburda: @mfaferek93 
Thanks for the suggestion! After revisiting the workflow I switched to running a single `colcon test` step that covers lint, unit, and integration tests for every package. That way we don’t need to maintain any naming conventions or update the CI when new packages/tests appear everything runs automatically. If we ever split the steps again I agree using `-R` would be the safer approach, but for now the all-in-one run keeps things simple and robust.

PR #27, opened 2025-11-24
[#24] Add sphinx documentation and sphinx-needs requirements 

### LINKED ISSUE #24: Add sphinx-needs requirements to track progress on specfication implmentation
# Feature request / General issue

Add sphinx-needs requirements to track progress on specfication implmentation

### Proposal

We would like to introduce a sphinx-needs requirements to track implementation of SOVD endpoints and also to mark which endpoints are extensions for ROS 2 purposes.

### Motivation

We would like to keep full SOVD-spec compatibility but also add some custom extensions. The best would be to track them with requirements as a code to map code/tests to the requirements. This way we can generate documentation page and traceability matrix to show what we already implemented and what is missing.

bburda:
# Pull Request

## Summary

Add sphinx documentation and sphinx-needs requirements.
Also adds generation of docs on the CI.

---

## Issue

- closes #24 

---

## Type

- [ ] Bug fix
- [x] New feature or tests
- [ ] Breaking change
- [x] Documentation only

---

## Testing

All in README:
- how to build docs
- how to run docs
- how to generate traceability

---

## Checklist

### DIFF
--- a/.github/workflows/docs.yml
+++ b/.github/workflows/docs.yml
@@ -1,71 +1,75 @@
 name: Documentation
 
 on:
-    push:
-        branches: [main]
-        paths:
-            - "docs/**"
-            - ".github/workflows/docs.yml"
-    pull_request:
-        branches: [main]
-        paths:
-            - "docs/**"
-            - ".github/workflows/docs.yml"
-    workflow_dispatch:
+  push:
+    branches: [main]
+    paths:
+      - "docs/**"
+      - "src/**"
+      - "scripts/**"
+      - ".github/workflows/docs.yml"
+  pull_request:
+    branches: [main]
+    paths:
+      - "docs/**"
+      - "src/**"
+      - "scripts/**"
+      - ".github/workflows/docs.yml"
+  workflow_dispatch:
 
 # Sets permissions of the GITHUB_TOKEN to allow deployment to GitHub Pages
 permissions:
-    contents: read
-    pages: write
-    id-token: write
+  contents: read
+  pages: write
+  id-token: write
 
 concurrency:
-    group: "pages"
-    cancel-in-progress: false
+  group: "pages"
+  cancel-in-progress: false
 
 jobs:
-    build-docs:
-        runs-on: ubuntu-24.04
-        steps:
-            - name: Checkout repository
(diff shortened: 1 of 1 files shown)


docs/conf.py
```diff
@@ -0,0 +1,133 @@
...
+}
+
+# -- Options for intersphinx extension ---------------------------------------
+intersphinx_mapping = {
+    "python": ("https://docs.python.org/3", None),
+    "sphinx": ("https://www.sphinx-doc.org/en/master", None),
+}
+
+# -- Options for PlantUML ----------------------------------------------------
+plantuml = "java -Djava.awt.headless=true -jar /usr/share/plantuml/plantuml.jar"
```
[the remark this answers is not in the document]
bburda: @copilot This needs to be done this way to ensure that CI correctly invoke plantuml for generation of diagrams.

.github/workflows/docs.yml
```diff
@@ -0,0 +1,71 @@
+name: Documentation
```
mfaferek93, line 1: @bburda 
I noticed that the docs workflow only triggers when files in `docs/**` are modified, but it doesn't watch the test files in `src/` that contain the @verifies annotations.

This means when someone adds a new test case or updates an existing one with @verifies REQ_XXX tags, the documentation build won't run automatically. The verification.rst file (which is generated by scanning the test files) will show outdated information until someone happens to edit something in the `docs/ folder`, right?

Potential fix: Update the workflow trigger paths in .github/workflows/docs.yml to also watch the test files and the verification script:
```
  on:
      push:
          branches: [main]
          paths:
              - "docs/**"
              - "src/**/test/**"              # Watch test files for @verifies changes
              - "scripts/generate_verification.py"  # Watch the generator script itself
              - ".github/workflows/docs.yml"
```
This way, whenever someone modifies a test (adding/changing @verifies annotations), the docs will automatically regenerate and stay in sync with the code.

bburda: @mfaferek93 Yes, you are right. Actually we want to look on the whole src/** because changes in design docs or tests should also rebuild docs. Updated.

mfaferek93 (approved):

PR #118, opened 2026-01-17
[#109] Discovery: Add YAML Manifest as a static source of entity discovery

### LINKED ISSUE #109: Discovery: Add Manifest YAML as production source of truth
## Summary

We want a production mode where discovery is stable and explicitly defined. This requires a Manifest YAML that declares stable IDs, intended structure, and relations.

Manifest should allow:
- Stable IDs for Areas, Components, Apps, and optional Functions
- Explicit relations (App belongs to Component, App depends on App, Function includes Apps or Components)
- Linking runtime ROS nodes to declared Apps and Runtime processes

---

- Design a Manifest YAML schema that represents Area, Component, App, Function and relations.
- Implement parsing and validation with clear error messages.
- Implement runtime linking logic (bind ROS node name and namespace to Manifest App entries).
- Define what happens when runtime has nodes not in Manifest

---

The goal is an adoption path where dev mode can be heuristic, but production can have full control and stable IDs. Manifest should be the source of truth when it exists.

bburda:
# Pull Request

## Summary

Adds YAML manifest-based discovery as a static source for entity discovery, implementing three discovery modes: runtime_only (existing behavior), manifest_only (static YAML definitions), and hybrid (combines manifest definitions with runtime linking).

Changes:

- Introduces manifest parser, validator, and manager for YAML-based entity definitions
- Adds App and Function discovery models with runtime linking support
- Refactors discovery architecture with strategy pattern (RuntimeDiscoveryStrategy, HybridDiscoveryStrategy)
- Adds HTTP handlers for /apps and /functions endpoints
- Updates API response format for list endpoints to include {items: [], total_count: N}

---

## Issue

- closes #109 
---

## Type

- [ ] Bug fix
- [x] New feature or tests
- [ ] Breaking change
- [ ] Documentation only

---

## Testing

Unit and Integration Testing

---

## Checklist

### DIFF
--- a/src/ros2_medkit_gateway/test/test_discovery_hybrid.test.py
+++ b/src/ros2_medkit_gateway/test/test_discovery_hybrid.test.py
@@ -472,9 +472,16 @@ class TestDiscoveryHybridMode(unittest.TestCase):
         self.assertEqual(response.status_code, 200)
 
         app = response.json()
-        # Should be located on temp-sensor-hw
-        if 'component_id' in app:
-            self.assertEqual(app['component_id'], 'temp-sensor-hw')
+        # Should be located on temp-sensor-hw via HATEOAS link
+        self.assertIn('_links', app, 'App response should contain _links')
+        self.assertIn(
+            'is-located-on', app['_links'],
+            'App should have is-located-on link when component is specified'
+        )
+        self.assertEqual(
+            app['_links']['is-located-on'],
+            '/api/v1/components/temp-sensor-hw'
+        )
 
     def test_app_depends_on_relationship(self):
         """


src/ros2_medkit_gateway/include/ros2_medkit_gateway/discovery/manifest/manifest_manager.hpp
```diff
@@ -0,0 +1,222 @@
...
+
+  // === Manifest Loading ===
+
+  /**
+   * @brief Load manifest from file
+   * @param file_path Path to manifest YAML file
+   * @param strict If true, fail on any validation error; if false, warn only
+   * @return true if loaded and validated successfully
+   */
+  bool load_manifest(const std::string & file_path, bool strict = true);
```
mfaferek93, line 63: Question about `strict=false` behavior with ERROR-level validation issues

I'm wondering if allowing strict=false to bypass ERROR-level validation (like R006 broken refs, R010 duplicate bindings, R011 circular deps) might be problematic?

Looking at the code, in non-strict mode:
- Broken references (R006, R007)  - API returns result which links that 404
- Duplicate ROS bindings (R010) - RuntimeLinker binds same node to multiple Apps (confusing is_online state)
- Circular dependencies (R011) - Currently just data, but if someone adds dependency traversal later - infinite loop

Would it make sense to:
- Always fail on ERRORs (they indicate a fundamentally broken manifest)
- Only allow strict=false to skip WARNINGs (missing optional refs, etc.)

Or is there a specific use case where you'd want to load a manifest with broken references?

src/ros2_medkit_gateway/config/examples/minimal_manifest.yaml
```diff
@@ -0,0 +1,33 @@
...
+  - id: robot-controller
+    name: "Robot Controller"
+    type: "controller"
+    area: robot
+
+# Apps mapping to ROS 2 nodes
+apps:
+  - id: example-node
+    name: "Example Node"
+    is_located_on: robot-controller
```
mfaferek93, line 30: I noticed Component uses `area: engine` but App uses `is_located_on: engine-ecu`. Would it be cleaner to just use  `component: engine-ecu` here? Feels more intuitive since you immediately know it points to a Component, and it's consistent with how Component -> Area works.

Not a blocker, just wondering if there's a reason for the different styles?

src/ros2_medkit_gateway/src/discovery/manifest/manifest_parser.cpp
```diff
@@ -0,0 +1,267 @@
...
+  return comp;
+}
+
+App ManifestParser::parse_app(const YAML::Node & node) const {
+  App app;
+  app.id = get_string(node, "id");
+  app.name = get_string(node, "name", app.id);
+  app.translation_id = get_string(node, "translation_id");
+  app.description = get_string(node, "description");
+  app.component_id = get_string(node, "component");
```
mfaferek93: demo_nodes_manifest.yaml, minimal_manifest.yaml, etc. all use `is_located_on`
This means apps won't have their component_id populated and `is-located-on links` will be broken, right?

src/ros2_medkit_gateway/test/test_discovery_hybrid.test.py
```diff
@@ -0,0 +1,640 @@
...
+                )
+
+    def test_app_component_relationship(self):
+        """Test app is_located_on links to correct component."""
+        response = requests.get(f'{self.BASE_URL}/apps/engine-temp-sensor', timeout=5)
+        self.assertEqual(response.status_code, 200)
+
+        app = response.json()
+        # Should be located on temp-sensor-hw
+        if 'component_id' in app:
```
mfaferek93: This soft assertion masks Bug - if `component_id` is missing (because parser doesn't recognize `is_located_on`), the test passes silently.

It should be a hard assertion:
```
self.assertIn('component_id', app)
self.assertEqual(app['component_id'], 'temp-sensor-hw')
```

src/ros2_medkit_gateway/config/examples/minimal_manifest.yaml
mfaferek93: Current state:
- Component -> area: engine (points to Area) ✅
- App -> is_located_on: engine-ecu (points to Component)
- Function -> hosted_by: [app1] (points to Apps)

Proposed (consistent entity-type naming):
- Component -> area: engine ✅ no change
- App -> component: engine-ecu 
- Function -> apps: [app1] or keep hosts:

Interestingly, the parser already looks for `component` but the examples use `is_located_on` so this might actually be a bug rather than just a style question? 🤔

src/ros2_medkit_gateway/test/test_discovery_hybrid.test.py
```diff
@@ -0,0 +1,640 @@
...
+        from ament_index_python.packages import get_package_prefix
+        pkg_prefix = get_package_prefix('ros2_medkit_gateway')
+        workspace = os.path.dirname(os.path.dirname(pkg_prefix))
+        build_dir = os.path.join(workspace, 'build', 'ros2_medkit_gateway')
+        if os.path.exists(build_dir):
+            return {
+                'GCOV_PREFIX': build_dir,
+                'GCOV_PREFIX_STRIP': str(build_dir.count(os.sep)),
+            }
+    except Exception:
```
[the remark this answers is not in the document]
bburda: Added explanatory comments to all empty except blocks explaining why they pass silently:

Coverage env setup is optional - tests should work without coverage
Gateway/apps endpoint retry loops - expected during startup, will retry after sleep

src/ros2_medkit_gateway/include/ros2_medkit_gateway/discovery/manifest/manifest_manager.hpp
bburda: Great catch! Implemented your suggestion - now:

- ERRORs (R006 broken refs, R010 duplicate bindings, R011 circular deps) always fail regardless of strict flag
- strict=false only allows skipping WARNINGs (missing optional refs, etc.)

This ensures fundamentally broken manifests never load, preventing runtime issues like 404 links or confusing is_online states.

src/ros2_medkit_gateway/config/examples/minimal_manifest.yaml
bburda: Good question! After checking the SOVD OpenAPI spec `is-located-on` is the normative field name in the API response schema:

```yaml
is-located-on:
  description: A reference to the component where the app is located (only available for Apps)
  type: string
  format: uri-reference
```

So we're keeping `is_located_on` in YAML manifests to stay consistent with SOVD specification. The internal model uses component_id but YAML/API use is_located_on/is-located-on.

src/ros2_medkit_gateway/src/discovery/manifest/manifest_parser.cpp
bburda: Fixed! Parser now correctly reads is_located_on from YAML. Also fixed `parent_component_id` parsing (was looking for `parent_component`).

src/ros2_medkit_gateway/test/test_discovery_hybrid.test.py
bburda: You're absolutely right! Fixed.

src/ros2_medkit_gateway/include/ros2_medkit_gateway/discovery/manifest/manifest_manager.hpp
mfaferek93: Thank you!

src/ros2_medkit_gateway/config/examples/minimal_manifest.yaml
mfaferek93: I see, thanks for the explanation!

PR #129, opened 2026-01-22
[#125] feat: SOVD-compatible API contract + ros2_medkit extensions (x-medkit)

### LINKED ISSUE #125: SOVD-compatible API contract + ros2_medkit extensions (x-medkit)
## Summary

I want to align `ros2_medkit_gateway` API with the SOVD specification and make ros2_medkit specific data explicitly extensible:

1) **Fix requests/responses and endpoint paths** so they follow SOVD conventions (resource collections, naming, payload shapes, status codes, and error model).  
2) **Add and document ros2_medkit-specific extensions** in a SOVD-friendly way, e.g. by introducing an `x-medkit: { ... }` object that carries all extra/vendor data without polluting the standard SOVD fields.

---

### A) SOVD compliance pass

- **Endpoints and paths**
  - Normalize resource paths to the SOVD-style hierarchy:
    - `/{entity-collection}/{entity-id}/{resource-collection}/{resource-id}`
  - Ensure collection vs. single-resource semantics are consistent (list vs. get by id).

- **Request/response payload shape**
  - Align response schemas with SOVD resource representations where we already map to SOVD Entities/resources.
  - Normalize optional query params and filtering behavior where relevant (and document what is supported).

- **HTTP status codes and error model**
  - Standardize status codes for common cases (200/201/204, 400/401/403/404/405/409/415, 5xx).
  - Align error responses to a consistent structure (e.g., `error_code`, `message`, `parameters`, optional `vendor_code`), and ensure it is used across all endpoints.

### B) ros2_medkit extension mechanism (`x-medkit`)

- **Principle:** keep SOVD “core” fields clean and compatible, put all ros2_medkit extras into a namespaced extension object.
- Example response shape:
  ```json
  {
    "id": "RearWindows",
    "data": { "PositionLeft": 100 },
    "x-medkit": {
      "ros": { "node": "...", "topic": "...", "type": "..." },
      "diagnostics": { "sampling": "...", "last_seen": "..." }
    }
  }
  ```

- **Documentation**
  - Define `x-medkit` as an explicit extension in OpenAPI schemas (so clients can validate it or safely ignore it).
  - Add docs that clearly separate:
    - “SOVD-compatible core”
    - “ros2_medkit extensions (x-medkit)”

### C) Definition of Done

- [ ] Implemented changes + regression tests for key endpoints.
- [ ] Documentation updates:
  - [ ] “SOVD compatibility” section
  - [ ] “Extensions: x-medkit” section

---

- The goal is that **a generic SOVD client can consume the “core” API** without knowing anything about ros2_medkit, while **ros2_medkit-aware tools** can rely on `x-medkit` for richer metadata and workflow-specific details.
- This also makes it easier to evolve ros2_medkit without diverging from SOVD semantics.

bburda:
# Pull Request

## Summary

Implements SOVD API compliance with custom ros2_medkit extensions under the x-medkit field. It represents a significant architectural shift to align with industry standards while preserving ROS 2-specific functionality.

Changes:

- Introduces SOVD-compliant error handling with standardized error codes and the GenericError schema
- Implements the x-medkit extension pattern to separate SOVD-compliant fields from ros2_medkit-specific extensions
- Adds comprehensive entity resource model with thread-safe caching, capabilities validation, and operation aggregation

---

## Issue

- closes #125 

---

## Type

- [ ] Bug fix
- [x] New feature or tests
- [x] Breaking change
- [ ] Documentation only

---

## Testing

---

## Checklist

### DIFF
--- a/src/ros2_medkit_gateway/src/http/handlers/handler_context.cpp
+++ b/src/ros2_medkit_gateway/src/http/handlers/handler_context.cpp
@@ -64,7 +64,7 @@ tl::expected<void, std::string> HandlerContext::validate_entity_id(const std::st
 
 tl::expected<std::string, std::string>
 HandlerContext::get_component_namespace_path(const std::string & component_id) const {
-  const auto& cache = node_->get_thread_safe_cache();
+  const auto & cache = node_->get_thread_safe_cache();
   auto component = cache.get_component(component_id);
   if (component) {
     return component->namespace_path;
@@ -73,7 +73,7 @@ HandlerContext::get_component_namespace_path(const std::string & component_id) c
 }
 
 EntityInfo HandlerContext::get_entity_info(const std::string & entity_id) const {
-  const auto& cache = node_->get_thread_safe_cache();
+  const auto & cache = node_->get_thread_safe_cache();
   EntityInfo info;
   info.id = entity_id;
 
--- a/src/ros2_medkit_gateway/src/http/handlers/operation_handlers.cpp
+++ b/src/ros2_medkit_gateway/src/http/handlers/operation_handlers.cpp
@@ -541,7 +541,7 @@ void OperationHandlers::handle_list_executions(const httplib::Request & req, htt
     }
 
     // Find entity to get namespace (O(1) lookups)
-    const auto& cache = ctx_.node()->get_thread_safe_cache();
+    const auto & cache = ctx_.node()->get_thread_safe_cache();
     std::string namespace_path;
     bool entity_found = false;
 
--- a/src/ros2_medkit_gateway/src/http/handlers/discovery/area_handlers.cpp
+++ b/src/ros2_medkit_gateway/src/http/handlers/discovery/area_handlers.cpp
@@ -29,7 +29,7 @@ void AreaHandlers::handle_list_areas(const httplib::Request & req, httplib::Resp
   (void)req;  // Unused parameter
 
   try {
-    const auto& cache = ctx_.node()->get_thread_safe_cache();
+    const auto & cache = ctx_.node()->get_thread_safe_cache();
     const auto areas = cache.get_areas();
 
     // Build items array with EntityReference format
@@ -167,7 +167,7 @@ void AreaHandlers::handle_area_components(const httplib::Request & req, httplib:
       return;
     }
 
-    const auto& cache = ctx_.node()->get_thread_safe_cache();
+    const auto & cache = ctx_.node()->get_thread_safe_cache();
 
     // Check if area exists (O(1) lookup)
     if (!cache.has_area(area_id)) {
--- a/src/ros2_medkit_gateway/src/http/handlers/discovery/component_handlers.cpp
+++ b/src/ros2_medkit_gateway/src/http/handlers/discovery/component_handlers.cpp
@@ -34,7 +34,7 @@ void ComponentHandlers::handle_list_components(const httplib::Request & req, htt
   (void)req;  // Unused parameter
 
   try {
-    const auto& cache = ctx_.node()->get_thread_safe_cache();
+    const auto & cache = ctx_.node()->get_thread_safe_cache();
     const auto components = cache.get_components();
 
     // Build items array with EntityReference format


src/ros2_medkit_gateway/src/http/handlers/handler_context.cpp
```diff
@@ -161,20 +174,26 @@ bool HandlerContext::is_origin_allowed(const std::string & origin) const {
...
-  res.status = status;
-  json error_json = {{"error", error}};
-  // Merge extra fields into the error object
-  for (const auto & [key, value] : extra_fields.items()) {
-    error_json[key] = value;
+  error_json["message"] = message;
+
+  // Use x-medkit extension for additional parameters
+  if (!parameters.empty()) {
+    error_json["x-medkit"] = parameters;
```
mfaferek93: extra parameters are placed under x-medkit field, but the SOVD GenericError schema (section 7.4.2) expects these to be in a parameters field. 
Should we rename this to match the spec? The x-medkit extension pattern makes sense for success responses, but for errors I think we want strict SOVD compliance?

src/ros2_medkit_gateway/include/ros2_medkit_gateway/http/error_codes.hpp
```diff
@@ -0,0 +1,105 @@
+// Copyright 2025 bburda
```
mfaferek93, line 1: 2026 :P

src/ros2_medkit_gateway/include/ros2_medkit_gateway/models/thread_safe_entity_cache.hpp
```diff
@@ -0,0 +1,295 @@
...
+ * - Multiple readers can access concurrently (shared lock)
+ * - Writers get exclusive access (unique lock)
+ * - Readers never block each other
+ * - Writers block all readers and other writers
+ *
+ * Usage:
+ * - Discovery thread calls update_*() methods (writer)
+ * - HTTP handlers call get_*() methods (reader)
+ */
+class ThreadSafeEntityCache {
```
mfaferek93, line 91: I noticed that ThreadSafeEntityCache and AggregationService are fully implemented and tested, but they're not actually wired into the handlers or gateway_node yet. The handlers still use the legacy EntityCache from get_entity_cache(). Is this intentional or was the integration missed? If intentional, might be worth adding a TODO comment or tracking issue.

src/ros2_medkit_gateway/src/http/handlers/operation_handlers.cpp
```diff
@@ -146,108 +169,293 @@ void OperationHandlers::handle_list_operations(const httplib::Request & req, htt
...
-    // Find entity and operation - check components first, then apps
+    // Find entity and operation
     bool entity_found = false;
-    std::string entity_type = "component";
     std::optional<ServiceInfo> service_info;
     std::optional<ActionInfo> action_info;
+    std::string entity_type = "component";
 
     // Try to find in components
     for (const auto & component : cache.components) {
```
mfaferek93: The operation lookup in handle_get_operation() does nested loops through entities and their operations (O(n*m) complexity). Meanwhile, ThreadSafeEntityCache has find_operation_owner() that provides O(1) lookup via the operation index. Once the new cache is integrated, we could simplify this significantly. For now it's fine with current scale, but maybe it is worth adding related TODO

src/ros2_medkit_gateway/src/http/handlers/handler_context.cpp
bburda: Yh, there is even plan to resign from x-medkit extension 😄 Fixed.

src/ros2_medkit_gateway/include/ros2_medkit_gateway/http/error_codes.hpp
bburda: Thanks, I'm mentally still in 2025 😄

src/ros2_medkit_gateway/include/ros2_medkit_gateway/models/thread_safe_entity_cache.hpp
bburda: Well spotted! I remembered to test this thoroughly and add the include, but completely forgot to replace the old Entity Cache with it. Fixed.

src/ros2_medkit_gateway/src/http/handlers/operation_handlers.cpp
bburda: Fixed!

PR #160, opened 2026-02-01
[#158] Fixes the dual-path routing bug

### LINKED ISSUE #158: [BUG] Dual-path routing bug: App IDs resolved via /components/{id}
# Bug report

## Steps to reproduce

- Run ros2_medkit_gateway in runtime-only discovery mode.
- Call a component endpoint using a valid synthetic component ID, e.g.:
- curl /api/v1/components/powertrain_engine_component/data

Call the same endpoint type using an App ID (an individual ROS 2 node identifier), e.g.:
- `curl /api/v1/components/temp_sensor/data`

## Expected behavior

In runtime-only mode:
- `/components/{id}` should accept only synthetic component IDs.
- Individual ROS 2 nodes (Apps) must be addressed via /apps/{id}.

Using an App ID under `/components/{id}` should return a clear error (e.g., 400/404) with a message such as:
- Invalid entity type: expected component id, got app id
- or Component not found: <id>

## Actual behavior

In runtime-only discovery mode, /components/{id}/... endpoints incorrectly accept App IDs.

Example (this should fail but currently works):
- `curl /api/v1/components/temp_sensor/data`

The current dual-path pattern allows /components/{app_id}/..., which functions but is semantically incorrect.

## Environment

ros2_medkit version: main branch
ROS 2 distro: ROS 2 Jazzy
OS: Ubuntu 24.04

## Additional information

In runtime-only discovery mode:
- Synthetic components are created to group Apps (ROS 2 nodes) by namespace.
- Therefore, `/components/{id}` should resolve only to synthetic component IDs, not to individual app/node IDs.

Affected routes (all dual-path routes in rest_server.cpp):
- `/components/{id}/data`
- `/components/{id}/operations`
- `/components/{id}/configurations`
- `/components/{id}/faults`

bburda:
# Pull Request

## Summary

Fixes the dual-path routing bug (https://github.com/selfpatch/ros2_medkit/issues/158) by enforcing semantic entity-type validation so /components/{id} no longer accepts App IDs (and similarly for other collection routes).

Changes:

- Add route-path-based entity type extraction and use it to reject mismatched entity IDs with a 400 error.
- Update handler entity lookup to support type-restricted lookups (avoids cross-collection ID collisions).
- Extend unit + integration tests (including polling improvements for action execution status).

---

## Issue

- closes #158 

---

## Type

- [x] Bug fix
- [ ] New feature or tests
- [ ] Breaking change
- [ ] Documentation only

---

## Testing

colcon test 

---

## Checklist

### DIFF
--- a/src/ros2_medkit_gateway/src/http/handlers/config_handlers.cpp
+++ b/src/ros2_medkit_gateway/src/http/handlers/config_handlers.cpp
@@ -491,10 +491,12 @@ void ConfigHandlers::handle_set_configuration(const httplib::Request & req, http
     entity_id = req.matches[1];
     param_id = req.matches[2];
 
-    // Validate entity ID and type for this route
-    auto entity_opt = ctx_.validate_entity_for_route(req, res, entity_id);
-    if (!entity_opt) {
-      return;  // Error response already sent
+    // Validate entity_id format first
+    auto entity_validation = ctx_.validate_entity_id(entity_id);
+    if (!entity_validation) {
+      HandlerContext::send_error(res, StatusCode::BadRequest_400, ERR_INVALID_PARAMETER, "Invalid entity ID",
+                                 {{"details", entity_validation.error()}, {"entity_id", entity_id}});
+      return;
     }
 
     if (param_id.empty() || param_id.length() > MAX_AGGREGATED_PARAM_ID_LENGTH) {
@@ -503,7 +505,7 @@ void ConfigHandlers::handle_set_configuration(const httplib::Request & req, http
       return;
     }
 
-    // Parse request body
+    // Parse request body before checking entity existence
     json body;
     try {
       body = json::parse(req.body);
@@ -523,6 +525,12 @@ void ConfigHandlers::handle_set_configuration(const httplib::Request & req, http
       HandlerContext::send_error(res, StatusCode::BadRequest_400, ERR_INVALID_REQUEST, "Missing 'data' field",
                                  {{"details", "Request body must contain 'data' field"}});
       return;
+    }
+
+    // Now validate entity exists and matches route type
+    auto entity_opt = ctx_.validate_entity_for_route(req, res, entity_id);
+    if (!entity_opt) {
+      return;  // Error response already sent
     }
 
     // Get aggregated configurations info


src/ros2_medkit_gateway/include/ros2_medkit_gateway/http/http_utils.hpp
```diff
@@ -32,6 +34,37 @@ inline std::string api_path(const std::string & endpoint) {
...
+ *
+ * @param path Request path (e.g., "/api/v1/components/my_component/data")
+ * @return Expected entity type, or UNKNOWN if path doesn't match entity routes
+ */
+inline SovdEntityType extract_entity_type_from_path(const std::string & path) {
+  // Path format: /api/v1/{entity_type}/{id}/...
+  // Check for each entity type prefix after API base path
+  const std::string base = std::string(API_BASE_PATH) + "/";
+
+  if (path.find(base + "components/") == 0 || path.find(base + "components") == 0) {
```
mfaferek93: a path like /api/v1/componentsx would incorrectly match because `find("/api/v1/components")` returns 0. Unlikely to hit in practice since no such routes exist, but safer to use exact match for the no-slash case:                                                                                             
`if (path.find(base + "components/") == 0 || path == base + "components") {`                                      
Same applies to apps (line 54), areas (line 57), and functions (line 60).

src/ros2_medkit_gateway/include/ros2_medkit_gateway/http/handlers/handler_context.hpp
```diff
@@ -172,6 +176,19 @@ class HandlerContext {
...
+   *
+   * Ensures semantic correctness - /components/{id} should only accept
+   * component IDs, not app IDs. This prevents the dual-path routing bug
+   * where /components/{app_id} would incorrectly succeed.
+   *
+   * @param entity Entity information (from get_entity_info)
+   * @param expected_type Expected type based on route (from extract_entity_type_from_path)
+   * @return std::nullopt if valid, error message if type mismatch
+   */
+  static std::optional<std::string> validate_entity_type(const EntityInfo & entity, SovdEntityType expected_type);
```
mfaferek93: it looks like it's not being used anywhere - handlers do inline validation instead. Either use it to reduce duplication in the handlers, or remove it to avoid dead code.

src/ros2_medkit_gateway/src/http/handlers/data_handlers.cpp
```diff
@@ -47,10 +47,26 @@ void DataHandlers::handle_list_data(const httplib::Request & req, httplib::Respo
...
-    if (!entity_ref) {
-      HandlerContext::send_error(res, StatusCode::NotFound_404, ERR_ENTITY_NOT_FOUND, "Entity not found",
-                                 {{"entity_id", entity_id}});
+
+    // Get expected type from route and look up entity in that specific collection
+    auto expected_type = extract_entity_type_from_path(req.path);
+    auto entity_info = ctx_.get_entity_info(entity_id, expected_type);
+    if (entity_info.type == EntityType::UNKNOWN) {
+      // Check if entity exists in ANY collection (for better error message)
+      auto any_entity = ctx_.get_entity_info(entity_id);
```
mfaferek93: this validation block is repeated 3 times in this file (lines 52, 156, 301) and similar pattern in fault_handlers.
Consider extracting to a helper or using the validate_entity_type() method you added to HandlerContext.

src/ros2_medkit_gateway/src/http/handlers/config_handlers.cpp
```diff
@@ -236,6 +237,18 @@ void ConfigHandlers::handle_list_configurations(const httplib::Request & req, ht
       return;
     }
 
+    // Validate entity type matches the route path
+    auto expected_type = extract_entity_type_from_path(req.path);
+    if (expected_type != SovdEntityType::UNKNOWN && entity_ref->type != expected_type) {
```
mfaferek93: this uses a different validation pattern than data_handlers/fault_handlers (direct type comparison vs two-step   lookup with better error message). Both work, but the inconsistency might confuse future readers.

src/ros2_medkit_gateway/include/ros2_medkit_gateway/http/http_utils.hpp
```diff
@@ -32,6 +34,37 @@ inline std::string api_path(const std::string & endpoint) {
...
+    return SovdEntityType::COMPONENT;
+  }
+  if (path.find(base + "apps/") == 0 || path.find(base + "apps") == 0) {
+    return SovdEntityType::APP;
+  }
+  if (path.find(base + "areas/") == 0 || path.find(base + "areas") == 0) {
+    return SovdEntityType::AREA;
+  }
+  if (path.find(base + "functions/") == 0 || path.find(base + "functions") == 0) {
```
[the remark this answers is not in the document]
bburda: Applied the suggested fix with the `matches_collection` lambda. The function now properly requires a segment boundary (either exact match or `/` after collection name), preventing false positives like `/api/v1/componentship`.

src/ros2_medkit_gateway/test/test_integration.test.py
```diff
@@ -4248,3 +4258,163 @@ def test_115_list_function_data_invalid_id(self):
...
+        apps_response = requests.get(f'{self.BASE_URL}/apps', timeout=10)
+        self.assertEqual(apps_response.status_code, 200)
+
+        apps = apps_response.json().get('items', [])
+        if len(apps) == 0:
+            self.skipTest('No apps available for testing')
+
+        # Use the first available app
+        app_id = apps[0]['id']
+
```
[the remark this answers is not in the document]
bburda: Implemented the suggested fix for all affected tests (116-120). Each test now:
1. Fetches component IDs first
2. Finds an app ID that doesn't collide with any component ID
3. Skips the test if no non-colliding app ID is available

src/ros2_medkit_gateway/include/ros2_medkit_gateway/http/handlers/handler_context.hpp
```diff
@@ -172,6 +176,19 @@ class HandlerContext {
...
+
   /**
```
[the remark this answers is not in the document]
bburda: Removed `validate_entity_type` from both header and implementation. The handlers already implement their own entity-type validation inline, making this helper redundant.

src/ros2_medkit_gateway/include/ros2_medkit_gateway/http/http_utils.hpp
bburda: Right, fixed!

src/ros2_medkit_gateway/include/ros2_medkit_gateway/http/handlers/handler_context.hpp
bburda: Reduced duplicated code 😄

src/ros2_medkit_gateway/src/http/handlers/data_handlers.cpp
bburda: Yh, fixed.

src/ros2_medkit_gateway/src/http/handlers/config_handlers.cpp
bburda: Now all handlers uses same pattern.

PR #188, opened 2026-02-13
Add caching to build_component_topic_map with graph invalidation

### LINKED ISSUE #184: Add caching to build_component_topic_map()
## Summary

In `src/ros2_medkit_gateway/src/native_topic_sampler.cpp` (line 492), there is a TODO comment:

```cpp
// TODO(optimization): Cache the map and invalidate on graph changes
```

The `build_component_topic_map()` function is called on every data request to map components to their associated topics. This involves iterating over the full entity cache each time, which is unnecessary overhead when the ROS 2 graph hasn't changed.

---

1. Add a `std::unordered_map<std::string, std::vector<TopicInfo>>` cache member to `NativeTopicSampler`
2. Populate it on first call to `build_component_topic_map()`
3. Invalidate the cache when the ROS 2 graph changes (subscribe to graph events via `rclcpp::Node::get_graph_event()` or check a graph change counter)
4. Return the cached map on subsequent calls if the graph hasn't changed

```cpp
// Pseudocode:
const auto& NativeTopicSampler::get_component_topic_map() {
  auto current_graph_id = node_->get_graph_guard_condition();
  if (cached_graph_id_ != current_graph_id || topic_map_cache_.empty()) {
    topic_map_cache_ = build_component_topic_map();
    cached_graph_id_ = current_graph_id;
  }
  return topic_map_cache_;
}
```

---

- File: `src/ros2_medkit_gateway/src/native_topic_sampler.cpp` (line 492)
- Header: `include/ros2_medkit_gateway/native_topic_sampler.hpp`
- The node already tracks graph changes in `DiscoveryManager` - the same pattern can be reused
- Thread safety: Access the cache through the existing `get_thread_safe_cache()` pattern
- Measure the impact with a simple benchmark: time `build_component_topic_map()` with >10 nodes

ganeshkarthik016:
# Pull Request

## Summary

Adds caching to build_component_topic_map() in NativeTopicSampler to avoid rebuilding the full component-topic map on every request.

The cache is invalidated when the ROS 2 graph changes, reducing unnecessary iteration over the full entity cache and improving performance when the graph is stable.

---

## Issue

- closes #184

---

## Type

- [ ] Bug fix
- [x] New feature or tests
- [ ] Breaking change
- [ ] Documentation only

---

## Testing
Build verified locally.

Manually reviewed logic to ensure:
- Cache is rebuilt when graph change count differs
- Cached map is reused when graph is unchanged
- Thread safety maintained via mutex protection
- Relies on existing CI to verify build and integration.

No functional behavior changes expected.

---

## Checklist

### DIFF
--- a/src/ros2_medkit_gateway/src/native_topic_sampler.cpp
+++ b/src/ros2_medkit_gateway/src/native_topic_sampler.cpp
@@ -491,12 +491,28 @@ NativeTopicSampler::get_component_topic_map()
 {
   std::lock_guard<std::mutex> lock(topic_map_mutex_);
 
-  // Use ROS 2 graph change counter
-  auto current_graph_change = node_->count_publishers("");
+  // Compute a lightweight fingerprint of the ROS graph (topics+types)
+  // Note: ROS 2 does not expose a simple graph change counter; using
+  // get_topic_names_and_types() and hashing the result is a reliable
+  // way to detect graph changes that affect topic/component mapping.
+  auto all_topics = node_->get_topic_names_and_types();
 
-  if (topic_map_cache_.empty() || cached_graph_change_count_ != current_graph_change) {
+  size_t current_graph_fingerprint = 1469598103934665603ULL;  // FNV-1a seed-ish
+  std::hash<std::string> hasher;
+  // Incorporate topic count
+  current_graph_fingerprint ^= all_topics.size() + 0x9e3779b97f4a7c15ULL + (current_graph_fingerprint << 6) + (current_graph_fingerprint >> 2);
+  for (const auto & kv : all_topics) {
+    const auto & topic = kv.first;
+    const auto & types = kv.second;
+    current_graph_fingerprint ^= hasher(topic) + 0x9e3779b97f4a7c15ULL + (current_graph_fingerprint << 6) + (current_graph_fingerprint >> 2);
+    if (!types.empty()) {
+      current_graph_fingerprint ^= hasher(types[0]) + 0x9e3779b97f4a7c15ULL + (current_graph_fingerprint << 6) + (current_graph_fingerprint >> 2);
+    }
+  }
+
+  if (topic_map_cache_.empty() || cached_graph_change_count_ != current_graph_fingerprint) {
     topic_map_cache_ = build_component_topic_map();
-    cached_graph_change_count_ = current_graph_change;
+    cached_graph_change_count_ = current_graph_fingerprint;
   }
 
   // Return a copy to avoid callers holding a reference to internal cache


src/ros2_medkit_gateway/src/native_topic_sampler.cpp
```diff
@@ -486,11 +486,30 @@ std::map<std::string, ComponentTopics> NativeTopicSampler::build_component_topic
   RCLCPP_DEBUG(node_->get_logger(), "Built topic map for %zu components", component_map.size());
   return component_map;
 }
+const std::map<std::string, ComponentTopics>&
+NativeTopicSampler::get_component_topic_map()
+{
+  std::lock_guard<std::mutex> lock(topic_map_mutex_);
+
+  // Use ROS 2 graph change counter
+  auto current_graph_change = node_->count_publishers("");
```
mfaferek93: the `count_publishers("")` doesn't track graph changes, it counts publishers on an empty topic name (which probably always returns 0). The result is the cache never invalidates properly. ROS 2 doesn't expose a simple graph change counter unfortunately :/

src/ros2_medkit_gateway/src/native_topic_sampler.cpp
```diff
@@ -486,11 +486,30 @@ std::map<std::string, ComponentTopics> NativeTopicSampler::build_component_topic
   RCLCPP_DEBUG(node_->get_logger(), "Built topic map for %zu components", component_map.size());
   return component_map;
 }
+const std::map<std::string, ComponentTopics>&
```
mfaferek93: Returning `const& to topic_map_cache_` I believe is risky, the mutex lock drops when the function returns, so the caller holds a reference to data that another thread could replace at any moment. Safer to just return by value (it's a map copy, but it only happens on graph changes so it's fine), or keep the lock alive with a different pattern.

src/ros2_medkit_gateway/include/ros2_medkit_gateway/native_topic_sampler.hpp
```diff
@@ -280,16 +280,22 @@ class NativeTopicSampler {
...
   std::shared_ptr<ros2_medkit_serialization::JsonSerializer> serializer_;
+
+  // Getter for cached component-topic map
+  const std::map<std::string, ComponentTopics>&
+  get_component_topic_map();
+
+  // --- Component topic map cache ---
+  std::map<std::string, ComponentTopics> topic_map_cache_;
+  size_t cached_graph_change_count_{0};
+  mutable std::mutex topic_map_mutex_;
```
mfaferek93, line 298: mutable on the mutex isn't needed here since `get_component_topic_map()` isn't a const method, right?

src/ros2_medkit_gateway/src/native_topic_sampler.cpp
```diff
@@ -486,11 +486,30 @@ std::map<std::string, ComponentTopics> NativeTopicSampler::build_component_topic
...
+    cached_graph_change_count_ = current_graph_change;
+  }
+
+  return topic_map_cache_;
+}
+
 
 ComponentTopics NativeTopicSampler::get_component_topics(const std::string & component_fqn) {
   // Build full map and extract for this component
   // TODO(optimization): Cache the map and invalidate on graph changes
```
mfaferek93: The old TODO comment is still here, can be removed now since this PR implements it :)

src/ros2_medkit_gateway/include/ros2_medkit_gateway/native_topic_sampler.hpp
```diff
@@ -280,16 +280,22 @@ class NativeTopicSampler {
    */
   static bool is_system_topic(const std::string & topic_name);
 
- private:
-  /**
-   * @brief Get the message type for a topic from the graph
-   */
+private:
```
mfaferek93: Small nit
`private:` label lost its indentation and there's an extra blank line before the closing brace. Probably just an editor thing, but worth reverting to keep the diff minimal.

bburda (changes requested):
Hi @ganeshkarthik016, thanks for picking up this optimization! Caching the component-topic map is the right direction. I see you've already addressed some of @mfaferek93's initial feedback. Nice iteration.
However, there are still several issues that need to be resolved before this can be merged.

bburda (changes requested):
Also please resolve conflicts so we can run CI.

src/ros2_medkit_gateway/include/ros2_medkit_gateway/native_topic_sampler.hpp
```diff
@@ -278,18 +278,24 @@ class NativeTopicSampler {
    * @param topic_name Full topic path
    * @return true if this is a system topic that should be filtered
    */
-  static bool is_system_topic(const std::string & topic_name);
+    static bool is_system_topic(const std::string & topic_name);
```
bburda, line 281: Indentation broken: `is_system_topic` uses 4-space indent, project convention is 2-space (Google clang-format). Please run `clang-format`.

src/ros2_medkit_gateway/include/ros2_medkit_gateway/native_topic_sampler.hpp
```diff
@@ -278,18 +278,24 @@ class NativeTopicSampler {
...
 
- private:
-  /**
-   * @brief Get the message type for a topic from the graph
-   */
-  std::string get_topic_type(const std::string & topic_name);
+    private:
```
bburda, line 283: `private:` label and `get_topic_type` have wrong indentation (4-space). Also, the Doxygen `@brief` comment for `get_topic_type()` was removed, please restore it.

src/ros2_medkit_gateway/include/ros2_medkit_gateway/native_topic_sampler.hpp
```diff
@@ -278,18 +278,24 @@ class NativeTopicSampler {
...
+    std::string get_topic_type(const std::string & topic_name);
 
   rclcpp::Node * node_;
 
-  /// Native JSON serializer for topic deserialization
   std::shared_ptr<ros2_medkit_serialization::JsonSerializer> serializer_;
```
bburda, line 288: The `/// Native JSON serializer for topic deserialization` doc comment on `serializer_` was removed. Please keep existing documentation.

src/ros2_medkit_gateway/include/ros2_medkit_gateway/native_topic_sampler.hpp
```diff
@@ -278,18 +278,24 @@ class NativeTopicSampler {
...
+
+  // Getter for cached component-topic map
+  // Returns a copy to avoid exposing internal data while the mutex is released.
```
bburda, line 291: New `get_component_topic_map()` comment says "Returns a copy" which is good, but the method should have a proper Doxygen `@brief` block per project conventions (see other methods in this header)

src/ros2_medkit_gateway/src/native_topic_sampler.cpp
```diff
@@ -486,11 +486,43 @@ std::map<std::string, ComponentTopics> NativeTopicSampler::build_component_topic
   RCLCPP_DEBUG(node_->get_logger(), "Built topic map for %zu components", component_map.size());
   return component_map;
 }
+std::map<std::string, ComponentTopics>
```
bburda, line 489: Missing blank line between `build_component_topic_map()` closing brace (line 487) and `get_component_topic_map()`. Separate function definitions with a blank line per project style.

src/ros2_medkit_gateway/src/native_topic_sampler.cpp
```diff
@@ -486,11 +486,43 @@ std::map<std::string, ComponentTopics> NativeTopicSampler::build_component_topic
...
+NativeTopicSampler::get_component_topic_map()
+{
+  std::lock_guard<std::mutex> lock(topic_map_mutex_);
```
bburda, line 492: `lock_guard` holds `topic_map_mutex_` for the entire duration including `build_component_topic_map()` (L514), which makes multiple expensive ROS 2 graph calls (`get_publishers_info_by_topic` + `get_subscriptions_info_by_topic` per topic). This blocks all concurrent callers. Better pattern: check fingerprint under lock → release lock → build outside lock → re-acquire lock → double-check → swap.

src/ros2_medkit_gateway/src/native_topic_sampler.cpp
```diff
@@ -486,11 +486,43 @@ std::map<std::string, ComponentTopics> NativeTopicSampler::build_component_topic
...
+
+  // Compute a lightweight fingerprint of the ROS graph (topics+types)
+  // Note: ROS 2 does not expose a simple graph change counter; using
+  // get_topic_names_and_types() and hashing the result is a reliable
+  // way to detect graph changes that affect topic/component mapping.
+  auto all_topics = node_->get_topic_names_and_types();
```
bburda, line 498: `get_component_topic_map()` calls it here (L498) for fingerprinting, then `build_component_topic_map()` calls it AGAIN internally (L450). This doubles the graph query cost on every miss. Refactor `build_component_topic_map()` to accept the already-fetched topic list, or share the query result.

src/ros2_medkit_gateway/src/native_topic_sampler.cpp
```diff
@@ -486,11 +486,43 @@ std::map<std::string, ComponentTopics> NativeTopicSampler::build_component_topic
...
+
+  size_t current_graph_fingerprint = 1469598103934665603ULL;  // FNV-1a seed-ish
```
bburda, line 500: Magic number `1469598103934665603ULL` and `0x9e3779b97f4a7c15ULL` (L503, 507, 509) - these should be named constants (`kFnvOffsetBasis`, `kHashCombineMagic`) or just use `std::hash`

src/ros2_medkit_gateway/src/native_topic_sampler.cpp
```diff
@@ -486,11 +486,43 @@ std::map<std::string, ComponentTopics> NativeTopicSampler::build_component_topic
...
+  std::hash<std::string> hasher;
+  // Incorporate topic count
+  current_graph_fingerprint ^= all_topics.size() + 0x9e3779b97f4a7c15ULL + (current_graph_fingerprint << 6) + (current_graph_fingerprint >> 2);
```
bburda, line 503: The fingerprint only hashes topic names+types, but `build_component_topic_map()` also queries per-topic publisher/subscriber info. Cache won't invalidate when: a new node publishes to existing topic, a subscriber leaves, or node names change. Must also hash `count_publishers()` + `count_subscribers()` per topic into the fingerprint.

src/ros2_medkit_gateway/src/native_topic_sampler.cpp
```diff
@@ -486,11 +486,43 @@ std::map<std::string, ComponentTopics> NativeTopicSampler::build_component_topic
...
+  for (const auto & kv : all_topics) {
+    const auto & topic = kv.first;
+    const auto & types = kv.second;
+    current_graph_fingerprint ^= hasher(topic) + 0x9e3779b97f4a7c15ULL + (current_graph_fingerprint << 6) + (current_graph_fingerprint >> 2);
+    if (!types.empty()) {
+      current_graph_fingerprint ^= hasher(types[0]) + 0x9e3779b97f4a7c15ULL + (current_graph_fingerprint << 6) + (current_graph_fingerprint >> 2);
+    }
+  }
+
+  if (topic_map_cache_.empty() || cached_graph_change_count_ != current_graph_fingerprint) {
```
bburda, line 513: Condition: `topic_map_cache_.empty()` - after the very first call, the cache will never be empty even if the graph truly has zero topics (empty map is still cached). Consider using `std::optional` or a bool flag `cache_valid_` instead.

PR #227, opened 2026-02-19
Refactor the gateway integration test suite into a dedicated `ros2_medkit_integration_tests` package

### LINKED ISSUE #139: Refactor integration tests for multiple discovery modes
## Summary

Refactor integration test suite to support running the same tests across different discovery modes (runtime, manifest-only, hybrid). Currently we have separate test files for different modes which leads to code duplication and inconsistent coverage.

---

## Proposed solution

1. **Create a parameterized test base class** that can run against different discovery configurations:
   ```python
   class DiscoveryModeTestBase(unittest.TestCase):
       discovery_mode = None  # Set by subclass or pytest parametrize
       expected_capabilities = {}  # What features are available in this mode
   ```

2. **Define capability flags per discovery mode**:
   ```python
   DISCOVERY_CAPABILITIES = {
       "runtime": {
           "dynamic_nodes": True,
           "configurations": True,  # Requires running nodes
           "operations": True,
           "faults": True,
           "data_read": True,
           "data_write": True,
       },
       "manifest_only": {
           "dynamic_nodes": False,
           "configurations": False,  # No running nodes
           "operations": False,      # Can list, but not execute
           "faults": False,
           "data_read": False,
           "data_write": False,
       },
       "hybrid": {
           # Manifest structure + runtime data where available
       }
   }
   ```

3. **Use skip decorators based on capabilities**:
   ```python
   @requires_capability("configurations")
   def test_get_configuration_value(self):
       # Only runs if discovery mode supports configurations
       ...

   @requires_capability("operations")
   def test_execute_operation(self):
       # Only runs if discovery mode supports operation execution
       ...
   ```

4. **Consolidate test files**:
   - Merge `test_integration.test.py` and `test_discovery_manifest.test.py`
   - Use pytest parametrize or separate launch files per mode
   - Single source of truth for API contract tests

5. **Launch file structure**:
   ```
   test/
     test_api.test.py           # Unified test file
     launch/
       test_runtime.launch.py   # Runtime discovery mode
       test_manifest.launch.py  # Manifest-only mode
       test_hybrid.launch.py    # Hybrid mode
   ```

---

## Additional context

- Current pain points:
  - `test_integration.test.py` focuses on runtime mode
  - `test_discovery_manifest.test.py` duplicates many tests for manifest mode
  - Inconsistent expected status codes between modes

- Benefits:
  - Single set of API contract tests
  - Clear documentation of what works in each mode
  - Easier to add new discovery modes
  - Reduced maintenance burden

- Implementation notes:
  - Consider using pytest-launch-testing for better parametrization
  - Each test should document which capability it requires
  - Tests that work in all modes should not require any capabilities

### LINKED ISSUE #222: [BUG] Flaky test_101_update_execution on Humble - action creation returns 500 due to concurrent goal race condition
# Bug report

### Steps to reproduce
1. Push to `main` branch (or any branch triggering CI)
2. Observe the `build-and-test (humble, ubuntu:jammy)` job
3. `test_101_update_execution` in `test_integration.test.py` fails intermittently

The test can be reproduced locally on ROS 2 Humble by running integration tests:
```bash
source /opt/ros/humble/setup.bash
colcon test --packages-select ros2_medkit_gateway --ctest-args -R test_integration
```

### Expected behavior

`test_101_update_execution` should pass consistently on all supported ROS 2 distros. The test creates a new action execution via `POST /apps/{app_id}/operations/long_calibration/executions` and expects HTTP 202 (Accepted).

### Actual behavior

On ROS 2 Humble, the action execution creation sometimes returns HTTP 500 instead of 202.

**Assertion error** (line 3697 of `test_integration.test.py`):
```
AssertionError: 500 != 202 : Expected 202 for action creation, got 500
```

**Root cause analysis:**

The test (`test_101`) creates a second action execution (Fibonacci with `order=10`) while a previous execution from `test_100` (Fibonacci with `order=5`) may still be in progress on the same action server (`/powertrain/engine/long_calibration`).

Gateway logs show the `SendGoal` request IS dispatched and the demo action server DOES receive and start executing the goal:
```
[gateway_node-1] SendGoal request type: example_interfaces/action/Fibonacci_SendGoal_Request, JSON: {"goal":{"order":10},...}
[gateway_node-1] Sending action goal: /powertrain/engine/long_calibration (type: example_interfaces/action/Fibonacci)
[demo_long_calibration_action-9] Received calibration goal request with order 10
[demo_long_calibration_action-9] Executing calibration...
```

However, unlike successful executions, there is **no** subsequent `Action goal accepted with ID: ...` log line from the gateway. The gateway returns 500 to the test client, suggesting an exception or timeout in the action client's `SendGoal` response handling on Humble.

For comparison, a later test (`test_41`) successfully creates the same action goal on the same action server and the gateway correctly logs `Action goal accepted with ID: ...` and returns 202.

**Flakiness pattern:**
- Fails only on **Humble** (`ubuntu:jammy`) - passes consistently on **Jazzy** and **Rolling**
- The same Humble job also failed on a previous run ([#22103100298](https://github.com/selfpatch/ros2_medkit/actions/runs/22103100298)) on the `feat/91-rate-limiting` branch
- Likely a race condition in ROS 2 Humble's `rclcpp_action` client when sending a new goal while a previous goal on the same action server is still executing

### Environment
- ros2_medkit version: `2531c67` (main, 2026-02-17)
- ROS 2 distro: **Humble** (fails) / Jazzy, Rolling (pass)
- OS: Ubuntu 22.04 (Jammy) in GitHub Actions CI
- CI run: https://github.com/selfpatch/ros2_medkit/actions/runs/22112666500/job/63912660542

### Additional information

**Failing job summary:**
- Job: `build-and-test (humble, ubuntu:jammy)` - **failure**
- Job: `build-and-test (jazzy, ubuntu:noble)` - success
- Job: `build-and-test (rolling, ubuntu:noble)` - success
- Job: `coverage` - success

**Test results:** 1260 total tests, 1 integration test failure (`test_101_update_execution`), 7 skipped

**Possible fixes:**
1. Add a small delay or wait for the previous action to complete before creating a new execution in `test_101`
2. Cancel the previous execution explicitly at the start of `test_101`
3. Use a separate action server / different operation for `test_101` to avoid contention
4. Investigate if the gateway's action client needs retry logic or better error handling for concurrent goals on Humble

bburda:
# Pull Request

## Summary

Refactor the gateway integration test suite into a dedicated `ros2_medkit_integration_tests` package with a shared Python test library (`ros2_medkit_test_utils`), two-tier test organization (features + scenarios), CMake labels, and structural fixes for flaky tests.

**What changed:**

- **New package `ros2_medkit_integration_tests`** - contains all integration tests, demo nodes, launch files, config YAMLs, and a shared Python test library
- **Shared test library `ros2_medkit_test_utils`** (4 modules) - eliminates ~1,200 lines of duplicated code: `get_coverage_env()` (9 copies), health polling (9 variants), demo node launch boilerplate, HTTP helpers
- **Monolithic `test_integration.test.py` (4,990 lines, 141 tests) split** into 16 feature test files (~90 tests) and 10 scenario test files (~50 tests) - each file launches its own gateway with only the demo nodes it needs
- **CMake labels** (`integration;feature`, `integration;scenario`) enable `colcon test --ctest-args -L feature` filtering
- **Gateway package cleaned up** - removed integration test files, demo node build targets, and test-only dependencies; unit tests (26 GTest targets) stay unchanged

**Key library modules:**
- `constants.py` - shared API paths, ports, timeouts (replaces 50+ duplicated lines)
- `coverage.py` - single `get_coverage_env()` (replaces 9 copies, ~270 lines)
- `launch_helpers.py` - `create_test_launch()` factory (replaces ~500 lines of inline launch boilerplate)
- `gateway_test_case.py` - `GatewayTestCase` base class with health polling, HTTP helpers, discovery waiters, assertion helpers

---

## Issue

- closes #222
- closes #139

---

## Type

- [ ] Bug fix
- [x] New feature or tests
- [ ] Breaking change
- [ ] Documentation only

---

## Testing

**Build verification:**
```bash
colcon build --symlink-install
# All 7 packages build successfully
```

**Gateway unit tests (unchanged):**
```bash
colcon test --packages-select ros2_medkit_gateway
colcon test-result --verbose
# 26/26 GTest targets pass
```

**Integration test registration:**
```bash
cd build/ros2_medkit_integration_tests
ctest -N                  # 26 tests registered (16 feature + 10 scenario)
ctest -N -L feature       # 16 feature tests
ctest -N -L scenario      # 10 scenario tests
ctest -N -L integration   # 26 total
```

**Run integration tests:**
```bash
colcon test --packages-select ros2_medkit_integration_tests
colcon test-result --verbose
```

**Coverage still works from separate package** - `get_coverage_env()` resolves `ros2_medkit_gateway`'s build dir via `ament_index_python`, gateway binary writes `.gcda` to `GCOV_PREFIX`, CI `lcov --capture --directory build` scans the entire build directory.

**Issue #222 (flaky test_101):** The race condition is resolved structurally - `test_scenario_action_lifecycle.test.py` replaces the problematic test_100 + test_101 sequence. Each test creates its own execution (no concurrent action goals from previous tests).

**Issue #139 (multi-discovery-mode):** Two new scenario files (`test_scenario_discovery_manifest.test.py`, `test_scenario_discovery_hybrid.test.py`) test manifest-only and hybrid modes. Shared assertion helpers (`assert_entity_exists`, `assert_entity_has_capabilities`, etc.) in `GatewayTestCase` support mode-specific validation. Runtime mode is tested implicitly by all feature tests (default mode).

---

## Checklist


src/ros2_medkit_integration_tests/test/scenarios/test_scenario_fault_inspection.test.py
```diff
@@ -0,0 +1,203 @@
...
+    def _get_fault_detail(self):
+        """Wait for a lidar fault, then GET the full detail response.
+
+        Returns the parsed JSON body, or skips the test on timeout.
+        """
+        fault_code = self.poll_endpoint_until(
+            f'{self.LIDAR_ENDPOINT}/faults',
+            lambda d: next(
+                (
+                    item.get('faultCode')
```
mfaferek93: **Bug:** The poll condition uses `faultCode` (camelCase) but the listing API returns `fault_code` (snake_case) - confirmed in `fault_manager.cpp:61`. This means the condition never matches, `poll_endpoint_until` times out, and all 6 tests in this file silently skip via `skip_on_timeout=True`.

```python
# Current (never matches):
item.get('faultCode')

# Should be:
item.get('fault_code')
```

src/ros2_medkit_integration_tests/package.xml
```diff
@@ -0,0 +1,32 @@
...
+  <depend>rclcpp_action</depend>
+  <depend>std_msgs</depend>
+  <depend>std_srvs</depend>
+  <depend>sensor_msgs</depend>
+  <depend>rcl_interfaces</depend>
+  <depend>example_interfaces</depend>
+  <depend>ros2_medkit_msgs</depend>
+
+  <!-- Test dependencies (Python integration tests) -->
+  <test_depend>launch_testing_ament_cmake</test_depend>
```
mfaferek93, line 25: Missing a few test dependencies that could bite on minimal build environments:

```xml
<test_depend>launch_testing</test_depend>
<test_depend>launch_ros</test_depend>
<test_depend>ament_index_python</test_depend>
```

`launch_testing_ament_cmake` is the CMake glue, but the Python packages (`launch_testing`, `launch_ros`) and `ament_index_python` (used in `coverage.py`) are separate.

src/ros2_medkit_integration_tests/ros2_medkit_test_utils/gateway_test_case.py
```diff
@@ -0,0 +1,765 @@
...
+                        f'Missing areas: {missing_areas}, '
+                        f'Missing apps: {missing_apps}'
+                    )
+            except requests.exceptions.RequestException:
+                # Connection errors expected while nodes are starting; retry.
+                pass
+            time.sleep(DISCOVERY_INTERVAL)
+
+        # Deadline reached -- warn but continue; individual tests may fail.
+        print('Warning: Discovery timeout, some tests may fail')
```
mfaferek93: `_wait_for_discovery()` only prints a warning on timeout, while `_wait_for_gateway_health()` raises `SkipTest`. This asymmetry means a slow discovery causes a wall of confusing cascading failures instead of a clean skip. Would it make sense to align these?

```python
# Current:
print('Warning: Discovery timeout, some tests may fail')

# Suggested:
raise unittest.SkipTest(
    f'Discovery incomplete after {DISCOVERY_TIMEOUT}s — '
    f'found {len(discovered_apps)} apps, need {cls.MIN_EXPECTED_APPS}'
)
```

src/ros2_medkit_integration_tests/test/scenarios/test_scenario_discovery_hybrid.test.py
```diff
@@ -0,0 +1,469 @@
...
+
+        self.assertGreater(
+            len(online_apps), 0,
+            'No apps are online - runtime linking may have failed'
+        )
+
+    def test_16_app_has_runtime_topics(self):
+        """Online app has topics from runtime discovery."""
+        # Wait a bit for runtime linking
+        time.sleep(3)
```
mfaferek93: These tests (16-17, 22-23) wrap assertions in `if response.status_code == 200:` guards, which means they pass silently when the endpoint isn't ready. Compare with test_15 which correctly uses `poll_endpoint_until()`. Could we do the same here?

```python
# Current (silently passes without asserting anything):
response = requests.get(f'{self.BASE_URL}/apps/engine-temp-sensor/data', timeout=5)
if response.status_code == 200:
    data = response.json()
    if 'items' in data and data['items']:
        # ...assert...

# Suggested — poll until data arrives, then assert unconditionally:
data = self.poll_endpoint_until(
    '/apps/engine-temp-sensor/data',
    lambda d: d.get('items'),
    timeout=15.0,
)
topic_names = [t.get('name', '') for t in data]
self.assertTrue(any('temperature' in name for name in topic_names))
```

Same pattern in `test_scenario_discovery_manifest.test.py:364` (`test_25_discovery_stats`).

src/ros2_medkit_integration_tests/test/scenarios/test_scenario_discovery_manifest.test.py
```diff
@@ -0,0 +1,402 @@
...
+        self.assert_entity_not_found('apps', 'nonexistent')
+
+    def test_14_app_online_status(self):
+        """Apps with running demo nodes should exist in the listing.
+
+        In manifest-only mode apps always exist from the manifest.
+        Online status depends on runtime linking timing.
+        """
+        # Wait a bit for runtime linking
+        time.sleep(5)
```
mfaferek93: Bare `time.sleep(5)` instead of polling. Flaky on slow CI, wasteful on fast machines. The hybrid test_15 already uses `poll_endpoint_until()` correctly - could this follow the same pattern?

Same issue in `test_scenario_discovery_hybrid.test.py:298` (`time.sleep(3)`).

src/ros2_medkit_integration_tests/test/scenarios/test_scenario_subscriptions.test.py
```diff
@@ -13,150 +13,107 @@
...
+            pass
 
         if not cls.app_id:
             raise unittest.SkipTest('Demo temp sensor app not discovered')
 
-        # Phase 3: Wait for data availability on the discovered app
+        # Wait for data availability on the discovered app
         # Skip ROS 2 system topics that don't have continuous data flow
         system_topics = {'/parameter_events', '/rosout'}
         deadline = time.time() + 15.0
```
mfaferek93: Minor: `time.time()` here while the base class consistently uses `time.monotonic()` (immune to NTP clock adjustments). Worth aligning for consistency.

src/ros2_medkit_integration_tests/test/scenarios/test_scenario_subscriptions.test.py
```diff
@@ -223,16 +181,15 @@ def test_01_create_subscription_returns_201_with_correct_schema(self):
         self._delete_subscription(sub['id'])
```
mfaferek93: If an assertion fails before `_delete_subscription()`, the subscription leaks and can affect subsequent tests. `addCleanup()` would guarantee cleanup regardless:

```python
sub = self._create_subscription()
self.addCleanup(self._delete_subscription, sub['id'])
```

(I noticed `addCleanup` / `tearDown` aren't used anywhere in the package currently.)

src/ros2_medkit_integration_tests/ros2_medkit_test_utils/gateway_test_case.py
```diff
@@ -0,0 +1,765 @@
...
+        Minimum number of apps that must be discovered before tests run.
+        Set to 0 to skip discovery waiting entirely.
+    REQUIRED_AREAS : set of str
+        Area IDs that must be discovered. Empty set skips area checking.
+    REQUIRED_APPS : set of str
+        App IDs that must be discovered. Empty set skips app checking.
+
+    """
+
+    PORT = DEFAULT_PORT
```
mfaferek93: Nit: `PORT` class attribute is never read by any test (confirmed with grep). Subclasses override `BASE_URL` directly. Could just remove `PORT` to avoid confusion.

src/ros2_medkit_integration_tests/test/features/test_health.test.py
```diff
@@ -0,0 +1,151 @@
...
+        @verifies REQ_INTEROP_010
+        """
+        data = self.get_json('/')
+        self.assertIn('name', data)
+        self.assertIn('version', data)
+        self.assertIn('endpoints', data)
+        self.assertIn('capabilities', data)
+
+        self.assertEqual(data['name'], 'ROS 2 Medkit Gateway')
+        self.assertEqual(data['version'], '0.1.0')
```
mfaferek93: `assertEqual(data['version'], '0.1.0')` will break on every version bump. A format check (e.g. semver regex) would be more resilient.

src/ros2_medkit_integration_tests/test/features/test_configuration_api.test.py
```diff
@@ -0,0 +1,271 @@
...
+        verify_response = requests.get(
+            f'{self.BASE_URL}/apps/temp_sensor/configurations/min_temp',
+            timeout=10
+        )
+        self.assertEqual(verify_response.status_code, 200)
+        verify_data = verify_response.json()
+        self.assertEqual(verify_data['x-medkit']['parameter']['value'], 80.0)
+
+        # Reset the value back to default using SOVD "data" field
+        requests.put(
```
mfaferek93: If the assertion fails before this reset PUT, `min_temp` stays at `80.0`. `addCleanup()` would guarantee the reset regardless of test outcome.

src/ros2_medkit_integration_tests/test/features/test_auth.test.py
```diff
@@ -279,14 +250,15 @@ def test_10_unsupported_grant_type(self):
...
     """Integration tests for RBAC authorization."""
 
-    BASE_URL = 'http://127.0.0.1:8085/api/v1'
+    BASE_URL = AUTH_BASE_URL
 
     @classmethod
     def setUpClass(cls):
         """Wait for gateway to be ready and get tokens."""
+        cls._wait_for_gateway_health()
         time.sleep(1)
```
mfaferek93: If the auth endpoint isn't ready, `response.status_code != 200` and the token is never stored. Then `_auth_header()` returns `Bearer ""` and auth tests fail with misleading errors. Maybe add a check after the loop?

```python
missing = [r for r in ('admin', 'operator', 'viewer', 'configurator') if r not in cls.tokens]
if missing:
    raise unittest.SkipTest(f'Could not acquire tokens for: {", ".join(missing)}')
```

src/ros2_medkit_integration_tests/test/scenarios/test_scenario_bulk_data_upload.test.py
mfaferek93: **Line 371** — `self.test_component_id` defaults to `None` (line 108). If no component is discovered, URL becomes `.../components/None/bulk-data/firmware`. A quick guard would help:

```python
if not self.test_component_id:
    self.skipTest('No component discovered')
```

src/ros2_medkit_integration_tests/demo_nodes/long_calibration_action.cpp
mfaferek93: The graceful shutdown rewrite is solid — joinable thread with atomic flag, `prepare_shutdown()` ordering, and the `set_terminate` workaround for the rclcpp_action race.

src/ros2_medkit_integration_tests/test/scenarios/test_scenario_fault_inspection.test.py
bburda: Fixed in e0e39e0.

src/ros2_medkit_integration_tests/package.xml
bburda: Fixed in e0e39e0 - added all three as `<test_depend>`.

src/ros2_medkit_integration_tests/ros2_medkit_test_utils/gateway_test_case.py
bburda: Both `_wait_for_gateway_health()` and `_wait_for_discovery()` now raise `AssertionError` on timeout. Fixed in b12dbc4.

src/ros2_medkit_integration_tests/test/scenarios/test_scenario_discovery_hybrid.test.py
bburda: Fixed in e0e39e0 - replaced `if status == 200` guards with `poll_endpoint_until()` in the hybrid test.

src/ros2_medkit_integration_tests/test/scenarios/test_scenario_discovery_manifest.test.py
bburda: Fixed in e0e39e0 - replaced bare `time.sleep()` with `poll_endpoint_until()` in manifest and hybrid tests.

src/ros2_medkit_integration_tests/test/scenarios/test_scenario_subscriptions.test.py
bburda: Fixed in b12dbc4 - all timing now uses `time.monotonic()`.

src/ros2_medkit_integration_tests/test/scenarios/test_scenario_subscriptions.test.py
bburda: Fixed in f554a46 - added `addCleanup` to `test_12`. All subscription-creating tests now use it.

src/ros2_medkit_integration_tests/ros2_medkit_test_utils/gateway_test_case.py
bburda: Removed in f554a46.

src/ros2_medkit_integration_tests/test/features/test_health.test.py
bburda: Fixed in e0e39e0 - uses semver regex (`assertRegex`) instead of exact match.

src/ros2_medkit_integration_tests/test/features/test_configuration_api.test.py
bburda: Fixed in f554a46 - added `addCleanup` with DELETE to reset `min_temp` to default.

src/ros2_medkit_integration_tests/test/features/test_auth.test.py
bburda: Fixed in b12dbc4 - added a check after the token acquisition loop.

src/ros2_medkit_integration_tests/CMakeLists.txt
```diff
@@ -0,0 +1,99 @@
...
+  demo_light_controller
+  demo_calibration_service
+  demo_long_calibration_action
+  demo_lidar_sensor
+  DESTINATION lib/${PROJECT_NAME}
+)
+
+# === Python test utilities ===
+
+ament_python_install_package(ros2_medkit_test_utils)
```
mfaferek93, line 66: **Blocker:** `ament_python_install_package(ros2_medkit_test_utils)` is called but there is no `find_package(ament_cmake_python REQUIRED)` (lines 11-19). Also missing `<buildtool_depend>ament_cmake_python</buildtool_depend>` in `package.xml`. Build will fail at cmake configure with "Unknown CMake command".

src/ros2_medkit_integration_tests/test/features/test_data_read.test.py
```diff
@@ -0,0 +1,575 @@
...
+        Apps are ROS 2 nodes. The temp_sensor app publishes temperature data.
+
+        @verifies REQ_INTEROP_018
+        """
+        data = self.poll_endpoint('/apps/temp_sensor/data')
+        self.assertIn('items', data)
+        items = data['items']
+        self.assertIsInstance(items, list)
+
+        if len(items) > 0:
```
mfaferek93: **Major:** Five `if len(items) > 0:` guards (lines 68, 90, 109, 128, 439) silently skip all structural assertions when items is empty. Tests pass with zero assertions. Use `self.assertGreater(len(items), 0)` before the loop or `poll_endpoint_until` with a non-empty condition.

src/ros2_medkit_integration_tests/test/features/test_operations_api.test.py
```diff
@@ -0,0 +1,509 @@
...
+        # Call the service
+        response = requests.post(
+            f'{self.BASE_URL}/components/powertrain/operations/{operation_id}/executions',
+            json={'parameters': {}},
+            timeout=30
+        )
+
+        # Should return 200 for sync or 400/500 if service is unavailable
+        self.assertIn(
+            response.status_code, [200, 400, 500, 503],
```
mfaferek93: **Major:** Accepts `[200, 400, 500, 503]`. HTTP 500 is a server bug, not a valid outcome. Remove 500 from the list.

src/ros2_medkit_integration_tests/test/scenarios/test_scenario_action_lifecycle.test.py
```diff
@@ -0,0 +1,370 @@
...
+        response = requests.post(
+            f'{self.BASE_URL}/components/powertrain/operations/'
+            f'{operation_id}/executions',
+            json={'parameters': {}},
+            timeout=30,
+        )
+
+        # Should return 200 for sync or 400/500/503 if service unavailable
+        self.assertIn(
+            response.status_code, [200, 400, 500, 503],
```
mfaferek93: **Major:** Same issue: accepts `[200, 400, 500, 503]`. Remove 500.

src/ros2_medkit_integration_tests/test/features/test_configuration_api.test.py
```diff
@@ -0,0 +1,281 @@
...
+
+        # Verify the value was actually reset by reading it
+        get_response = requests.get(
+            f'{self.BASE_URL}/apps/temp_sensor/configurations/min_temp',
+            timeout=10
+        )
+        self.assertEqual(get_response.status_code, 200)
+        get_data = get_response.json()
+        # The value should be reset to default (85.0)
+        _ = get_data['x-medkit']['parameter']['value']
```
mfaferek93: **Major:** `_ = get_data['x-medkit']['parameter']['value']` discards the reset value without asserting it. The comment says "should be reset to default (85.0)" but nothing verifies this. Replace with `self.assertEqual(...)`.

src/ros2_medkit_integration_tests/test/scenarios/test_scenario_fault_lifecycle.test.py
```diff
@@ -0,0 +1,192 @@
...
+
+    def test_02_delete_all_faults_for_component(self):
+        """DELETE /components/{id}/faults clears all faults for component.
+
+        @verifies REQ_INTEROP_014
+        """
+        # Get the perception component (lidar_sensor is under /perception)
+        data = self.get_json('/components')
+        self.assertGreater(len(data['items']), 0)
+        component_id = data['items'][0]['id']
```
mfaferek93: **Major:** `component_id = data['items'][0]['id']` picks an arbitrary first component. May not be the one with lidar faults. DELETE passes vacuously. Use a known component or verify faults are actually gone after DELETE.

src/ros2_medkit_integration_tests/test/scenarios/test_scenario_fault_lifecycle.test.py
```diff
@@ -0,0 +1,192 @@
...
+
+    def test_03_delete_all_faults_for_app(self):
+        """DELETE /apps/{id}/faults clears all faults for app.
+
+        @verifies REQ_INTEROP_014
+        """
+        # Get an app
+        data = self.get_json('/apps')
+        self.assertGreater(len(data['items']), 0)
+        app_id = data['items'][0]['id']
```
mfaferek93: **Major:** Same pattern: `app_id = data['items'][0]['id']` picks first app. Should be `'lidar_sensor'` directly since that's the only app with faults.

src/ros2_medkit_integration_tests/ros2_medkit_test_utils/gateway_test_case.py
```diff
@@ -0,0 +1,859 @@
...
+                )
+                if response.status_code == 200:
+                    data = response.json()
+                    items = data.get('items', [])
+                    if len(items) > 0:
+                        return items[0].get('id')
+            except requests.exceptions.RequestException:
+                pass
+            time.sleep(1)
+        return None
```
mfaferek93: **Major:** `wait_for_fault_with_rosbag()` returns `None` on timeout. Every other `wait_for_*` method raises `AssertionError`. Inconsistent API — callers have to manually check and skip.

src/ros2_medkit_integration_tests/test/features/test_bulk_data_api.test.py
```diff
@@ -0,0 +1,223 @@
...
+        data = response.json()
+        self.assertIn('error_code', data)
+
+    def test_bulk_data_list_descriptors_structure(self):
+        """GET /apps/{app}/bulk-data/rosbags returns BulkDataDescriptor[].
+
+        @verifies REQ_INTEROP_072
+        """
+        # Wait for fault to be generated (lidar sensor has invalid params)
+        time.sleep(3)
```
mfaferek93: **Minor:** Bare `time.sleep(3)` to wait for fault generation. Same pattern that was systematically replaced elsewhere. Use `poll_endpoint_until` or `wait_for_fault`.

src/ros2_medkit_integration_tests/test/scenarios/test_scenario_subscriptions.test.py
```diff
@@ -13,150 +13,107 @@
...
+                apps = r.json().get('items', [])
+                for app in apps:
+                    if 'temp_sensor' in app.get('id', '').lower() or \
+                       'engine' in app.get('id', '').lower():
+                        cls.app_id = app['id']
                         break
-            except Exception:
-                pass
-            time.sleep(DISCOVERY_CHECK_INTERVAL)
+        except Exception:
```
mfaferek93: **Minor:** `except Exception: pass` (also at lines 120, 152, 351, 392) swallows all errors including `KeyError`, `TypeError`, etc. Narrow to `requests.exceptions.RequestException`.

src/ros2_medkit_integration_tests/test/features/test_sse.test.py
```diff
@@ -0,0 +1,126 @@
...
+                'no-cache'
+            )
+
+            # Close the connection (we just wanted to check headers)
+            response.close()
+        except requests.exceptions.ReadTimeout:
+            # Timeout is expected since SSE keeps connection open
+            pass
+
+    def test_sse_stream_sends_keepalive(self):
```
mfaferek93: **Minor:** `test_sse_stream_sends_keepalive` only verifies no connection error during 1-second window. Never checks any keepalive data. Name is misleading — rename to `test_sse_stream_connects_without_error` or actually verify keepalive content.

src/ros2_medkit_integration_tests/test/features/test_sse.test.py
```diff
@@ -0,0 +1,126 @@
...
+    )
+
+
+class TestSse(GatewayTestCase):
+    """SSE (Server-Sent Events) fault stream tests."""
+
+    MIN_EXPECTED_APPS = 1
+    REQUIRED_APPS = {'temp_sensor'}
+
+    def test_sse_stream_endpoint_returns_correct_headers(self):
```
mfaferek93, line 47: **Minor:** If `ReadTimeout` fires before headers arrive, the except clause passes the test with zero header assertions.

src/ros2_medkit_integration_tests/test/scenarios/test_scenario_subscriptions.test.py
```diff
@@ -358,35 +305,33 @@ def test_07_empty_list_after_delete(self):
...
-        """
-        The SSE events endpoint returns text/event-stream content type.
+        """The SSE events endpoint returns text/event-stream content type.
 
         @verifies REQ_INTEROP_090
         """
         sub = self._create_subscription(interval='slow', duration=30)
-        events_url = f'http://localhost:8080{sub["event_source"]}'
+        self.addCleanup(self._delete_subscription, sub['id'])
+        events_url = f'http://localhost:{DEFAULT_PORT}{sub["event_source"]}'
```
mfaferek93: **Minor:** SSE URLs hardcode `http://localhost:{DEFAULT_PORT}` (also at lines 334, 382). If port is overridden (as in auth/rate-limit tests), these point to the wrong server. Derive from `self.BASE_URL`.

src/ros2_medkit_integration_tests/test/features/test_configuration_api.test.py
```diff
@@ -0,0 +1,281 @@
...
+
+        This test modifies min_temp and resets it within the same test.
+
+        @verifies REQ_INTEROP_050
+        """
+        # Register cleanup to reset value even if assertions fail
+        self.addCleanup(
+            lambda: requests.put(
+                f'{self.BASE_URL}/apps/temp_sensor/configurations/min_temp',
+                json={'data': 85.0},
```
mfaferek93: **Minor:** Cleanup resets `min_temp` via PUT to hardcoded `85.0`. If default changes, cleanup silently sets the wrong value. Use DELETE instead (like the cleanup at line 171).

src/ros2_medkit_integration_tests/test/features/test_entity_routing.test.py
```diff
@@ -0,0 +1,203 @@
...
+
+    def test_component_route_rejects_app_id_data(self):
+        """Component /data rejects app IDs.
+
+        In runtime-only discovery mode, /components/{id} should only accept
+        synthetic component IDs, not individual ROS 2 node (app) IDs.
+
+        @verifies REQ_INTEROP_003
+        """
+        app_id = self._find_app_only_id()
```
mfaferek93, line 92: **Minor:** `_find_app_only_id()` called 4 times (lines 85, 116, 133, 150), making 8 identical HTTP requests. Cache the result in `setUpClass`.

src/ros2_medkit_integration_tests/ros2_medkit_test_utils/gateway_test_case.py
```diff
@@ -0,0 +1,859 @@
...
+
+    def create_execution(self, entity_endpoint, operation_id, *, input_data=None):
```
mfaferek93, line 676: **Minor:** `create_execution` calls `response.json()` without checking status code. Non-JSON error response will raise unhelpful `JSONDecodeError`.

src/ros2_medkit_integration_tests/CMakeLists.txt
bburda: Fixed in 02b8163 - added `find_package(ament_cmake_python REQUIRED)` and `<buildtool_depend>ament_cmake_python</buildtool_depend>`.

src/ros2_medkit_integration_tests/test/features/test_data_read.test.py
bburda: Fixed in 02b8163 - all 5 `if len(items) > 0` guards replaced with `self.assertGreater(len(items), 0, ...)` followed by unconditional structural assertions.

src/ros2_medkit_integration_tests/test/features/test_operations_api.test.py
bburda: Fixed in 02b8163 - replaced multi-status list with `assertEqual(response.status_code, 200)`. The calibration service is always available in this test setup.

src/ros2_medkit_integration_tests/test/scenarios/test_scenario_action_lifecycle.test.py
bburda: Fixed in 02b8163 - same fix as operations_api. Also fixed the pause/resume test which accepted `[400, 501]` - now asserts exactly 400.

src/ros2_medkit_integration_tests/test/features/test_configuration_api.test.py
bburda: Fixed in 02b8163 - replaced `_ = ...` with `self.assertEqual(..., 85.0)`. Also added numeric prefixes (test_01_ through test_08_) to enforce execution order, since the gateway's `cache_default_values()` captures current runtime values on first access - without ordering, the cache could be populated after a test already modified the parameter.

src/ros2_medkit_integration_tests/test/scenarios/test_scenario_fault_lifecycle.test.py
bburda: Fixed in 02b8163 - hardcoded `component_id = 'perception'` (the namespace component that groups lidar_sensor).

src/ros2_medkit_integration_tests/test/scenarios/test_scenario_fault_lifecycle.test.py
bburda: Fixed in 02b8163 - hardcoded `app_id = 'lidar_sensor'` directly.

src/ros2_medkit_integration_tests/ros2_medkit_test_utils/gateway_test_case.py
bburda: Fixed in 02b8163 - `wait_for_fault_with_rosbag()` now raises `AssertionError` on timeout, consistent with all other `wait_for_*` methods.

src/ros2_medkit_integration_tests/test/features/test_bulk_data_api.test.py
bburda: Fixed in 02b8163 - replaced `time.sleep(3)` with `poll_endpoint_until` that waits for rosbag items to appear.

src/ros2_medkit_integration_tests/test/scenarios/test_scenario_subscriptions.test.py
bburda: Fixed in 02b8163 - all 5 `except Exception` narrowed to `except requests.exceptions.RequestException`.

src/ros2_medkit_integration_tests/test/features/test_sse.test.py
bburda: Fixed in 02b8163 - renamed to `test_sse_stream_connects_without_error`.

src/ros2_medkit_integration_tests/test/features/test_sse.test.py
bburda: Fixed in 02b8163 - added `headers_verified` flag set inside the try block, with `self.assertTrue(headers_verified, ...)` after the except.

src/ros2_medkit_integration_tests/test/scenarios/test_scenario_subscriptions.test.py
bburda: Fixed in 02b8163 - all 3 SSE URLs now derived from `self.BASE_URL` with `sub["event_source"].removeprefix(API_BASE_PATH)`.

src/ros2_medkit_integration_tests/test/features/test_configuration_api.test.py
bburda: Fixed in 02b8163 - cleanup now uses DELETE instead of PUT with hardcoded value.

src/ros2_medkit_integration_tests/test/features/test_entity_routing.test.py
bburda: Fixed in 02b8163 - result cached in `_cached_app_only_id` class variable, reused across all 4 test methods.

src/ros2_medkit_integration_tests/ros2_medkit_test_utils/gateway_test_case.py
bburda: Fixed in 02b8163 - `create_execution()` now asserts `response.status_code in (200, 201, 202)` before calling `.json()`.

PR #231, opened 2026-02-22
feat: software updates plugin system with 8 SOVD endpoints

### LINKED ISSUE #206: Implement Software Updates management endpoints
## Summary

Software updates endpoints provide a standardized interface for managing update packages at the **server level** (not per-entity). The workflow covers: listing available updates, registering new packages, preparing (download + verify), executing (install), and monitoring progress.

This is a server-level feature - all endpoints are under `/updates`, not under `/{entity-path}/updates`.

---

### 1. `GET /api/v1/updates`

List all available/registered software update packages.

**Query parameters:**

| `origin` | `string` | No | Filter by origin: `remote` or `proximity` |
| `target-version` | `string` | No | Filter by target version string |

**Response `200 OK`:**

```json
{
  "items": ["autonomous", "ADAS-v2-03-2154", "CAM-v1-18-8643"]
}
```

Returns a **string array** of update package identifiers.

---

### 2. `GET /api/v1/updates/{update-package-id}`

Read full metadata for a specific update package.

**Path parameters:**

| `{update-package-id}` | `string` | Yes | Update package identifier |

**Response `200 OK`:**

```json
{
  "id": "ADAS-v2-03-2154",
  "update_name": "ADAS System Update v2.03",
  "automated": true,
  "origins": ["remote", "proximity"],
  "notes": "Fixes lane-keeping edge case in heavy rain",
  "preconditions": "Vehicle must be stationary",
  "execution_conditions": "Engine off, battery > 80%",
  "duration": 1800,
  "size": 524288,
  "updated_components": ["adas_controller", "camera_front"],
  "affected_components": ["adas_controller", "camera_front", "lidar_top"]
}
```

| `id` | `string` | Yes | Update package identifier |
| `update_name` | `string` | Yes | Human-readable name |
| `update_translation_id` | `string` | No | Translation key |
| `automated` | `boolean` | Yes | Whether this update can run fully automatically |
| `origins` | `string[]` | Yes | Supported origins: `remote` (from OTA server), `proximity` (from local diagnostic tester) |
| `notes` | `string` | No | Additional description |
| `user_activity` | `string` | No | What the user needs to do during the update |
| `preconditions` | `string` | No | Requirements before update can start |
| `execution_conditions` | `string` | No | Conditions that must hold during execution |
| `duration` | `integer` | No | Estimated duration in seconds |
| `size` | `integer` | No | Package size in kilobytes |
| `added_components` | `string[]` | No | New components introduced by this update |
| `removed_components` | `string[]` | No | Components removed by this update |
| `updated_components` | `string[]` | No | Components that will be updated |
| `affected_components` | `string[]` | No | Components affected during the update process (may experience downtime) |

**Error:** `404` if update package doesn't exist.

---

### 3. `POST /api/v1/updates`

Register a new update package.

**Request body:** Manufacturer-specific content type (e.g., JSON metadata or binary package).

For a ROS 2 implementation, a JSON body is practical:

```json
{
  "id": "gateway-v1.5.0",
  "update_name": "Gateway Update v1.5.0",
  "automated": false,
  "origins": ["proximity"],
  "notes": "Adds cyclic subscription support",
  "updated_components": ["ros2_medkit_gateway"]
}
```

**Response `201 Created`** with `Location` header pointing to the new update resource.

**Error:** `400` if required fields are missing or ID already exists.

---

### 4. `PUT /api/v1/updates/{update-package-id}/prepare`

Trigger preparation phase: download package, verify integrity, check dependencies.

**Response `202 Accepted`** with `Location` header pointing to the status resource.

---

### 5. `PUT /api/v1/updates/{update-package-id}/execute`

Begin the actual update installation. Should only be called after preparation succeeds.

**Response `202 Accepted`** with `Location` header pointing to the status resource.

---

### 6. `PUT /api/v1/updates/{update-package-id}/automated`

Trigger a fully automated update (prepare + execute in one step). Only available for packages where `automated: true`.

**Response `202 Accepted`** with `Location` header.

**Error:** `400` if the update doesn't support automated execution.

---

### 7. `GET /api/v1/updates/{update-package-id}/status`

Poll the current status of an update.

**Response `200 OK`:**

```json
{
  "status": "inProgress",
  "progress": 45,
  "sub_progress": [
    { "name": "Downloading package", "progress": 90 },
    { "name": "Flashing ECU", "progress": 0 }
  ]
}
```

| `status` | `string` | Yes | See status values below |
| `progress` | `integer` | No | Overall percentage 0–100 |
| `sub_progress` | `object[]` | No | Detailed per-step progress |
| `error` | `GenericError` | No | Error details when status is `failed` |

**Status values:**

| Status | Description |
|--------|-------------|
| `pending` | Queued, not yet started |
| `inProgress` | Update is running |
| `completed` | Successfully finished |
| `failed` | Error occurred |

---

### 8. `DELETE /api/v1/updates/{update-package-id}`

Remove an update package.

**Response `204 No Content`**

**Error responses:**

|--------|-----------|------|
| `404` | `resource-not-found` | Package doesn't exist |
| `405` | `invalid-request` | Cannot delete - update is currently executing or was automatically triggered |

---

### ROS 2 context

Updates could map to different deployment models:
- **Docker deployments:** Pull new container image, restart container
- **Native deployments:** apt/snap package upgrade, colcon workspace rebuild
- **Initial implementation:** Store metadata only, delegate execution to shell scripts

### Architecture

- Create an `UpdateManager` class at the gateway level (not per-entity)
- Store update metadata in memory or SQLite (for persistence across restarts)
- Execution delegates to a pluggable backend (strategy pattern)
- Progress tracking via polling

### Route registration

```cpp
srv->Get(api_path("/updates"), handler);
srv->Get((api_path("/updates") + R"(/([^/]+)$)"), handler);
srv->Post(api_path("/updates"), handler);
srv->Put((api_path("/updates") + R"(/([^/]+)/prepare$)"), handler);
srv->Put((api_path("/updates") + R"(/([^/]+)/execute$)"), handler);
srv->Put((api_path("/updates") + R"(/([^/]+)/automated$)"), handler);
srv->Get((api_path("/updates") + R"(/([^/]+)/status$)"), handler);
srv->Delete((api_path("/updates") + R"(/([^/]+)$)"), handler);
```

bburda:
# Pull Request

## Summary

Add a plugin-based software updates subsystem implementing 8 SOVD-compliant REST endpoints for managing software update packages with async prepare/execute lifecycle.

- **Plugin architecture**: `UpdateBackend` abstract interface loaded at runtime via `dlopen`/`dlsym` from `.so` shared libraries
- **8 REST endpoints**: GET/POST /updates, GET/DELETE /updates/{id}, PUT prepare/execute/automated, GET status - with async operations returning 202 + Location header for status polling
- **Production hardening**: typed `UpdateError` with `UpdateErrorCode` enum replacing string-based errors, TOCTOU-safe deletion with `Deleting` sentinel phase, exception safety in async threads, thread-safety documentation
- **Full test coverage**: 17 GTest unit tests for UpdateManager + 32 integration tests covering CRUD, async lifecycle, automated mode, error cases, failure/exception handling, and 501 no-backend mode
- **Documentation**: API reference (8 endpoints), changelog, configuration guide, requirements traceability (REQ_INTEROP_082-094)
- **Verification script**: enhanced to detect `@verifies` tags before test definitions and generate clickable GitHub links with line numbers

---

## Issue

- closes #206

---

## Type

- [ ] Bug fix
- [x] New feature or tests
- [ ] Breaking change
- [ ] Documentation only

---

## Testing

- **Unit tests**: `colcon test --ctest-args -R test_update_manager` - 17 tests covering CRUD, prepare/execute workflow, automated mode, status polling, concurrent operations, deadlock-free destruction, backend failure/exception handling, and error codes
- **Integration tests**: `colcon test --packages-select ros2_medkit_integration_tests --ctest-args -R test_updates` - 32 tests against running gateway with and without plugin backend
- **Linters**: clang-format, flake8 pass on all new and modified files
- **Build**: Clean build with no warnings on Jazzy
- **Docs**: Sphinx build succeeds with 0 warnings, all GitHub links verified

---

## Checklist

### DIFF
--- a/src/ros2_medkit_gateway/include/ros2_medkit_gateway/updates/update_backend.hpp
+++ b/src/ros2_medkit_gateway/include/ros2_medkit_gateway/updates/update_backend.hpp
@@ -50,6 +50,20 @@ struct UpdateStatusInfo {
 /// Internal phase tracking for update lifecycle
 enum class UpdatePhase { None, Preparing, Prepared, Executing, Executed, Failed, Deleting };
 
+/// Error codes for backend return values
+enum class UpdateBackendError {
+  NotFound,       // Package does not exist
+  AlreadyExists,  // Duplicate ID on registration
+  InvalidInput,   // Malformed metadata
+  Internal        // Unexpected error
+};
+
+/// Typed error for backend return values
+struct UpdateBackendErrorInfo {
+  UpdateBackendError code;
+  std::string message;
+};
+
 /**
  * @brief Thread-safe reporter for update progress.
  *
@@ -92,10 +106,13 @@ class UpdateProgressReporter {
  *
  * @par Thread Safety
  * - CRUD methods (list_updates, get_update, register_update, delete_update)
- *   are called while UpdateManager holds its mutex. They will not be called
- *   concurrently with each other, but may overlap with prepare/execute running
- *   in a background thread. If the backend shares state between CRUD and
- *   async methods, it must provide its own synchronization.
+ *   are called WITHOUT holding UpdateManager's mutex. They may be called
+ *   concurrently with each other and with prepare/execute running in a
+ *   background thread. If the backend shares state, it must provide its own
+ *   synchronization.
+ *   Note: delete_update has a partial guard - UpdateManager checks/sets a
+ *   Deleting sentinel under lock before calling delete_update, but the
+ *   backend call itself runs outside the lock.
  * - prepare() and execute() run in a background std::async thread. They may
  *   run concurrently with CRUD calls from the HTTP thread.
  * - The UpdateProgressReporter passed to prepare/execute is already
@@ -111,28 +128,30 @@ class UpdateBackend {
   // ---- CRUD (plugin owns all metadata storage) ----
 
   /// List all registered update package IDs, optionally filtered
-  virtual tl::expected<std::vector<std::string>, std::string> list_updates(const UpdateFilter & filter) = 0;
+  virtual tl::expected<std::vector<std::string>, UpdateBackendErrorInfo> list_updates(const UpdateFilter & filter) = 0;
 
   /// Get full metadata for a specific update package as JSON
-  virtual tl::expected<nlohmann::json, std::string> get_update(const std::string & id) = 0;
+  virtual tl::expected<nlohmann::json, UpdateBackendErrorInfo> get_update(const std::string & id) = 0;
 
   /// Register a new update package from JSON metadata
-  virtual tl::expected<void, std::string> register_update(const nlohmann::json & metadata) = 0;
+  virtual tl::expected<void, UpdateBackendErrorInfo> register_update(const nlohmann::json & metadata) = 0;
 
   /// Delete an update package
-  virtual tl::expected<void, std::string> delete_update(const std::string & id) = 0;
+  virtual tl::expected<void, UpdateBackendErrorInfo> delete_update(const std::string & id) = 0;
 
   // ---- Async operations (called in background thread by UpdateManager) ----
 
   /// Prepare an update (download, verify, check dependencies).
   /// Reporter is optional - plugin may call reporter.set_progress() etc.
-  virtual tl::expected<void, std::string> prepare(const std::string & id, UpdateProgressReporter & reporter) = 0;
+  virtual tl::expected<void, UpdateBackendErrorInfo> prepare(const std::string & id,
+                                                             UpdateProgressReporter & reporter) = 0;
 
   /// Execute an update (install). Only called after prepare succeeds.
-  virtual tl::expected<void, std::string> execute(const std::string & id, UpdateProgressReporter & reporter) = 0;
+  virtual tl::expected<void, UpdateBackendErrorInfo> execute(const std::string & id,
+                                                             UpdateProgressReporter & reporter) = 0;
 
   /// Check whether a package supports automated mode (prepare + execute)
-  virtual tl::expected<bool, std::string> supports_automated(const std::string & id) = 0;
+  virtual tl::expected<bool, UpdateBackendErrorInfo> supports_automated(const std::string & id) = 0;
 
   /// Optional: receive plugin-specific configuration from YAML
   virtual void configure(const nlohmann::json & /* config */) {
--- a/src/ros2_medkit_gateway/src/http/handlers/update_handlers.cpp
+++ b/src/ros2_medkit_gateway/src/http/handlers/update_handlers.cpp
@@ -60,10 +60,7 @@ json UpdateHandlers::status_to_json(const UpdateStatusInfo & status) {
     }
   }
   if (status.error_message.has_value()) {
-    json err;
-    err["error_code"] = ERR_INTERNAL_ERROR;
-    err["message"] = *status.error_message;
-    j["error"] = err;
+    j["error"] = *status.error_message;
   }
   return j;
 }
@@ -134,6 +131,20 @@ void UpdateHandlers::handle_register_update(const httplib::Request & req, httpli
       return;
     }
 
+    // Validate id field exists before calling backend
+    if (!body.contains("id") || !body["id"].is_string() || body["id"].get<std::string>().empty()) {
+      HandlerContext::send_error(res, 400, ERR_INVALID_REQUEST, "Missing required field: id");
+      return;
+    }
+    auto id = body["id"].get<std::string>();
+
+    // Validate id format to prevent CRLF injection in Location header
+    auto id_validation = ctx_.validate_entity_id(id);
+    if (!id_validation) {
+      HandlerContext::send_error(res, 400, ERR_INVALID_REQUEST, id_validation.error());
+      return;
+    }
+
     auto result = update_mgr_->register_update(body);
     if (!result) {
       switch (result.error().code) {
@@ -147,12 +158,6 @@ void UpdateHandlers::handle_register_update(const httplib::Request & req, httpli
       return;
     }
 
-    // Validate id field exists before using it for Location header
-    if (!body.contains("id") || !body["id"].is_string() || body["id"].get<std::string>().empty()) {
-      HandlerContext::send_error(res, 400, ERR_INVALID_REQUEST, "Missing required field: id");
-      return;
-    }
-    auto id = body["id"].get<std::string>();
     json response = {{"id", id}};
     HandlerContext::send_json(res, response);
     res.status = 201;
--- a/src/ros2_medkit_gateway/src/updates/update_manager.cpp
+++ b/src/ros2_medkit_gateway/src/updates/update_manager.cpp
@@ -23,6 +23,9 @@ UpdateManager::UpdateManager(std::unique_ptr<UpdateBackend> backend, void * plug
 }
 
 UpdateManager::~UpdateManager() {
+  // Signal background tasks to stop accepting new work
+  stopped_ = true;
+
   // Collect all valid futures, then wait OUTSIDE the lock to avoid
   // deadlock (async tasks also acquire mutex_ during execution).
   std::vector<std::future<void>> futures;
@@ -54,7 +57,7 @@ tl::expected<std::vector<std::string>, UpdateError> UpdateManager::list_updates(
   }
   auto result = backend_->list_updates(filter);
   if (!result) {
-    return tl::make_unexpected(UpdateError{UpdateErrorCode::Internal, result.error()});
+    return tl::make_unexpected(UpdateError{UpdateErrorCode::Internal, result.error().message});
   }
   return *result;
 }
@@ -65,7 +68,7 @@ tl::expected<nlohmann::json, UpdateError> UpdateManager::get_update(const std::s
   }
   auto result = backend_->get_update(id);
   if (!result) {
-    return tl::make_unexpected(UpdateError{UpdateErrorCode::NotFound, result.error()});
+    return tl::make_unexpected(UpdateError{UpdateErrorCode::NotFound, result.error().message});
   }
   return *result;
 }
@@ -76,10 +79,15 @@ tl::expected<void, UpdateError> UpdateManager::register_update(const nlohmann::j
   }
   auto result = backend_->register_update(metadata);
   if (!result) {
-    if (result.error().find("already exists") != std::string::npos) {
-      return tl::make_unexpected(UpdateError{UpdateErrorCode::AlreadyExists, result.error()});
+    const auto & err = result.error();
+    switch (err.code) {
+      case UpdateBackendError::AlreadyExists:
+        return tl::make_unexpected(UpdateError{UpdateErrorCode::AlreadyExists, err.message});
+      case UpdateBackendError::InvalidInput:
+        return tl::make_unexpected(UpdateError{UpdateErrorCode::InvalidRequest, err.message});
+      default:
+        return tl::make_unexpected(UpdateError{UpdateErrorCode::Internal, err.message});
     }
-    return tl::make_unexpected(UpdateError{UpdateErrorCode::InvalidRequest, result.error()});
   }
   return {};
 }
@@ -89,10 +97,12 @@ tl::expected<void, UpdateError> UpdateManager::delete_update(const std::string &
     return tl::make_unexpected(UpdateError{UpdateErrorCode::NoBackend, "No update backend loaded"});
   }
 
+  bool had_state = false;
   {
     std::lock_guard<std::mutex> lock(mutex_);
     auto it = states_.find(id);
     if (it != states_.end()) {
+      had_state = true;
       if (is_task_active(id)) {
         return tl::make_unexpected(
             UpdateError{UpdateErrorCode::InProgress, "Cannot delete update while operation is in progress"});
@@ -103,6 +113,10 @@ tl::expected<void, UpdateError> UpdateManager::delete_update(const std::string &
       }
       // Mark as deleting so no new operations can start on this package
       it->second->phase = UpdatePhase::Deleting;
+    } else {
+      // Create sentinel to prevent concurrent start_prepare
+      states_[id] = std::make_unique<PackageState>();
+      states_[id]->phase = UpdatePhase::Deleting;
     }
   }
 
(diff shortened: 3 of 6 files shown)


docs/api/rest.rst
```diff
@@ -672,6 +672,145 @@ fault manager and cannot be deleted via this endpoint.
 - **404 Not Found**: Entity, category, or bulk-data ID not found
 - **405 Method Not Allowed**: Delete attempted on areas or functions
 
+Software Updates
+----------------
+
+Manage software update packages with an async prepare/execute lifecycle.
+The updates feature requires a backend plugin to be loaded (see :doc:`/config/server`).
```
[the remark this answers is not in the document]
bburda: Won't do. We need to develop e2e solution first to provide good tutorial.

src/ros2_medkit_gateway/include/ros2_medkit_gateway/updates/update_backend.hpp
```diff
@@ -0,0 +1,142 @@
...
+ *
+ * For runtime loading, the .so must export:
+ *   extern "C" UpdateBackend* create_update_backend();
+ *
+ * @par Thread Safety
+ * - CRUD methods (list_updates, get_update, register_update, delete_update)
+ *   are called while UpdateManager holds its mutex. They will not be called
+ *   concurrently with each other, but may overlap with prepare/execute running
+ *   in a background thread. If the backend shares state between CRUD and
+ *   async methods, it must provide its own synchronization.
```
mfaferek93: The docstring here says CRUD methods are called while UpdateManager holds its mutex, but looking at `update_manager.cpp` - `list_updates` (line 51), `get_update` (line 62), and `register_update` (line 73) all call the backend without any mutex. A third-party backend author trusting this doc could end up with data races. Worth either adding the lock or fixing the comment.

src/ros2_medkit_gateway/src/updates/update_manager.cpp
```diff
@@ -0,0 +1,387 @@
...
+      // Wait for completed task before proceeding (clean up future)
+      if (it->second->active_task.valid()) {
+        it->second->active_task.wait();
+      }
+      // Mark as deleting so no new operations can start on this package
+      it->second->phase = UpdatePhase::Deleting;
+    }
+  }
+
+  auto result = backend_->delete_update(id);
```
mfaferek93, line 123: There's a TOCTOU race in `delete_update()` - if no `PackageState` exists for the id, the Deleting sentinel is skipped (line 95 check falls through) and we go straight to `backend_->delete_update()` here without the lock. A concurrent `start_prepare()` could sneak in between, create a PackageState and launch an async task for a package that's about to be deleted.

src/ros2_medkit_gateway/src/http/handlers/update_handlers.cpp
```diff
@@ -0,0 +1,338 @@
...
+  try {
+    json body;
+    try {
+      body = json::parse(req.body);
+    } catch (const json::parse_error &) {
+      HandlerContext::send_error(res, 400, ERR_INVALID_REQUEST, "Invalid JSON body");
+      return;
+    }
+
+    auto result = update_mgr_->register_update(body);
```
mfaferek93, line 148: The backend call `register_update(body)` here happens before validating that `body["id"]` exists (line 151). If a permissive backend accepts metadata without an id, the data gets persisted but the handler returns 400 - leaving things in an inconsistent state. Moving the id validation before the backend call should fix it.

src/ros2_medkit_gateway/src/updates/update_manager.cpp
```diff
@@ -0,0 +1,387 @@
...
+  return *result;
+}
+
+tl::expected<void, UpdateError> UpdateManager::register_update(const nlohmann::json & metadata) {
+  if (!backend_) {
+    return tl::make_unexpected(UpdateError{UpdateErrorCode::NoBackend, "No update backend loaded"});
+  }
+  auto result = backend_->register_update(metadata);
+  if (!result) {
+    if (result.error().find("already exists") != std::string::npos) {
```
mfaferek93: The error classification here relies on string-matching backend error messages (`find("already exists")` here, `find("not found")` on line 120). This works with the current backends but it's fragile - a different backend phrasing breaks the classification. Maybe worth adding error codes to the `UpdateBackend` interface, similar to `UpdateErrorCode` already used by `UpdateManager`.

src/ros2_medkit_gateway/src/updates/update_manager.cpp
```diff
@@ -0,0 +1,387 @@
...
+  auto pkg = backend_->get_update(id);
+  if (!pkg) {
+    return tl::make_unexpected(UpdateError{UpdateErrorCode::NotFound, pkg.error()});
+  }
+
+  auto it = states_.find(id);
+  if (it == states_.end() || !it->second || it->second->phase != UpdatePhase::Prepared) {
+    return tl::make_unexpected(UpdateError{UpdateErrorCode::NotPrepared, "Package must be prepared before execution"});
+  }
+  if (it->second->phase == UpdatePhase::Deleting) {
```
mfaferek93: Small thing - this Deleting phase check is dead code. Line 174 already returns `NotPrepared` for any phase other than `Prepared`, which includes `Deleting`. Safe to remove.

src/ros2_medkit_gateway/src/http/handlers/update_handlers.cpp
```diff
@@ -0,0 +1,338 @@
...
+      break;
+    case UpdateStatus::Failed:
+      j["status"] = "failed";
+      break;
+  }
+  if (status.progress.has_value()) {
+    j["progress"] = *status.progress;
+  }
+  if (status.sub_progress.has_value()) {
+    j["sub_progress"] = json::array();
```
mfaferek93, line 57: `sub_progress` is snake_case here, but the status values use camelCase (`inProgress` on line 44). If we're targeting SOVD compliance, this should probably be `subProgress`. Same question for `error_code` / `error_message` on lines 64-65.

src/ros2_medkit_gateway/src/http/handlers/update_handlers.cpp
```diff
@@ -0,0 +1,338 @@
...
+    // Validate id field exists before using it for Location header
+    if (!body.contains("id") || !body["id"].is_string() || body["id"].get<std::string>().empty()) {
+      HandlerContext::send_error(res, 400, ERR_INVALID_REQUEST, "Missing required field: id");
+      return;
+    }
+    auto id = body["id"].get<std::string>();
+    json response = {{"id", id}};
+    HandlerContext::send_json(res, response);
+    res.status = 201;
+    res.set_header("Location", api_path("/updates/" + id));
```
mfaferek93, line 164: The `id` from the POST body JSON goes straight into the Location header without character validation. `validate_entity_id()` is used for URL-extracted ids (e.g., line 106) but not here. Might be worth running the same validation to prevent any CRLF header injection shenanigans.

src/ros2_medkit_gateway/src/http/rest_server.cpp
```diff
@@ -983,6 +987,40 @@ void RESTServer::setup_routes() {
...
+    });
+    srv->Post(api_path("/updates"), [this](const httplib::Request & req, httplib::Response & res) {
+      update_handlers_->handle_register_update(req, res);
+    });
+    // Specific routes before generic /{id}
+    srv->Get((api_path("/updates") + R"(/([^/]+)/status$)"),
+             [this](const httplib::Request & req, httplib::Response & res) {
+               update_handlers_->handle_get_status(req, res);
+             });
+    srv->Put((api_path("/updates") + R"(/([^/]+)/prepare$)"),
```
mfaferek93, line 1003: `prepare`, `execute`, `automated` are registered as PUT, but these are non-idempotent action triggers (they start background tasks). The rest of the codebase uses POST for that (e.g., `handle_create_execution` on line 311). Unless SOVD spec mandates PUT here, POST would be more consistent.

src/ros2_medkit_gateway/test/test_update_manager.cpp
```diff
@@ -0,0 +1,405 @@
...
+      break;
+    }
+    std::this_thread::sleep_for(std::chrono::milliseconds(10));
+  }
+  EXPECT_EQ(status.status, UpdateStatus::Failed);
+  ASSERT_TRUE(status.error_message.has_value());
+  EXPECT_NE(status.error_message->find("download failed"), std::string::npos);
+}
+
+// @verifies REQ_INTEROP_091
```
mfaferek93, line 464: There are `PrepareFailureSetsFailedStatus` (line 357) and `PrepareExceptionSetsFailedStatus` (here) tests for prepare, but no matching ones for execute. The mocks (`MockFailingBackend::execute`, `MockThrowingBackend::execute`) are already wired up - just needs the test cases.

src/ros2_medkit_gateway/test/test_update_manager.cpp
```diff
@@ -0,0 +1,405 @@
...
+}
+
+// @verifies REQ_INTEROP_092
+TEST_F(UpdateManagerTest, ExecuteAfterPrepare) {
+  json pkg = {{"id", "test-pkg"}, {"update_name", "Test"}, {"automated", false}};
+  (void)manager_->register_update(pkg);
+
+  (void)manager_->start_prepare("test-pkg");
+  // Wait for prepare to complete
+  for (int i = 0; i < 100; ++i) {
```
mfaferek93: Some polling loops are missing explicit timeout assertions. `PrepareAndPollStatus` (line 181) does it right with `ASSERT_TRUE(found) << "Timed out"`, but this loop and `AutomatedCompletes` (line 235) just loop and then check status - if they time out you'd get a confusing `status != Completed` instead of a clear timeout message.

src/ros2_medkit_gateway/src/updates/update_manager.cpp
```diff
@@ -0,0 +1,387 @@
...
+
+#include <dlfcn.h>
+
+namespace ros2_medkit_gateway {
+
+UpdateManager::UpdateManager(std::unique_ptr<UpdateBackend> backend, void * plugin_handle)
+  : backend_(std::move(backend)), plugin_handle_(plugin_handle) {
+}
+
+UpdateManager::~UpdateManager() {
```
mfaferek93, line 25: Nit: the destructor collects futures under lock, then waits and calls `backend_.reset()` outside the lock. In theory a new task could start between lock release and `backend_.reset()`. In practice the HTTP server shuts down first so it's fine, but a `stopped_` atomic flag would make it bulletproof.

src/ros2_medkit_gateway/src/http/handlers/update_handlers.cpp
bburda: `sub_progress` is the SOVD-specified field name (ISO 17978-3, §7.18 status response schema). The camelCase values like `inProgress` are enum values, not field names - SOVD uses snake_case for JSON field names throughout. No change needed.

src/ros2_medkit_gateway/src/http/rest_server.cpp
bburda: SOVD spec mandates PUT for prepare, execute, and automated (ISO 17978-3, §7.18). These are defined as idempotent state transitions in the spec - re-triggering a prepare on an already-prepared package is a no-op/409, not a new task creation. POST is reserved for register (creating a new resource).

src/ros2_medkit_gateway/include/ros2_medkit_gateway/updates/update_backend.hpp
bburda: Fixed. Docstring now states CRUD methods are called WITHOUT holding the mutex, with a note about the delete_update sentinel guard.

src/ros2_medkit_gateway/src/updates/update_manager.cpp
bburda: Fixed. delete_update now creates a Deleting sentinel in the else branch when no PackageState exists, preventing concurrent start_prepare. On backend failure, the sentinel is cleaned up (erased) rather than set to Failed.

src/ros2_medkit_gateway/src/http/handlers/update_handlers.cpp
bburda: Fixed. ID existence check + validate_entity_id() now run before the backend call.

src/ros2_medkit_gateway/src/updates/update_manager.cpp
bburda: Fixed. Added UpdateBackendError enum and UpdateBackendErrorInfo struct to the backend interface. All backends return typed errors, UpdateManager switches on error codes instead of string matching.

src/ros2_medkit_gateway/src/updates/update_manager.cpp
bburda: Removed.

src/ros2_medkit_gateway/src/http/handlers/update_handlers.cpp
bburda: Fixed. validate_entity_id() now runs on the POST body id before any backend call or header use.

src/ros2_medkit_gateway/test/test_update_manager.cpp
bburda: Added ExecuteFailureSetsFailedStatus and ExecuteExceptionSetsFailedStatus tests with dedicated MockExecuteFailingBackend and MockExecuteThrowingBackend (prepare succeeds, execute fails/throws).

src/ros2_medkit_gateway/test/test_update_manager.cpp
bburda: Fixed. All polling loops now use deadline-based waits with ASSERT_TRUE(found) << "Timed out..." messages.

src/ros2_medkit_gateway/src/updates/update_manager.cpp
bburda: Added. stopped_ atomic<bool> is set in the destructor before collecting futures. start_prepare/start_execute/start_automated all check it early and return Internal error if shutting down.

PR #245, opened 2026-03-03
feat(logging): add Log Entries REST API endpoints

### LINKED ISSUE #208: Implement Logging (application log query) endpoints
## Summary

Logging endpoints expose structured log entries from entities and allow configuring logging verbosity per entity. In ROS 2, all node logs are published to the `/rosout` topic, making them centrally available to the gateway. The gateway subscribes to `/rosout`, stores entries in a ring buffer, and serves them via these endpoints.

---

### 1. `GET /api/v1/{entity-path}/logs`

Query structured log entries from an entity.

**Applies to entity types:** Components, Apps

**Path parameters:**

| `{entity-path}` | URL segment | Yes | e.g., `apps/temp_sensor` |

**Query parameters:**

| `severity` | `string` | No | Minimum severity filter. Only entries at this level or higher are returned. |
| `context` | `string` | No | Filter by context identifier (e.g., logger name sub-component) |

**Severity levels (ordered lowest to highest):**

| Level | Description | ROS 2 Equivalent |
|-------|-------------|------------------|
| `debug` | Detailed debugging info | `DEBUG` (10) |
| `info` | General information | `INFO` (20) |
| `warning` | Warning conditions | `WARN` (30) |
| `error` | Error conditions | `ERROR` (40) |
| `fatal` | Critical failures | `FATAL` (50) |

When `severity=warning` is specified, only `warning`, `error`, and `fatal` entries are returned.

**Response `200 OK`:**

```json
{
  "items": [
    {
      "id": "log_001",
      "timestamp": "2026-02-14T10:30:00Z",
      "severity": "warning",
      "message": "Sensor calibration drift detected: offset=0.23",
      "context": {
        "node": "temp_sensor",
        "logger": "temp_sensor.calibration"
      }
    },
    {
      "id": "log_002",
      "timestamp": "2026-02-14T10:30:05Z",
      "severity": "error",
      "message": "Failed to read sensor: timeout after 5000ms",
      "context": {
        "node": "temp_sensor",
        "logger": "temp_sensor.driver"
      }
    }
  ]
}
```

Each item is a `LogEntry`:

| `id` | `string` | Yes | Unique log entry identifier (server-generated, monotonically increasing) |
| `timestamp` | `string` (ISO 8601) | Yes | When the log event occurred |
| `severity` | `string` | Yes | One of: `debug`, `info`, `warning`, `error`, `fatal` |
| `message` | `string` | Yes | Human-readable log message |
| `context` | `object` | No | Structured context (key-value pairs). Typically includes `node` name and `logger` name. |

**Error responses:**

|--------|-----------|------|
| `400` | `invalid-parameter` | Unknown severity value |
| `404` | `entity-not-found` | Entity doesn't exist |

---

### 2. `GET /api/v1/{entity-path}/logs/configuration`

Read the current logging configuration for an entity.

**Response `200 OK`:**

```json
{
  "severity_filter": "info",
  "max_entries": 10000
}
```

| `severity_filter` | `string` | Yes | Current minimum severity level for log collection |
| `max_entries` | `integer` | Yes | Maximum number of log entries retained in the ring buffer |

---

### 3. `PUT /api/v1/{entity-path}/logs/configuration`

Update the logging configuration for an entity.

**Request body:**

```json
{
  "severity_filter": "warning",
  "max_entries": 5000
}
```

| `severity_filter` | `string` | No | New minimum severity. One of: `debug`, `info`, `warning`, `error`, `fatal` |
| `max_entries` | `integer` | No | New max entries. Must be > 0. |

Both fields are optional - only provided fields are updated.

**Response `204 No Content`**

**Error responses:**

|--------|-----------|------|
| `400` | `invalid-parameter` | Unknown severity value, max_entries ≤ 0 |
| `404` | `entity-not-found` | Entity doesn't exist |

---

### ROS 2 log collection

1. **Subscribe to `/rosout`** - the standard ROS 2 topic for centralized logging (message type: `rcl_interfaces/msg/Log`)
2. **Filter by node name** - each `/rosout` message includes `name` (node name). Match this against the entity's bound ROS 2 node (`App.bound_fqn` or `Component.fqn`).
3. **Store in a ring buffer** per entity - configurable `max_entries`, oldest entries evicted when full.

**`rcl_interfaces/msg/Log` fields → `LogEntry` mapping:**

| ROS 2 Log Field | LogEntry Field |
|-----------------|----------------|
| `stamp` | `timestamp` |
| `level` (uint8) | `severity` (mapped: 10→debug, 20→info, 30→warning, 40→error, 50→fatal) |
| `msg` | `message` |
| `name` | `context.node` |
| `function` | `context.function` (optional) |
| `file` | `context.file` (optional) |
| `line` | `context.line` (optional) |

### Logging configuration mapping

Setting `severity_filter` for an entity can map to:
1. **Gateway-side filter** - only store entries at or above the configured level (simplest)
2. **ROS 2 logger level** - call `rcl_interfaces/srv/SetLoggerLevel` service to change the actual node's log level (changes what the node emits, not just what's stored)

Recommended: implement both - gateway-side filter for what's stored, and optionally set the ROS 2 logger level for efficiency.

### Architecture

- Create a `LogManager` class that subscribes to `/rosout`
- Per-entity ring buffers (keyed by entity ID → node name mapping)
- Thread-safe access (multiple HTTP requests may query simultaneously)
- `LogEntry.id` can be a simple monotonically increasing integer counter

### Route registration

```cpp
srv->Get((api_path("/apps") + R"(/([^/]+)/logs$)"), handler);
srv->Get((api_path("/apps") + R"(/([^/]+)/logs/configuration$)"), handler);
srv->Put((api_path("/apps") + R"(/([^/]+)/logs/configuration$)"), handler);
// Same for /components/
```

**Important:** Register `/logs/configuration` routes **before** the generic `/logs$` route.

### Tests

- Unit test: query logs returns stored entries
- Unit test: severity filter only returns matching entries
- Unit test: read configuration returns current settings
- Unit test: update configuration changes severity filter
- Unit test: ring buffer evicts old entries when full
- Unit test: invalid severity → 400
- Integration test: launch demo nodes, collect /rosout logs, query via endpoint

bburda:
# Pull Request

## Summary

Adds Log Entries REST API endpoints for Apps and Components, backed by a per-node `/rosout` ring buffer with configurable severity filter and buffer capacity.

- `GET /apps/{id}/logs`, `GET /components/{id}/logs` - query buffered log entries (severity and context filters supported)
- `GET/PUT /apps/{id}/logs/configuration`, `GET/PUT /components/{id}/logs/configuration` - read and update per-entity log config
- `LogProvider` plugin interface for third-party log backends
- `Capability::LOGS` advertised in App and Component discovery responses

---

## Issue

- closes #208

---

## Type

- [ ] Bug fix
- [x] New feature or tests
- [ ] Breaking change
- [ ] Documentation only

---

## Testing

- `test_log_manager` - 19 unit tests: severity mapping, FQN normalisation, ring-buffer eviction, config updates, prefix vs exact matching, plugin delegation
- `test_log_handlers` - 9 unit tests: missing-matches guard paths for all three handler methods
- `test_logging_api` integration test - 15 tests covering GET /logs (schema, severity filter, 404), GET/PUT /logs/configuration (defaults, updates, 400 validation), and capability advertisement

All `ros2_medkit_gateway` unit tests pass. One pre-existing flaky fault_manager test (`FaultEventPublishingTest.NewFaultPublishesConfirmedEvent`) is unrelated to these changes.

---

## Checklist

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -115,7 +115,7 @@ jobs:
     runs-on: ubuntu-latest
     container:
       image: ubuntu:noble
-    timeout-minutes: 30
+    timeout-minutes: 40
     defaults:
       run:
         shell: bash


src/ros2_medkit_gateway/src/plugins/plugin_manager.cpp
```diff
@@ -228,6 +266,26 @@ std::vector<IntrospectionProvider *> PluginManager::get_introspection_providers(
...
+  // disable_plugin() only nulls pointers/resets unique_ptrs and does not resize plugins_, so the
+  // race is benign on all current platforms. A proper fix would require a shared_mutex or
+  // copying the observer list at subscription time.
+  std::vector<LogProvider *> result;
+  for (const auto & lp : plugins_) {
+    if (lp.log_provider) {
+      result.push_back(lp.log_provider);
+    }
+  }
+  return result;
```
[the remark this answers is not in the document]
bburda: The race described here is identical to the pre-existing race in `get_introspection_providers()` and `get_update_provider()` — both read `plugins_` from the ROS 2 executor thread while `disable_plugin()` can run from the HTTP handler thread. The pattern is already documented at lines 274–280. Fixing the synchronization consistently across all providers is a broader refactor that goes beyond this PR's scope. Left as-is to match the existing pattern.

src/ros2_medkit_gateway/src/plugins/plugin_manager.cpp
```diff
@@ -83,6 +93,15 @@ size_t PluginManager::load_plugins(const std::vector<PluginConfig> & configs) {
...
       lp.introspection_provider = result->introspection_provider;
+      // LogProvider: discovered via dynamic_cast across the dlopen boundary.
+      // Limitation: with -fvisibility=hidden the RTTI symbol for LogProvider may exist in
+      // both the gateway binary and the .so, causing dynamic_cast to silently return nullptr
+      // even when the plugin does implement LogProvider. The safe fix is an extern "C" query
+      // function (like get_update_provider_ptr()), but that requires an ABI change deferred
+      // to a future release. For now, LogProvider plugins must be built with matching visibility
+      // settings (e.g. -fvisibility=default or explicit attribute((visibility("default"))) on
+      // the LogProvider vtable).
+      lp.log_provider = dynamic_cast<LogProvider *>(result->plugin.get());
```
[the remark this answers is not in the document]
bburda: Fixed. Added `extern "C" LogProvider* get_log_provider(GatewayPlugin*)` discovery function in plugin_loader.cpp, mirroring the existing UpdateProvider/IntrospectionProvider mechanism. Result stored in `GatewayPluginLoadResult::log_provider`, used in plugin_manager.cpp instead of `dynamic_cast`.

src/ros2_medkit_gateway/src/http/handlers/log_handlers.cpp
```diff
@@ -0,0 +1,153 @@
...
+
+  // Components use prefix matching (all nodes under the component namespace);
+  // Apps use exact matching (single node FQN).
+  const bool prefix_match = (entity.type == EntityType::COMPONENT);
+
+  // Optional query parameters
+  const std::string min_severity = req.get_param_value("severity");
+  const std::string context_filter = req.get_param_value("context");
+
+  auto logs = log_mgr->get_logs({entity.fqn}, prefix_match, min_severity, context_filter, entity_id);
```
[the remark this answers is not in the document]
bburda: Fixed. Added validation in `handle_get_logs`: if `?severity=` is present and not one of {debug, info, warning, error, fatal}, returns 400 ERR_INVALID_PARAMETER before calling `get_logs()`.

src/ros2_medkit_gateway/src/http/handlers/log_handlers.cpp
```diff
@@ -0,0 +1,153 @@
...
+  }
+
+  std::optional<std::string> severity_filter;
+  std::optional<size_t> max_entries;
+
+  if (body.contains("severity_filter") && body["severity_filter"].is_string()) {
+    severity_filter = body["severity_filter"].get<std::string>();
+  }
+  if (body.contains("max_entries") && body["max_entries"].is_number_unsigned()) {
+    max_entries = body["max_entries"].get<size_t>();
```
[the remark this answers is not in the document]
bburda: Fixed. Added explicit type checks in `handle_put_logs_configuration`: `severity_filter` must be a string (400 otherwise), and `max_entries` must be an integer > 0 (400 otherwise).

src/ros2_medkit_gateway/src/log_manager.cpp
```diff
@@ -0,0 +1,307 @@
...
+  entry.line = msg->line;
+
+  // Notify all LogProvider observers — they may forward to OTel, DB, etc.
+  bool suppress_buffer = false;
+  if (plugin_mgr_) {
+    for (auto * observer : plugin_mgr_->get_log_observers()) {
+      if (observer->on_log_entry(entry)) {
+        suppress_buffer = true;
+      }
+    }
```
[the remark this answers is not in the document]
bburda: Fixed. Each `observer->on_log_entry(entry)` call in `on_rosout()` is now wrapped in try/catch(std::exception&) and catch(...) with RCLCPP_WARN logging, so a misbehaving plugin cannot crash the subscription callback.

src/ros2_medkit_gateway/src/http/handlers/log_handlers.cpp
```diff
@@ -0,0 +1,174 @@
...
+  const auto entity_id = req.matches[1].str();
+  auto entity_opt = ctx_.validate_entity_for_route(req, res, entity_id);
+  if (!entity_opt) {
+    return;
+  }
+  const auto & entity = *entity_opt;
+
+  auto * log_mgr = ctx_.node()->get_log_manager();
+  if (!log_mgr) {
+    HandlerContext::send_error(res, 503, ERR_INTERNAL_ERROR, "LogManager not available");
```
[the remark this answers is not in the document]
bburda: Fixed. All three 503 paths in log_handlers.cpp now use ERR_SERVICE_UNAVAILABLE.

src/ros2_medkit_gateway/include/ros2_medkit_gateway/http/handlers/log_handlers.hpp
```diff
@@ -0,0 +1,77 @@
...
+  /**
+   * @brief Handle GET /{entity-path}/logs/configuration - get log configuration.
+   */
+  void handle_get_logs_configuration(const httplib::Request & req, httplib::Response & res);
+
+  /**
+   * @brief Handle PUT /{entity-path}/logs/configuration - update log configuration.
+   *
+   * Body (JSON, all fields optional):
+   *   severity_filter  - Minimum severity to store/return (debug/info/warning/error/fatal)
```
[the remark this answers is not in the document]
bburda: Fixed. Changed to "return in query results" - the default ring buffer stores all entries regardless of severity_filter; the filter only applies at query time.

docs/api/rest.rst
```diff
@@ -487,6 +487,120 @@ Query and manage faults.
...
+
+   .. code-block:: json
+
+      {
+        "severity_filter": "warning",
+        "max_entries": 500
+      }
+
+   ``severity_filter`` — minimum severity to retain/return (``debug`` | ``info`` | ``warning`` |
+   ``error`` | ``fatal``). Entries below this level are excluded from queries. Default: ``debug``.
```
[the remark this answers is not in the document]
bburda: Fixed. Updated to "return in query results" in both the docs and the log_handlers.hpp comment.

src/ros2_medkit_gateway/src/log_manager.cpp
```diff
@@ -0,0 +1,360 @@
...
+      }
+      return items;
+    } catch (const std::exception & e) {
+      RCLCPP_ERROR(node_->get_logger(), "LogProvider::get_logs threw: %s", e.what());
+      return json::array();
+    } catch (...) {
+      RCLCPP_ERROR(node_->get_logger(), "LogProvider::get_logs threw unknown exception");
+      return json::array();
+    }
+  }
```
mfaferek93, line 232: When `get_logs()` on the plugin throws, you return an empty `json::array()` and log the error - but the HTTP layer will happily serve this as 200 `{"items":[]}`. The client has literally no way to know the plugin crashed vs "there are just no logs".

For the full-ingestion mode where the plugin IS the only source of truth, this is silent data loss from the client's perspective. Worth returning 503 here, or at least propagating via `tl::expected` so the handler can decide.

src/ros2_medkit_gateway/src/log_manager.cpp
```diff
@@ -0,0 +1,360 @@
...
+  if (auto * provider = effective_provider()) {
+    try {
+      return provider->get_config(entity_id);
+    } catch (const std::exception & e) {
+      RCLCPP_ERROR(node_->get_logger(), "LogProvider::get_config threw: %s", e.what());
+    } catch (...) {
+      RCLCPP_ERROR(node_->get_logger(), "LogProvider::get_config threw unknown exception");
+    }
+  }
+
```
mfaferek93, line 326: If plugin `get_config()` throws, you fall through to the local `configs_` map. But when `manages_ingestion() == true`, that map is always empty, so the client gets default config `{"severity_filter":"debug","max_entries":100}` with no indication anything went wrong.

This is arguably worse than the `get_logs()` case because the client might use the returned config to make decisions ("oh severity is debug, my filter must be working").

src/ros2_medkit_gateway/src/plugins/plugin_manager.cpp
```diff
@@ -228,6 +261,26 @@ std::vector<IntrospectionProvider *> PluginManager::get_introspection_providers(
   return result;
 }
 
+LogProvider * PluginManager::get_log_provider() const {
+  return first_log_provider_;
+}
+
+std::vector<LogProvider *> PluginManager::get_log_observers() const {
```
mfaferek93, line 275: The comment is honest about the race, which I appreciate, but worth noting that the new observer path is called on **every** `/rosout` message - much higher frequency than the existing introspection provider reads. If `disable_plugin()` resets the `unique_ptr` (line ~148), the object is destroyed, and the observer loop at `log_manager.cpp:167` may be holding a raw pointer to freed memory.

Not asking you to fix this in this PR since it's pre-existing, but maybe worth a follow-up issue?

src/ros2_medkit_gateway/src/http/handlers/discovery_handlers.cpp
```diff
@@ -489,13 +489,14 @@ void DiscoveryHandlers::handle_get_component(const httplib::Request & req, httpl
...
-    std::vector<Cap> caps = {Cap::DATA,   Cap::OPERATIONS,    Cap::CONFIGURATIONS,
-                             Cap::FAULTS, Cap::SUBCOMPONENTS, Cap::HOSTS};
+    std::vector<Cap> caps = {Cap::DATA, Cap::OPERATIONS,    Cap::CONFIGURATIONS, Cap::FAULTS,
+                             Cap::LOGS, Cap::SUBCOMPONENTS, Cap::HOSTS};
     if (!comp.depends_on.empty()) {
       caps.push_back(Cap::DEPENDS_ON);
     }
     auto comp_caps = CapabilityBuilder::build_capabilities("components", comp.id, caps);
     append_plugin_capabilities(comp_caps, "components", comp.id, SovdEntityType::COMPONENT, ctx_.node());
+    response["capabilities"] = comp_caps;
```
mfaferek93, line 501: You're setting capabilities both at `response["capabilities"]` (this line) AND via `ext.add("capabilities", comp_caps)` which ends up in `response["x-medkit"]["capabilities"]`. Is this duplication intentional for backward compat? Apps only have it at root. A quick comment here would save the next person some head-scratching.

src/ros2_medkit_gateway/src/gateway_node.cpp
```diff
@@ -420,6 +424,18 @@ GatewayNode::GatewayNode() : Node("ros2_medkit_gateway") {
...
   }
 
+  // Initialize log manager (subscribes to /rosout, delegates to plugin if available)
+  static constexpr int kMinBufferSize = 1;
+  static constexpr int kMaxBufferSize = 100000;
+  auto raw_buffer_size = get_parameter("logs.buffer_size").as_int();
+  auto clamped =
+      std::clamp(raw_buffer_size, static_cast<int64_t>(kMinBufferSize), static_cast<int64_t>(kMaxBufferSize));
+  if (clamped != raw_buffer_size) {
+    RCLCPP_WARN(get_logger(), "logs.buffer_size %ld clamped to %ld", raw_buffer_size, clamped);
```
mfaferek93: Nit: `raw_buffer_size` is `int64_t` but `%ld` is the format specifier for `long`, which is only guaranteed to be 32 bits on some platforms. Technically UB (and will break on Windows if you ever go there). `PRId64` from `<cinttypes>` would be correct, or just cast to `long`.

src/ros2_medkit_gateway/src/http/handlers/log_handlers.cpp
```diff
@@ -0,0 +1,181 @@
...
+// ---------------------------------------------------------------------------
+// handle_put_logs_configuration
+// ---------------------------------------------------------------------------
+
+void LogHandlers::handle_put_logs_configuration(const httplib::Request & req, httplib::Response & res) {
+  if (req.matches.size() < 2) {
+    HandlerContext::send_error(res, 400, ERR_INVALID_REQUEST, "Invalid request");
+    return;
+  }
+
```
mfaferek93, line 128: Minor thought: unknown fields in the JSON body are silently ignored. If someone sends `{"severityFilter": "error"}` (camelCase) or `{"max_entry": 50}` (typo), the PUT returns 204 and nothing changes. The client thinks it worked.

Not a blocker, but a `RCLCPP_DEBUG` or even a 400 for unrecognized keys would save debugging time.

src/ros2_medkit_gateway/src/log_manager.cpp
bburda: Fixed. `get_logs()` now returns `tl::expected<json, std::string>`. When the plugin throws, the handler returns 503 `ERR_SERVICE_UNAVAILABLE` with the error message instead of silently serving 200 with empty items. Added `MockThrowingLogPlugin` test to verify.

src/ros2_medkit_gateway/src/log_manager.cpp
bburda: Fixed. `get_config()` now returns `tl::expected<LogConfig, std::string>`. When the plugin throws, the handler returns 503 `ERR_SERVICE_UNAVAILABLE` instead of silently returning default config. Added test coverage for this path.

src/ros2_medkit_gateway/src/plugins/plugin_manager.cpp
bburda: Fixed in this PR. Added `std::shared_mutex` to PluginManager - reader methods (`get_log_observers`, `get_introspection_providers`, `get_update_provider`, etc.) take `shared_lock`, lifecycle methods that call `disable_plugin()` take `unique_lock`. This fixes the race for all provider access paths, not just log observers. Added 3 concurrency tests (concurrent reads, reads+lifecycle, shutdown while reading).

src/ros2_medkit_gateway/src/http/handlers/discovery_handlers.cpp
bburda: Added clarifying comment. The duplication is intentional: root-level `capabilities` is the SOVD standard location, `x-medkit.capabilities` is for vendor extension tooling that only reads the x-medkit block. Apps don't duplicate because their x-medkit block doesn't include capabilities.

src/ros2_medkit_gateway/src/gateway_node.cpp
bburda: Fixed. Using `PRId64` from `<cinttypes>` for portable `int64_t` formatting.

src/ros2_medkit_gateway/src/http/handlers/log_handlers.cpp
bburda: Fixed. Added `RCLCPP_DEBUG` for unrecognized keys in the PUT body. This will help catch typos like `severityFilter` vs `severity_filter` during debugging.

src/ros2_medkit_gateway/src/http/handlers/log_handlers.cpp
```diff
@@ -0,0 +1,181 @@
...
+      return;
+    }
+    const auto val = me.get<long long>();
+    if (val <= 0) {
+      HandlerContext::send_error(res, 400, ERR_INVALID_PARAMETER, "max_entries must be greater than 0");
+      return;
+    }
+    if (val > kMaxEntriesCap) {
+      HandlerContext::send_error(res, 400, ERR_INVALID_PARAMETER, "max_entries exceeds maximum allowed value of 10000");
+      return;
```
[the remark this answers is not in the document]
bburda: Fixed. Wrapped `me.get<long long>()` in `try/catch(nlohmann::json::exception)` - returns 400 `ERR_INVALID_PARAMETER` with "max_entries value out of range" for values that overflow signed 64-bit.

src/ros2_medkit_gateway/src/gateway_node.cpp
```diff
@@ -89,6 +89,10 @@ GatewayNode::GatewayNode() : Node("ros2_medkit_gateway") {
   declare_parameter("sse.max_clients", 10);         // Limit concurrent SSE connections to prevent resource exhaustion
   declare_parameter("sse.max_subscriptions", 100);  // Maximum active cyclic subscriptions across all entities
 
+  // Log management parameters
+  declare_parameter("logs.buffer_size",
+                    200);  // Ring buffer capacity per node; entries exceed this are dropped (oldest first)
```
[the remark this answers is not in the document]
bburda: Fixed.

PR #286, opened 2026-03-17
feat: SOVD diagnostic scripts with plugin-extensible ScriptProvider

### LINKED ISSUE #200: Implement Scripts - upload, list, get, delete
## Summary

Scripts are diagnostic programs that can be uploaded to the server and later executed on entities. This issue covers the script **management** endpoints (upload, list, read metadata, delete). Script **execution** is covered in a separate issue.

---

### 1. `POST /api/v1/{entity-path}/scripts`

Upload a new diagnostic script to an entity.

**Applies to entity types:** Components, Apps

**Path parameters:**

| `{entity-path}` | URL segment | Yes | e.g., `components/engine_ecu` or `apps/temp_sensor` |

**Request:** `Content-Type: multipart/form-data`

The multipart body should contain:
- **`file`** - the script binary/text content
- **`metadata`** (optional) - JSON with script metadata:

```json
{
  "name": "Sensor Calibration Check",
  "description": "Validates sensor calibration values against known thresholds",
  "parameters_schema": {
    "type": "object",
    "properties": {
      "threshold": { "type": "number", "default": 0.1 }
    }
  }
}
```

**Response `201 Created`:**

```json
{
  "id": "script_001",
  "name": "Sensor Calibration Check",
  "href": "/api/v1/apps/temp_sensor/scripts/script_001"
}
```

**Error responses:**

|--------|-----------|------|
| `400` | `invalid-request` | Missing file part, invalid metadata |
| `404` | `entity-not-found` | Entity doesn't exist |
| `415` | `invalid-request` | Unsupported content type (not multipart/form-data) |

---

### 2. `GET /api/v1/{entity-path}/scripts`

List all uploaded scripts for an entity.

**Response `200 OK`:**

```json
{
  "items": [
    {
      "id": "script_001",
      "name": "Sensor Calibration Check",
      "description": "Validates sensor calibration values",
      "href": "/api/v1/apps/temp_sensor/scripts/script_001"
    },
    {
      "id": "script_002",
      "name": "Network Diagnostics",
      "description": "Tests DDS communication paths",
      "href": "/api/v1/apps/temp_sensor/scripts/script_002"
    }
  ]
}
```

Each item is a `ScriptDescription`:

| `id` | `string` | Yes | Script identifier |
| `name` | `string` | Yes | Human-readable name |
| `description` | `string` | No | Script description |
| `href` | `string` | Yes | URI to this script resource |
| `proximity_proof_required` | `boolean` | No | Whether execution requires proof of co-location |
| `parameters_schema` | `object` | No | JSON Schema for script input parameters |

---

### 3. `GET /api/v1/{entity-path}/scripts/{script-id}`

Read metadata for a specific script.

**Path parameters:**

| `{script-id}` | `string` | Yes | Script identifier |

**Response `200 OK`:** Returns a single `ScriptDescription` (same fields as list items, potentially with more detail).

**Error responses:**

|--------|-----------|------|
| `404` | `resource-not-found` | Script doesn't exist on this entity |

---

### 4. `DELETE /api/v1/{entity-path}/scripts/{script-id}`

Delete an uploaded script. Cannot delete a script that is currently being executed.

**Path parameters:**

| `{script-id}` | `string` | Yes | Script identifier |

**Response `204 No Content`** - script deleted.

**Error responses:**

|--------|-----------|------|
| `404` | `resource-not-found` | Script doesn't exist |
| `409` | `invalid-request` | Script is currently executing (has a running execution) |

---

### Architecture

- Create a `ScriptManager` class that stores uploaded scripts
- Storage: temporary directory on the filesystem with metadata JSON sidecar files
- Scripts do not survive server restart (unless using persistent storage)
- The manager needs thread-safe access to the script registry

### Security considerations

- Script upload should require `admin` or `operator` role (integrate with auth middleware)
- Store scripts in a controlled directory, not user-specified paths
- Consider file size limits (configurable, e.g., 10MB max)

### cpp-httplib multipart handling

The library has built-in multipart/form-data parsing - access files via `req.files`:

```cpp
auto file_it = req.files.find("file");
if (file_it != req.files.end()) {
  const auto& file = file_it->second;
  // file.content, file.filename, file.content_type
}
```

### Route registration

```cpp
srv->Post((api_path("/components") + R"(/([^/]+)/scripts$)"), handler);
srv->Get((api_path("/components") + R"(/([^/]+)/scripts$)"), handler);
srv->Get((api_path("/components") + R"(/([^/]+)/scripts/([^/]+)$)"), handler);
srv->Delete((api_path("/components") + R"(/([^/]+)/scripts/([^/]+)$)"), handler);
// Same for /apps/
```

### Tests

- Unit test: upload script → 201
- Unit test: list scripts → returns uploaded scripts
- Unit test: get script metadata → correct data
- Unit test: delete script → 204
- Unit test: delete running script → 409
- Unit test: get nonexistent script → 404

### LINKED ISSUE #201: Implement Script Execution lifecycle endpoints
## Summary

Script execution endpoints allow clients to start, monitor, control, and clean up diagnostic script runs. This is separate from script management (upload/list/delete) - execution requires a script to already be uploaded.

---

### 1. `POST /api/v1/{entity-path}/scripts/{script-id}/executions`

Start a new execution of an uploaded script.

**Applies to entity types:** Components, Apps

**Path parameters:**

| `{entity-path}` | URL segment | Yes | e.g., `apps/temp_sensor` |
| `{script-id}` | `string` | Yes | ID of the uploaded script to execute |

**Request body:**

```json
{
  "execution_type": "now",
  "parameters": {
    "threshold": 0.1,
    "timeout": 30
  },
  "proximity_response": "optional-proof-token"
}
```

| `execution_type` | `string` | Yes | When to run. See values below. |
| `parameters` | `object` | No | Input parameters for the script (must match `parameters_schema` if defined) |
| `proximity_response` | `string` | No | Co-location proof token (if script requires proximity) |

**`execution_type` values:**

| Value | Description |
|-------|-------------|
| `now` | Execute immediately |
| `now_and_on_restart` | Execute now and also schedule for server restart |
| `on_restart` | Schedule for next server restart only |
| `once_on_restart` | Execute once on next restart, then remove schedule |

**Response `202 Accepted`:**

```json
{
  "id": "exec_001",
  "status": "prepared",
  "href": "/api/v1/apps/temp_sensor/scripts/script_001/executions/exec_001"
}
```

The `Location` header also points to the execution status resource.

**Error responses:**

|--------|-----------|------|
| `400` | `invalid-parameter` | Invalid `execution_type`, parameters don't match schema |
| `404` | `resource-not-found` | Script doesn't exist |
| `409` | `invalid-request` | Script already has a running execution (if single-execution only) |

---

### 2. `GET /api/v1/{entity-path}/scripts/{script-id}/executions/{execution-id}`

Poll the status of a script execution.

**Path parameters:**

| `{script-id}` | `string` | Yes | Script identifier |
| `{execution-id}` | `string` | Yes | Execution identifier |

**Response `200 OK`:**

```json
{
  "id": "exec_001",
  "status": "running",
  "progress": 45,
  "started_at": "2026-02-14T10:30:00Z"
}
```

| `id` | `string` | Yes | Execution identifier |
| `status` | `string` | Yes | Current status (see values below) |
| `progress` | `integer` | No | Percentage 0–100 |
| `started_at` | `string` (ISO 8601) | No | When execution started |
| `completed_at` | `string` (ISO 8601) | No | When execution completed/failed/terminated |
| `parameters` | `object` | No | Output parameters (when `completed`) |
| `error` | `GenericError` | No | Error details (when `failed`) |

**`status` values:**

| Status | Description |
|--------|-------------|
| `prepared` | Created but not yet running |
| `running` | Currently executing |
| `completed` | Successfully finished (output in `parameters`) |
| `failed` | Execution failed (details in `error`) |
| `terminated` | Manually stopped by client |

**Error responses:**

|--------|-----------|------|
| `404` | `resource-not-found` | Execution doesn't exist |

---

### 3. `PUT /api/v1/{entity-path}/scripts/{script-id}/executions/{execution-id}`

Send a termination action to a running script execution.

**Request body:**

```json
{
  "action": "stop"
}
```

| Value | Description |
|-------|-------------|
| `stop` | Request graceful termination (SIGTERM equivalent) |
| `forced_termination` | Force immediate termination (SIGKILL equivalent) |

**Response `200 OK`:** Returns updated execution status.

**Error responses:**

|--------|-----------|------|
| `400` | `invalid-parameter` | Unknown action value |
| `404` | `resource-not-found` | Execution doesn't exist |
| `409` | `invalid-request` | Execution is not in a state that can be terminated (already completed/failed/terminated) |

---

### 4. `DELETE /api/v1/{entity-path}/scripts/{script-id}/executions/{execution-id}`

Remove a completed, failed, or terminated execution resource. Cannot remove a running execution.

**Response `204 No Content`**

**Error responses:**

|--------|-----------|------|
| `404` | `resource-not-found` | Execution doesn't exist |
| `409` | `invalid-request` | Execution is still running |

---

### Architecture

- The `ScriptManager` (from issue #06) should have an `ExecutionManager` component
- Script execution spawns a subprocess (e.g., Python interpreter) with configurable timeout
- Track execution status in memory; clean up terminated processes
- Consider resource limits: max concurrent executions per entity, CPU/memory limits

### ROS 2 context

In a ROS 2 environment, scripts could be:
- Python scripts using `rclpy` to interact with the ROS 2 graph
- Shell scripts for system diagnostics
- Launch file fragments for diagnostic sequences

### Route registration

```cpp
srv->Post((api_path("/apps") + R"(/([^/]+)/scripts/([^/]+)/executions$)"), handler);
srv->Get((api_path("/apps") + R"(/([^/]+)/scripts/([^/]+)/executions/([^/]+)$)"), handler);
srv->Put((api_path("/apps") + R"(/([^/]+)/scripts/([^/]+)/executions/([^/]+)$)"), handler);
srv->Delete((api_path("/apps") + R"(/([^/]+)/scripts/([^/]+)/executions/([^/]+)$)"), handler);
// Same for /components/
```

### Tests

- Unit test: start execution → 202
- Unit test: poll status → transitions through prepared → running → completed
- Unit test: terminate running execution → terminated
- Unit test: delete completed execution → 204
- Unit test: delete running execution → 409
- Unit test: start execution for nonexistent script → 404

bburda:
# Pull Request

## Summary

Implement SOVD-compliant diagnostic scripts feature (ISO 17978-3, §7.15) with a plugin-extensible architecture. Clients can upload, manage, and execute diagnostic scripts on Components and Apps entities. The gateway ships with a built-in DefaultScriptProvider (manifest + filesystem + subprocess), but plugins can replace the entire backend via dlopen (same pattern as UpdateProvider).

**Key components:**
- **ScriptProvider** interface - abstract CRUD + execution lifecycle, typed C++ structs
- **ScriptManager** - thin pass-through (like UpdateManager) with set_backend()
- **ScriptHandlers** - 8 HTTP handlers with SOVD error mapping (404/409/400/413/429/500/501)
- **DefaultScriptProvider** - manifest-defined scripts (immutable) + filesystem uploads + subprocess execution (python3/bash/sh) with positional/named/flag args, env vars, JSON stdin fallback, timeout enforcement
- **Plugin integration** - PLUGIN_API_VERSION bumped to 4, get_script_provider() extern C
- **16 routes** via RouteRegistry (8 SOVD endpoints x components/apps)
- **SCRIPTS capability** added to entity capabilities for Components and Apps

**SOVD endpoints implemented:**
- `POST /{entity}/scripts` - Upload script (201 + Location)
- `GET /{entity}/scripts` - List scripts
- `GET /{entity}/scripts/{id}` - Get metadata
- `DELETE /{entity}/scripts/{id}` - Delete script (204, 409 for manifest)
- `POST /{entity}/scripts/{id}/executions` - Start execution (202 + Location)
- `GET /{entity}/scripts/{id}/executions/{eid}` - Poll status
- `PUT /{entity}/scripts/{id}/executions/{eid}` - Terminate (stop/forced_termination)
- `DELETE /{entity}/scripts/{id}/executions/{eid}` - Remove execution (204)

---

## Issue

- closes #200
- closes #201

---

## Type

- [ ] Bug fix
- [x] New feature or tests
- [ ] Breaking change
- [ ] Documentation only

---

## Testing

**Unit tests (34 new):**
- ScriptManager with MockScriptProvider (14 tests) - CRUD + execution lifecycle pass-through
- DefaultScriptProvider (20 tests) - manifest loading, filesystem CRUD, subprocess execution (python/bash/sh), positional/named/flag args, env vars, stdin JSON, timeout, concurrency limits

**Integration tests (27 new):**
- TestScriptsNotConfigured (8) - all endpoints return 501 without scripts_dir
- TestScriptsCRUD (9) - upload/list/get/delete lifecycle, multipart parsing, entity validation
- TestScriptsExecution (9) - Python/shell execution with polling, termination, status transitions

Run: `./scripts/test.sh` (unit) and `./scripts/test.sh integ` (integration)

---

## Checklist


src/ros2_medkit_integration_tests/test/features/test_scripts_api.test.py
```diff
@@ -0,0 +1,770 @@
...
+from ros2_medkit_test_utils.constants import ALLOWED_EXIT_CODES, API_BASE_PATH, get_test_port
+from ros2_medkit_test_utils.gateway_test_case import GatewayTestCase
+from ros2_medkit_test_utils.launch_helpers import create_demo_nodes, get_coverage_env
+
+
+PORT_WITH_SCRIPTS = get_test_port(0)
+PORT_NO_SCRIPTS = get_test_port(1)
+
+SCRIPTS_DIR = tempfile.mkdtemp(prefix='medkit_scripts_test_')
+
```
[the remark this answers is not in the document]
bburda: Fixed in 9bf2e6c - replaced `mkdtemp` with `TemporaryDirectory` so cleanup is automatic.

src/ros2_medkit_gateway/include/ros2_medkit_gateway/http/error_codes.hpp
```diff
@@ -129,6 +129,16 @@ constexpr const char * ERR_X_MEDKIT_ENTITY_MISMATCH = "x-medkit-entity-mismatch"
...
 constexpr const char * ERR_X_MEDKIT_UNSUPPORTED_PROTOCOL = "x-medkit-unsupported-protocol";
 
+// Script error codes
+constexpr const char * ERR_SCRIPT_NOT_FOUND = "resource-not-found";
+constexpr const char * ERR_SCRIPT_ALREADY_EXISTS = "x-medkit-script-already-exists";
+constexpr const char * ERR_SCRIPT_MANAGED = "x-medkit-managed-script";
+constexpr const char * ERR_SCRIPT_RUNNING = "x-medkit-script-running";
+constexpr const char * ERR_SCRIPT_NOT_RUNNING = "x-medkit-script-not-running";
+constexpr const char * ERR_SCRIPT_CONCURRENCY_LIMIT = "x-medkit-concurrency-limit";
+constexpr const char * ERR_SCRIPT_UNSUPPORTED_TYPE = "invalid-parameter";
```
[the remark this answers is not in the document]
bburda: Fixed in 1113346 - removed `ERR_SCRIPT_NOT_FOUND` and `ERR_SCRIPT_UNSUPPORTED_TYPE`, now reusing `ERR_RESOURCE_NOT_FOUND` and `ERR_INVALID_PARAMETER` directly. Kept the vendor-specific `x-medkit-*` codes (no existing equivalents).

src/ros2_medkit_gateway/include/ros2_medkit_gateway/default_script_provider.hpp
```diff
@@ -0,0 +1,180 @@
...
+
+  // Execution tracking
+  std::unordered_map<std::string, std::unique_ptr<ExecutionState>> executions_;
+  mutable std::mutex exec_mutex_;
+  int active_execution_count_ = 0;
+
+  // Counter for ID generation
+  std::atomic<int> id_counter_{0};
+  std::atomic<int> exec_id_counter_{0};
+};
```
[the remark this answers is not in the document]
bburda: Fixed in 9585a15 - moved `exec_mutex_` before `executions_` and `fs_mutex_` before `manifest_scripts_` so mutexes outlive the data they protect.

src/ros2_medkit_gateway/src/default_script_provider.cpp
```diff
@@ -0,0 +1,986 @@
...
+  std::lock_guard<std::mutex> lock(exec_mutex_);
+
+  auto it = executions_.find(execution_id);
+  if (it == executions_.end()) {
+    return tl::make_unexpected(
+        ScriptBackendErrorInfo{ScriptBackendError::NotFound, "Execution not found: " + execution_id});
+  }
+
+  return state_to_info(*it->second);
+}
```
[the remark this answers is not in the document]
bburda: Fixed in 9585a15 - all three execution methods (get/control/delete) now validate `state.entity_id == entity_id && state.script_id == script_id` before allowing access. Added `CrossEntityExecutionAccessRejected` unit test.

src/ros2_medkit_gateway/src/default_script_provider.cpp
```diff
@@ -0,0 +1,986 @@
...
+    return tl::make_unexpected(
+        ScriptBackendErrorInfo{ScriptBackendError::NotFound, "Execution not found: " + execution_id});
+  }
+
+  auto & state = *it->second;
+
+  if (state.status != "running") {
+    return tl::make_unexpected(ScriptBackendErrorInfo{ScriptBackendError::NotRunning,
+                                                      "Execution is not running (status: " + state.status + ")"});
+  }
```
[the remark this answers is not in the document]
bburda: Fixed in 9585a15 - same ownership validation as get_execution. See above.

src/ros2_medkit_gateway/src/default_script_provider.cpp
```diff
@@ -0,0 +1,986 @@
...
+
+    auto & state = *it->second;
+    if (state.status == "running") {
+      return tl::make_unexpected(ScriptBackendErrorInfo{ScriptBackendError::AlreadyRunning,
+                                                        "Cannot delete running execution. Stop it first."});
+    }
+
+    owned = std::move(it->second);
+    executions_.erase(it);
+  }
```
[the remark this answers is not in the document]
bburda: Fixed in 9585a15 - same ownership validation. All three execution methods now validate entity_id and script_id match.

src/ros2_medkit_gateway/src/http/handlers/script_handlers.cpp
```diff
@@ -0,0 +1,456 @@
...
+  if (!check_backend(res)) {
+    return;
+  }
+
+  try {
+    auto entity_id = req.matches[1].str();
+    auto entity = ctx_.validate_entity_for_route(req, res, entity_id);
+    if (!entity) {
+      return;
+    }
```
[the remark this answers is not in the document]
bburda: Routes are registered via RouteRegistry with fixed regex patterns, so matches are guaranteed by cpp-httplib routing. Adding defensive checks would be inconsistent with other handlers (data, operations, faults, etc.) which all rely on routing guarantees. The `is_valid_resource_id` validation added in 809c103 provides the safety net for extracted values.

src/ros2_medkit_gateway/src/http/handlers/script_handlers.cpp
```diff
@@ -0,0 +1,456 @@
...
+    if (!result) {
+      send_script_error(res, result.error());
+      return;
+    }
+
+    json items = json::array();
+    for (const auto & info : *result) {
+      items.push_back(script_info_to_json(info, base_path));
+    }
+    HandlerContext::send_json(res, json{{"items", items}});
```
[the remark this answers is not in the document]
bburda: Fixed in 9c599a4 - added `test_script_handlers.cpp` with 18 unit tests covering all 501 paths (9 tests) and error mapping + response codes (9 tests: 404, 409, 413, 429, 500, 201, 202, 204).

docs/api/rest.rst
```diff
@@ -1189,6 +1189,112 @@ Subscribe to a specific configuration parameter:
...
+      :widths: 25 15 10 50
+
+      * - Attribute
+        - Type
+        - Conv
+        - Description
+      * - ``execution_type``
+        - string
+        - M
+        - When to run: ``now``, ``on_restart``, ``now_and_on_restart``, ``once_on_restart``
```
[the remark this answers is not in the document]
bburda: Fixed in 9bf2e6c - added a note clarifying that the built-in backend supports only `now`, and other execution types require a plugin-provided ScriptProvider.

src/ros2_medkit_gateway/src/default_script_provider.cpp
```diff
@@ -0,0 +1,986 @@
...
+
+  std::filesystem::remove_all(dir, ec);
+  if (ec) {
+    return tl::make_unexpected(
+        ScriptBackendErrorInfo{ScriptBackendError::Internal, "Failed to delete script: " + ec.message()});
+  }
+
+  return {};
+}
```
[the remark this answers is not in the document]
bburda: Fixed in 9585a15 - `delete_script` now checks for running executions under `exec_mutex_` and returns `AlreadyRunning` (409) per SOVD spec. Added `DeleteScriptWithRunningExecutionBlocked` unit test.

src/ros2_medkit_gateway/src/script_manager.cpp
```diff
@@ -0,0 +1,73 @@
...
+                                                                                     const std::string & execution_id,
+                                                                                     const std::string & action) {
+  return backend_->control_execution(entity_id, script_id, execution_id, action);
+}
+
+tl::expected<void, ScriptBackendErrorInfo> ScriptManager::delete_execution(const std::string & entity_id,
+                                                                           const std::string & script_id,
+                                                                           const std::string & execution_id) {
+  return backend_->delete_execution(entity_id, script_id, execution_id);
+}
```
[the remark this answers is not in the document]
bburda: By design - the handler layer calls `check_backend()` before any ScriptManager method, same pattern as UpdateHandlers. ScriptManager is a thin pass-through and assumes callers gate on `has_backend()`. Adding redundant null checks would diverge from the UpdateManager pattern. The 18 handler unit tests (9c599a4) verify the 501 guard works correctly.

src/ros2_medkit_gateway/src/http/handlers/script_handlers.cpp
```diff
@@ -0,0 +1,456 @@
...
+    }
+
+    std::string entity_type_segment = (req.path.find("/components/") != std::string::npos) ? "components" : "apps";
+    auto base_path = "/" + entity_type_segment + "/" + entity_id;
+
+    auto result = script_mgr_->list_scripts(entity_id);
+    if (!result) {
+      send_script_error(res, result.error());
+      return;
+    }
```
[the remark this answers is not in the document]
bburda: Fixed in 1113346 - all 8 handler methods now call `validate_collection_access(*entity, ResourceCollection::SCRIPTS)` after `validate_entity_for_route()`, matching the pattern used by bulk-data and other handlers.

src/ros2_medkit_gateway/src/default_script_provider.cpp
```diff
@@ -0,0 +1,1014 @@
...
+#include <unistd.h>
+
+#include <algorithm>
+#include <chrono>
+#include <cstring>
+#include <fstream>
+#include <iomanip>
+#include <sstream>
+#include <thread>
+#include <vector>
```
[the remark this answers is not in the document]
bburda: Fixed in 423de80 - added `#include <cerrno>` and `#include <sys/select.h>` explicitly.

src/ros2_medkit_gateway/src/default_script_provider.cpp
```diff
@@ -0,0 +1,1014 @@
...
+      if (stderr_open && FD_ISSET(stderr_fd, &read_fds)) {
+        auto n = read(stderr_fd, buf, sizeof(buf));
+        if (n > 0) {
+          if (state_ptr->stderr_data.size() < kMaxOutputSize) {
+            auto to_append = std::min(static_cast<size_t>(n), kMaxOutputSize - state_ptr->stderr_data.size());
+            state_ptr->stderr_data.append(buf, to_append);
+          }
+        } else {
+          stderr_open = false;
+        }
```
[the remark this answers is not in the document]
bburda: Fixed in 423de80 - read() now distinguishes `n == 0` (EOF), `errno == EINTR` (retry on next select iteration), and real errors (close fd).

src/ros2_medkit_gateway/include/ros2_medkit_gateway/default_script_provider.hpp
```diff
@@ -0,0 +1,181 @@
...
+
+  ~ExecutionState();
+
+  // Non-copyable, non-movable (due to atomic + threads)
+  ExecutionState() = default;
+  ExecutionState(const ExecutionState &) = delete;
+  ExecutionState & operator=(const ExecutionState &) = delete;
+  ExecutionState(ExecutionState &&) = delete;
+  ExecutionState & operator=(ExecutionState &&) = delete;
+};
```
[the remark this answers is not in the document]
bburda: Fixed in 423de80 - added `~DefaultScriptProvider() override` that SIGKILL-s all running process groups, moves states out of the map, and joins threads outside the lock for bounded shutdown.

src/ros2_medkit_gateway/src/default_script_provider.cpp
```diff
@@ -0,0 +1,1014 @@
...
+    // Wildcard matches everything
+    if (filter == "*") {
+      return true;
+    }
+
+    // Check for glob pattern ending with /* (e.g. "components/*")
+    if (filter.size() >= 2 && filter.substr(filter.size() - 2) == "/*") {
+      // The prefix before /* is informational (entity type hint) - since entity_id
+      // alone does not carry type info, a trailing wildcard matches any entity
+      return true;
```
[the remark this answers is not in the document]
bburda: By design for now - entity_id alone does not carry type information. The comment in the code explains this limitation. Fixing properly requires passing entity type through the provider API, which is a larger change for a future PR. The current behavior is documented.

src/ros2_medkit_integration_tests/test/features/test_scripts_api.test.py
```diff
@@ -0,0 +1,771 @@
...
+        self.assertEqual(data['vendor_code'], 'x-medkit-script-running')
+
+        # Cleanup: stop the execution
+        requests.put(
+            f'{self.BASE_URL}/apps/temp_sensor/scripts/{script_id}'
+            f'/executions/{exec_id}',
+            json={'action': 'stop'},
+            timeout=5,
+        )
+
```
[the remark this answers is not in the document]
bburda: Fixed in 423de80 - replaced both `time.sleep(0.5)` calls with polling loops that wait up to 10s for execution status to become "running" before attempting stop/delete.

src/ros2_medkit_gateway/src/script_manager.cpp
```diff
@@ -0,0 +1,73 @@
...
+  backend_ = backend;
+}
+
+bool ScriptManager::has_backend() const {
+  return backend_ != nullptr;
+}
+
+tl::expected<std::vector<ScriptInfo>, ScriptBackendErrorInfo>
+ScriptManager::list_scripts(const std::string & entity_id) {
+  return backend_->list_scripts(entity_id);
```
mfaferek93: **Blocker**: Every method blindly dereferences `backend_` without null check. `backend_` is initialized to `nullptr` in the header. Handler layer's `check_backend()` prevents crash in practice, but `UpdateManager` guards every method with `if (!backend_) return NoBackend`. ScriptManager should follow the same pattern.

Add null guard returning `ScriptBackendError::Internal` (or a dedicated `NoBackend` value) at the top of every method.

src/ros2_medkit_gateway/src/default_script_provider.cpp
```diff
@@ -0,0 +1,1050 @@
...
+            ScriptBackendErrorInfo{ScriptBackendError::Internal, "Failed to read script metadata"});
+      }
+    }
+  }
+
+  return tl::make_unexpected(ScriptBackendErrorInfo{ScriptBackendError::NotFound, "Script not found: " + script_id});
+}
+
+tl::expected<ScriptUploadResult, ScriptBackendErrorInfo>
+DefaultScriptProvider::upload_script(const std::string & entity_id, const std::string & filename,
```
mfaferek93, line 225: **Major**: Once `scripts_dir` is configured, any user with CONFIGURATOR role can upload arbitrary scripts that execute with full gateway privileges. No sandboxing, no signing, no `allow_uploads` toggle.

Feature is disabled by default (`scripts_dir = ""`) and RBAC gates properly, so this isn't RCE-by-default. But for hardened deployments that only want pre-deployed manifest scripts, there's no way to enable execution without also enabling uploads.

Consider adding `allow_uploads: false` config toggle.

src/ros2_medkit_gateway/src/http/handlers/script_handlers.cpp
```diff
@@ -0,0 +1,490 @@
...
+
+void ScriptHandlers::handle_get_script(const httplib::Request & req, httplib::Response & res) {
+  if (!check_backend(res)) {
+    return;
+  }
+
+  try {
+    auto entity_id = req.matches[1].str();
+    auto script_id = req.matches[2].str();
+    if (!is_valid_resource_id(script_id)) {
```
mfaferek93, line 247: **Major (test gap)**: `is_valid_resource_id` is correctly wired here and in all other handlers (lines 267, 301, 367, 371, 406, 410, 459, 464). The function rejects non-`[a-zA-Z0-9_-]` chars, empty strings, and >256 chars.

But zero tests exercise this validation with adversarial inputs: path traversal (`../../../etc/passwd`), empty IDs, shell metacharacters, 257-char strings. This is the primary input validation boundary and should have explicit test coverage.

src/ros2_medkit_gateway/src/http/handlers/script_handlers.cpp
```diff
@@ -0,0 +1,490 @@
...
+
+    json body;
+    try {
+      body = json::parse(req.body);
+    } catch (const json::parse_error &) {
+      HandlerContext::send_error(res, 400, ERR_INVALID_REQUEST, "Invalid JSON body");
+      return;
+    }
+
+    if (!body.contains("action") || !body["action"].is_string() || body["action"].get<std::string>().empty()) {
```
mfaferek93, line 449: **Major (test gap)**: Validation here is correct (checks missing, non-string, empty `action`). But no unit test and no integration test exercises this path.

Add tests for `PUT .../executions/{eid}` with `{}`, `{"action": ""}`, `{"action": "bogus"}` - all should return 400.

src/ros2_medkit_gateway/include/ros2_medkit_gateway/script_manager.hpp
```diff
@@ -0,0 +1,51 @@
...
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+#pragma once
+
+#include "ros2_medkit_gateway/providers/script_provider.hpp"
+
+namespace ros2_medkit_gateway {
+
+class ScriptManager {
```
mfaferek93, line 31: **Major**: Pure delegation wrapper - 8 methods that each `return backend_->foo(args)`. No mutex, no null guard, no state, no error mapping. `UpdateManager` owns async state, manages lifecycle transitions, has its own mutex, maps error codes.

Either eliminate this class (handlers call ScriptProvider directly) or add actual value: null guards, mutex, logging. If lifecycle management is planned for a follow-up, the null guard and mutex are needed now regardless.

src/ros2_medkit_gateway/include/ros2_medkit_gateway/script_manager.hpp
```diff
@@ -0,0 +1,51 @@
...
+  get_execution(const std::string & entity_id, const std::string & script_id, const std::string & execution_id);
+  tl::expected<ExecutionInfo, ScriptBackendErrorInfo> control_execution(const std::string & entity_id,
+                                                                        const std::string & script_id,
+                                                                        const std::string & execution_id,
+                                                                        const std::string & action);
+  tl::expected<void, ScriptBackendErrorInfo>
+  delete_execution(const std::string & entity_id, const std::string & script_id, const std::string & execution_id);
+
+ private:
+  ScriptProvider * backend_ = nullptr;
```
mfaferek93, line 111: **Major**: No mutex protecting `backend_`. `set_backend()` writes this while HTTP threads can concurrently read it. `UpdateManager` has `mutable std::mutex mutex_`. In practice `set_backend()` is init-only, but the pattern is inconsistent.

Also: class is copyable/movable (no deleted copy/move). `UpdateManager` explicitly deletes them.

src/ros2_medkit_gateway/include/ros2_medkit_gateway/default_script_provider.hpp
```diff
@@ -0,0 +1,182 @@
...
+  std::string generate_execution_id();
+
+  ScriptsConfig config_;
+  mutable std::mutex fs_mutex_;
+  std::map<std::string, ScriptEntryConfig> manifest_scripts_;
+
+  // Execution tracking - exec_mutex_ declared before executions_ so the mutex
+  // outlives the map (ExecutionState destructors join threads that lock it).
+  mutable std::mutex exec_mutex_;
+  std::unordered_map<std::string, std::unique_ptr<ExecutionState>> executions_;
```
mfaferek93, line 176: **Major**: Completed/failed/terminated executions stay in this map forever. No TTL, no max-entries, no periodic cleanup. Each can hold up to 1MB stdout+stderr (per `kMaxOutputSize`). Unbounded memory growth proportional to execution count.

Add `max_execution_history` or `execution_ttl_sec` config with eviction of oldest completed entries.

src/ros2_medkit_gateway/src/default_script_provider.cpp
```diff
@@ -0,0 +1,1050 @@
...
+  setpgid(child_pid, child_pid);  // Also set in parent (race-free with child's setpgid(0,0))
+
+  // Close pipe ends we don't use
+  close(stdout_pipe[1]);
+  close(stderr_pipe[1]);
+
+  // Write stdin data and close
+  if (pass_stdin_json) {
+    close(stdin_pipe[0]);  // Close read end in parent
+    auto written = write(stdin_pipe[1], stdin_data.data(), stdin_data.size());
```
mfaferek93, line 546: **Minor**: Blocking `write()` to pipe with `(void)written` ignoring partial writes. POSIX pipe buffer is ~64KB. True deadlock is unlikely since the monitor thread drains stdout/stderr separately, but possible if the child fills both output pipes before reading stdin.

Enforce a size limit on parameters JSON or use non-blocking write with select loop.

src/ros2_medkit_gateway/src/http/handlers/script_handlers.cpp
```diff
@@ -0,0 +1,490 @@
...
+    }
+
+    auto result = script_mgr_->upload_script(entity_id, file_part.filename, file_part.content, metadata);
+    if (!result) {
+      send_script_error(res, result.error());
+      return;
+    }
+
+    std::string entity_type_segment = (req.path.find("/components/") != std::string::npos) ? "components" : "apps";
+    auto script_path = "/api/v1/" + entity_type_segment + "/" + entity_id + "/scripts/" + result->id;
```
mfaferek93: **Major**: Hardcoded `/api/v1/` instead of `api_path()`. Line 90 in this same file correctly uses `api_path()` for href generation. Same issue on line 348 for execution Location header.

```cpp
auto script_path = api_path("/" + entity_type_segment + "/" + entity_id + "/scripts/" + result->id);
```

src/ros2_medkit_gateway/src/default_script_provider.cpp
```diff
@@ -0,0 +1,1050 @@
...
+  std::lock_guard<std::mutex> lock(fs_mutex_);
+
+  if (config_.scripts_dir.empty()) {
+    return tl::make_unexpected(
+        ScriptBackendErrorInfo{ScriptBackendError::Internal, "Scripts directory not configured"});
+  }
+
+  // Check file size
+  auto max_bytes = static_cast<size_t>(config_.max_file_size_mb) * 1024 * 1024;
+  if (content.size() > max_bytes) {
```
mfaferek93, line 241: **Major (test gap)**: This size check has zero test coverage. `FileTooLarge` is only tested via mock injection at the handler level, not through an actual oversized upload. Add a provider-level test with small `max_file_size_mb` and content exceeding the limit.

src/ros2_medkit_gateway/include/ros2_medkit_gateway/providers/script_provider.hpp
```diff
@@ -0,0 +1,104 @@
...
+
+struct ExecutionRequest {
+  std::string execution_type;  // now, on_restart, etc.
+  std::optional<nlohmann::json> parameters;
+  std::optional<std::string> proximity_response;
+};
+
+enum class ScriptBackendError {
+  NotFound,
+  AlreadyExists,
```
mfaferek93: **Major**: `AlreadyExists` is defined here and handled in `send_script_error` (maps to 409), but `DefaultScriptProvider` never returns it - unreachable dead code. Either remove or document as reserved for plugins.

`NotRunning` (line 60) IS returned at `default_script_provider.cpp:753,759` but has zero test coverage. Add a test: start execution -> wait for completion -> `control_execution("stop")` -> expect `NotRunning`.

src/ros2_medkit_gateway/src/default_script_provider.cpp
```diff
@@ -0,0 +1,1050 @@
...
+          continue;
+        }
+      }
+    }
+  }
+
+  return result;
+}
+
+tl::expected<ScriptInfo, ScriptBackendErrorInfo> DefaultScriptProvider::get_script(const std::string & entity_id,
```
mfaferek93, line 173: **Minor**: `get_script` reads uploaded script metadata from filesystem without holding `fs_mutex_`. `list_scripts` (line 94) and `delete_script` (line 328) do hold it. Concurrent `get_script` + `delete_script` could read a partially-deleted directory.

src/ros2_medkit_gateway/src/default_script_provider.cpp
```diff
@@ -0,0 +1,1050 @@
...
+    std::lock_guard<std::mutex> lock(exec_mutex_);
+    for (const auto & [exec_id, exec_state] : executions_) {
+      if (exec_state->entity_id == entity_id && exec_state->script_id == script_id && exec_state->status == "running") {
+        return tl::make_unexpected(ScriptBackendErrorInfo{ScriptBackendError::AlreadyRunning,
+                                                          "Cannot delete script with running execution: " + exec_id});
+      }
+    }
+  }
+
+  std::lock_guard<std::mutex> lock(fs_mutex_);
```
mfaferek93: **Minor**: TOCTOU - `exec_mutex_` is released at end of block above, then `fs_mutex_` acquired here. A new execution for this script could start in this window. The script gets deleted from disk while a subprocess still references it.

Re-check running executions after acquiring `fs_mutex_`, or mark script as "deleting".

src/ros2_medkit_gateway/src/http/handlers/script_handlers.cpp
```diff
@@ -0,0 +1,490 @@
...
+    if (!entity) {
+      return;
+    }
+
+    if (auto err = HandlerContext::validate_collection_access(*entity, ResourceCollection::SCRIPTS)) {
+      HandlerContext::send_error(res, 400, ERR_COLLECTION_NOT_SUPPORTED, *err);
+      return;
+    }
+
+    auto file_it = req.files.find("file");
```
mfaferek93, line 203: **Minor**: Accessing `req.files` without verifying `Content-Type: multipart/form-data`. Wrong content type returns confusing "Missing required multipart field: file" instead of 415 Unsupported Media Type.

src/ros2_medkit_gateway/src/http/handlers/script_handlers.cpp
```diff
@@ -0,0 +1,490 @@
...
+    if (!entity) {
+      return;
+    }
+
+    if (auto err = HandlerContext::validate_collection_access(*entity, ResourceCollection::SCRIPTS)) {
+      HandlerContext::send_error(res, 400, ERR_COLLECTION_NOT_SUPPORTED, *err);
+      return;
+    }
+
+    std::string entity_type_segment = (req.path.find("/components/") != std::string::npos) ? "components" : "apps";
```
mfaferek93: **Minor**: `entity_type_segment` derivation via `req.path.find("/components/")` is duplicated 4 times (lines 150, 211, 244, 346). Fragile if new entity types are added. Extract to a helper.

src/ros2_medkit_gateway/src/http/handlers/script_handlers.cpp
```diff
@@ -0,0 +1,490 @@
...
+    if (!result) {
+      send_script_error(res, result.error());
+      return;
+    }
+
+    json items = json::array();
+    for (const auto & info : *result) {
+      items.push_back(script_info_to_json(info, base_path));
+    }
+    HandlerContext::send_json(res, json{{"items", items}});
```
mfaferek93: **Minor**: Collection response lacks `_links` (self, parent). Other collection handlers (discovery subareas, subcomponents, hosts) consistently include HATEOAS links. Same gap in `handle_get_script` and `handle_get_execution`.

src/ros2_medkit_gateway/src/default_script_provider.cpp
```diff
@@ -0,0 +1,1050 @@
...
+    timeout_thread.join();
+  }
+}
+
+// --- Constructor ---
+
+DefaultScriptProvider::DefaultScriptProvider(const ScriptsConfig & config) : config_(config) {
+  // Populate manifest_scripts_ from config entries
+  for (const auto & entry : config_.entries) {
+    manifest_scripts_[entry.id] = entry;
```
mfaferek93, line 55: **Minor**: Manifest `entry.path` stored from YAML without validation. No check that path exists, is a regular file, or is within an expected directory. Validate during construction.

src/ros2_medkit_gateway/src/default_script_provider.cpp
```diff
@@ -0,0 +1,1050 @@
...
+    }
+    depth++;
+  }
+  if (depth != 2) {
+    return {};  // Must be exactly entity_id/script_id
+  }
+  return canonical_path;
+}
+
+std::string DefaultScriptProvider::generate_id() {
```
mfaferek93, line 892: **Minor**: Sequential IDs (`script_0`, `script_1`) are trivially predictable and enumerable. For an HTTP-facing API, consider UUIDs or a random component.

src/ros2_medkit_gateway/src/default_script_provider.cpp
```diff
@@ -0,0 +1,1050 @@
...
+        }
+        std::this_thread::sleep_for(std::chrono::milliseconds(100));
+      }
+
+      // Timeout reached - mark and send SIGTERM to process group
+      state_ptr->timed_out.store(true);
+      if (state_ptr->pid.load() <= 0) {
+        return;
+      }
+      kill(-child_pid, SIGTERM);  // Negative PID = process group
```
mfaferek93, line 714: **Minor**: Timeout thread captures `child_pid` by value. TOCTOU gap between `pid.load() > 0` check (line 682) and `kill(-child_pid, SIGTERM)` here. Mitigated by pid.load() checks and process group kill making PGID collision unlikely. Practical risk very low but exists.

src/ros2_medkit_gateway/src/auth/auth_config.cpp
```diff
@@ -43,6 +43,13 @@ const std::unordered_map<UserRole, std::unordered_set<std::string>> & AuthConfig
            "GET:/api/v1/components/*/configurations/*",
            "GET:/api/v1/components/*/faults",
            "GET:/api/v1/components/*/faults/*",
+           // Scripts: read-only (list, details, execution status)
+           "GET:/api/v1/components/*/scripts",
```
mfaferek93, line 156: **Minor (pre-existing)**: Scripts correctly adds both `components/*` and `apps/*` patterns here. But the existing permissions above (data, operations, faults, configurations - lines 36-45) only have `components/*` - no `apps/*`. This means VIEWER/OPERATOR/CONFIGURATOR can't access non-script resources via app paths. Track as separate issue.

src/ros2_medkit_gateway/include/ros2_medkit_gateway/providers/script_provider.hpp
```diff
@@ -0,0 +1,104 @@
...
+  FileTooLarge,
+  Internal
+};
+
+struct ScriptBackendErrorInfo {
+  ScriptBackendError code;
+  std::string message;
+};
+
+class ScriptProvider {
```
mfaferek93, line 89: **Minor**: No thread safety documentation. `UpdateProvider` has extensive `@par Thread Safety` docs explaining which methods run on which threads. Plugin implementors won't know synchronization requirements.

mfaferek93 (changes requested):

src/ros2_medkit_gateway/src/script_manager.cpp
bburda: Fixed in 3879e32 (null guards in all 8 methods) and e61c538 (try/catch wrapping + logging for plugin safety). ScriptManager now catches exceptions from plugin backends and returns Internal error with RCLCPP_ERROR logging.

src/ros2_medkit_gateway/src/default_script_provider.cpp
bburda: Fixed in 3bfb922 - added `scripts.allow_uploads` config toggle (default `true`). When `false`, POST upload returns 400 "Script uploads are disabled by configuration". Hardened deployments can use manifest-only scripts without exposing upload API. Also documented in rest.rst and gateway_params.yaml.

src/ros2_medkit_gateway/src/http/handlers/script_handlers.cpp
bburda: Fixed in be7b7d4 - added handler unit tests for adversarial inputs: path traversal (`../../../etc/passwd`), shell metacharacters (`test;rm`), overlong strings (257 chars), plus positive test for valid IDs with underscore/hyphen.

src/ros2_medkit_gateway/src/http/handlers/script_handlers.cpp
bburda: Fixed in be7b7d4 - added handler unit tests for `{}` (missing action), `{"action": ""}` (empty action), `{"action": "bogus"}` (invalid action), and non-JSON body. All return 400.

src/ros2_medkit_gateway/include/ros2_medkit_gateway/script_manager.hpp
bburda: Addressed in e61c538 - ScriptManager is now a safety/observability layer, not a pure passthrough. Added: (1) null guards returning Internal error (3879e32), (2) try/catch wrapping all backend calls for plugin safety - third-party providers may throw (e61c538), (3) RCLCPP_ERROR logging with entity/script/execution IDs (e61c538), (4) Doxygen documentation with thread safety contract (e61c538). The class serves as the integration seam for swapping default/plugin backends - same architectural role as UpdateManager.

src/ros2_medkit_gateway/include/ros2_medkit_gateway/script_manager.hpp
bburda: `set_backend()` is init-only - called in GatewayNode constructor before HTTP threads start. Adding a mutex for a write-once field would be premature. Copy/move constructors are now deleted (3879e32). UpdateManager also has no mutex on its `backend_` field for the same reason - it is set once during init.

src/ros2_medkit_gateway/include/ros2_medkit_gateway/default_script_provider.hpp
bburda: Fixed in 86a1fb9 - added `scripts.max_execution_history` config (default 100). Oldest completed entries are evicted when limit is exceeded. Eviction runs in start_execution after registering new execution. Thread joins happen outside the lock (same pattern as delete_execution). Added unit test.

src/ros2_medkit_gateway/src/default_script_provider.cpp
bburda: Fixed in 86a1fb9 - added 64KB stdin size limit (`kMaxStdinSize`). Parameters JSON exceeding this returns 400 InvalidInput before fork. This matches the POSIX pipe buffer size and prevents the blocking write scenario. Added unit test.

src/ros2_medkit_gateway/src/http/handlers/script_handlers.cpp
bburda: Fixed in 3879e32 - replaced hardcoded `/api/v1/` on lines 212 and 348 with `api_path()` helper, consistent with line 90.

src/ros2_medkit_gateway/src/default_script_provider.cpp
bburda: Fixed in be7b7d4 - added `UploadExceedingFileSizeRejected` provider-level test: config with `max_file_size_mb=0`, actual upload attempt returns `FileTooLarge`.

src/ros2_medkit_gateway/include/ros2_medkit_gateway/providers/script_provider.hpp
bburda: Fixed both: (1) `AlreadyExists` documented as "Reserved for plugin implementations" in 3879e32 - kept for plugins like SQLite registry that reject duplicate IDs. (2) `NotRunning` test added in be7b7d4 - `ControlCompletedExecutionReturnsNotRunning` verifies stopping a completed execution returns NotRunning error.

src/ros2_medkit_gateway/src/default_script_provider.cpp
bburda: Already fixed in 075eac0 - `get_script` and `resolve_script` now hold `fs_mutex_`.

src/ros2_medkit_gateway/src/default_script_provider.cpp
bburda: Acknowledged - the exec_mutex_/fs_mutex_ gap is a best-effort check (script deletion while an execution is starting is a rare race). The execution will complete normally since fork already copied the script. Acceptable for initial release.

src/ros2_medkit_gateway/src/http/handlers/script_handlers.cpp
bburda: Acknowledged. cpp-httplib returns empty `req.files` when Content-Type is not multipart - our check `req.files.find("file") == req.files.end()` catches this and returns 400 "Missing file in multipart body". The error message could be more precise for non-multipart requests - good follow-up improvement.

src/ros2_medkit_gateway/src/http/handlers/script_handlers.cpp
bburda: Acknowledged - valid DRY observation. Will extract a helper in a follow-up. The duplication is contained within one file and is straightforward.

src/ros2_medkit_gateway/src/http/handlers/script_handlers.cpp
bburda: Acknowledged. Scripts list response follows the `{"items": [...]}` SOVD pattern. Adding `_links` (self, parent) is a follow-up aligned with HATEOAS improvements across all collection handlers.

src/ros2_medkit_gateway/src/default_script_provider.cpp
bburda: Acknowledged. Manifest path validation at startup would catch misconfiguration early. Currently fails at execution time with a clear error. Good follow-up for manifest loading feature.

src/ros2_medkit_gateway/src/default_script_provider.cpp
bburda: Acknowledged. Script IDs are sequential (`script_0`) while execution IDs now use epoch prefix (`exec_1710...`). Script ID predictability is mitigated by the ownership validation (entity_id/script_id match required). UUID would be better - good follow-up.

src/ros2_medkit_gateway/src/default_script_provider.cpp
bburda: The timeout thread captures `child_pid` by value (not the atomic member), so it uses the original PID throughout its lifetime. The `pid.store(-1)` in the monitor thread only affects subsequent `control_execution` calls. The TOCTOU gap is between the `kill(child_pid, 0)` liveness check and `kill(-child_pid, SIGTERM)` - this is inherent to signal-based process management and acceptable.

src/ros2_medkit_gateway/src/auth/auth_config.cpp
bburda: Good catch on the pre-existing RBAC patterns. The existing roles (viewer, operator, configurator) were missing wildcard patterns for sub-resources in other collections too. Scripts added the correct patterns. Fixing pre-existing patterns is out of scope for this PR.

src/ros2_medkit_gateway/include/ros2_medkit_gateway/providers/script_provider.hpp
bburda: ScriptManager has Doxygen docs with thread safety notes (added in e61c538). ScriptProvider interface should also have `@par Thread Safety` - will add to align with UpdateProvider pattern.

src/ros2_medkit_gateway/src/default_script_provider.cpp
bburda: Fixed in 075eac0 - `get_script` and `resolve_script` now hold `fs_mutex_`.

src/ros2_medkit_gateway/src/default_script_provider.cpp
bburda: Fixed in 87563ff - `delete_script` now holds `exec_mutex_` throughout and acquires `fs_mutex_` inside the same scope. Lock ordering (exec_mutex_ -> fs_mutex_) is consistent across all methods.

src/ros2_medkit_gateway/src/http/handlers/script_handlers.cpp
bburda: Fixed in 0b4e855 - `handle_upload_script` now validates `Content-Type: multipart/form-data` before accessing `req.files`. Returns 400 with clear error message. Added `UploadRejectsWrongContentType` unit test.

src/ros2_medkit_gateway/src/http/handlers/script_handlers.cpp
bburda: Fixed in 0b4e855 - extracted `entity_type_from_path(req)` private static helper. All 4 occurrences replaced.

src/ros2_medkit_gateway/src/http/handlers/script_handlers.cpp
bburda: Fixed in 0b4e855 - `handle_list_scripts` now includes `_links` with `self` and `parent` hrefs, matching the pattern used by discovery collection endpoints.

src/ros2_medkit_gateway/src/default_script_provider.cpp
bburda: Fixed in 87563ff - constructor now validates each manifest entry path with `std::filesystem::exists()` and `is_regular_file()`, logging `RCLCPP_WARN` for missing or invalid paths at startup.

src/ros2_medkit_gateway/src/default_script_provider.cpp
bburda: Fixed in 87563ff - IDs now use `steady_clock` millisecond timestamp prefix plus atomic counter (e.g., `script_1710752400123_0`, `exec_1710752400456_0`). Non-predictable and unique across gateway restarts.

src/ros2_medkit_gateway/src/default_script_provider.cpp
bburda: Fixed in 87563ff - added `state_ptr->pid.load() > 0` check before the grace period `kill(child_pid, 0)` probe, closing the narrow window where pid could be reaped between the outer check and the probe.

src/ros2_medkit_gateway/src/auth/auth_config.cpp
bburda: Fixed in 1300147 - comprehensive RBAC pattern fix for ALL collections and ALL entity types (areas, apps, components, functions). Added missing sub-resource patterns for data, operations, configurations, faults, logs, bulk-data, cyclic-subscriptions, locks, updates, and docs across VIEWER, OPERATOR, and CONFIGURATOR roles.

src/ros2_medkit_gateway/include/ros2_medkit_gateway/providers/script_provider.hpp
bburda: Fixed in 87563ff - added `@par Thread Safety` documentation to `ScriptProvider` interface with contract for implementations. Also added `@param`/`@return` docs to all 8 virtual methods. ScriptManager already had Doxygen docs from e61c538.

PR #301, opened 2026-03-22
fix(gateway): support namespaced fault manager integration

### LINKED ISSUE #141: Gateway cannot resolve Fault Manager GetFault service when Fault Manager runs in a namespace
## Summary

When an additional ROS 2 namespace is applied to the `Fault Manager` node, ros2_medkit_gateway is no longer able to locate the `GetFault` service (e.g., it fails to find/call the expected GetFault/get_faults service). As a result, the system becomes fragile: introducing a namespace (often done for composition, multi-robot setups, or deployment conventions) can silently break Gateway-to-Fault Manager integration and be difficult to diagnose.

## Proposed solution

Update `ros2_medkit_gateway` configuration so it can resolve the `Fault Manager` services even when the Fault Manager runs under a non-root namespace.

Concretely:
- Add an explicit, configurable parameter in the Gateway (e.g., fault_manager_namespace or full service name overrides like get_faults_service_name).
- Use that configuration to construct/resolve the fully-qualified service name (FQN), instead of assuming a fixed path such as /fault_manager/....
- Optionally provide a safe default that preserves current behavior (root namespace), while allowing deployments to set a namespace without breaking service discovery.

This ensures that adding a namespace does not introduce a hard-to-trace integration failure and makes the dependency between Gateway and Fault Manager explicit and operationally visible.

eclipse0922:
# Pull Request

## Summary

Support running the gateway against a fault manager deployed under an additional ROS 2 namespace.

This change:
- resolves fault manager service clients through a shared namespace-aware helper
- applies the same namespace resolution to fault event subscribers used by SSE and triggers
- adds regression tests and updates config/message docs for the new gateway parameters

---

## Issue

- closes #141

---

## Type

- [x] Bug fix
- [x] New feature or tests
- [ ] Breaking change
- [ ] Documentation only

---

## Testing

Ran in the devcontainer:

```bash
colcon build --symlink-install --packages-select ros2_medkit_gateway
source install/setup.bash
colcon test --event-handlers console_direct+ --return-code-on-test-failure --packages-select ros2_medkit_gateway
colcon test-result --verbose --test-result-base build/ros2_medkit_gateway
```

Additional focused verification:

```bash
source install/setup.bash
build/ros2_medkit_gateway/test_fault_manager
```

---

## Checklist

### DIFF
--- a/src/ros2_medkit_gateway/config/gateway_params.yaml
+++ b/src/ros2_medkit_gateway/config/gateway_params.yaml
@@ -92,15 +92,16 @@ ros2_medkit_gateway:
     topic_sample_timeout_sec: 2.0
 
     # Fault Manager integration
-    # Namespace prefix used to resolve Fault Manager services.
-    # Examples:
-    #   ""        -> /fault_manager/list_faults
-    #   "robot1"  -> /robot1/fault_manager/list_faults
-    #   "/robot1" -> /robot1/fault_manager/list_faults
-    fault_manager_namespace: ""
+    fault_manager:
+      # Namespace prefix used to resolve Fault Manager services.
+      # Examples:
+      #   ""        -> /fault_manager/list_faults
+      #   "robot1"  -> /robot1/fault_manager/list_faults
+      #   "/robot1" -> /robot1/fault_manager/list_faults
+      namespace: ""
 
-    # Timeout (in seconds) for Fault Manager service calls
-    fault_service_timeout_sec: 5.0
+      # Timeout (in seconds) for Fault Manager service calls
+      service_timeout_sec: 5.0
 
     # NOTE: Native-only implementation
     # The gateway uses native rclcpp APIs for all ROS 2 interactions:
--- a/src/ros2_medkit_gateway/src/fault_manager_paths.cpp
+++ b/src/ros2_medkit_gateway/src/fault_manager_paths.cpp
@@ -45,7 +45,7 @@ std::string get_fault_manager_namespace(rclcpp::Node * node) {
     throw std::invalid_argument("Fault manager namespace resolution requires a valid node");
   }
 
-  constexpr auto kParameterName = "fault_manager_namespace";
+  constexpr auto kParameterName = "fault_manager.namespace";
   std::string namespace_value;
   if (!node->get_parameter(kParameterName, namespace_value)) {
     return "";
@@ -59,7 +59,7 @@ std::string get_fault_manager_namespace(rclcpp::Node * node) {
     }
 
     RCLCPP_WARN(node->get_logger(),
-                "Ignoring invalid fault_manager_namespace '%s': character '%c' is not allowed. "
+                "Ignoring invalid fault_manager.namespace '%s': character '%c' is not allowed. "
                 "Falling back to root fault manager path.",
                 namespace_value.c_str(), c);
     return "";
--- a/docs/config/server.rst
+++ b/docs/config/server.rst
@@ -175,17 +175,17 @@ Configure how the gateway connects to the fault manager services and event topic
      - Type
      - Default
      - Description
-   * - ``fault_manager_namespace``
+   * - ``fault_manager.namespace``
      - string
      - ``""``
      - Optional namespace prefix for fault manager service and event topic resolution.
        Examples: ``""`` -> ``/fault_manager/list_faults``, ``"robot1"`` -> ``/robot1/fault_manager/list_faults``.
-   * - ``fault_service_timeout_sec``
+   * - ``fault_manager.service_timeout_sec``
      - float
      - ``5.0``
      - Timeout for fault manager service calls such as ``list_faults`` and ``get_snapshots``.
 
-When ``fault_manager_namespace`` is set, the gateway also subscribes to the matching
+When ``fault_manager.namespace`` is set, the gateway also subscribes to the matching
 fault event topic (for example ``/robot1/fault_manager/events`` instead of the default
 ``/fault_manager/events``).
 
@@ -195,8 +195,9 @@ Example:
 
    ros2_medkit_gateway:
      ros__parameters:
-       fault_manager_namespace: "robot1"
-       fault_service_timeout_sec: 5.0
+       fault_manager:
+         namespace: "robot1"
+         service_timeout_sec: 5.0
 
 Logging Configuration
 ---------------------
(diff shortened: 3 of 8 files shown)


src/ros2_medkit_gateway/src/fault_manager_paths.cpp
```diff
@@ -0,0 +1,71 @@
...
+
+std::string get_fault_manager_namespace(rclcpp::Node * node) {
+  if (!node) {
+    throw std::invalid_argument("Fault manager namespace resolution requires a valid node");
+  }
+
+  constexpr auto kParameterName = "fault_manager_namespace";
+  if (!node->has_parameter(kParameterName)) {
+    node->declare_parameter<std::string>(kParameterName, "");
+  }
```
bburda, line 52: `has_parameter()` + conditional `declare_parameter()` is a lazy-declare anti-pattern in this codebase - the parameter will not appear in `ros2 param list` or `--show-args` until `FaultManager` is constructed, and `ros2 param set` will reject the parameter before that point.

Consider moving `declare_parameter("fault_manager_namespace", "")` to `GatewayNode::GatewayNode()` alongside all other parameter declarations, and using only `get_parameter()` here.

src/ros2_medkit_gateway/src/fault_manager_paths.cpp
```diff
@@ -0,0 +1,71 @@
...
+  while (normalized.size() > 1 && normalized.back() == '/') {
+    normalized.pop_back();
+  }
+
+  if (normalized == "/") {
+    return "";
+  }
+
+  return normalized;
+}
```
bburda, line 41: `normalize_fault_manager_namespace` handles slashes but does not validate characters against ROS 2 naming rules. A misconfigured value like `"robot 1"` (space), `"~/trick"` (tilde), or control characters will pass normalization and then crash `create_client` at runtime with an unhelpful error from the rmw layer.

Consider adding a character allowlist check after normalization (letters, digits, underscore, forward slash) with `RCLCPP_WARN` + fallback to empty string on invalid input.

src/ros2_medkit_gateway/src/http/handlers/sse_fault_handler.cpp
```diff
@@ -27,18 +28,18 @@ namespace handlers {
 
 SSEFaultHandler::SSEFaultHandler(HandlerContext & ctx, std::shared_ptr<SSEClientTracker> client_tracker)
   : ctx_(ctx), client_tracker_(std::move(client_tracker)) {
+  const auto fault_events_topic = build_fault_manager_events_topic(ctx_.node());
```
bburda, line 31: Good change - hardcoded path replaced with `build_fault_manager_events_topic()`. However, unlike `TriggerFaultSubscriber` which has two dedicated namespace tests (`TriggerFaultSubscriberUsesConfiguredFaultManagerNamespace`, `TriggerFaultSubscriberForwardsNamespacedFaultEvents`), `SSEFaultHandler` has no test verifying it subscribes to the namespaced topic. If this wiring broke, SSE fault streaming would silently fail in namespaced deployments while all other tests pass.

src/ros2_medkit_gateway/config/gateway_params.yaml
```diff
@@ -91,6 +91,17 @@ ros2_medkit_gateway:
     # Valid range: 0.1-30.0
     topic_sample_timeout_sec: 2.0
 
+    # Fault Manager integration
+    # Namespace prefix used to resolve Fault Manager services.
+    # Examples:
+    #   ""        -> /fault_manager/list_faults
+    #   "robot1"  -> /robot1/fault_manager/list_faults
+    #   "/robot1" -> /robot1/fault_manager/list_faults
+    fault_manager_namespace: ""
```
bburda: Nit: other multi-parameter features use dot-notation grouping (`server.host`, `sse.max_clients`, `logs.buffer_size`, `auth.enabled`, `triggers.enabled`). These two are flat top-level keys. Consider `fault_manager.namespace` / `fault_manager.service_timeout_sec` for consistency - it would also give a natural home for future fault-manager-related parameters.

src/ros2_medkit_gateway/src/fault_manager_paths.cpp
eclipse0922: Moved the parameter declaration to GatewayNode and removed the lazy declare path from FaultManager.

The gateway now declares `fault_manager_namespace` and `fault_service_timeout_sec` up front, so they show up in parameter introspection as expected. `FaultManager` only reads the value now, with a fallback for plain `rclcpp::Node` unit-test construction.

eclipse0922: Updated this to follow the codebase pattern.

`fault_manager.namespace` and `fault_manager.service_timeout_sec` are now declared in `GatewayNode::GatewayNode()`, so they show up in parameter introspection up front. `FaultManager` no longer conditionally declares the parameter and only reads the value, with a fallback retained for plain `rclcpp::Node` unit-test construction.

src/ros2_medkit_gateway/src/http/handlers/sse_fault_handler.cpp
eclipse0922: Added direct coverage for this path.

There is now a `GatewayNode` test that sets `fault_manager.namespace=robot5` and verifies the gateway subscribes on `/robot5/fault_manager/events` instead of `/fault_manager/events`. That covers both the SSE fault handler and the trigger fault subscriber wiring through the real node setup.

src/ros2_medkit_gateway/config/gateway_params.yaml
eclipse0922: Applied this suggestion.

The config surface now uses `fault_manager.namespace` and `fault_manager.service_timeout_sec`, and I updated the gateway config defaults, code paths, docs, and tests to match.

src/ros2_medkit_gateway/src/fault_manager_paths.cpp
eclipse0922: Added validation here.

After normalization, invalid characters now trigger a warning and the code falls back to the root fault manager path instead of passing a malformed name down to client/subscription creation. I also added regression coverage for both `///` and an invalid value like `robot 1`.

PR #310, opened 2026-03-25
fix: merge pipeline fixes, flat entity tree, trigger resolution

### LINKED ISSUE #260: Merge pipeline architecture improvements
## Summary

Several architecture improvements for the merge pipeline to reduce lock contention, improve extensibility, and fix subtle merge semantics.

---

### 1. Double-buffering for HybridDiscovery mutex
`HybridDiscoveryStrategy::refresh()` holds the mutex during the entire `pipeline_.execute()` call, which includes synchronous ROS 2 graph introspection. All HTTP handlers calling `discover_*()` block during refresh, causing request latency spikes.

**File:** `src/ros2_medkit_gateway/src/discovery/hybrid_discovery.cpp:48-55`

**Fix:** Execute pipeline without lock, then swap result under lock.

### 2. Replace dynamic_cast for RuntimeLayer statistics
`MergePipeline::execute()` uses `dynamic_cast<RuntimeLayer*>` to get `last_filtered_count()`. This couples the pipeline to a concrete type, violating OCP.

**File:** `src/ros2_medkit_gateway/src/discovery/merge_pipeline.cpp:418`

**Fix:** Add `virtual size_t filtered_count() const { return 0; }` to `DiscoveryLayer` base class.

### 3. Fix merge_bool sticky OR for `external` field
`merge_bool()` uses `target = target || source` for `MergeWinner::BOTH`. Once any layer sets a boolean to `true`, it can never be reverted. Correct for `is_online` but wrong for `external`.

**File:** `src/ros2_medkit_gateway/src/discovery/merge_pipeline.cpp:89-103`

**Fix:** Move `external` to METADATA field group where manifest is AUTHORITATIVE by default.

### 4. Make MergePipeline accessors private
`get_last_report()` and `get_linking_result()` are public but only called through `HybridDiscoveryStrategy`. Direct calls would be a data race (no internal mutex).

**File:** `src/ros2_medkit_gateway/include/ros2_medkit_gateway/discovery/merge_pipeline.hpp:67-85`

**Fix:** Make them private with `friend class HybridDiscoveryStrategy`.

---

Found during self-review of PR #258. Item 1 (mutex contention) is the most impactful for production - it affects request latency during every discovery refresh cycle.

### LINKED ISSUE #261: Test coverage gaps in merge pipeline and discovery
## Summary

The merge pipeline has strong test coverage (1117-line test file) but several gaps were identified during self-review.

---

### 1. Area-specific field group merge tests
All merge tests use Component or App entities. Area field merge (IDENTITY, HIERARCHY, METADATA) is untested.

**File:** `src/ros2_medkit_gateway/test/test_merge_pipeline.cpp`

### 2. PluginLayer entity ID validation tests
`PluginLayer::discover()` calls `validate_entities()` which drops entities with invalid IDs. No tests exist for empty, oversized (>256 chars), or special-character IDs.

**File:** `src/ros2_medkit_gateway/src/discovery/layers/plugin_layer.cpp:26-45`

### 3. handle_health discovery block unit tests
Unit tests only cover `nullptr` GatewayNode path. The discovery info JSON construction (mode, strategy, pipeline report, linking stats) is never unit-tested.

**File:** `src/ros2_medkit_gateway/test/test_health_handlers.cpp`

### 4. Gap-fill positive assertion
`test_discovery_gap_fill.test.py` sets `allow_heuristic_apps: true` but only checks that areas/components are absent. If gap-fill stopped producing apps, tests would still pass.

**File:** `src/ros2_medkit_integration_tests/test/features/test_discovery_gap_fill.test.py`

### 5. merge_bool ENRICHMENT vs ENRICHMENT edge case
No test where two ENRICHMENT layers both provide `is_online` with conflicting values. The OR semantics should be explicitly tested.

**File:** `src/ros2_medkit_gateway/test/test_merge_pipeline.cpp`

---

Found during self-review of PR #258. These can all be addressed in a single PR.

### LINKED ISSUE #305: Trigger notifications should resolve ROS identifiers to manifest entity IDs
## Summary

Trigger notifications from ROS-derived sources (faults, logs) use raw ROS identifiers that don't match manifest entity IDs. This means triggers created on manifest entities (e.g., `apps/diagnostic-bridge`) never fire - they only work on runtime-discovered entities with ROS node names (e.g., `apps/diagnostic_bridge`).

**Affected collections:**
- **Faults:** `TriggerFaultSubscriber` extracts entity_id from `reporting_sources[0]` last path segment (e.g., `/bridge/diagnostic_bridge` -> `diagnostic_bridge`)
- **Logs:** `LogManager::on_rosout()` uses `normalize_fqn()` which strips leading `/` (e.g., `/bridge/diagnostic_bridge` -> `bridge/diagnostic_bridge`)
- **Data:** Not affected - entity_id comes from the trigger creation request

**Impact:** In hybrid discovery mode, users must know to target runtime entities (underscore names) instead of manifest entities (hyphen names). In manifest-only mode, fault/log triggers cannot work at all since the runtime entities don't exist.

Add an entity ID resolution step in the notification path (either in each subscriber or in `ResourceChangeNotifier::notify()`) that maps ROS node FQNs to their linked manifest entity IDs via the entity cache's `RosBinding.node_name` matching - the same resolution that `RuntimeDiscoveryStrategy` already does for node linking.

- `trigger_fault_subscriber.cpp` lines 44-50
- `log_manager.cpp` line 198, 238
- Discovered while implementing selfpatch/selfpatch_demos#43 (triggers showcase)
- selfpatch/selfpatch_demos#52 works around this by targeting runtime entities

### LINKED ISSUE #307: [BUG] Hybrid discovery merge pipeline does not suppress runtime entities linked to manifest entities
# Bug report

### Steps to reproduce

1. Configure gateway in hybrid mode with a manifest that defines entities with hyphenated IDs (e.g., `lidar-sim`, `fault-manager`)
2. Ensure `create_synthetic_components: false` in config (though the bug occurs regardless of this setting)
3. Start the gateway with ROS 2 nodes whose names match manifest `ros_binding` entries
4. Query entity collection endpoints (`GET /areas`, `/components`, `/apps`)

Reproducible with any selfpatch_demos demo (sensor_diagnostics, moveit_pick_place, turtlebot3_integration).

### Expected behavior

After runtime linking matches runtime nodes to manifest entities via `ros_binding`, the merge pipeline should emit only the manifest-defined entities. Linked runtime entities should be suppressed/merged into their manifest counterparts.

For sensor_diagnostics demo:
- Areas: 4 (sensors, processing, diagnostics, bridge)
- Components: 8 (lidar-unit, imu-unit, gps-unit, camera-unit, compute-unit, gateway, fault-manager, diagnostic-bridge-unit)
- Apps: 8 (lidar-sim, imu-sim, gps-sim, camera-sim, anomaly-detector, medkit-gateway, medkit-fault-manager, diagnostic-bridge)

### Actual behavior

Both manifest entities AND linked runtime entities appear in API responses, creating duplicates:

**Areas: 6 instead of 4**
```
sensors, processing, diagnostics, bridge, fault_manager, root
```
Extra: `fault_manager` (synthetic area from runtime), `root` (synthetic area)

**Components: 17 instead of 8**
```
lidar-unit, imu-unit, gps-unit, camera-unit, compute-unit, gateway, fault-manager, diagnostic-bridge-unit,
lidar_sim, camera_sim, imu_sim, gps_sim, anomaly_detector, diagnostic_bridge, ros2_medkit_gateway, _param_client_node, fault_manager
```
The bottom 9 are runtime-created synthetic entities with underscored IDs that should have been merged.

**Apps: 17 instead of 8**
Same pattern - every manifest app has a corresponding unmerged runtime duplicate.

Gateway logs show linking succeeds but duplicates still appear:
```
Runtime linking: 8 linked, 9 unlinked, 1 orphan nodes
MergePipeline: 43 entities from 4 layers, 4 enriched, 0 conflicts
```

The merge pipeline likely compares entity IDs directly (underscored != hyphenated) instead of using the `ros_binding` linkage to identify duplicates.

### Environment
- ros2_medkit version: main (v0.4.0)
- ROS 2 distro: Jazzy
- OS: Ubuntu 24.04 (Docker)

### Additional information

- `create_synthetic_components: false` does not prevent this - it only changes the grouping strategy, not whether runtime entities are created
- The graph_provider plugin is NOT the source of duplicates - it only enriches existing entities, does not create new ones
- Discovered while adding entity count assertions to selfpatch_demos smoke tests (selfpatch/selfpatch_demos#53)

### LINKED ISSUE #309: Support flat entity tree without areas (component/subcomponent/app only)
## Summary

The README describes two deployment patterns:
1. **Large robot/vehicle** - full hierarchy: areas/subareas/components/subcomponents/apps
2. **Simple robot** - flat tree: components/subcomponents/apps (e.g., a TurtleBot as a single component with subcomponents hosting apps)

The entity model structs already support this - `Component.area` can be empty, `parent_component_id` enables nesting, and the manifest parser treats `areas:` as optional. However, this flat mode has gaps:

**No example manifest without areas.** Even `minimal_manifest.yaml` defines an area. Users have no reference for a flat component-only setup.

**`runtime_only` always synthesizes areas.** `RuntimeDiscoveryStrategy::discover_areas()` extracts areas from namespaces and always creates at least a "root" area. There is no way to run runtime discovery with a flat component tree.

**No integration test coverage** for a manifest without areas. It's unclear if the full handler chain (discovery, data, operations, faults, etc.) works correctly when `EntityCache` has zero areas.

1. Add an example manifest without areas (e.g., `config/examples/flat_robot_manifest.yaml`) showing a TurtleBot-style setup: one top-level component with subcomponents and apps, no areas at all.

2. Add an integration test that boots the gateway in `manifest_only` mode with the flat manifest and verifies:
   - `GET /areas` returns empty `{"items": []}`
   - `GET /components` returns the top-level component
   - `GET /components/{id}/subcomponents` returns children
   - Data, operations, faults, configs all work on apps within the flat tree

3. Consider a `runtime_only` config option to suppress synthetic area creation (or make the "root" area implicit/hidden), so runtime discovery can also produce a flat tree.

Related issues:
- #307 - hybrid discovery merge pipeline duplicates (affects hybrid mode usability)
- #260 - merge pipeline architecture improvements
- #196 - `GET /apps/{id}/belongs-to` endpoint (hierarchy navigation)

The SOVD spec models areas as optional organizational grouping. A simple robot with 5-10 nodes should not need artificial area wrappers to use the gateway effectively.

bburda:
# Pull Request

## Summary

Combined PR addressing 5 issues in the discovery/merge pipeline area:

**#307 (Bug fix):** Hybrid discovery produced duplicate entities (both manifest and runtime synthetic) because merge_entities() merges by ID, and manifest entities use hyphenated IDs while runtime uses underscored. Post-linking suppression now filters runtime-origin apps, components, and areas whose namespaces are covered by linked manifest entities.

**#260 (Architecture):** Four merge pipeline improvements - execute pipeline outside mutex in HybridDiscovery (reduces request latency during refresh), replace dynamic_cast with virtual filtered_count(), move external field from STATUS to METADATA field group (fixes sticky OR semantics), document thread-safety on MergePipeline accessors.

**#309 (Feature):** Flat entity tree support - manifests can now omit areas entirely, using components as top-level entities with subcomponents for hardware modules. New `create_synthetic_areas` parameter for runtime discovery. Example manifest, integration test, and documentation.

**#305 (Feature):** Trigger notifications now resolve ROS node FQNs to manifest entity IDs via injectable NodeToEntityFn callback. TriggerFaultSubscriber and LogManager both use the resolver with last-segment fallback for backward compatibility.

**#261 (Tests):** Comprehensive test coverage additions - area field group merge, merge_bool ENRICHMENT edge case, PluginLayer validation, health endpoint discovery block, hybrid suppression integration test with exact count assertions, flat entity tree integration test.

Additional fixes found during review: deadlock in stop_rest_server (join while holding mutex), ROS_DOMAIN_ID collision, root-namespace over-suppression, TOCTOU in cache update, unbounded LogManager buffers, linker starvation when allow_heuristic_apps=false.

---

## Issue

- closes #307
- closes #260
- closes #261
- closes #309
- closes #305

---

## Type

- [x] Bug fix
- [x] New feature or tests
- [ ] Breaking change
- [x] Documentation only

---

## Testing

- 2124 unit tests pass (0 errors, 0 failures)
- 2 new integration tests: flat entity tree (manifest_only, exact counts), hybrid suppression (exact entity counts match manifest, no duplicates, all apps online)
- Docs build clean with -W (warnings as errors)

Verify locally:
```bash
./scripts/test.sh          # Unit tests
./scripts/test.sh integ    # Integration tests
.venv/bin/sphinx-build -b html docs docs/_build/html -W  # Docs
```

---

## Checklist

### DIFF
--- a/src/ros2_medkit_gateway/test/test_log_manager.cpp
+++ b/src/ros2_medkit_gateway/test/test_log_manager.cpp
@@ -879,8 +879,8 @@ TEST_F(LogManagerResolverTest, ResolverReturnsEmptyFallsBackToLastSegment) {
 
 // @verifies REQ_INTEROP_061
 TEST_F(LogManagerBufferTest, BufferCapDropsNewNodesWhenFull) {
-  // LogManager was created with max_buffer_size_=3 in the fixture.
-  // Buffer cap = max_buffer_size_ * 10 = 30 distinct nodes.
+  // LogManager was created with buffer size 3 in the fixture.
+  // Buffer cap = size * 10 = 30 distinct nodes.
   // Create a fresh manager with buffer size 5 so cap = 50.
   mgr_.reset();
   mgr_ = std::make_unique<LogManager>(node_.get(), nullptr, /*max_buffer_size=*/5);


src/ros2_medkit_gateway/src/models/thread_safe_entity_cache.cpp
```diff
@@ -82,6 +84,11 @@ void ThreadSafeEntityCache::update_topic_types(std::unordered_map<std::string, s
   generation_.fetch_add(1, std::memory_order_release);
 }
 
+std::unordered_map<std::string, std::string> ThreadSafeEntityCache::get_node_to_app() const {
```
mfaferek93, line 87: **Minor**: Returns full `unordered_map<string,string>` by value under shared_lock. Called from the node_resolver lambda (`gateway_node.cpp:817`) on every fault event and rosout message. Each call copies the entire map just to look up one key.

Suggestion: add a `std::string resolve_node(const std::string& fqn) const` method that does the lookup under shared_lock and returns just the result string. Eliminates the per-event map copy entirely.

src/ros2_medkit_gateway/src/log_manager.cpp
```diff
@@ -168,10 +168,41 @@ void LogManager::on_rosout(const rcl_interfaces::msg::Log::ConstSharedPtr & msg)
...
+
+  // Notify triggers about the log event (resolve ROS node name to entity ID)
+  if (notifier_) {
+    std::string entity_id;
+
+    // Try resolving via node_to_app mapping (manifest/hybrid mode)
+    if (node_to_entity_resolver_) {
+      // entry.name has no leading '/' (rcl convention), but the mapping
+      // may use FQNs with leading '/' - try both forms
+      entity_id = node_to_entity_resolver_("/" + entry.name);
```
mfaferek93, line 192: **Minor**: LogManager's resolver usage (prepend `/` to `entry.name`, try both forms, fallback to last segment) has zero unit tests. `test_log_manager.cpp` has no `set_node_to_entity_resolver` calls. The TriggerFaultSubscriber resolver has 3 dedicated tests in `test_fault_manager.cpp`, but LogManager's 3 branches (match with `/` prefix, match bare name, empty fallback) are completely uncovered.

Add unit tests covering all 3 branches.

src/ros2_medkit_gateway/src/discovery/hybrid_discovery.cpp
```diff
@@ -46,9 +46,16 @@ std::vector<Function> HybridDiscoveryStrategy::discover_functions() {
 }
 
 void HybridDiscoveryStrategy::refresh() {
-  // Hold mutex for the full execute()+swap to prevent data races with add_layer()
+  // Execute pipeline WITHOUT lock - safe because:
+  // 1. add_layer() is construction-only (called in DiscoveryManager::initialize()
+  //    before EntityCache timer starts refresh cycles)
+  // 2. refresh() is only called from the ROS executor timer callback (single-threaded)
+  //    - concurrent refresh() calls cannot occur under the default executor model
+  auto new_result = pipeline_.execute();
```
mfaferek93, line 54: **Minor**: `pipeline_.execute()` runs without mutex. `add_layer()` (line 80) also has no lock. Safety relies on `add_layer()` being construction-only - enforced only by the `@warning` comment on the header.

Consider adding a `debug_assert` or `bool initialized_` flag that trips if `add_layer()` is called after the first `execute()`.

src/ros2_medkit_gateway/src/log_manager.cpp
```diff
@@ -168,10 +168,41 @@ void LogManager::on_rosout(const rcl_interfaces::msg::Log::ConstSharedPtr & msg)
 
   if (!suppress_buffer) {
     std::lock_guard<std::mutex> lock(buffers_mutex_);
-    auto & buf = buffers_[entry.name];
-    buf.push_back(std::move(entry));
-    if (buf.size() > max_buffer_size_) {
-      buf.pop_front();
+    // Cap the number of distinct node buffers to prevent unbounded growth.
+    // Uses max_buffer_size_ * 10 as the cap (e.g., 200 entries -> 2000 distinct nodes).
+    if (buffers_.find(entry.name) == buffers_.end() && buffers_.size() >= max_buffer_size_ * 10) {
```
mfaferek93, line 173: **Minor**: Buffer cap at `max_buffer_size_ * 10` distinct nodes has no unit test. Key behaviors to verify: cap reached -> new nodes silently dropped, existing buffers still accept entries, per-node limit still applies within each buffer.

src/ros2_medkit_gateway/src/discovery/merge_pipeline.cpp
```diff
@@ -478,6 +480,53 @@ MergeResult MergePipeline::execute() {
 
     linking_result_ = linker_->get_last_result();
     result.linking_result = linking_result_;
+
+    // Suppress runtime-origin components/areas that duplicate manifest entities (#307)
+    // Build set of namespaces covered by linked manifest apps
+    std::set<std::string> linked_namespaces;
+    for (const auto & [fqn, app_id] : linking_result_.node_to_app) {
+      auto last_slash = fqn.rfind('/');
```
mfaferek93: **Minor**: `rfind('/')` for namespace extraction doesn't handle trailing-slash FQNs (e.g., `/sensors/lidar_node/`). Would produce `/sensors/lidar_node` as namespace instead of `/sensors`. No test covers this edge case. Either strip trailing `/` before extraction, or add a test confirming trailing slashes don't occur in practice.

src/ros2_medkit_gateway/src/models/thread_safe_entity_cache.cpp
```diff
@@ -82,6 +84,11 @@ void ThreadSafeEntityCache::update_topic_types(std::unordered_map<std::string, s
...
+  std::shared_lock lock(mutex_);
+  return node_to_app_;
+}
```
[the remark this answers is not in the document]
bburda: Fixed in 851d373b. Added `resolve_node_to_app(fqn)` keyed lookup method to ThreadSafeEntityCache that does a single find under shared_lock. The gateway_node resolver lambda now uses it instead of copying the full map.

src/ros2_medkit_gateway/test/test_health_handlers.cpp
```diff
@@ -265,3 +274,107 @@ TEST_F(HealthHandlersTest, HandleRootTlsEnabledIncludesTlsMetadataBlock) {
...
+  }
+  socklen_t len = sizeof(addr);
+  if (getsockname(sock, reinterpret_cast<sockaddr *>(&addr), &len) != 0) {
+    close(sock);
+    return 0;
+  }
+  int port = ntohs(addr.sin_port);
+  close(sock);
+  return port;
+}
```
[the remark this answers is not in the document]
bburda: Acknowledged. The TOCTOU window is minimal (microseconds between close and GatewayNode bind) and cpp-httplib doesn't support port 0 binding. Adding a retry loop would add complexity for negligible benefit in a test helper. Leaving as-is.

src/ros2_medkit_gateway/test/test_health_handlers.cpp
```diff
@@ -265,3 +274,107 @@ TEST_F(HealthHandlersTest, HandleRootTlsEnabledIncludesTlsMetadataBlock) {
...
+  // Default mode is runtime_only
+  EXPECT_EQ(disc["mode"].get<std::string>(), "runtime_only");
+
+  ASSERT_TRUE(disc.contains("strategy"));
+  EXPECT_TRUE(disc["strategy"].is_string());
+  EXPECT_FALSE(disc["strategy"].get<std::string>().empty());
+
+  // In runtime_only mode, pipeline and linking are not present (only in hybrid mode).
+  // Verify the block structure is well-formed JSON with at least mode + strategy.
+  EXPECT_EQ(disc.size(), 2u);  // mode + strategy only
```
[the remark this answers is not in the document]
bburda: Fixed in 851d373b. Replaced `EXPECT_EQ(disc.size(), 2u)` with `EXPECT_FALSE(disc.contains("pipeline"))` and `EXPECT_FALSE(disc.contains("linking"))`.

src/ros2_medkit_gateway/src/discovery/merge_pipeline.cpp
```diff
@@ -478,6 +480,53 @@ MergeResult MergePipeline::execute() {
...
+      if (last_slash != std::string::npos && last_slash > 0) {
+        linked_namespaces.insert(fqn.substr(0, last_slash));
+      }
+      // Skip root namespace "/" - too broad, would suppress all root-namespace entities
+    }
```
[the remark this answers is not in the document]
bburda: C++17 structured bindings don't trigger -Wunused-variable with GCC or Clang (confirmed: CI passed with -Wall -Wextra -Wpedantic). No change needed.

src/ros2_medkit_gateway/src/discovery/merge_pipeline.cpp
```diff
@@ -14,10 +14,10 @@
...
 #include "ros2_medkit_gateway/discovery/merge_pipeline.hpp"
 
-#include "ros2_medkit_gateway/discovery/layers/runtime_layer.hpp"
 #include "ros2_medkit_gateway/providers/introspection_provider.hpp"
 
 #include <array>
+#include <set>
 #include <type_traits>
 #include <unordered_map>
 #include <unordered_set>
```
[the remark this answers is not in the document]
bburda: Fixed in 851d373b. Added explicit `#include <algorithm>` for `std::remove_if`.

src/ros2_medkit_gateway/src/discovery/merge_pipeline.cpp
```diff
@@ -478,6 +480,53 @@ MergeResult MergePipeline::execute() {
...
+    });
+    result.components.erase(comp_it, result.components.end());
+
+    // Remove runtime areas whose namespace is covered
+    auto area_it = std::remove_if(result.areas.begin(), result.areas.end(), [&](const Area & area) {
+      if (area.source == "manifest") {
+        return false;
+      }
+      return manifest_area_ns.count(area.namespace_path) > 0 || linked_namespaces.count(area.namespace_path) > 0;
+    });
```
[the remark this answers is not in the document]
bburda: Fixed in 851d373b. Changed from `source == "manifest"` allowlist to explicit runtime source check: only `heuristic`, `topic`, and `synthetic` sources are suppressed. Plugin-provided entities are now preserved.

src/ros2_medkit_gateway/src/gateway_node.cpp
```diff
@@ -807,6 +809,34 @@ GatewayNode::GatewayNode(const rclcpp::NodeOptions & options) : Node("ros2_medki
...
+      // Try with normalized FQN (strip leading /)
+      if (!ros_fqn.empty() && ros_fqn[0] == '/') {
+        auto normalized = ros_fqn.substr(1);
+        auto it2 = mapping.find(normalized);
+        if (it2 != mapping.end()) {
+          return it2->second;
+        }
+      }
+      return "";
+    };
```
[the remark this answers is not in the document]
bburda: Fixed in 851d373b (same as the thread_safe_entity_cache comment). Lambda now calls `resolve_node_to_app(fqn)` - single lookup under shared_lock, no map copy.

src/ros2_medkit_gateway/src/models/thread_safe_entity_cache.cpp
bburda: Fixed in 851d373b. Added `resolve_node_to_app(fqn)` that does a single `find()` under shared_lock and returns the result string directly. No more per-event map copies.

src/ros2_medkit_gateway/src/log_manager.cpp
bburda: Fixed in 851d373b. Added 3 unit tests in LogManagerResolverTest covering all branches: slash-prefix match, bare name match, and empty-fallback to last segment.

src/ros2_medkit_gateway/src/discovery/hybrid_discovery.cpp
bburda: Fixed in 851d373b. Added `bool executed_{false}` flag to MergePipeline, set on first `execute()`. `add_layer()` now asserts `!executed_` to catch post-initialization calls at debug time.

src/ros2_medkit_gateway/src/log_manager.cpp
bburda: Fixed in 851d373b. Added `BufferCapDropsNewNodesWhenFull` test that verifies: cap reached -> new nodes silently dropped, existing buffers still accept entries.

src/ros2_medkit_gateway/src/discovery/merge_pipeline.cpp
bburda: Fixed in 851d373b. Added trailing slash strip before rfind in the namespace extraction loop. Defensive - ROS 2 normalizes FQNs but the guard is cheap insurance.

PR #323, opened 2026-03-29
fix(gateway): prevent configurations endpoint deadlock on nodes without parameter service

### LINKED ISSUE #318: Gateway deadlocks when querying configurations on nodes without parameter service
# Bug report

### Steps to reproduce
1. Start gateway in runtime_only mode on a system with micro-ROS nodes (e.g., Yahboom ROSMASTER M3 Pro)
2. Gateway discovers nodes like `YB_Node` (micro-ROS bridge) and `autostart_node` which don't have ROS 2 parameter service
3. Send GET request to `/api/v1/apps/YB_Node/configurations`
4. Request blocks for 4-6 seconds waiting for parameter service response
5. During that time, ALL other HTTP requests queue behind it (httplib is single-threaded)
6. If gateway queries its own node's configurations (`/apps/ros2_medkit_gateway/configurations`), it permanently deadlocks - it can't respond to its own parameter service call while the HTTP thread is blocked

### Expected behavior

Gateway should return quickly (< 0.5s) with empty configurations or 503 for nodes without parameter service. It should never deadlock.

### Actual behavior

- `GET /apps/YB_Node/configurations` blocks for 4-6 seconds, returns 503
- `GET /apps/ros2_medkit_gateway/configurations` causes permanent deadlock - gateway process alive but all HTTP endpoints unresponsive
- `GET /components/root/configurations` also causes permanent deadlock
- Once deadlocked, only `kill -9` recovers the gateway
- Cascading effect: any configurations request during the 4-6s block causes subsequent requests to timeout, amplifying into full gateway freeze

### Environment
- ros2_medkit version: main branch (commit ~00b243e)
- ROS 2 distro: Humble
- OS: Ubuntu 22.04 on Jetson Orin Nano (aarch64), L4T R36.4.3
- Robot: Yahboom ROSMASTER M3 Pro with micro-ROS nodes

### Additional information

This is the root cause of all gateway stability issues observed during M3 Pro testing. Web UI triggers this by loading entity detail which fetches configurations.

`parameter_service_timeout_sec` parameter exists (default 2.0s) but reducing it to 0.5s did not help - the blocking still occurs.

Possible fixes:
- Make parameter service calls non-blocking (async with callback, don't block httplib thread)
- Cache negative results - if a node doesn't respond to parameter service once, don't try again for N seconds
- Detect self-query deadlock - never query gateway's own node configurations via parameter service
- Use a thread pool for parameter service calls so they don't block the HTTP server thread

mfaferek93:
## Summary

Prevent gateway HTTP deadlock when configurations endpoint queries nodes without ROS 2 parameter service (e.g., micro-ROS bridges, external nodes).

**Root cause:** ConfigurationManager used a single param_operations_mutex_ held for the entire parameter operation including blocking wait_for_service() calls. One slow/unavailable node blocked all configuration requests. Self-queries (gateway querying its own parameters) caused permanent deadlock.

**Solution:**
- **Negative cache** - after a node parameter service fails to respond, subsequent requests return instantly with SERVICE_UNAVAILABLE for a configurable TTL (default 60s). Hard cap at 500 entries with oldest-entry eviction.
- **Self-query guard** - gateway uses direct node_->get_parameters() for its own node instead of IPC. Covers all 5 operations (list, get, set, reset, reset_all).
- **Narrower spin_mutex_ scope** - serializes only ROS 2 IPC calls, not JSON building or cache lookups. Single pre-created param_node_ for fast DDS service discovery.
- **Idempotent shutdown** with atomic flag + post-shutdown guard on all public methods
- **Parameter validation** via FloatingPointRange descriptors
- **Explicit config_mgr_->shutdown()** in GatewayNode destructor

## Test plan

- [x] 22 unit tests pass (including new: negative cache, self-query, cross-method cache, concurrent)
- [x] 58 integration tests pass (100%) including test_scenario_discovery_manifest
- [x] Tested on Jetson Orin Nano (aarch64) with ROSMASTER M3 Pro
- [x] Self-query returns 200 in 0.4s (was: permanent deadlock)
- [x] Micro-ROS nodes return instant SERVICE_UNAVAILABLE (was: 4-6s block + gateway freeze)

Fixes #318

### DIFF
--- a/src/ros2_medkit_gateway/src/configuration_manager.cpp
+++ b/src/ros2_medkit_gateway/src/configuration_manager.cpp
@@ -228,6 +228,9 @@ ParameterResult ConfigurationManager::list_parameters(const std::string & node_n
     {
       std::lock_guard<std::mutex> spin_lock(spin_mutex_);
 
+      // Cache default values first (gives node extra time for DDS service discovery)
+      cache_default_values(node_name);
+
       if (!client->wait_for_service(get_service_timeout())) {
         result.success = false;
         result.error_message = "Parameter service not available for node: " + node_name;
@@ -237,8 +240,6 @@ ParameterResult ConfigurationManager::list_parameters(const std::string & node_n
                     node_name.c_str(), negative_cache_ttl_sec_);
         return result;
       }
-
-      cache_default_values(node_name);
 
       auto param_names = client->list_parameters({}, 0);
       parameters = client->get_parameters(param_names.names);


src/ros2_medkit_gateway/src/configuration_manager.cpp
```diff
@@ -14,95 +14,189 @@
...
-    RCLCPP_DEBUG(node_->get_logger(), "Got param client for node: '%s'", node_name.c_str());
-
     if (!client->wait_for_service(get_service_timeout())) {
       result.success = false;
       result.error_message = "Parameter service not available for node: " + node_name;
       result.error_code = ParameterErrorCode::SERVICE_UNAVAILABLE;
-      RCLCPP_WARN(node_->get_logger(), "Parameter service not available for node: '%s'", node_name.c_str());
+      mark_node_unavailable(node_name);
+      RCLCPP_WARN(node_->get_logger(), "Parameter service not available for node: '%s' (cached for %.0fs)",
+                  node_name.c_str(), negative_cache_ttl_sec_);
```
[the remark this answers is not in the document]
mfaferek93: fixed

src/ros2_medkit_gateway/include/ros2_medkit_gateway/configuration_manager.hpp
```diff
@@ -53,6 +56,11 @@ struct ParameterResult {
 class ConfigurationManager {
  public:
   explicit ConfigurationManager(rclcpp::Node * node);
+  ~ConfigurationManager();
+
+  /// Clean up thread-local param nodes before ROS 2 context shutdown.
```
bburda: This docstring says "thread-local param nodes" but the implementation uses a single shared `param_node_` protected by `spin_mutex_`. No `thread_local` storage anywhere in the final code.

Same issue on line 96 (`get_param_client` docstring): "Creates a per-thread param node to avoid executor conflicts" - also not what the code does.

Suggest:
```cpp
/// Clean up shared param node and cached clients before ROS 2 context shutdown.
```
and for `get_param_client`:
```cpp
/// Get or create a cached SyncParametersClient for the given node.
```

src/ros2_medkit_gateway/include/ros2_medkit_gateway/configuration_manager.hpp
```diff
@@ -104,28 +129,41 @@ class ConfigurationManager {
...
 
-  /// Mutex to serialize all parameter operations.
-  /// SyncParametersClient internally spins param_node_ via spin_node_until_future_complete(),
-  /// which is not thread-safe. Concurrent HTTP requests would cause:
-  /// "Node '/_param_client_node' has already been added to an executor"
-  mutable std::mutex param_operations_mutex_;
+  /// Internal node for parameter client operations.
+  /// Created once in constructor - must be in DDS graph early for fast service discovery.
+  /// SyncParametersClient requires a node NOT in an executor to spin internally.
+  std::shared_ptr<rclcpp::Node> param_node_;
```
bburda: `spin_mutex_` (line 162) protects spin operations on `param_node_` (line 154), but it's declared after it. C++ destroys members in reverse declaration order, so `spin_mutex_` would be destroyed before `param_node_`.

In practice, the destructor's `shutdown()` call mitigates this (it acquires the mutex and resets `param_node_` to null before member destruction begins), so it's not a crash risk. But declaring the mutex first makes the intent clearer:
```cpp
mutable std::mutex spin_mutex_;
```

src/ros2_medkit_gateway/src/configuration_manager.cpp
```diff
@@ -21,28 +21,140 @@ using namespace std::chrono_literals;
...
   options.start_parameter_services(false);
   options.start_parameter_event_publisher(false);
-  // Don't inherit node name/namespace from global arguments (--ros-args -r __node:=...)
-  // Without this, the internal node would have the same name as the main gateway node
   options.use_global_arguments(false);
   param_node_ = std::make_shared<rclcpp::Node>("_param_client_node", options);
 
   // Get configurable timeout for parameter services (default 2.0 seconds)
+  // For systems with micro-ROS nodes without parameter service, set lower (e.g., 0.5s)
   service_timeout_sec_ = node_->declare_parameter("parameter_service_timeout_sec", 2.0);
```
bburda: The docs specify valid ranges (`0.1-10.0` and `0-3600`) but these aren't enforced at the ROS 2 parameter layer. A user can set `parameter_service_timeout_sec: 0` or a negative value without any validation error.

Consider using `rcl_interfaces::msg::FloatingPointRange` in a `ParameterDescriptor`:
```cpp
rcl_interfaces::msg::ParameterDescriptor timeout_desc;
timeout_desc.description = "Timeout for parameter service calls";
rcl_interfaces::msg::FloatingPointRange range;
range.from_value = 0.1;
range.to_value = 10.0;
timeout_desc.floating_point_range.push_back(range);
service_timeout_sec_ = node_->declare_parameter("parameter_service_timeout_sec", 2.0, timeout_desc);
```

src/ros2_medkit_gateway/src/configuration_manager.cpp
```diff
@@ -21,28 +21,140 @@ using namespace std::chrono_literals;
...
+
+  RCLCPP_INFO(node_->get_logger(), "ConfigurationManager initialized (timeout=%.1fs, negative_cache=%.0fs)",
+              service_timeout_sec_, negative_cache_ttl_sec_);
+}
+
+ConfigurationManager::~ConfigurationManager() {
+  shutdown();
+}
+
+void ConfigurationManager::shutdown() {
```
bburda, line 60: Two concerns with `shutdown()`:

1. **No idempotency guard.** It's called from the destructor, and could also be called explicitly (the docstring says "must be called before rclcpp::shutdown()"). The second call re-acquires both locks to clear already-empty collections. Consider:
```cpp
std::atomic<bool> shutdown_{false};

void ConfigurationManager::shutdown() {
  if (shutdown_.exchange(true)) return;
  // ...
}
```

2. **No post-shutdown guard on public methods.** After `shutdown()` resets `param_node_`, a concurrent or subsequent call to `get_param_client()` would create `SyncParametersClient` on a null node. With the atomic flag above, public methods could return a clean error:
```cpp
if (shutdown_.load()) {
  return {false, {}, "ConfigurationManager is shut down", ParameterErrorCode::INTERNAL_ERROR};
}
```

src/ros2_medkit_gateway/src/configuration_manager.cpp
```diff
@@ -21,28 +21,140 @@ using namespace std::chrono_literals;
...
+  auto elapsed = std::chrono::steady_clock::now() - it->second;
+  return elapsed < std::chrono::duration<double>(negative_cache_ttl_sec_);
+}
+
+void ConfigurationManager::mark_node_unavailable(const std::string & node_name) {
+  std::unique_lock<std::shared_mutex> lock(negative_cache_mutex_);
+  unavailable_nodes_[node_name] = std::chrono::steady_clock::now();
+
+  // Lazy cleanup: remove expired entries to prevent unbounded growth
+  if (unavailable_nodes_.size() > 100) {
```
bburda: The lazy cleanup only removes *expired* entries. With `parameter_service_negative_cache_sec` up to 3600s, all entries within the TTL window are fresh and survive cleanup - the map keeps growing.

Consider adding a hard cap (e.g., 500) that evicts the oldest entry regardless of TTL:
```cpp
if (unavailable_nodes_.size() > kMaxNegativeCacheSize) {
  auto oldest = std::min_element(unavailable_nodes_.begin(), unavailable_nodes_.end(),
    [](const auto& a, const auto& b) { return a.second < b.second; });
  unavailable_nodes_.erase(oldest);
}
```

The same applies to `param_clients_` and `default_values_` maps - they create one entry per unique node name and never evict (only cleared in `shutdown()`).

src/ros2_medkit_gateway/src/configuration_manager.cpp
```diff
@@ -198,7 +320,42 @@ ParameterResult ConfigurationManager::get_parameter(const std::string & node_nam
 
 ParameterResult ConfigurationManager::set_parameter(const std::string & node_name, const std::string & param_name,
                                                     const json & value) {
-  std::lock_guard<std::mutex> op_lock(param_operations_mutex_);
+  // Self-query guard: set parameter directly on own node
+  if (is_self_node(node_name)) {
+    ParameterResult result;
+    try {
+      auto current_value = node_->get_parameter(param_name).get_parameter_value();
```
bburda, line 374: The self-query `set_parameter` path calls `node_->get_parameter(param_name)` without checking if the parameter exists. If it doesn't, ROS 2 throws `ParameterNotDeclaredException`, caught by the generic `catch` as `INTERNAL_ERROR` (HTTP 500).

Meanwhile, `get_own_parameter()` (line 132) properly checks `has_parameter()` first and returns `NOT_FOUND` (HTTP 404). This makes the API inconsistent: GET returns 404 for a missing param, PUT returns 500.

Suggest adding the same guard:
```cpp
if (!node_->has_parameter(param_name)) {
  result.success = false;
  result.error_message = "Parameter not found: " + param_name;
  result.error_code = ParameterErrorCode::NOT_FOUND;
  return result;
}
```

src/ros2_medkit_gateway/src/configuration_manager.cpp
```diff
@@ -482,14 +624,55 @@ void ConfigurationManager::cache_default_values(const std::string & node_name) {
 }
 
 ParameterResult ConfigurationManager::reset_parameter(const std::string & node_name, const std::string & param_name) {
-  std::lock_guard<std::mutex> op_lock(param_operations_mutex_);
+  // Self-query guard: reset via direct access
+  if (is_self_node(node_name)) {
+    ParameterResult result;
+    std::lock_guard<std::mutex> lock(defaults_mutex_);
```
bburda: `defaults_mutex_` is held for the entire self-query block, including the `node_->set_parameter()` call at line 645. The IPC path (below, line ~700) carefully copies the default value under lock, then releases before doing I/O:
```cpp
default_param = param_it->second;
}  // lock released
auto client = get_param_client(node_name);
```

The self-query path should follow the same pattern. `node_->set_parameter()` is local and fast, so this isn't a production risk right now, but if a parameter change callback ever does anything non-trivial (e.g., logging that acquires another lock), the held mutex blocks all default value lookups.

src/ros2_medkit_gateway/test/test_configuration_manager.cpp
```diff
@@ -316,6 +316,48 @@ TEST_F(TestConfigurationManager, test_error_messages_informative) {
...
+// ==================== SPIN SERIALIZATION TESTS ====================
+
+TEST_F(TestConfigurationManager, test_concurrent_queries_no_crash) {
+  // Concurrent queries to different nonexistent nodes must not crash
+  // (spin_mutex_ serializes ROS 2 IPC to prevent executor conflicts).
+  // Second+ calls hit negative cache and return instantly.
+  std::vector<std::thread> threads;
+  for (int i = 0; i < 3; ++i) {
+    threads.emplace_back([this, i]() {
+      (void)config_manager_->list_parameters("/concurrent_node_" + std::to_string(i));
```
bburda: Results are discarded with `(void)`. Since the test is named `no_crash`, survival is the minimum bar, but verifying the results would catch silent failures:
```cpp
std::vector<ParameterResult> results(3);
std::vector<std::thread> threads;
for (int i = 0; i < 3; ++i) {
  threads.emplace_back([this, i, &results]() {
    results[i] = config_manager_->list_parameters("/concurrent_node_" + std::to_string(i));
  });
}
for (auto & t : threads) t.join();
for (int i = 0; i < 3; ++i) {
  EXPECT_FALSE(results[i].success);
  EXPECT_EQ(results[i].error_code, ParameterErrorCode::SERVICE_UNAVAILABLE);
}
```

src/ros2_medkit_gateway/include/ros2_medkit_gateway/configuration_manager.hpp
mfaferek93: Fixed. Docstrings updated to match actual implementation.

src/ros2_medkit_gateway/include/ros2_medkit_gateway/configuration_manager.hpp
mfaferek93: Fixed. spin_mutex_ now declared before param_node_.

src/ros2_medkit_gateway/src/configuration_manager.cpp
mfaferek93: Fixed. Added FloatingPointRange descriptors for both parameters.

src/ros2_medkit_gateway/src/configuration_manager.cpp
mfaferek93: Fixed both. Atomic shutdown_ flag for idempotency + SHUT_DOWN error code guard on all public methods.

src/ros2_medkit_gateway/src/configuration_manager.cpp
mfaferek93: Fixed. Hard cap (500) with oldest-entry eviction via min_element.

src/ros2_medkit_gateway/src/configuration_manager.cpp
mfaferek93: Fixed. Added has_parameter() check - returns NOT_FOUND (404) consistently.

src/ros2_medkit_gateway/src/configuration_manager.cpp
mfaferek93: Fixed. Copy default_param under lock, release before set_parameter().

src/ros2_medkit_gateway/test/test_configuration_manager.cpp
mfaferek93: Fixed. Results verified (SERVICE_UNAVAILABLE). Added cross-method cache test.

PR #327, opened 2026-03-30
feat: SOVD entity model alignment and multi-instance peer aggregation

### LINKED ISSUE #316: feat(gateway): multi-instance entity aggregation with peer discovery
## Summary

Support for aggregating entities from multiple gateway instances running on the same device. One gateway acts as the primary endpoint, forwarding requests to peer gateways and merging their responses into a unified view.

This enables scenarios where multiple gateway instances manage different subsystems on the same hardware (e.g., separate partitions, containers, or processes) while presenting a single API surface to clients.

### Peer discovery (two modes)

**1. Static configuration (gateway_params.yaml):**
```yaml
aggregation:
  enabled: true
  peers:
    - url: "http://localhost:8081"
      name: "subsystem_b"
      health_check_interval_sec: 10
```

**2. mDNS auto-discovery:**
- Each gateway announces itself as `_medkit._tcp.local` via mDNS
- Peers auto-discovered on the local network/device
- Static config takes precedence (override)
- Lightweight implementation via single-header mdns library (cross-platform)

### Aggregation behavior

**Discovery merging:**
- Primary gateway merges its own entity tree with peer entity trees
- `/areas`, `/components`, `/apps`, `/functions` return combined results
- Each entity tagged with source peer in `x-medkit` metadata
- Routing table maps `entity_id -> peer_url` for request forwarding

**Request routing:**
- If requested entity is local, handle normally
- If entity belongs to a peer, forward the request via HTTP and return the response
- Entity ID collisions across peers resolved by prefixing with peer name

**Global endpoint fan-out:**
- `/faults` - merge faults from all peers
- `/updates` - merge updates from all peers
- `/logs` - merge logs from all peers
- `/plugins` - merge plugin lists from all peers
- `/scripts` - merge scripts from all peers

**Health monitoring:**
- Periodic health check on each peer
- Peer down: mark its entities as offline, return partial results
- Peer recovery: re-discover and re-merge entities

### Implementation scope

**New files:**
- `src/aggregation/peer_client.cpp` - HTTP client for peer communication
- `src/aggregation/aggregation_manager.cpp` - merge logic, routing table, health monitoring
- `src/aggregation/mdns_discovery.cpp` - mDNS announce + browse

**Modified files:**
- `gateway_node.cpp` - initialize aggregation manager if enabled
- `rest_server.cpp` - intercept requests for remote entities
- `discovery_manager.cpp` - merge peer entities into cache

## Additional context

- HTTP between peers initially; transport could be swapped later (gRPC, shared memory, etc.)
- mDNS library candidate: [mjansson/mdns](https://github.com/mjansson/mdns) (MIT, header-only, cross-platform)
- Aggregation is opt-in (disabled by default, zero overhead when off)
- Fleet-level aggregation across devices is a separate concern - this is device-level onl

### LINKED ISSUE #320: Relationship endpoints return 404 in runtime_only mode
# Bug report

### Steps to reproduce
1. Start gateway in runtime_only mode
2. `GET /api/v1/areas/root/related-components` returns 404
3. `GET /api/v1/components/root/related-apps` returns 404
4. `GET /api/v1/areas/{any}/hosts` returns 404

### Expected behavior

Relationship endpoints should return the connected entities. Areas should list their components via `related-components`. Components should list their apps via `related-apps`.

### Actual behavior

All relationship navigation endpoints return 404:
```
GET /areas/root/related-components -> 404
GET /areas/imu/related-components  -> 404
GET /areas/joy/related-components  -> 404
GET /components/root/related-apps  -> 404
GET /areas/root/hosts              -> 404
```

`is-located-on` works for apps (200), so the relationship exists in one direction but not the reverse.

### Environment
- ros2_medkit version: main branch
- ROS 2 distro: Humble
- OS: Ubuntu 22.04 on Jetson Orin Nano (aarch64)

### Additional information

This breaks SOVD entity tree navigation. Client can't walk the hierarchy from area -> component -> app via standard SOVD relationship endpoints.

May be specific to runtime_only mode - in hybrid mode with manifest these might work since relationships are explicitly defined.

### LINKED ISSUE #325: Rethink runtime discovery entity mapping to align with SOVD entity model
## Summary

Runtime discovery maps ROS 2 namespaces to the wrong SOVD entity types. Namespaces become Areas and Components, but per the SOVD spec (ISO 17978-3, Clause 5) they should map to Functions.

## Problem

Current mapping:
- ROS 2 namespace -> **Area** (wrong - Area = logical grouping like domains/zones, deployment-specific)
- ROS 2 namespace -> **Component** (wrong - Component = HW or SW platform like ECU, HPC, OS)
- ROS 2 node -> **App** (correct)
- Topic prefix -> **Area/Component** (wrong - same issue)

Correct SOVD semantics:
- **Area** - deployment topology (domains, zones, SDV architecture). Not inferrable from ROS 2 graph. For small robots with 1-3 ECUs, areas are just noise. Areas make sense at fleet/building level (e.g., floor_1, zone_a) not per-device.
- **Component** - execution platform. Should be extracted from the actual host system (OS, hostname, architecture), not from namespace groupings. A single-device deployment has one Component (the OS) with all Apps running on it. Multi-ECU setups define Components in manifest.
- **Function** - "Functional view spanning multiple components." THIS is where namespace grouping fits: `/diagnostics/gateway` + `/diagnostics/fault_manager` -> Function "diagnostics".
- **App** - ROS 2 node from graph introspection (not topic-based discovery).

## Proposed changes

1. **Deprecate synthetic areas** - `create_synthetic_areas` defaults to false, eventually removed. Areas only from manifest. Empty areas collection is valid for single-device deployments.

2. **Change Component discovery** - instead of per-namespace synthetic components, extract actual host info (OS, hostname). Single default Component for runtime_only mode. Manifest defines Component hierarchy for multi-ECU setups. Subcomponents model SW-on-HW (e.g., Component: Nvidia Orin -> Subcomponent: Ubuntu 24.04 -> Apps).

3. **Map namespaces to Functions** - repurpose existing namespace grouping logic to create Functions. `/sensors/*` nodes -> Function "sensors". Works in both runtime_only and hybrid modes.

4. **Remove topic-based entity creation** - `topic_only_policy` should not create Areas or Components from topic prefixes. Apps come from ROS 2 node graph only.

5. **Simplify config** - most gap-fill flags (`allow_heuristic_areas`, `allow_heuristic_components`, `create_synthetic_areas`, `topic_only_policy`) become unnecessary. Currently 5+ flags needed to suppress noise in hybrid mode with a complete manifest.

## Context

Found during selfpatch_demos smoke test work (issues #307, #311, #312) where synthetic entities leaked into API responses. The root cause is incorrect entity mapping, not suppression logic.

bburda:
# Pull Request

## Summary

Aligns the runtime discovery entity model with the SOVD spec (ISO 17978-3) and adds multi-instance peer aggregation with mDNS auto-discovery.

**Entity model changes:**
- Namespaces now create Function entities (not Areas/Components) per SOVD entity semantics
- Single default Component from host system info (hostname, OS, arch) via HostInfoProvider
- Synthetic Areas deprecated (default off) - Areas only from manifest
- Functions and Areas support faults, logs, data collections (x-medkit extension, aggregated from child entities)

**Multi-instance aggregation:**
- Static peer config (`peer_urls`/`peer_names`) and mDNS auto-discovery (opt-in)
- Entity merge: Area/Function merge by ID, Component/App prefix with `__` on collision
- Transparent request forwarding via `validate_entity_for_route()` - zero handler changes
- Global fan-out for `/faults` with partial results on peer failure
- Peer status in `/health` endpoint
- StreamProxy abstraction (SSE implementation) for future streaming proxy

**Security hardening:**
- Thread-safe aggregation (shared_mutex, per-operation client mutex)
- mDNS disabled by default, self-discovery filter, peer name sanitization
- Peer URL validation (blocks loopback, link-local, cloud metadata for mDNS peers)
- Response body limits (10MB), SSE buffer limit (1MB), entity count limit (10000/peer)
- Response header allowlist, collision prefix stripping on forwarded paths

---

## Issue

- closes #316
- closes #320
- closes #325

---

## Type

- [ ] Bug fix
- [x] New feature or tests
- [x] Breaking change
- [x] Documentation only

---

## Testing

**Unit tests:** 2296 tests, 0 failures
- `test_host_info_provider` (16), `test_runtime_discovery` (11), `test_function_resource_collections` (38)
- `test_peer_client` (23 - mock server), `test_entity_merger` (21), `test_aggregation_manager` (26)
- `test_mdns_discovery` (15), `test_stream_proxy` (18 - mock server)

**Integration tests:**
- `test_entity_model_runtime.test.py` - namespace -> Function, default Component, App linking
- `test_peer_aggregation.test.py` - two-gateway setup, merged entities, forwarding, peer health
- All existing integration tests updated and passing

**Breaking changes:**
- `/areas` returns empty in `runtime_only` mode (was namespace-based synthetic areas)
- `/components` returns single host component (was per-namespace synthetic components)
- `/functions` now populated from namespaces (was empty)
- Migration guide added to `docs/config/aggregation.rst`

---

## Checklist

### DIFF
--- a/src/ros2_medkit_gateway/src/discovery/runtime_discovery.cpp
+++ b/src/ros2_medkit_gateway/src/discovery/runtime_discovery.cpp
@@ -130,7 +130,9 @@ std::vector<App> RuntimeDiscoveryStrategy::discover_apps() {
     if (name_namespaces[name].size() > 1 && ns != "/") {
       std::string ns_prefix = ns.substr(1);  // Remove leading '/'
       std::replace(ns_prefix.begin(), ns_prefix.end(), '/', '_');
-      app.id = ns_prefix + "_" + name;
+      app.id = ns_prefix;
+      app.id += "_";
+      app.id += name;
       RCLCPP_DEBUG(node_->get_logger(), "Name collision detected for '%s', using namespaced ID '%s'", name.c_str(),
                    app.id.c_str());
     } else {


src/ros2_medkit_gateway/README.md
```diff
@@ -1460,15 +1455,20 @@ In addition to standard ROS 2 node discovery, the gateway supports **topic-based
...
 
 ```
-/powertrain/engine/temp_sensor  → Area: powertrain, Component: temp_sensor
-/chassis/brakes/pressure_sensor → Area: chassis, Component: pressure_sensor
-/body/lights/controller         → Area: body, Component: controller
-/standalone_node                → Area: root, Component: standalone_node
+/powertrain/engine/temp_sensor  -> Area: powertrain, Function: engine, App: temp_sensor
+/chassis/brakes/pressure_sensor -> Area: chassis, Function: brakes, App: pressure_sensor
+/body/lights/controller         -> Area: body, Function: lights, App: controller
+/standalone_node                -> Area: root, Function: root, App: standalone_node
```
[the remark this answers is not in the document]
bburda: Fixed - Areas description updated to manifest-only in commits 2396ff6b and b5ea656e.

src/ros2_medkit_gateway/README.md
```diff
@@ -1460,15 +1455,20 @@ In addition to standard ROS 2 node discovery, the gateway supports **topic-based
...
-### Area Organization
+### Entity Organization
 
-The gateway organizes nodes into "areas" based on their namespace:
+In runtime discovery mode, the gateway maps the ROS 2 graph to the SOVD entity model:
+
+- **Component**: A single host-derived Component is created from `HostInfoProvider` (hostname, OS, architecture). All Apps belong to this Component.
+- **App**: Each discovered ROS 2 node becomes an App entity.
+- **Function**: Namespace prefixes create Function entities that group Apps sharing a namespace (e.g., `/powertrain/engine/temp_sensor` and `/powertrain/engine/rpm_sensor` both belong to Function `engine`).
+- **Area**: Top-level namespace segments create Area entities that organize Functions (e.g., `/powertrain/engine/*` nodes create Area `powertrain` containing Function `engine`).
```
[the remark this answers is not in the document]
bburda: Fixed in 2396ff6b - README now states Areas are manifest-only.

src/ros2_medkit_gateway/src/aggregation/aggregation_manager.cpp
```diff
@@ -0,0 +1,314 @@
...
+                    max_entities_per_peer);
+      }
+      continue;
+    }
+
+    EntityMerger merger(peer->name());
+    merged.areas = merger.merge_areas(merged.areas, result->areas);
+    merged.functions = merger.merge_functions(merged.functions, result->functions);
+    merged.components = merger.merge_components(merged.components, result->components);
+    merged.apps = merger.merge_apps(merged.apps, result->apps);
```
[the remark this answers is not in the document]
bburda: Fixed in 9ffa3b7b - merge_apps now runs before merge_functions, and Function hosts are remapped using the collision rename table.

src/ros2_medkit_gateway/src/aggregation/entity_merger.cpp
```diff
@@ -0,0 +1,199 @@
...
+  }
+
+  for (const auto & remote_func : remote) {
+    auto it = local_index.find(remote_func.id);
+    if (it != local_index.end()) {
+      // Collision: merge by combining hosts lists
+      auto & merged = result[it->second];
+
+      std::unordered_set<std::string> host_set(merged.hosts.begin(), merged.hosts.end());
+      for (const auto & host : remote_func.hosts) {
```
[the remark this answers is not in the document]
bburda: Fixed in 9ffa3b7b - Function hosts remapped after App collision prefixing.

src/ros2_medkit_gateway/src/aggregation/entity_merger.cpp
```diff
@@ -0,0 +1,202 @@
...
+        if (host_set.insert(host).second) {
+          merged.hosts.push_back(host);
+        }
```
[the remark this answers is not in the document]
bburda: Fixed in 9ffa3b7b - app_id_remap built from routing table, Function hosts updated before merge.

src/ros2_medkit_gateway/src/aggregation/aggregation_manager.cpp
```diff
@@ -0,0 +1,314 @@
...
+  for (auto & peer : peers_) {
+    peer->check_health();
+  }
+}
+
+std::vector<PeerClient *> AggregationManager::healthy_peers() {
+  std::shared_lock<std::shared_mutex> lock(mutex_);
+
+  std::vector<PeerClient *> result;
+  for (auto & peer : peers_) {
```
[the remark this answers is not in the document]
bburda: Fixed - healthy_peers() replaced with healthy_peer_count(). peers_ now uses shared_ptr with snapshot pattern, no raw pointer exposure.

src/ros2_medkit_gateway/src/aggregation/aggregation_manager.cpp
```diff
@@ -0,0 +1,314 @@
...
+  // Strip peer prefix from entity ID in the path if present.
+  // When entity ID collision causes renaming (e.g., camera_driver -> peer_b__camera_driver),
+  // the peer only knows the entity by its original ID (camera_driver), so we must strip
+  // the prefix before forwarding.
+  std::string forwarded_path = req.path;
+  std::string prefix = peer_name + EntityMerger::SEPARATOR;
+  auto prefix_pos = forwarded_path.find(prefix);
+  if (prefix_pos != std::string::npos) {
+    forwarded_path.erase(prefix_pos, prefix.size());
+  }
```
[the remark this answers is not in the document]
bburda: Fixed in 2396ff6b - prefix stripping now anchored to path segment boundaries (checks preceding '/').

docs/tutorials/multi-instance.rst
```diff
@@ -0,0 +1,314 @@
...
+
+.. code-block:: json
+
+   {
+     "items": [
+       {"id": "robot", "source": "runtime"},
+       {"id": "subsystem_b__robot", "source": "peer:subsystem_b"}
+     ]
+   }
+
```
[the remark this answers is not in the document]
bburda: Fixed in 2396ff6b - tutorial now correctly states Components merge by ID, only Apps get prefixed.

src/ros2_medkit_gateway/CMakeLists.txt
```diff
@@ -316,6 +329,7 @@ if(BUILD_TESTING)
     "test/*.cpp" "test/*.h" "test/*.hpp"
   )
   list(FILTER _format_files EXCLUDE REGEX ".*/vendored/.*")
+  list(FILTER _format_files EXCLUDE REGEX ".*/vendored/.*")
```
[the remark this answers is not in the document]
bburda: Fixed - removing duplicate filter line.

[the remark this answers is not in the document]
bburda: Fixed - removing duplicate filter line.

src/ros2_medkit_gateway/src/aggregation/peer_client.cpp
```diff
@@ -0,0 +1,477 @@
...
+  // Copy response from peer - only forward safe headers
+  static const std::set<std::string> allowed_headers = {"content-type", "etag", "cache-control", "last-modified"};
+
+  res.status = result->status;
+  res.body = result->body;
+  for (const auto & header : result->headers) {
+    std::string lower_name = header.first;
+    std::transform(lower_name.begin(), lower_name.end(), lower_name.begin(), ::tolower);
+    if (allowed_headers.count(lower_name) > 0 || lower_name.find("x-medkit") == 0) {
+      res.set_header(header.first, header.second);
```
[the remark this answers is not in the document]
bburda: Fixed - using lambda with unsigned char cast.

src/ros2_medkit_gateway/src/aggregation/peer_client.cpp
```diff
@@ -0,0 +1,400 @@
...
+  // Forward Authorization header if present
+  if (req.has_header("Authorization")) {
+    headers.emplace("Authorization", req.get_header_value("Authorization"));
+  }
+
+  httplib::Result result{nullptr, httplib::Error::Unknown};
+  const std::string & path = req.path;
+  const std::string content_type = req.get_header_value("Content-Type");
+
+  if (req.method == "GET") {
```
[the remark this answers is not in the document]
bburda: Fixed - forwarded request now includes query parameters.

src/ros2_medkit_gateway/src/aggregation/aggregation_manager.cpp
```diff
@@ -0,0 +1,314 @@
...
+    error_body["message"] = "Peer '" + peer_name + "' is not known to this gateway";
+    res.set_content(error_body.dump(), "application/json");
+    return;
+  }
+
+  // Strip peer prefix from entity ID in the path if present.
+  // When entity ID collision causes renaming (e.g., camera_driver -> peer_b__camera_driver),
+  // the peer only knows the entity by its original ID (camera_driver), so we must strip
+  // the prefix before forwarding.
+  std::string forwarded_path = req.path;
```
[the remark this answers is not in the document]
bburda: Fixed - query string now forwarded with the request.

src/ros2_medkit_gateway/src/aggregation/peer_client.cpp
```diff
@@ -0,0 +1,400 @@
...
+    func.description = xm.value("description", "");
+  }
+  if (j.contains("translationId")) {
+    func.translation_id = j["translationId"].get<std::string>();
+  }
+  if (j.contains("tags") && j["tags"].is_array()) {
+    func.tags = j["tags"].get<std::vector<std::string>>();
+  }
+  return func;
+}
```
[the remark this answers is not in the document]
bburda: Entity IDs are now validated before URL construction (9ffa3b7b). Function hosts are remapped after App collision prefixing. The fetch_entities detail requests populate the full entity data including hosts.

src/ros2_medkit_gateway/src/aggregation/entity_merger.cpp
```diff
@@ -0,0 +1,182 @@
...
+    added.source = peer_source();
+
+    if (local_ids.count(remote_app.id) > 0) {
+      // Collision: prefix the remote entity ID
+      added.id = prefix_id(remote_app.id);
+      added.name = peer_name_ + SEPARATOR + remote_app.name;
+    }
+
+    routing_table_[added.id] = peer_name_;
+    result.push_back(added);
```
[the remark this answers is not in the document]
bburda: App.component_id is remapped to peer_name by design - the peer Component is merged by ID (not prefixed). Function hosts are now remapped after App collision prefixing (9ffa3b7b).

src/ros2_medkit_gateway/src/aggregation/entity_merger.cpp
```diff
@@ -0,0 +1,202 @@
...
+    local_ids.insert(app.id);
+  }
+
+  for (const auto & remote_app : remote) {
+    App added = remote_app;
+    added.source = peer_source();
+    // Remap component_id to peer name - in the aggregation model, the peer
+    // IS a component/subcomponent and all its apps belong to it.
+    added.component_id = peer_name_;
+
```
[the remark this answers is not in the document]
bburda: By design - remote Apps are assigned to the peer-named Component because the remote Component merges by ID with the local one. The peer name acts as the aggregation-level component grouping.

src/ros2_medkit_gateway/src/aggregation/entity_merger.cpp
```diff
@@ -0,0 +1,202 @@
...
+      // Merged components do NOT go into routing table - they are combined local+remote
+    } else {
+      // No collision: add remote component with source tagged
+      Component added = remote_comp;
+      added.source = peer_source();
+      result.push_back(added);
+
+      // Remote-only components get a routing entry
+      routing_table_[added.id] = peer_name_;
+    }
```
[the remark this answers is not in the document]
bburda: By design - Component merge-by-ID combines both local and remote metadata. Remote-specific capabilities are accessible through the merged component. This matches SOVD's component-as-physical-unit model.

src/ros2_medkit_gateway/src/aggregation/sse_stream_proxy.cpp
```diff
@@ -0,0 +1,207 @@
...
+        if (buffer.size() > kMaxSSEBufferSize) {
+          return false;  // Disconnect - peer sending malformed stream
+        }
+
+        // Process complete events (delimited by double newline)
+        size_t pos = 0;
+        while (true) {
+          auto boundary = buffer.find("\n\n", pos);
+          if (boundary == std::string::npos) {
+            break;
```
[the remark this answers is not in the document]
bburda: Our SSE peers are ros2_medkit gateways using the same cpp-httplib which sends LF line endings. CRLF handling could be added if we need to proxy non-httplib SSE sources in the future.

src/ros2_medkit_gateway/src/aggregation/peer_client.cpp
```diff
@@ -0,0 +1,477 @@
...
+    result = client_->Get(path, headers);
+  } else if (req.method == "POST") {
+    result = client_->Post(path, headers, req.body, content_type);
+  } else if (req.method == "PUT") {
+    result = client_->Put(path, headers, req.body, content_type);
+  } else if (req.method == "DELETE") {
+    result = client_->Delete(path, headers);
+  } else if (req.method == "PATCH") {
+    result = client_->Patch(path, headers, req.body, content_type);
+  }
```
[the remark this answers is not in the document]
bburda: SOVD spec only defines GET/POST/PUT/DELETE/PATCH. Unsupported methods returning 502 is acceptable - the request would also fail at the peer. No action needed.

src/ros2_medkit_integration_tests/test/features/test_peer_aggregation.test.py
```diff
@@ -0,0 +1,325 @@
...
+                        f'  {label}: waiting for {missing} '
+                        f'(have {found_ids})'
+                    )
+            except requests.exceptions.RequestException:
+                pass
+            time.sleep(1.0)
+        raise AssertionError(
+            f'{label}: apps not discovered after {DISCOVERY_TIMEOUT}s. '
+            f'Missing: {required_apps}'
+        )
```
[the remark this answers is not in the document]
bburda: The peer aggregation test uses a two-gateway setup with distinct DDS domains, which requires custom launch logic beyond what GatewayTestCase provides. The polling pattern is consistent with other tests.

src/ros2_medkit_gateway/src/aggregation/peer_client.cpp
```diff
@@ -0,0 +1,400 @@
...
+    for (auto & app : entities.apps) {
+      app.source = peer_source;
+    }
+  }
+
+  // Fetch functions
+  {
+    auto result = cli.Get(std::string(API_PREFIX) + "/functions");
+    if (!result) {
+      return tl::unexpected<std::string>("Failed to connect to peer '" + name_ + "' at " + url_);
```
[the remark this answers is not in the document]
bburda: fetch_entities makes detail GET requests for each function (peer_client.cpp:363-373), which returns the hosts array. The list endpoint is only used to discover function IDs.

src/ros2_medkit_gateway/src/aggregation/peer_client.cpp
```diff
@@ -0,0 +1,603 @@
...
+
+    // Fetch subareas for each top-level area (list endpoint filters them out)
+    size_t top_level_count = entities.areas.size();
+    for (size_t i = 0; i < top_level_count; ++i) {
+      auto sub_result = cli.Get(std::string(API_PREFIX) + "/areas/" + entities.areas[i].id + "/subareas");
+      if (sub_result && sub_result->status == 200 && sub_result->body.size() <= MAX_PEER_RESPONSE_SIZE) {
+        auto sub_json = nlohmann::json::parse(sub_result->body, nullptr, false);
+        if (!sub_json.is_discarded()) {
+          auto subareas = parse_collection<Area>(sub_json, parse_area);
+          for (auto & sub : subareas) {
```
mfaferek93: **Blocker**

`push_back` into `entities.areas` while iterating it by index (line 322-336). If `push_back` triggers reallocation, the next `entities.areas[i].id` on line 325 reads freed memory - undefined behavior.

Same pattern for subcomponents at lines 380-399.

Fix: collect subareas in a separate `std::vector<Area> sub_areas`, then `std::move` them into `entities.areas` after the loop completes.

src/ros2_medkit_gateway/src/aggregation/mdns_discovery.cpp
```diff
@@ -0,0 +1,563 @@
...
+/**
+ * @brief mDNS callback for browse/query responses
+ *
+ * Called by mdns_query_recv() for each record in a response. We look for
+ * SRV records that match our service type and extract the hostname and port
+ * to build a peer URL.
+ */
+int browse_callback(int /*sock*/, const struct sockaddr * from, size_t addrlen, mdns_entry_type_t entry,
+                    uint16_t /*query_id*/, uint16_t rtype, uint16_t /*rclass*/, uint32_t /*ttl*/, const void * data,
+                    size_t size, size_t name_offset, size_t /*name_length*/, size_t record_offset, size_t record_length,
```
mfaferek93: **Major**

The `ttl` parameter is declared but unused (`uint32_t /*ttl*/`). TTL=0 records (RFC 6762 goodbye) are never detected. `on_removed_` is stored (line 312) and wired to `remove_discovered_peer()` in gateway_node.cpp:978, but never called.

Result: peers that leave the network stay in the peer list forever until gateway restart.

Fix: check `ttl == 0` in browse_callback, call `(*ctx->on_removed)(peer_name)` for goodbye records.

mfaferek93 (commented):
test

src/ros2_medkit_gateway/src/discovery/runtime_discovery.cpp
```diff
@@ -197,9 +118,6 @@ std::vector<App> RuntimeDiscoveryStrategy::discover_apps() {
     app.is_online = true;
```
mfaferek93: **Major** (re: line 115)

`app.id = name` uses the bare node name without namespace. Two nodes in different namespaces with the same name (e.g., `/sensors/driver` and `/actuators/driver`) both produce `id = "driver"`. The second overwrites the first in the entity cache.

The `bound_fqn` field (line 119) stores the qualified name, but `id` is used for cache lookups and URL routing.

Fix: use namespace-qualified ID (e.g., `sensors__driver`) or detect and disambiguate collisions.

mfaferek93 (commented):
_deleted - ignore_

mfaferek93 (commented):
test

src/ros2_medkit_gateway/src/aggregation/peer_client.cpp
mfaferek93, line 333: **Blocker**

`push_back` into `entities.areas` while iterating it by index (line 322-336). If `push_back` triggers reallocation, the next `entities.areas[i].id` on line 325 reads freed memory - undefined behavior.

Same pattern for subcomponents at lines 380-399.

Fix: collect subareas in a separate `std::vector<Area> sub_areas`, then `std::move` them into `entities.areas` after the loop completes.

src/ros2_medkit_gateway/src/aggregation/peer_client.cpp
```diff
@@ -0,0 +1,603 @@
...
+      return tl::unexpected<std::string>("Peer '" + name_ + "' returned " + std::to_string(entities.areas.size()) +
+                                         " areas (max " + std::to_string(MAX_ENTITIES_PER_COLLECTION) + ")");
+    }
+    for (auto & area : entities.areas) {
+      area.source = peer_source;
+    }
+
+    // Fetch subareas for each top-level area (list endpoint filters them out)
+    size_t top_level_count = entities.areas.size();
+    for (size_t i = 0; i < top_level_count; ++i) {
```
mfaferek93: **Major**

Sub-entity fetching has two gaps:

1. No `is_valid_entity_id` check on subareas/subcomponents (top-level entities are checked at line 309-313). Sub-entity IDs flow directly into URL construction at line 388 (`"/components/" + sub.id`) - path traversal possible.

2. No `MAX_ENTITIES_PER_COLLECTION` check after subs are added. A malicious peer: 999 areas (under limit) x 1000 subareas each = 999K entries + 999K HTTP detail requests.

Fix: apply `is_valid_entity_id` to subs, enforce cumulative count limit.

src/ros2_medkit_gateway/src/aggregation/mdns_discovery.cpp
mfaferek93, line 98: **Major**

The `ttl` parameter is declared but unused (`uint32_t /*ttl*/`). TTL=0 records (RFC 6762 goodbye) are never detected. `on_removed_` is stored (line 312) and wired to `remove_discovered_peer()` in gateway_node.cpp:978, but never called.

Result: peers that leave the network stay in the peer list forever until gateway restart.

Fix: check `ttl == 0` in browse_callback, call `(*ctx->on_removed)(peer_name)` for goodbye records.

src/ros2_medkit_gateway/src/aggregation/mdns_discovery.cpp
```diff
@@ -0,0 +1,563 @@
...
+    fd_set readfds;
+    FD_ZERO(&readfds);
+    FD_SET(sock, &readfds);
+
+    int ret = select(sock + 1, &readfds, nullptr, nullptr, &tv);
+    if (ret < 0) {
+      if (config_.on_log) {
+        config_.on_log("announce_loop: select() error (errno=" + std::to_string(errno) + ")");
+      }
+      break;
```
mfaferek93, line 467: **Major**

`break` on any `select()` error, including EINTR. A single signal (SIGCHLD from ROS 2 process manager, profiling timer, etc.) permanently kills mDNS announcement.

Fix:
```cpp
if (ret < 0) {
    if (errno == EINTR) continue;
    // ... log and break for real errors
    break;
}
```

src/ros2_medkit_gateway/src/aggregation/mdns_discovery.cpp
```diff
@@ -0,0 +1,563 @@
...
+    fd_set readfds;
+    FD_ZERO(&readfds);
+    FD_SET(sock, &readfds);
+
+    int ret = select(sock + 1, &readfds, nullptr, nullptr, &tv);
+    if (ret < 0) {
+      if (config_.on_log) {
+        config_.on_log("browse_loop: select() error (errno=" + std::to_string(errno) + ")");
+      }
+      break;
```
mfaferek93, line 580: **Major**

Same EINTR issue as the announce loop. `break` without `errno == EINTR` check.

src/ros2_medkit_gateway/src/aggregation/sse_stream_proxy.cpp
```diff
@@ -0,0 +1,207 @@
...
+SSEStreamProxy::SSEStreamProxy(const std::string & peer_url, const std::string & path, const std::string & peer_name)
+  : peer_url_(peer_url), path_(path), peer_name_(peer_name) {
+}
+
+SSEStreamProxy::~SSEStreamProxy() {
+  close();
+}
+
+void SSEStreamProxy::open() {
+  if (connected_.load()) {
```
mfaferek93: **Major**

Guard checks `connected_` which is only set to `true` inside `reader_loop()` (line 65), after the thread starts. Two rapid `open()` calls both see `connected_ == false`, both assign to `reader_thread_`. Assigning to a joinable `std::thread` calls `std::terminate`.

Also affects `open()` after failed connection (thread exited but `connected_` is false, `reader_thread_` still joinable).

Fix: guard on `reader_thread_.joinable()` + add a mutex, or use a state machine (`idle -> starting -> connected -> stopping`).

src/ros2_medkit_gateway/src/aggregation/sse_stream_proxy.cpp
```diff
@@ -0,0 +1,207 @@
...
+  callback_ = std::move(cb);
+}
+
+void SSEStreamProxy::reader_loop() {
+  httplib::Client client(peer_url_);
+  // Set reasonable timeouts - SSE connections are long-lived.
+  // Use a long read timeout instead of 0 (which causes immediate timeout in
+  // cpp-httplib when SO_RCVTIMEO is set to zero). 24 hours allows SSE streams
+  // to stay open indefinitely while still having a finite timeout for cleanup.
+  client.set_read_timeout(86400, 0);
```
mfaferek93: **Major**

`read_timeout(86400, 0)` - the `should_stop_` flag is only checked in the chunked callback. If the peer is alive but silent (no data, no TCP FIN), `close()` blocks until the 24h timeout expires.

Fix: store `httplib::Client` as a member, call `client_.stop()` from `close()` to abort the in-flight request. Or use a shorter timeout (30-60s) with a reconnect loop.

src/ros2_medkit_gateway/src/aggregation/sse_stream_proxy.cpp
```diff
@@ -0,0 +1,207 @@
...
+
+bool SSEStreamProxy::is_connected() const {
+  return connected_.load();
+}
+
+void SSEStreamProxy::on_event(std::function<void(const StreamEvent &)> cb) {
+  callback_ = std::move(cb);
+}
+
+void SSEStreamProxy::reader_loop() {
```
mfaferek93, line 62: **Major**

`reader_loop()` runs a single `client.Get()` and exits on any failure. Connection drop, HTTP error, buffer overflow - all result in permanent disconnect with no recovery. SSE connections are expected to be auto-resumable.

Fix: wrap connection logic in `while (!should_stop_)` with exponential backoff. Consider `Last-Event-ID` header for resumption.

src/ros2_medkit_gateway/src/aggregation/sse_stream_proxy.cpp
```diff
@@ -0,0 +1,207 @@
...
+  httplib::Client client(peer_url_);
+  // Set reasonable timeouts - SSE connections are long-lived.
+  // Use a long read timeout instead of 0 (which causes immediate timeout in
+  // cpp-httplib when SO_RCVTIMEO is set to zero). 24 hours allows SSE streams
+  // to stay open indefinitely while still having a finite timeout for cleanup.
+  client.set_read_timeout(86400, 0);
+  client.set_connection_timeout(5, 0);
+
+  connected_.store(true);
```
mfaferek93: **Major**

`connected_.store(true)` is set before `client.Get()` is called. If connection fails (DNS, TCP, HTTP 503), `is_connected()` briefly returns true.

Fix: set `connected_` inside the response header callback (line 71-73) after the `response.status == 200` check.

src/ros2_medkit_gateway/src/http/handlers/fault_handlers.cpp
```diff
@@ -574,7 +665,7 @@ void FaultHandlers::handle_clear_all_faults(const httplib::Request & req, httpli
     // Validate entity ID and type for this route
     auto entity_opt = ctx_.validate_entity_for_route(req, res, entity_id);
     if (!entity_opt) {
-      return;  // Error response already sent
+      return;  // Response already sent (error or forwarded to peer)
     }
     auto entity_info = *entity_opt;
```
mfaferek93, line 673: **Major** (re: line 712)

`handle_clear_all_faults_global` only clears local FaultManager. Peer faults (visible via `handle_list_all_faults` which does fan-out at line 270) will reappear on the next GET /faults.

Fix: forward DELETE to all healthy peers via fan-out, or document as intentionally local-only with `x-medkit.local_only: true`.

mfaferek93, line 673: **Major** (re: line 679)

`handle_clear_all_faults` uses `fault_mgr->list_faults(entity_info.namespace_path)`. For Function entities, `namespace_path` is `""` (empty string). `list_faults("")` returns the entire fault database. This handler will clear ALL faults when called on any Function entity.

Compare with `handle_list_faults` (line 387-433) which correctly uses FQN-based source filtering for Functions.

Fix: mirror the FQN-based filtering from `handle_list_faults` instead of using namespace_path.

src/ros2_medkit_gateway/src/discovery/runtime_discovery.cpp
mfaferek93, line 143: **Major** (re: line 115)

`app.id = name` uses the bare node name without namespace. Two nodes in different namespaces with the same name (e.g., `/sensors/driver` and `/actuators/driver`) both produce `id = "driver"`. The second overwrites the first in the entity cache.

The `bound_fqn` field (line 119) stores the qualified name, but `id` is used for cache lookups and URL routing.

Fix: use namespace-qualified ID (e.g., `sensors__driver`) or detect and disambiguate collisions.

src/ros2_medkit_gateway/src/aggregation/aggregation_manager.cpp
```diff
@@ -0,0 +1,592 @@
...
+  {
+    std::shared_lock<std::shared_mutex> lock(mutex_);
+    for (const auto & peer : peers_) {
+      if (peer->is_healthy()) {
+        snapshot.push_back(peer);
+      }
+    }
+  }
+
+  for (auto & peer : snapshot) {
```
mfaferek93, line 374: **Major**

`fetch_and_merge_peer_entities()` iterates peers in a plain `for` loop - sequential. Each `fetch_entities()` makes 4+ HTTP requests plus N detail fetches. For 10 peers at 2s timeout: worst case 80+ seconds.

Meanwhile, `fan_out_get()` (line 520) and `check_all_health()` (line 283) correctly use `std::async` for parallelism.

Fix: fetch from all peers in parallel with `std::async`, collect results, then merge sequentially.

src/ros2_medkit_gateway/src/aggregation/mdns_discovery.cpp
```diff
@@ -0,0 +1,563 @@
...
+  answer.ttl = 120;
+
+  // SRV additional: maps instance name to hostname and port
+  mdns_record_t additional{};
+  additional.name.str = instance.c_str();
+  additional.name.length = instance.size();
+  additional.type = MDNS_RECORDTYPE_SRV;
+  additional.data.srv.priority = 0;
+  additional.data.srv.weight = 0;
+  additional.data.srv.port = static_cast<uint16_t>(config->port);
```
mfaferek93, line 290: **Major**

`Config::port` is `int` (header line 49), cast to `uint16_t` via `static_cast` at lines 262, 389, 462 without bounds check. Port 70000 silently becomes 4464 in SRV records. Port -1 becomes 65535.

The browse side validates incoming `srv.port < 1024` (line 146) but the announce side has zero validation.

Fix: validate in constructor: `if (config_.port < 1024 || config_.port > 65535)` log error and refuse to announce.

docs/glossary.rst
```diff
@@ -97,6 +97,7 @@ This glossary defines key terms used throughout ros2_medkit documentation.
       a subset of SOVD adapted for ROS 2 systems.
```
mfaferek93, line 100: **Major** (re: line 12)

Line 12 (Area): "In runtime-only mode, areas are derived from ROS 2 namespace prefixes." - removed behavior.
Line 25 (Component): "components are synthetic groups created by namespace aggregation." - removed behavior.
Line 62 (Function): "Available in manifest mode." - now also available in runtime mode.

These contradict the "Synthetic Component (Removed)" entry at line 99 in the same file.

src/ros2_medkit_gateway/CMakeLists.txt
```diff
@@ -630,6 +657,30 @@ if(BUILD_TESTING)
...
+  ament_add_gtest(test_mdns_discovery test/test_mdns_discovery.cpp)
+  target_link_libraries(test_mdns_discovery gateway_lib)
+
+  # Network utility tests (URL parsing, local address collection)
+  ament_add_gtest(test_network_utils test/test_network_utils.cpp)
+  target_link_libraries(test_network_utils gateway_lib)
+
   # Apply coverage flags to test targets
   if(ENABLE_COVERAGE)
     set(_test_targets
```
mfaferek93, line 686: **Major**

The `_test_targets` coverage list is missing 4 targets that are defined above:
- `test_entity_resource_model` (line 458)
- `test_function_resource_collections` (line 462)
- `test_manifest_manager` (line 419)
- `test_x_medkit` (line 440)

These tests compile and run but don't get `--coverage` flags when `ENABLE_COVERAGE=ON`.

src/ros2_medkit_gateway/src/aggregation/aggregation_manager.cpp
```diff
@@ -0,0 +1,592 @@
...
+  }
+
+  // Resolve the host to check the actual address
+  struct addrinfo hints {};
+  hints.ai_family = AF_UNSPEC;
+  hints.ai_socktype = SOCK_STREAM;
+  hints.ai_flags = AI_NUMERICHOST;  // Try numeric first (fast, no DNS)
+
+  struct addrinfo * result = nullptr;
+  int ret = getaddrinfo(host.c_str(), nullptr, &hints, &result);
```
mfaferek93, line 164: **Minor**

`is_valid_peer_url()` resolves hostname via `getaddrinfo()` at validation time, but `PeerClient` creates a new `httplib::Client` with the original hostname - resolving again. DNS rebinding possible between validation and connection.

Not exploitable for mDNS peers (they use numeric IPs from `inet_ntop`), only for static peers configured with hostnames.

Fix: pin the resolved IP at validation time and use it for all subsequent connections.

src/ros2_medkit_gateway/src/aggregation/peer_client.cpp
```diff
@@ -0,0 +1,603 @@
...
+  return healthy_.load();
+}
+
+void PeerClient::ensure_client() {
+  if (!client_) {
+    client_ = std::make_unique<httplib::Client>(url_);
+    client_->set_connection_timeout(timeout_ms_ / 1000, (timeout_ms_ % 1000) * 1000);
+    client_->set_read_timeout(timeout_ms_ / 1000, (timeout_ms_ % 1000) * 1000);
+    // Note: cpp-httplib Client does not expose set_payload_max_length (server-only).
+    // Response size is enforced post-download in forward_request() and
```
mfaferek93, line 267: **Minor**

Comment honestly acknowledges: "Response size is enforced post-download." cpp-httplib Client doesn't expose `set_payload_max_length`. A fast peer on local network can push hundreds of MB before the 10MB check fires.

Consider using a content receiver (like SSEStreamProxy does) to abort early once limit is exceeded.

src/ros2_medkit_gateway/src/aggregation/peer_client.cpp
```diff
@@ -0,0 +1,603 @@
...
+  if (result->body.size() > MAX_PEER_RESPONSE_SIZE) {
+    res.status = 502;
+    auto error_body = make_error_body(ERR_VENDOR_ERROR, "Response from peer '" + name_ + "' exceeds size limit",
+                                      ERR_X_MEDKIT_PEER_UNAVAILABLE);
+    res.set_content(error_body.dump(), "application/json");
+    return;
+  }
+
+  // Copy response from peer - only forward safe headers
+  static const std::set<std::string> allowed_headers = {"content-type", "etag", "cache-control", "last-modified"};
```
mfaferek93: **Minor**

Any header matching `x-medkit*` prefix is forwarded from peer responses. A malicious peer can inject arbitrary `x-medkit-*` headers. If any downstream consumer trusts these for auth/authz, this is exploitable.

Fix: allowlist specific `x-medkit-*` headers, or document they are informational-only.

src/ros2_medkit_gateway/src/http/handlers/fault_handlers.cpp
```diff
@@ -264,6 +265,21 @@ void FaultHandlers::handle_list_all_faults(const httplib::Request & req, httplib
         ext.add("clusters", result.data["clusters"]);
       }
 
+      // Fan-out to peers: faults are managed by FaultManager, not cached,
+      // so we need to query each peer's /faults endpoint and merge results.
+      if (auto * agg = ctx_.aggregation_manager()) {
+        auto fan_result = agg->fan_out_get(req.path, req.get_header_value("Authorization"));
```
mfaferek93, line 271: **Minor**

`agg->fan_out_get(req.path, ...)` - httplib's `req.path` excludes query string. Query params like `?status=confirmed`, `?include_muted=true` are silently dropped. Peers return unfiltered results while local faults are filtered.

Note: `peer_client.cpp` has a `path_with_query()` helper (line 73-89) that reconstructs path + query, but it's not used here.

Fix: pass full path + query string to `fan_out_get`.

src/ros2_medkit_gateway/src/http/handlers/fault_handlers.cpp
```diff
@@ -345,8 +367,15 @@ void FaultHandlers::handle_list_faults(const httplib::Request & req, httplib::Re
...
       ext.add("host_count", host_fqns.size());
+      // Include source app IDs for cross-referencing aggregated results
+      json source_ids = json::array();
+      for (const auto & fqn : host_fqns) {
+        source_ids.push_back(fqn);
+      }
+      ext.add("aggregation_sources", source_ids);
 
       response["x-medkit"] = ext.build();
       HandlerContext::send_json(res, response);
```
mfaferek93, line 383: **Minor** (re: line 387)

`handle_list_faults` uses FQN-based source matching for Components: collects app FQNs via `get_apps_for_component()`, then `filter_faults_by_sources()`.

`handle_clear_all_faults` (line 679) uses `fault_mgr->list_faults(entity_info.namespace_path)` for ALL entity types - namespace prefix matching.

For the same Component, list and clear may operate on different fault sets.

src/ros2_medkit_gateway/src/aggregation/aggregation_manager.cpp
```diff
@@ -0,0 +1,592 @@
...
+  }
+
+  // Strip peer prefix from entity ID in the path if present.
+  // When entity ID collision causes renaming (e.g., camera_driver -> peer_b__camera_driver),
+  // the peer only knows the entity by its original ID (camera_driver), so we must strip
+  // the prefix before forwarding.
+  // Anchor to path segment boundary: the prefix must appear right after '/' to avoid
+  // false matches inside other path segments (e.g., "v1" matching "/api/v1/").
+  std::string forwarded_path = req.path;
+  std::string prefix = peer_name + EntityMerger::SEPARATOR;
```
mfaferek93, line 521: **Minor**

`forwarded_path.find(prefix)` matches the peer prefix anywhere after a `/`. The entity ID in SOVD paths is always the 4th segment (`/api/v1/{type}/{entity_id}/...`). Matching at arbitrary positions could strip from wrong segments if the prefix appears elsewhere in the path.

Fix: parse path segments and strip only from the entity ID position.

src/ros2_medkit_gateway/src/aggregation/entity_merger.cpp
```diff
@@ -0,0 +1,203 @@
...
+  for (const auto & app : local) {
+    local_ids.insert(app.id);
+  }
+
+  for (const auto & remote_app : remote) {
+    App added = remote_app;
+    added.source = peer_source();
+    // Remap component_id to peer name - in the aggregation model, the peer
+    // IS a component/subcomponent and all its apps belong to it.
+    added.component_id = peer_name_;
```
mfaferek93: **Minor**

`added.component_id = peer_name_` set unconditionally for all remote apps. A remote app belonging to `perception_ecu` component will have its `component_id` changed to the peer name, breaking the component-app relationship.

Fix: only remap when the original component is not present in the merged result.

docs/tutorials/heuristic-apps.rst
mfaferek93, line 1: **Blocker** (re: line 5)

This tutorial still describes the removed entity model (synthetic per-namespace Components, namespace-derived Areas). Only lines 115-137 were updated. Needs full rewrite: single host Component, no Areas in runtime, Functions from namespaces.

docs/tutorials/manifest-discovery.rst
mfaferek93, line 1: **Major** (re: line 44)

Two stale cells in Runtime-Only column: Line 44 says Areas derived from namespaces (removed), Line 52 says no Functions in runtime (now added). Contradicts line 432.

docs/getting_started.rst
mfaferek93, line 1: **Major** (re: line 329)

Checkpoint says Areas/Components that don't exist in runtime-only mode. Fix: Functions + single host Component.

docs/glossary.rst
mfaferek93, line 1: **Major** (re: line 12)

Area/Component/Function definitions are stale. Contradicts Synthetic Component (Removed) entry at line 99.

src/ros2_medkit_gateway/src/http/handlers/discovery_handlers.cpp
mfaferek93, line 1: **Minor** (re: line 199)

Relationship handlers use validate_entity_id() instead of validate_entity_for_route(). Between-refresh entities return 404.

src/ros2_medkit_gateway/src/gateway_node.cpp
mfaferek93, line 1: **Minor** (re: line 893)

peer_scheme not validated. Misconfigured value silently rejects all mDNS peers.

mfaferek93, line 1: **Minor** (re: line 958)

collect_local_addresses() called once. Stale on network changes.

src/ros2_medkit_gateway/src/aggregation/mdns_discovery.cpp
mfaferek93, line 1: **Minor** (re: line 218)

Substring match for service name. RFC 6763 requires exact match.

mfaferek93, line 1: **Minor** (re: line 126)

Uses source IP instead of SRV target + A/AAAA. Document as intentional.

src/ros2_medkit_gateway/src/aggregation/sse_stream_proxy.cpp
mfaferek93, line 1: **Minor** (re: line 90)

Only detects \n\n boundaries, not \r\n\r\n.

mfaferek93, line 1: **Minor** (re: line 71)

No Content-Type validation. Non-SSE 200 responses accepted.

src/ros2_medkit_gateway/test/test_handler_context.cpp
mfaferek93, line 1: **Minor** (re: line 390)

Port reservation TOCTOU. Flaky on CI.

src/ros2_medkit_gateway/test/test_mdns_discovery.cpp
mfaferek93, line 1: **Minor** (re: line 28)

All tests run with announce=false, discover=false. Zero mDNS coverage.

src/ros2_medkit_gateway/test/test_aggregation_manager.cpp
mfaferek93, line 1: **Minor** (re: line 659)

No test for multi-peer overlapping entity IDs.

src/ros2_medkit_gateway/src/discovery/host_info_provider.cpp
mfaferek93, line 1: **Minor** (re: line 67)

sanitize_entity_id can return empty string. Fallback to unknown_host.

.github/copilot-instructions.md
mfaferek93, line 1: **Minor** (re: line 47)

Stale entity model description. Drives incorrect automated review.

src/ros2_medkit_gateway/src/aggregation/peer_client.cpp
bburda: Fixed in 53bd5c27 - sub-entities collected into separate vector, appended after iteration.

src/ros2_medkit_gateway/src/aggregation/peer_client.cpp
bburda: Fixed in 53bd5c27 - is_valid_entity_id check added for subareas and subcomponents.

src/ros2_medkit_gateway/src/aggregation/mdns_discovery.cpp
bburda: Fixed in 53bd5c27 - TTL=0 now triggers on_removed callback.

src/ros2_medkit_gateway/src/aggregation/mdns_discovery.cpp
bburda: Fixed in 53bd5c27 - errno==EINTR now continues instead of breaking.

src/ros2_medkit_gateway/src/aggregation/mdns_discovery.cpp
bburda: Fixed in 53bd5c27.

src/ros2_medkit_gateway/src/aggregation/sse_stream_proxy.cpp
bburda: Fixed in 53bd5c27 - open() now checks reader_thread_.joinable() instead of connected_. connected_ set only on first data chunk.

src/ros2_medkit_gateway/src/aggregation/sse_stream_proxy.cpp
bburda: Fixed in 53bd5c27 - reduced to 300s.

src/ros2_medkit_gateway/src/aggregation/sse_stream_proxy.cpp
bburda: Fixed in 53bd5c27 - reconnect loop with exponential backoff (1s-30s), should_stop_ checked between retries.

src/ros2_medkit_gateway/src/aggregation/sse_stream_proxy.cpp
bburda: Fixed in 53bd5c27 - connected_.store(false) set immediately after Get() returns.

src/ros2_medkit_gateway/src/http/handlers/fault_handlers.cpp
bburda: Fixed in 53bd5c27 - documented limitation with X-Medkit-Local-Only header on 204 response.

bburda: Fixed in 53bd5c27 - handle_clear_all_faults per-entity now uses FQN-based filtering for Functions/Components/Areas.

src/ros2_medkit_gateway/src/discovery/runtime_discovery.cpp
bburda: Fixed in 53bd5c27 - collision-only disambiguation: namespace prefix added when same bare name appears in 2+ namespaces.

src/ros2_medkit_gateway/src/aggregation/mdns_discovery.cpp
bburda: Fixed in 53bd5c27 - port range validated (0-65535), fallback to 8080 on invalid.

docs/glossary.rst
bburda: Fixed in 53bd5c27 - Area, Component, Function, App definitions updated.

docs/glossary.rst
bburda: Fixed in 53bd5c27.

src/ros2_medkit_gateway/CMakeLists.txt
bburda: Fixed in 53bd5c27 - 4 missing targets added.

docs/tutorials/heuristic-apps.rst
bburda: Fixed in 53bd5c27 - tutorial rewritten for current entity model.

docs/tutorials/manifest-discovery.rst
bburda: Fixed in 53bd5c27 - table cells updated.

docs/getting_started.rst
bburda: Fixed in 53bd5c27 - checkpoint mentions Functions + single host Component.

src/ros2_medkit_gateway/src/http/handlers/fault_handlers.cpp
bburda: Fixed - forward_request reconstructs full path with query parameters from req.params.

src/ros2_medkit_gateway/src/http/handlers/fault_handlers.cpp
bburda: Fixed in 53bd5c27 - same FQN-based filtering as handle_list_faults.

src/ros2_medkit_gateway/src/aggregation/aggregation_manager.cpp
bburda: Already fixed - anchored to path segment boundaries (checks preceding '/').

src/ros2_medkit_gateway/src/http/handlers/discovery_handlers.cpp
bburda: Fixed - discovery detail handlers use validate_entity_for_route for transparent forwarding.

src/ros2_medkit_gateway/src/gateway_node.cpp
bburda: Fixed in dd3ea7f2 - validated as 'http' or 'https', fallback to 'http' with RCLCPP_ERROR.

src/ros2_medkit_gateway/src/aggregation/sse_stream_proxy.cpp
bburda: Fixed in dd3ea7f2 - ResponseHandler validates Content-Type contains 'text/event-stream' before parsing.

bburda: Our SSE peers are ros2_medkit gateways using cpp-httplib which sends LF. Content-Type validation (dd3ea7f2) now rejects non-SSE responses, preventing misparse of other formats.

src/ros2_medkit_gateway/test/test_aggregation_manager.cpp
bburda: Fixed in dd3ea7f2 - test with local + peer_a + peer_b all having 'camera_driver', verifies distinct prefixed IDs and correct routing.

.github/copilot-instructions.md
bburda: Updated - ValidateResult type and handler pattern documented.

src/ros2_medkit_gateway/src/aggregation/aggregation_manager.cpp
bburda: Fixed in 10ef45a7 - parallelized with std::async, same pattern as fan_out_get and check_all_health.

src/ros2_medkit_gateway/src/aggregation/aggregation_manager.cpp
bburda: Documented in 10ef45a7 - inherent TOCTOU between validation-time and connection-time DNS. IP pinning breaks legitimate failover. 2s read timeout limits exposure.

src/ros2_medkit_gateway/src/aggregation/peer_client.cpp
bburda: Fixed in 10ef45a7 - forward_and_get_json GET requests now use ContentReceiver with size accumulator that aborts mid-stream at MAX_PEER_RESPONSE_SIZE.

src/ros2_medkit_gateway/src/aggregation/peer_client.cpp
bburda: Fixed in 10ef45a7 - replaced blanket x-medkit* prefix with explicit allowlist (x-medkit-local-only). Test updated to verify unknown x-medkit headers are blocked.

src/ros2_medkit_gateway/src/aggregation/entity_merger.cpp
bburda: Fixed in 10ef45a7 - removed forced component_id = peer_name_ remap. Remote apps keep their original component_id since remote components merge by ID.

src/ros2_medkit_gateway/src/gateway_node.cpp
bburda: Fixed in 10ef45a7 - local addresses refreshed every 60s in the mDNS on_found callback.

src/ros2_medkit_gateway/src/aggregation/mdns_discovery.cpp
bburda: Fixed in 10ef45a7 - replaced substring find() with exact == comparison per RFC 6763.

bburda: Documented in 10ef45a7 - comment explains source IP is used intentionally over SRV target resolution.

src/ros2_medkit_gateway/test/test_handler_context.cpp
bburda: Documented in 10ef45a7 - inherent TOCTOU with ephemeral ports, collisions astronomically unlikely.

src/ros2_medkit_gateway/test/test_mdns_discovery.cpp
bburda: Fixed in 10ef45a7 - added 3 tests: out-of-range port, negative port, and ephemeral socket test exercising actual socket creation path.

src/ros2_medkit_gateway/src/discovery/host_info_provider.cpp
bburda: Fixed in 10ef45a7 - falls back to 'unknown_host' when sanitized result is empty.

PR #329, opened 2026-03-31
fix(gateway): correct OpenAPI schemas to match handler implementations

### LINKED ISSUE #322: GET /api/v1 (API root) returns 404
# Bug report

### Steps to reproduce
1. Start gateway
2. `GET /api/v1` returns 404

### Expected behavior

API root should return server info, available endpoints, or redirect to `/api/v1/health`.

### Actual behavior

```
GET /api/v1 -> 404
```

### Environment
- ros2_medkit version: main branch
- ROS 2 distro: Humble
- OS: Ubuntu 22.04 on Jetson Orin Nano (aarch64)

### Additional information

Low priority. `GET /api/v1/health` and `GET /api/v1/docs` work fine.

### LINKED ISSUE #328: [BUG] OpenAPI spec: Lock and Trigger schemas use wrong models for request bodies
# Bug report

### Steps to reproduce

1. Generate a typed API client from the OpenAPI spec (GET /api/v1/docs)
2. Try to call POST /{entity}/locks (acquire lock) with the generated client
3. The generated Lock model requires `id`, `lock_expiration` (ISO datetime), `owned` - fields that only exist in the response, not the request

### Expected behavior

Request and response schemas should be separate when they differ. The acquire lock endpoint should have a dedicated request schema matching what the handler actually accepts.

### Actual behavior

Three endpoints use incorrect schemas for request bodies:

**1. POST /{entity}/locks (acquire lock)**
- OpenAPI spec uses `Lock` schema for request body
- `Lock` requires: `id` (string), `lock_expiration` (ISO 8601 datetime), `owned` (boolean)
- Gateway handler (`lock_handlers.cpp:150-156`) actually expects: `{"lock_expiration": <integer_seconds>, "scopes": [...]}`
- Generated clients fail with `KeyError: 'id'` when constructing the request

**2. PUT /{entity}/locks/{id} (extend lock)**
- Same issue - uses `Lock` schema for request body
- Should use `AcquireLockRequest` or a dedicated `ExtendLockRequest`

**3. POST /{entity}/triggers (create trigger)**
- `TriggerCreateRequest.trigger_condition` requires `condition_type` field
- Need to verify this matches what `trigger_handlers.cpp` actually parses

### Suggested fix

In `schema_builder.cpp`:
1. Add `AcquireLockRequest` schema with `lock_expiration` (integer, seconds) and `scopes` (optional array of strings)
2. Use `AcquireLockRequest` for POST /locks request body (keep `Lock` for response)
3. Use `AcquireLockRequest` (or dedicated schema) for PUT /locks/{id} request body
4. Verify `TriggerCreateRequest.trigger_condition` fields match handler expectations

### Environment
- ros2_medkit version: 0.4.0
- Spec version: 0.4.0

bburda:
# Pull Request

## Summary

Fixes OpenAPI spec schemas to match handler implementations, making the spec usable by generated API clients. Also adds an integration test that automatically catches future spec/handler mismatches by calling every endpoint using only information from the spec. Additionally fixes #322 (trailing slash routing).

---

## Issue

- closes #328
- closes #322

---

## Type

- [x] Bug fix
- [x] New feature or tests
- [ ] Breaking change
- [ ] Documentation only

---

## Testing

- 1645 unit tests pass (9 new schema tests + 3 route registry tests including trailing slash)
- New integration test `test_openapi_callability` passes - fetches spec from `/docs`, generates requests for every endpoint (GET/POST/PUT/DELETE), verifies no 400 responses from spec-conformant requests
- Pre-commit hooks pass (clang-format, flake8, ament-copyright)

### Schema fixes
- Add AcquireLockRequest/ExtendLockRequest (integer seconds, not Lock response schema)
- Add DataWriteRequest (type + data required)
- Add ExecutionUpdateRequest (capability enum)
- Add ScriptControlRequest (action enum: stop/forced_termination)
- Extract shared trigger_condition_schema() with additionalProperties
- ScriptMetadata.parameters_schema now nullable (type: [object, null])
- x-medkit fields: additionalProperties on all vendor extension objects
- LogConfiguration: both fields optional, severity_filter enum, bounds
- capability_generator uses $ref instead of inline LogConfiguration copy

### Route/handler fixes
- Lock endpoints document X-Client-Id header with minLength/maxLength schema
- Lock GET list/detail: X-Client-Id optional (for owned field)
- PUT /locks/{id} response: 200+Lock -> 204 No Content
- X-Client-Id: length (256 max) + control character validation
- RouteEntry::header_param() with optional schema parameter
- Optional trailing slash on all routes (/?$ regex anchor) - fixes #322

### Bug fixes
- ResourceChangeNotifier: fix -Wreorder (error_logger_ before worker_thread_)
- ChangeTypePropagation test: eliminate race (per-iteration notifier)

### Cross-repo follow-ups
- selfpatch/ros2_medkit_mcp#15: extend_lock -> _call_void for 204 response
- selfpatch/ros2_medkit_clients#17: regenerate openapi.yaml

---

## Checklist

### DIFF
--- a/src/ros2_medkit_integration_tests/test/features/test_docs_endpoint.test.py
+++ b/src/ros2_medkit_integration_tests/test/features/test_docs_endpoint.test.py
@@ -218,20 +218,35 @@ class TestDocsEndpoint(GatewayTestCase):
         config_path = config_paths[0]
         path_item = data['paths'][config_path]
 
-        # Verify GET response schema has correct field names
+        # Verify GET response schema references LogConfiguration component
         self.assertIn('get', path_item)
         get_schema = path_item['get']['responses']['200']['content']['application/json']['schema']
-        props = get_schema.get('properties', {})
-        self.assertIn('severity_filter', props, f'Expected severity_filter in properties: {props}')
-        self.assertIn('max_entries', props, f'Expected max_entries in properties: {props}')
-        self.assertNotIn('level', props, f'Should not contain old "level" field: {props}')
+        if '$ref' in get_schema:
+            # Entity-scoped docs use $ref to shared LogConfiguration schema
+            self.assertIn('LogConfiguration', get_schema['$ref'])
+            # Verify the referenced component schema has correct fields
+            schemas = data.get('components', {}).get('schemas', {})
+            if 'LogConfiguration' in schemas:
+                props = schemas['LogConfiguration'].get('properties', {})
+                self.assertIn('severity_filter', props)
+                self.assertIn('max_entries', props)
+                self.assertNotIn('level', props)
+        else:
+            # Inline schema (root docs)
+            props = get_schema.get('properties', {})
+            self.assertIn('severity_filter', props, f'Missing severity_filter: {props}')
+            self.assertIn('max_entries', props, f'Missing max_entries: {props}')
+            self.assertNotIn('level', props, f'Old "level" field present: {props}')
 
-        # Verify PUT request body schema
+        # Verify PUT request body schema references LogConfiguration
         self.assertIn('put', path_item)
         put_schema = path_item['put']['requestBody']['content']['application/json']['schema']
-        put_props = put_schema.get('properties', {})
-        self.assertIn('severity_filter', put_props)
-        self.assertNotIn('level', put_props)
+        if '$ref' in put_schema:
+            self.assertIn('LogConfiguration', put_schema['$ref'])
+        else:
+            put_props = put_schema.get('properties', {})
+            self.assertIn('severity_filter', put_props)
+            self.assertNotIn('level', put_props)
 
     def test_nonexistent_entity_docs_returns_404(self):
         """GET /apps/nonexistent_entity_xyz/docs returns 404.


src/ros2_medkit_integration_tests/test/features/test_openapi_callability.test.py
```diff
@@ -0,0 +1,408 @@
...
+    'area_id': 'areas',
+    'component_id': 'components',
+    'app_id': 'apps',
+    'function_id': 'functions',
+}
+
+# Business-logic 400s that are NOT spec bugs. These occur when the handler
+# rejects a request for reasons that cannot be expressed in the OpenAPI schema
+# (e.g. cross-resource validation, runtime state dependencies).
+_KNOWN_BUSINESS_400_PATTERNS = [
```
mfaferek93, line 200: **Minor**: 7-entry allowlist. At least 2-3 entries ("Resource URI must reference same entity", "Data collection requires resource path") could be eliminated by improving `_generate_resource_uri` to always match the route's entity type, rather than allowlisting the 400. Each entry should have a comment noting whether it's a spec limitation (can't fix) or a test-generation limitation (should fix).

src/ros2_medkit_gateway/src/http/handlers/lock_handlers.cpp
```diff
@@ -43,12 +43,20 @@ bool LockHandlers::check_locking_enabled(httplib::Response & res) {
...
 std::optional<std::string> LockHandlers::require_client_id(const httplib::Request & req, httplib::Response & res) {
+  static constexpr size_t kMaxClientIdLen = 256;
+
   auto client_id = req.get_header_value("X-Client-Id");
   if (client_id.empty()) {
     HandlerContext::send_error(res, 400, ERR_INVALID_PARAMETER, "Missing required X-Client-Id header",
                                json{{"details", "X-Client-Id header is required for lock operations"}});
     return std::nullopt;
   }
+  if (client_id.size() > kMaxClientIdLen) {
```
mfaferek93, line 54: **Minor**: Length validation added (good). No character validation though - control characters (`\x00-\x1f`, newlines) pass through. Verified: X-Client-Id is NOT logged and NOT returned in API responses (only used for `owned` boolean comparison), so the practical risk is zero today. But a pattern constraint would be defense-in-depth if future code logs or serializes it.

src/ros2_medkit_integration_tests/test/features/test_openapi_callability.test.py
bburda: Done in 221331a5. Annotated each entry as [spec limitation] or [test limitation]. Left the test-limitation entries as-is for now - the resource URI matching logic would need the full path context which adds significant complexity for marginal gain.

src/ros2_medkit_gateway/src/http/handlers/lock_handlers.cpp
bburda: Done in 221331a5. Added control character validation (rejects bytes < 0x20). Length validation was already there from the first commit.

src/ros2_medkit_integration_tests/test/features/test_openapi_callability.test.py
```diff
@@ -0,0 +1,408 @@
...
+
+                if resp.status_code == 400:
+                    try:
+                        err = resp.json()
+                        msg = err.get('message', resp.text[:200])
+                    except Exception:
+                        msg = resp.text[:200]
+
+                    # Skip known business-logic rejections
+                    if is_known_business_400(msg):
```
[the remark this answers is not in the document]
bburda: Done.

src/ros2_medkit_integration_tests/test/features/test_openapi_callability.test.py
```diff
@@ -0,0 +1,408 @@
...
+                        continue
+
+                    failures.append(
+                        f'{op_id}: 400 Bad Request - {msg}'
+                    )
+
```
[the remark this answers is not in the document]
bburda: Done.

src/ros2_medkit_gateway/src/openapi/route_registry.cpp
```diff
@@ -79,6 +79,17 @@ RouteEntry & RouteEntry::query_param(const std::string & name, const std::string
   return *this;
 }
 
+RouteEntry & RouteEntry::header_param(const std::string & name, const std::string & desc, bool required) {
+  nlohmann::json param;
+  param["name"] = name;
+  param["in"] = "header";
+  param["required"] = required;
+  param["description"] = desc;
+  param["schema"] = {{"type", "string"}};
```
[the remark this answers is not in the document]
bburda: Done.

mfaferek93 (approved):
LGTM!

PR #347, opened 2026-04-03
Abstract httplib from plugin API, vendor as build farm fallback

### LINKED ISSUE #341: [BUG] Humble build farm fails: cpp-httplib 0.10.x too old
# Bug report

### Steps to reproduce
1. Submit ros2_medkit to the ROS 2 Humble build farm (Ubuntu 22.04 Jammy)
2. Build farm installs `libcpp-httplib-dev 0.10.3` from system packages
3. `medkit_find_cpp_httplib()` rejects it (requires >= 0.14)
4. Build fails with `FATAL_ERROR`

### Expected behavior

Gateway should build on the Humble build farm without requiring manual source installs.

### Actual behavior

CMake fails during configuration:

```
CMake Error at .../ROS2MedkitCompat.cmake:99 (message):
  [MedkitCompat] Could not find cpp-httplib >= 0.14.

  The system libcpp-httplib-dev package on Ubuntu 22.04 provides 0.10.x which is too old.
```

The build farm only installs rosdep-resolved system packages. There is no mechanism to build cpp-httplib from source.

### Environment
- ros2_medkit version: 0.4.0
- ROS 2 distro: Humble
- OS: Ubuntu 22.04 (Jammy) on ROS build farm

### Additional information

Proposed fix: vendor cpp-httplib v0.14.3 (MIT, header-only, single file) in `ros2_medkit_cmake` as a fallback. The `medkit_find_cpp_httplib()` macro already has a three-tier resolution (pkg-config, cmake config, fatal error) - adding a vendored tier between cmake config and fatal error fixes the build farm without affecting Jazzy/Rolling where the system package is new enough.

bburda:
# Pull Request

## Summary

Removes cpp-httplib from the public plugin API so only `ros2_medkit_gateway` depends on it. Plugins now use thin `PluginRequest`/`PluginResponse` wrappers and declare routes declaratively via `get_routes()` instead of imperatively registering on `httplib::Server`.

This also vendors cpp-httplib 0.14.3 in the gateway package with a `VENDORED_DIR` cmake fallback, fixing the Humble build farm failure where the system package (0.10.x) is too old.

---

## Issue

- closes #341

---

## Type

- [x] Bug fix
- [x] New feature or tests
- [x] Breaking change
- [ ] Documentation only

---

## Testing

- Full colcon build (13 packages) passes
- 2422 unit tests pass (0 failures)
- New unit test `RegisterRoutesWrapsPluginHandlers` verifies end-to-end route registration and PluginRequest/PluginResponse wrapping
- New unit tests for `PluginRequest` (path_param, header, path, body) and `PluginResponse` (send_json, send_error)
- Integration test updated for demo plugin response format change (plaintext -> JSON)
- Linters pass

---

## Checklist

### Breaking changes

- `PLUGIN_API_VERSION` bumped 4 -> 5
- `GatewayPlugin::register_routes()` and `get_route_descriptions()` removed, replaced by `get_routes()` returning `vector<PluginRoute>`
- `PluginContext::send_json()` and `send_error()` static methods removed, now instance methods on `PluginResponse`
- Plugin route handlers receive `PluginRequest`/`PluginResponse` instead of `httplib::Request`/`httplib::Response`
- Plugin packages no longer need `libcpp-httplib-dev`, `libssl-dev`, or `medkit_find_cpp_httplib()` in CMake

### DIFF
--- a/src/ros2_medkit_gateway/CMakeLists.txt
+++ b/src/ros2_medkit_gateway/CMakeLists.txt
@@ -655,7 +655,7 @@ if(BUILD_TESTING)
   target_link_libraries(test_condition_evaluator gateway_lib)
 
   # Resource change notifier tests (async notification hub for triggers)
-  ament_add_gtest(test_resource_change_notifier test/test_resource_change_notifier.cpp TIMEOUT 600)
+  ament_add_gtest(test_resource_change_notifier test/test_resource_change_notifier.cpp TIMEOUT 180)
   target_link_libraries(test_resource_change_notifier gateway_lib)
 
   # Trigger store tests (SQLite persistence for triggers)
--- a/src/ros2_medkit_gateway/src/resource_change_notifier.cpp
+++ b/src/ros2_medkit_gateway/src/resource_change_notifier.cpp
@@ -105,6 +105,13 @@ void ResourceChangeNotifier::shutdown() {
     });
   }
 
+  // Synchronize with worker_loop()'s predicate check. Without this, the flag
+  // store above can land between the worker's predicate evaluation and its
+  // wait() call, losing the subsequent notify_one(). Acquiring queue_mutex_
+  // here guarantees the worker is either still outside the critical section
+  // (will observe the new flag) or already enqueued on queue_cv_ (notify will
+  // wake it).
+  { std::lock_guard<std::mutex> sync(queue_mutex_); }
   queue_cv_.notify_one();
   if (worker_thread_.joinable()) {
     worker_thread_.join();


src/ros2_medkit_gateway/src/vendored/cpp_httplib/httplib.h
```diff
@@ -0,0 +1,2272 @@
...
+
+  Server();
+
+  virtual ~Server();
+
+  virtual bool is_valid() const;
+
+  Server &Get(const std::string &pattern, Handler handler);
+  Server &Post(const std::string &pattern, Handler handler);
+  Server &Post(const std::string &pattern, HandlerWithContentReader handler);
```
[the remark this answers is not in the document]
bburda: Fixed - replaced with full upstream header (9370 lines with inline implementations) instead of Ubuntu stub-only system header (2272 lines).

src/ros2_medkit_gateway/test/test_plugin_manager.cpp
```diff
@@ -296,22 +316,53 @@ TEST(PluginManagerTest, ThrowOnSetContextDisablesPlugin) {
...
+
+  // Start server in background
+  std::thread server_thread([&srv]() {
+    srv.listen("127.0.0.1", 18321);
+  });
+  srv.wait_until_ready();
+
+  // Make request to the plugin route
+  httplib::Client cli("127.0.0.1", 18321);
+  auto res = cli.Get("/api/v1/apps/test_entity/x-test-route");
```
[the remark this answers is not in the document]
bburda: Fixed - switched to bind_to_any_port + listen_after_bind for ephemeral port allocation.

src/ros2_medkit_plugins/ros2_medkit_graph_provider/CMakeLists.txt
```diff
@@ -93,6 +83,11 @@ if(BUILD_TESTING)
 
   find_package(ament_cmake_gtest REQUIRED)
 
+  # The route test uses httplib::Server/Client directly for end-to-end HTTP testing.
+  # Find cpp-httplib so the test binary can link against the compiled implementation.
+  find_package(OpenSSL REQUIRED)
+  medkit_find_cpp_httplib()
+
```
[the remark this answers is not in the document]
bburda: Fixed - added VENDORED_DIR pointing to gateway's installed vendored copy (share/ros2_medkit_gateway/vendored/cpp_httplib). Installed to share/ (not include/) to avoid polluting downstream include paths.

src/ros2_medkit_gateway/src/plugins/plugin_manager.cpp
```diff
@@ -260,20 +263,43 @@ void PluginManager::register_transport(std::unique_ptr<SubscriptionTransportProv
...
-      lp.load_result.plugin->register_routes(server, api_prefix);
+      auto routes = lp.load_result.plugin->get_routes();
+      for (auto & route : routes) {
+        std::string full_pattern = api_prefix + "/" + route.pattern;
+        auto handler_fn = route.handler;  // capture by value for lambda
+        auto httplib_handler = [handler_fn](const httplib::Request & req, httplib::Response & res) {
+          PluginRequest plugin_req(&req);
+          PluginResponse plugin_res(&res);
+          handler_fn(plugin_req, plugin_res);
+        };
```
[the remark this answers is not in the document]
bburda: Fixed - route handler invocations are now wrapped in try/catch. Exceptions are logged with plugin name and route pattern, and a 500 SOVD GenericError is returned to the client.

docs/tutorials/plugin-system.rst
```diff
@@ -302,10 +307,10 @@ for the lower-level registry API.
...
    (exact-match check).
 
-PluginContext API (v4)
+PluginContext API (v5)
 ----------------------
 
-Version 4 of the plugin API introduced several new methods on ``PluginContext``.
+Version 5 of the plugin API introduced several new methods on ``PluginContext``.
 These methods have default no-op implementations, so existing plugins continue to
 compile without changes (though a rebuild is required to match the new
```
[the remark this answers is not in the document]
bburda: Fixed - reworded to clarify that plugins implementing custom routes require source changes for v5, while plugins without routes only need a rebuild.

src/ros2_medkit_gateway/src/plugins/plugin_manager.cpp
```diff
@@ -260,20 +263,57 @@ void PluginManager::register_transport(std::unique_ptr<SubscriptionTransportProv
...
       continue;
     }
     try {
-      lp.load_result.plugin->register_routes(server, api_prefix);
+      auto routes = lp.load_result.plugin->get_routes();
+      for (auto & route : routes) {
+        std::string full_pattern = api_prefix + "/" + route.pattern;
+        auto handler_fn = route.handler;  // capture by value for lambda
+        auto plugin_name = lp.load_result.plugin->name();
+        auto httplib_handler = [handler_fn, plugin_name, &full_pattern](const httplib::Request & req,
```
mfaferek93: **Blocker**

`full_pattern` is a local `std::string` in the for-loop body (line 276). The lambda captures it by reference (`&full_pattern`) but is registered with httplib and called later when HTTP requests arrive - by which time `full_pattern` is destroyed.

Only triggers in the error-path logging (lines 287, 292) when a plugin handler throws, making it a latent crash that's hard to diagnose in production.

Fix: change `&full_pattern` to `full_pattern` (capture by value).

src/ros2_medkit_gateway/include/ros2_medkit_gateway/plugins/plugin_http_types.hpp
```diff
@@ -0,0 +1,66 @@
...
+
+namespace ros2_medkit_gateway {
+
+/// Thin read-only wrapper over an HTTP request.
+/// Hides the underlying HTTP library (currently cpp-httplib) from plugin code.
+/// Constructed by the gateway per-request; plugins receive it by const reference.
+class PluginRequest {
+ public:
+  /// Construct from opaque pointer to httplib::Request.
+  explicit PluginRequest(const void * impl);
```
mfaferek93, line 28: **Major**

`PluginRequest` exposes `path_param()`, `header()`, `path()`, `body()` but no query parameter access. Any plugin needing pagination, filtering, or format selection (`?limit=10&offset=0`) is blocked.

Given the stated goal of enabling HTTP lib swaps, this is the #1 gap - query params are fundamental REST and the first plugin needing them will have to break the abstraction.

Fix: add `std::string query_param(const std::string& name) const;` and `bool has_query_param(const std::string& name) const;`.

src/ros2_medkit_gateway/include/ros2_medkit_gateway/plugins/plugin_http_types.hpp
```diff
@@ -0,0 +1,66 @@
...
+  const void * impl_;
+};
+
+/// Thin wrapper over an HTTP response.
+/// Hides the underlying HTTP library from plugin code.
+/// Constructed by the gateway per-request; plugins receive it by reference.
+class PluginResponse {
+ public:
+  /// Construct from opaque pointer to httplib::Response.
+  explicit PluginResponse(void * impl);
```
mfaferek93, line 53: **Major**

`PluginResponse` only exposes `send_json()` (always 200) and `send_error()`. A plugin cannot:
- Set a custom success status (201 Created, 204 No Content)
- Set response headers (Location, Cache-Control, custom X- headers)
- Send non-JSON content (plain text, binary, streaming)

If the goal is abstracting the HTTP lib so it can be swapped later, the response surface needs to be rich enough that plugins never reach through. Any non-SOVD plugin (health probes, Prometheus metrics, SSE streams) would need to bypass the abstraction entirely.

Fix: add `set_status(int)`, `set_header(string, string)`, `set_content(string, string)`.

src/ros2_medkit_gateway/include/ros2_medkit_gateway/plugins/plugin_http_types.hpp
```diff
@@ -0,0 +1,66 @@
...
+
+#pragma once
+
+#include <nlohmann/json.hpp>
+#include <string>
+
+namespace ros2_medkit_gateway {
+
+/// Thin read-only wrapper over an HTTP request.
+/// Hides the underlying HTTP library (currently cpp-httplib) from plugin code.
```
mfaferek93, line 23: **Minor**

Constructors take raw `void*`/`const void*`. Passing wrong pointer type = silent reinterpret. When the HTTP lib gets swapped (stated goal), these `static_cast<httplib::*>` in the .cpp are the migration points. void* means the compiler won't catch if you forget one.

Fix: make constructors private, friend-declare PluginManager. Or forward-declare `namespace httplib { class Request; class Response; }` and use typed pointers.

src/ros2_medkit_gateway/src/plugins/plugin_manager.cpp
```diff
@@ -260,20 +263,57 @@ void PluginManager::register_transport(std::unique_ptr<SubscriptionTransportProv
...
+            plugin_res.send_error(500, "x-medkit-plugin-error", "Internal plugin error");
+          } catch (...) {
+            RCLCPP_ERROR(rclcpp::get_logger("plugin_manager"), "Plugin '%s' handler threw unknown exception on %s",
+                         plugin_name.c_str(), full_pattern.c_str());
+            PluginResponse plugin_res(&res);
+            plugin_res.send_error(500, "x-medkit-plugin-error", "Internal plugin error");
+          }
+        };
+
+        if (route.method == "GET") {
```
mfaferek93, line 297: **Minor**

Exact string comparison: `route.method == "GET"`. A plugin returning `"get"` or `"Get"` silently skips registration (falls through to RCLCPP_WARN). No PATCH support either.

Fix: normalize to uppercase before matching, or use an enum.

src/ros2_medkit_gateway/src/plugins/plugin_http_types.cpp
```diff
@@ -0,0 +1,64 @@
...
+std::string PluginRequest::header(const std::string & name) const {
+  const auto & req = *static_cast<const httplib::Request *>(impl_);
+  return req.get_header_value(name);
+}
+
+std::string PluginRequest::path() const {
+  return static_cast<const httplib::Request *>(impl_)->path;
+}
+
+std::string PluginRequest::body() const {
```
mfaferek93: **Minor**

`body()` copies the entire request body string. For POST/PUT payloads this is wasteful. The `httplib::Request` outlives `PluginRequest` (both stack-scoped in the handler lambda), so returning `const std::string&` is safe.

Fix: return `const std::string&` for both `path()` and `body()`.

src/ros2_medkit_gateway/include/ros2_medkit_gateway/plugins/plugin_manager.hpp
```diff
@@ -97,10 +96,10 @@ class PluginManager {
 
   /**
    * @brief Register custom REST routes from all plugins
-   * @param server httplib server instance
+   * @param server Opaque pointer to httplib::Server (avoids httplib in header)
    * @param api_prefix API path prefix (e.g., "/api/v1")
    */
-  void register_routes(httplib::Server & server, const std::string & api_prefix);
+  void register_routes(void * server, const std::string & api_prefix);
```
mfaferek93: **Minor**

`void register_routes(void * server, ...)` - type safety regression from the old `httplib::Server&`. Only called internally by rest_server.cpp.

Fix: forward-declare `namespace httplib { class Server; }` and use `httplib::Server*`.

src/ros2_medkit_gateway/include/ros2_medkit_gateway/plugins/plugin_http_types.hpp
```diff
@@ -0,0 +1,66 @@
...
+/// Constructed by the gateway per-request; plugins receive it by const reference.
+class PluginRequest {
+ public:
+  /// Construct from opaque pointer to httplib::Request.
```
mfaferek93: **Minor**

Doxygen says "Construct from opaque pointer to httplib::Request" - names the implementation type directly in the abstraction header. Defeats the opacity for readers.

Fix: "Construct from opaque pointer to the underlying HTTP request implementation."

src/ros2_medkit_integration_tests/test/features/test_plugin_vendor_extensions.test.py
mfaferek93, line 1: **Minor**

4 locations still reference `register_routes()` which was replaced by `get_routes()` in API v5:
- gateway_params.yaml:319
- merge_pipeline.cpp:444
- introspection_provider.hpp:55
- test_plugin_vendor_extensions.test.py:20

src/ros2_medkit_gateway/src/plugins/plugin_manager.cpp
bburda: Fixed - full_pattern captured by value now, not by reference.

src/ros2_medkit_gateway/include/ros2_medkit_gateway/plugins/plugin_http_types.hpp
bburda: Acknowledged - no current plugin needs query params. Will add when a use case arises.

src/ros2_medkit_gateway/include/ros2_medkit_gateway/plugins/plugin_http_types.hpp
bburda: Acknowledged - current plugins are GET-only and use send_json (200) or send_error. Will add send_json(status, data) and set_header() when POST/PUT plugins need them.

src/ros2_medkit_gateway/include/ros2_medkit_gateway/plugins/plugin_http_types.hpp
bburda: Kept void* for PluginRequest/PluginResponse (plugins never construct these). Restored httplib::Server& with forward declaration for register_routes.

src/ros2_medkit_gateway/src/plugins/plugin_manager.cpp
bburda: Acknowledged - all current plugins use uppercase. Will normalize if needed.

src/ros2_medkit_gateway/src/plugins/plugin_http_types.cpp
bburda: Fixed - path() and body() now return const std::string&.

src/ros2_medkit_gateway/include/ros2_medkit_gateway/plugins/plugin_manager.hpp
bburda: Fixed - restored httplib::Server& with forward declaration in the header.

src/ros2_medkit_gateway/include/ros2_medkit_gateway/plugins/plugin_http_types.hpp
bburda: Fixed - changed to 'opaque HTTP request/response pointer (gateway-internal)'.

src/ros2_medkit_integration_tests/test/features/test_plugin_vendor_extensions.test.py
bburda: Fixed - updated gateway_params.yaml, introspection_provider.hpp, and merge_pipeline.cpp.

PR #350, opened 2026-04-03
fix(gateway): plugin config from YAML, thread-safety, CI stability

### LINKED ISSUE #349: [BUG] Plugin config from --params-file YAML not passed to plugins
# Bug report

### Steps to reproduce
1. Configure a plugin with config params in the YAML params file:
   ```yaml
   ros2_medkit_gateway:
     ros__parameters:
       plugins: ["my_plugin"]
       plugins.my_plugin.path: "/path/to/plugin.so"
       plugins.my_plugin.custom_key: "custom_value"
       plugins.my_plugin.mode: "testing"
   ```
2. Launch the gateway: `ros2 run ros2_medkit_gateway gateway_node --ros-args --params-file params.yaml`
3. The plugin's `configure()` method receives an empty JSON object `{}`

### Expected behavior

The plugin should receive `{"custom_key": "custom_value", "mode": "testing"}` in its `configure()` call, as specified in the YAML file.

### Actual behavior

The plugin receives `{}` (empty config). All plugin configuration values fall back to their defaults, regardless of what is in the YAML file.

**Root cause:** `extract_plugin_config()` in `gateway_node.cpp` reads from `get_node_options().parameter_overrides()`, which only contains parameters set programmatically via `NodeOptions::parameter_overrides(...)`. Parameters from `--params-file` are loaded into the ROS 2 global `rcl` context, NOT into `NodeOptions::parameter_overrides_`. The collection is always empty in production.

Unit tests pass because they set overrides programmatically via `options.parameter_overrides({...})`, which does populate the collection. The bug only manifests with `--params-file` (the production code path).

### Environment
- ros2_medkit version: main (3ee3484)
- ROS 2 distro: Jazzy (also affects Humble, Rolling)
- OS: Ubuntu 24.04

### Additional information

A failing test (`test_plugin_config.cpp`) that proves the bug is included in the fix PR.

bburda:
# Pull Request

## Summary

Originally a fix for `extract_plugin_config()` ignoring `--params-file` YAML params. Expanded to address multiple thread-safety and CI stability issues discovered during review and CI validation.

**Plugin config fix:**
- `extract_plugin_config()` read from `NodeOptions::parameter_overrides()` which is empty for `--params-file` params
- Added `declare_plugin_params_from_yaml()` (shared header `param_utils.hpp`) that reads from global rcl context

**Thread-safety fixes:**
- `NativeTopicSampler::sample_topic()`: dedicated callback group, sampling mutex, heap-allocated callback state (shared_ptr instead of stack references)
- Gateway switched from `SingleThreadedExecutor` to `MultiThreadedExecutor` - required because httplib threads call `create_generic_subscription` which races with executor iteration

**CI stability fixes:**
- Deferred trigger topic resolution: triggers created before publisher is discoverable now retry via 5s timer instead of silently having no subscription
- Demo node timer callback mutex applied to all 7 remaining nodes (was only on rpm_sensor)
- Dockerfile: re-add `libcpp-httplib-dev` to rosdep skip-keys for consistent vendored builds
- `sovd_service_interface`: VENDORED_DIR for humble + test stubs updated for PluginRequest/PluginResponse API

---

## Issue

- closes #349

---

## Type

- [x] Bug fix
- [ ] New feature or tests
- [ ] Breaking change
- [ ] Documentation only

---

## Testing

- `test_plugin_config` - proves YAML plugin params reach the gateway
- `NativeTopicSamplerThreadSafetyTest` - 3 tests: concurrent sampling, rapid timeout, timeout-no-crash
- `TriggerManagerTest` - 3 tests: deferred resolution create/retry/safety
- `test_triggers_late_publisher` - integration test for deferred trigger resolution
- All demo node SIGSEGV tests pass on humble (mutex fix)
- All 71 gateway unit tests pass

```
colcon test --packages-select ros2_medkit_gateway --ctest-args -LE "linter|integration"
```

---

## Checklist

### DIFF
--- a/src/ros2_medkit_gateway/src/main.cpp
+++ b/src/ros2_medkit_gateway/src/main.cpp
@@ -21,7 +21,14 @@ int main(int argc, char ** argv) {
 
   auto node = std::make_shared<ros2_medkit_gateway::GatewayNode>();
 
-  rclcpp::spin(node);
+  // MultiThreadedExecutor is required because cpp-httplib handler threads call
+  // Node::create_generic_subscription() for topic sampling. SingleThreadedExecutor
+  // does not synchronize external subscription creation with its internal iteration,
+  // causing non-deterministic SIGSEGV on rolling. All gateway callbacks are already
+  // protected by mutexes (EntityCache, LogManager, TriggerManager, etc.).
+  rclcpp::executors::MultiThreadedExecutor executor;
+  executor.add_node(node);
+  executor.spin();
 
   rclcpp::shutdown();
   return 0;


src/ros2_medkit_gateway/src/gateway_node.cpp
```diff
@@ -59,20 +62,82 @@ nlohmann::json parameter_to_json(const rclcpp::Parameter & param) {
...
 
-/// Extract per-plugin config from YAML parameter overrides.
+/// Declare plugin config parameters from the global --params-file YAML.
+///
+/// Parameters from --params-file go into the ROS 2 global rcl context,
+/// NOT into NodeOptions::parameter_overrides(). We must discover and
+/// declare them explicitly so list_parameters()/get_parameter() can find them.
+void declare_plugin_params_from_yaml(rclcpp::Node * node, const std::string & prefix, const std::string & path_key) {
+  auto rcl_ctx = node->get_node_base_interface()->get_context()->get_rcl_context();
+  rcl_params_t * global_params = nullptr;
```
mfaferek93: `global_params` leaks if an exception escapes outside the inner `try/catch` - e.g. from `std::string` construction of `yaml_node`/`pname` or `has_parameter()`. Low probability but trivial fix:

```cpp
auto cleanup = [](rcl_params_t* p) { if (p) rcl_yaml_node_struct_fini(p); };
std::unique_ptr<rcl_params_t, decltype(cleanup)> guard(global_params, cleanup);
```

Then remove the manual `rcl_yaml_node_struct_fini(global_params)` at the bottom.

src/ros2_medkit_gateway/src/gateway_node.cpp
```diff
@@ -59,20 +62,82 @@ nlohmann::json parameter_to_json(const rclcpp::Parameter & param) {
...
+      continue;
+    }
+    auto * node_p = &global_params->params[n];
+    for (size_t p = 0; p < node_p->num_params; ++p) {
+      std::string pname = node_p->parameter_names[p];
+      if (pname.rfind(prefix, 0) == 0 && pname != path_key && !node->has_parameter(pname)) {
+        // Determine the type from the rcl_variant_t so we can declare with the correct type
+        auto & val = node_p->parameter_values[p];
+        try {
+          if (val.string_value != nullptr) {
```
mfaferek93: Silent `catch(...)` hides config issues. At minimum log the param name so misconfigurations are debuggable:

```cpp
} catch (const std::exception& e) {
  RCLCPP_DEBUG(node->get_logger(), "Could not declare param '%s': %s", pname.c_str(), e.what());
}
```

src/ros2_medkit_gateway/test/test_plugin_config.cpp
```diff
@@ -0,0 +1,140 @@
...
+#include <rcl/arguments.h>
+#include <rcl_yaml_param_parser/parser.h>
+#include <rclcpp/rclcpp.hpp>
+
+namespace {
+
+/// Replicate the fix logic from gateway_node.cpp for testability.
+/// Declares plugin params from the global rcl context on a given node.
+void declare_plugin_params_from_yaml(rclcpp::Node * node, const std::string & prefix) {
+  auto rcl_ctx = node->get_node_base_interface()->get_context()->get_rcl_context();
```
mfaferek93: This is a copy of the production `declare_plugin_params_from_yaml()` that already diverges (no `path_key` param). If the prod function changes, this test still passes while prod is broken.

Consider extracting to a shared header (e.g. `param_utils.hpp`) so the test calls the real function. The function has no gateway-specific deps - just `rclcpp::Node*`.

src/ros2_medkit_gateway/src/gateway_node.cpp
```diff
@@ -59,20 +62,82 @@ nlohmann::json parameter_to_json(const rclcpp::Parameter & param) {
...
+  }
+
+  std::string node_name = node->get_name();
+  std::string node_fqn = node->get_fully_qualified_name();
+  for (size_t n = 0; n < global_params->num_nodes; ++n) {
+    std::string yaml_node = global_params->node_names[n];
+    if (yaml_node != node_name && yaml_node != node_fqn && yaml_node != "/**") {
+      continue;
+    }
+    auto * node_p = &global_params->params[n];
```
mfaferek93: `rcl_variant_t` array types (string_array, int_array, double_array, bool_array) are silently skipped. Not a blocker since no plugin currently uses arrays, but a `RCLCPP_WARN` when all scalar pointers are null would save future debugging time.

src/ros2_medkit_gateway/src/gateway_node.cpp
```diff
@@ -59,20 +62,82 @@ nlohmann::json parameter_to_json(const rclcpp::Parameter & param) {
   }
 }
 
-/// Extract per-plugin config from YAML parameter overrides.
+/// Declare plugin config parameters from the global --params-file YAML.
```
mfaferek93: PR description says "only a failing test, fix in follow-up" but this commit includes the fix. Update the PR body to reflect the actual scope.

src/ros2_medkit_gateway/src/gateway_node.cpp
bburda: Fixed - using `unique_ptr` with custom deleter now in `param_utils.hpp`.

src/ros2_medkit_gateway/src/gateway_node.cpp
bburda: Fixed - logging param name and exception message via `RCLCPP_DEBUG`.

src/ros2_medkit_gateway/test/test_plugin_config.cpp
bburda: Fixed - extracted to `param_utils.hpp`, test now calls the real function.

src/ros2_medkit_gateway/src/gateway_node.cpp
bburda: Fixed - added `RCLCPP_WARN` when all scalar pointers are null.

src/ros2_medkit_gateway/src/gateway_node.cpp
bburda: Fixed - PR description updated.

PR #363, opened 2026-04-11
fix(plugin_manager): declare YAML array params from --params-file

### LINKED ISSUE #362: [BUG] Plugin YAML array parameters are silently dropped
# Bug

### Steps to reproduce

1. Create a plugin config with one or more array-valued parameters under `plugins.<name>.<key>`, e.g.:

   ```yaml
   ros2_medkit_gateway:
     ros__parameters:
       plugins:
         - my_plugin
       plugins.my_plugin.path: /path/to/libmy_plugin.so
       plugins.my_plugin.string_list:
         - "a"
         - "b"
         - "c"
       plugins.my_plugin.int_list: [1, 2, 3]
   ```

2. Launch `gateway_node` with `--params-file` pointing at that yaml.
3. Observe the plugin config that gets passed to the plugin's `configure()` callback (either via plugin-side logging or the `Plugin 'my_plugin' config: N key(s)` line from `gateway_node.cpp`).

### Expected

`string_list` and `int_list` are visible to the plugin's `configure()` as JSON arrays, so the plugin can read them via `std::vector<std::string>` / `std::vector<int64_t>`. Scalar `path` is consumed separately by the plugin loader.

### Actual

Array-valued params are silently dropped. The gateway emits:

```
[WARN] [ros2_medkit_gateway]: Skipping param 'plugins.my_plugin.string_list': unsupported type (array?)
[WARN] [ros2_medkit_gateway]: Skipping param 'plugins.my_plugin.int_list': unsupported type (array?)
```

The plugin's `configure()` is invoked with a JSON object that is missing both keys.

### Root cause

`declare_plugin_params_from_yaml` in `src/ros2_medkit_gateway/include/ros2_medkit_gateway/param_utils.hpp` walks the parsed `rcl_params_t` values and only checks `string_value`, `bool_value`, `integer_value`, `double_value`. Anything else - including `string_array_value`, `bool_array_value`, `integer_array_value`, `double_array_value` - hits the final `else` branch which logs the warning and skips `declare_parameter`. Since the param is never declared, the subsequent `list_parameters()` in `extract_plugin_config` doesn't see it.

`parameter_to_json` in `src/gateway_node.cpp` already handles all four array types via `param.as_*_array()`, so the gap is purely on the YAML-declare side - once a param is declared it flows through to the plugin config correctly.

### Environment
- ros2_medkit version: main (commit 3377bb7 and newer)
- ROS 2 distro: Jazzy (also reproduces on Humble and Rolling - helper is distro-agnostic)
- OS: Ubuntu 24.04 / 22.04

### Fix

Add `string_array_value`, `bool_array_value`, `integer_array_value`, `double_array_value` branches to the `if / else if` chain, copying each into a `std::vector<T>` and calling `node->declare_parameter(pname, arr)`. Strictly additive change.

bburda:
# Pull Request

## Summary

`declare_plugin_params_from_yaml` in `src/ros2_medkit_gateway/include/ros2_medkit_gateway/param_utils.hpp` previously only handled scalar parameter types (`string`, `bool`, `integer`, `double`). YAML list values in plugin config were silently dropped with a "Skipping param: unsupported type (array?)" warning and the plugin's `configure()` never saw them.

This PR adds support for `string_array_value`, `bool_array_value`, `integer_array_value` and `double_array_value` so plugin configs can accept lists naturally, e.g.:

```yaml
plugins.my_plugin.peer_urls:
  - "http://host-a:8081"
  - "http://host-b:8081"
```

`parameter_to_json` in `src/gateway_node.cpp` already handles all four array types via the corresponding `rclcpp::Parameter::as_*_array()` getters, so once the param is declared it flows through to the plugin's JSON config unchanged.

---

## Issue

- closes #362

---

## Type

- [x] Bug fix
- [ ] New feature or tests
- [ ] Breaking change
- [ ] Documentation only

---

## Testing

Reproduced the original warning with a minimal plugin config that declares a `string_list` under `plugins.<name>.<key>`:

Before: gateway logs `Skipping param 'plugins.my_plugin.string_list': unsupported type (array?)` and the plugin's `configure()` receives a JSON object missing that key.

After: the param is declared, `list_parameters` sees it, `parameter_to_json` converts via `param.as_string_array()`, and the plugin receives a JSON array under the correct key.

Same verification with int / double / bool array types.

Strictly additive change - all existing scalar-type handling is untouched, so scalar params and plugins that don't use arrays see no behavior change. The unified "Skipping param" warning loses the `(array?)` hint because arrays are no longer the common case; anything that still hits the `else` branch is a genuinely unsupported type.

---

## Checklist

**Tests**: the helper is exercised indirectly through plugin-manager integration tests. A dedicated `declare_plugin_params_from_yaml` unit test would be a good follow-up but is out of scope for this fix - the change is small and type-symmetric with the existing scalar handling.

**Docs**: no public API changed (the helper is header-only and internal to gateway setup). No doc update needed.

### DIFF
--- a/src/ros2_medkit_gateway/include/ros2_medkit_gateway/param_utils.hpp
+++ b/src/ros2_medkit_gateway/include/ros2_medkit_gateway/param_utils.hpp
@@ -16,6 +16,7 @@
 
 #include <memory>
 #include <string>
+#include <vector>
 
 #include <rcl/arguments.h>
 #include <rcl_yaml_param_parser/parser.h>
@@ -64,11 +65,49 @@ inline void declare_plugin_params_from_yaml(rclcpp::Node * node, const std::stri
             node->declare_parameter(pname, static_cast<int64_t>(*val.integer_value));
           } else if (val.double_value != nullptr) {
             node->declare_parameter(pname, *val.double_value);
+          } else if (val.string_array_value != nullptr) {
+            // Preserve index alignment: map null elements to empty strings rather than
+            // dropping them, so plugin configs see the same length the YAML declared.
+            std::vector<std::string> arr;
+            arr.reserve(val.string_array_value->size);
+            for (size_t i = 0; i < val.string_array_value->size; ++i) {
+              const char * elem = val.string_array_value->data[i];
+              arr.emplace_back(elem != nullptr ? elem : "");
+            }
+            node->declare_parameter(pname, arr);
+          } else if (val.bool_array_value != nullptr) {
+            // Explicit loop (rather than range ctor) avoids UB when size==0 and
+            // values==nullptr - nullptr+0 pointer arithmetic is undefined in C++.
+            std::vector<bool> arr;
+            arr.reserve(val.bool_array_value->size);
+            for (size_t i = 0; i < val.bool_array_value->size; ++i) {
+              arr.push_back(val.bool_array_value->values[i]);
+            }
+            node->declare_parameter(pname, arr);
+          } else if (val.integer_array_value != nullptr) {
+            std::vector<int64_t> arr;
+            arr.reserve(val.integer_array_value->size);
+            for (size_t i = 0; i < val.integer_array_value->size; ++i) {
+              arr.push_back(val.integer_array_value->values[i]);
+            }
+            node->declare_parameter(pname, arr);
+          } else if (val.double_array_value != nullptr) {
+            std::vector<double> arr;
+            arr.reserve(val.double_array_value->size);
+            for (size_t i = 0; i < val.double_array_value->size; ++i) {
+              arr.push_back(val.double_array_value->values[i]);
+            }
+            node->declare_parameter(pname, arr);
           } else {
-            RCLCPP_WARN(node->get_logger(), "Skipping param '%s': unsupported type (array?)", pname.c_str());
+            RCLCPP_WARN(node->get_logger(), "Skipping param '%s': unsupported type", pname.c_str());
           }
+        } catch (const rclcpp::exceptions::ParameterAlreadyDeclaredException &) {
+          // Expected when a base class / earlier pass already declared the
+          // same parameter - ignore silently to keep idempotency.
         } catch (const std::exception & e) {
-          RCLCPP_DEBUG(node->get_logger(), "Could not declare param '%s': %s", pname.c_str(), e.what());
+          // Real errors (invalid type/value, RCL errors, bad YAML typos) must
+          // surface to the operator, not vanish into DEBUG.
+          RCLCPP_WARN(node->get_logger(), "Could not declare param '%s': %s", pname.c_str(), e.what());
         }
       }
     }
--- a/src/ros2_medkit_gateway/test/test_plugin_config.cpp
+++ b/src/ros2_medkit_gateway/test/test_plugin_config.cpp
@@ -26,10 +26,39 @@
 #include <cstdio>
 #include <fstream>
 #include <string>
+#include <utility>
 
 #include <rclcpp/rclcpp.hpp>
 
 #include "ros2_medkit_gateway/param_utils.hpp"
+
+namespace {
+
+/// RAII guard that removes a temp YAML file when it goes out of scope so
+/// aborted tests (ASSERT_* failures) do not leak files under /tmp.
+class ScopedYamlFile {
+ public:
+  ScopedYamlFile(std::string path, const std::string & content) : path_(std::move(path)) {
+    std::ofstream yaml(path_);
+    yaml << content;
+  }
+  ~ScopedYamlFile() {
+    std::remove(path_.c_str());
+  }
+  ScopedYamlFile(const ScopedYamlFile &) = delete;
+  ScopedYamlFile & operator=(const ScopedYamlFile &) = delete;
+  ScopedYamlFile(ScopedYamlFile &&) = delete;
+  ScopedYamlFile & operator=(ScopedYamlFile &&) = delete;
+
+  const std::string & path() const {
+    return path_;
+  }
+
+ private:
+  std::string path_;
+};
+
+}  // namespace
 
 /// Proves the bug and validates the fix using a lightweight rclcpp::Node.
 ///
@@ -39,21 +68,19 @@
 /// 4. Verifies NodeOptions::parameter_overrides() is empty (the bug)
 /// 5. Verifies declare_plugin_params_from_yaml() resolves the YAML values (the fix)
 TEST(PluginConfig, YamlPluginParamsReachGateway) {
-  // Write YAML with plugin config using the gateway node name
-  std::string yaml_path = "/tmp/test_plugin_config_" + std::to_string(getpid()) + ".yaml";
-  {
-    std::ofstream yaml(yaml_path);
-    yaml << "test_plugin_config_node:\n"
-         << "  ros__parameters:\n"
-         << "    plugins.my_plugin.custom_key: \"custom_value\"\n"
-         << "    plugins.my_plugin.mode: \"testing\"\n"
-         << "    plugins.my_plugin.timeout: 42\n"
-         << "    plugins.my_plugin.verbose: true\n"
-         << "    plugins.my_plugin.threshold: 3.14\n";
-  }
+  // Write YAML with plugin config using the gateway node name. ScopedYamlFile
+  // cleans up on scope exit so an ASSERT_* abort does not leak /tmp files.
+  ScopedYamlFile yaml("/tmp/test_plugin_config_" + std::to_string(getpid()) + ".yaml",
+                      "test_plugin_config_node:\n"
+                      "  ros__parameters:\n"
+                      "    plugins.my_plugin.custom_key: \"custom_value\"\n"
+                      "    plugins.my_plugin.mode: \"testing\"\n"
+                      "    plugins.my_plugin.timeout: 42\n"
+                      "    plugins.my_plugin.verbose: true\n"
+                      "    plugins.my_plugin.threshold: 3.14\n");
 
   // Init rclcpp with --params-file (simulates: ros2 run ... --ros-args --params-file ...)
-  const char * args[] = {"test", "--ros-args", "--params-file", yaml_path.c_str()};
+  const char * args[] = {"test", "--ros-args", "--params-file", yaml.path().c_str()};
   rclcpp::init(4, const_cast<char **>(args));
 
   // Create a lightweight node (no HTTP server, no DDS subscriptions)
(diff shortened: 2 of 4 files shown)


src/ros2_medkit_gateway/include/ros2_medkit_gateway/param_utils.hpp
```diff
@@ -64,8 +65,32 @@ inline void declare_plugin_params_from_yaml(rclcpp::Node * node, const std::stri
...
+          } else if (val.integer_array_value != nullptr) {
+            std::vector<int64_t> arr(
+                val.integer_array_value->values,
+                val.integer_array_value->values + val.integer_array_value->size);
+            node->declare_parameter(pname, arr);
+          } else if (val.double_array_value != nullptr) {
+            std::vector<double> arr(
+                val.double_array_value->values,
+                val.double_array_value->values + val.double_array_value->size);
+            node->declare_parameter(pname, arr);
```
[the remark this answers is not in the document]
bburda: Addressed: extended `test_plugin_config.cpp` with a new `YamlPluginArrayParamsReachGateway` test covering `string_array`, `bool_array`, `integer_array`, and `double_array` via `get_parameter(...).as_*_array()`.

src/ros2_medkit_gateway/include/ros2_medkit_gateway/param_utils.hpp
```diff
@@ -64,8 +65,32 @@ inline void declare_plugin_params_from_yaml(rclcpp::Node * node, const std::stri
             node->declare_parameter(pname, static_cast<int64_t>(*val.integer_value));
           } else if (val.double_value != nullptr) {
             node->declare_parameter(pname, *val.double_value);
+          } else if (val.string_array_value != nullptr) {
+            std::vector<std::string> arr;
+            arr.reserve(val.string_array_value->size);
+            for (size_t i = 0; i < val.string_array_value->size; ++i) {
+              if (val.string_array_value->data[i] != nullptr) {
+                arr.emplace_back(val.string_array_value->data[i]);
```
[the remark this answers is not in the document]
bburda: Fixed: null elements are now mapped to empty strings rather than dropped, so index alignment and array length match the YAML as declared.

src/ros2_medkit_gateway/include/ros2_medkit_gateway/param_utils.hpp
```diff
@@ -64,8 +65,32 @@ inline void declare_plugin_params_from_yaml(rclcpp::Node * node, const std::stri
...
+            node->declare_parameter(pname, arr);
+          } else if (val.integer_array_value != nullptr) {
+            std::vector<int64_t> arr(
+                val.integer_array_value->values,
+                val.integer_array_value->values + val.integer_array_value->size);
+            node->declare_parameter(pname, arr);
+          } else if (val.double_array_value != nullptr) {
+            std::vector<double> arr(
+                val.double_array_value->values,
+                val.double_array_value->values + val.double_array_value->size);
```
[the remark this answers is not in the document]
bburda: Fixed: replaced the pointer-range constructors with explicit index loops, so `size == 0` never triggers `nullptr + 0` pointer arithmetic.

src/ros2_medkit_gateway/include/ros2_medkit_gateway/param_utils.hpp
```diff
@@ -64,8 +65,41 @@ inline void declare_plugin_params_from_yaml(rclcpp::Node * node, const std::stri
...
+            for (size_t i = 0; i < val.double_array_value->size; ++i) {
+              arr.push_back(val.double_array_value->values[i]);
+            }
+            node->declare_parameter(pname, arr);
           } else {
-            RCLCPP_WARN(node->get_logger(), "Skipping param '%s': unsupported type (array?)", pname.c_str());
+            RCLCPP_WARN(node->get_logger(), "Skipping param '%s': unsupported type", pname.c_str());
           }
         } catch (const std::exception & e) {
           RCLCPP_DEBUG(node->get_logger(), "Could not declare param '%s': %s", pname.c_str(), e.what());
```
mfaferek93: DEBUG is suppressed by default, so any real error here just silently vanishes. This catch now covers four additional array-param paths, and `declare_parameter` throws a bunch of `std::exception` subclasses users actually want to see: `InvalidParameterTypeException`, `InvalidParameterValueException`, `RCLError`. A typo in their YAML gets zero feedback.

Suggestion - separate the expected case and WARN on everything else:

```cpp
} catch (const rclcpp::exceptions::ParameterAlreadyDeclaredException &) {
  // expected when a base class already declared the param, stay quiet
} catch (const std::exception & e) {
  RCLCPP_WARN(node->get_logger(), "Could not declare param '%s': %s", pname.c_str(), e.what());
}
```

src/ros2_medkit_gateway/test/test_plugin_config.cpp
```diff
@@ -98,6 +98,59 @@ TEST(PluginConfig, YamlPluginParamsReachGateway) {
...
+  ASSERT_TRUE(node->has_parameter("plugins.arr_plugin.thresholds"));
+  auto thresholds = node->get_parameter("plugins.arr_plugin.thresholds").as_double_array();
+  ASSERT_EQ(thresholds.size(), 3u);
+  EXPECT_DOUBLE_EQ(thresholds[0], 0.5);
+  EXPECT_DOUBLE_EQ(thresholds[1], 1.5);
+  EXPECT_DOUBLE_EQ(thresholds[2], 2.5);
+
+  node.reset();
+  rclcpp::shutdown();
+  std::remove(yaml_path.c_str());
```
mfaferek93: This line only runs on the success path - any `ASSERT_*` above it aborts the TEST function and leaks the `/tmp` YAML file. PID suffix avoids collisions, but it's still a slow-growing pile. Pre-existing in `YamlPluginParamsReachGateway` too, worth fixing both in one sweep.

Simplest fix - a local RAII scope guard:

```cpp
struct ScopeExit { std::function<void()> fn; ~ScopeExit() { fn(); } };
ScopeExit cleanup{[&] { std::remove(yaml_path.c_str()); }};
```

Or convert the file to a gtest fixture with `TearDown()`.

src/ros2_medkit_gateway/test/test_plugin_config.cpp
```diff
@@ -98,6 +98,59 @@ TEST(PluginConfig, YamlPluginParamsReachGateway) {
...
+  rclcpp::init(4, const_cast<char **>(args));
+
+  auto node = std::make_shared<rclcpp::Node>("test_plugin_config_arrays_node");
+  ros2_medkit_gateway::declare_plugin_params_from_yaml(node.get(), "plugins.arr_plugin.");
+
+  ASSERT_TRUE(node->has_parameter("plugins.arr_plugin.peer_urls"));
+  auto peer_urls = node->get_parameter("plugins.arr_plugin.peer_urls").as_string_array();
+  ASSERT_EQ(peer_urls.size(), 2u);
+  EXPECT_EQ(peer_urls[0], "http://host-a:8081");
+  EXPECT_EQ(peer_urls[1], "http://host-b:8081");
```
mfaferek93, line 151: The production code has load-bearing special handling for null string elements (`elem != nullptr ? elem : ""`) to preserve index alignment - the comment in `param_utils.hpp` explicitly calls out that plugins rely on this. Nothing here tests it. A refactor that silently drops nulls instead of substituting `""` would pass CI.

YAML can't really express a null string inside a list, but at minimum add a case that exercises empty-string elements to lock in related semantics:

```yaml
plugins.arr_plugin.labels: ["", "mid", ""]
```

Ideally a separate unit test that builds `rcl_variant_t` directly with `data[i] = nullptr`, but that's a bigger lift.

src/ros2_medkit_gateway/test/test_plugin_config.cpp
```diff
@@ -98,6 +98,59 @@ TEST(PluginConfig, YamlPluginParamsReachGateway) {
...
+  EXPECT_EQ(timeouts[0], 100);
+  EXPECT_EQ(timeouts[1], 200);
+  EXPECT_EQ(timeouts[2], 300);
+
+  ASSERT_TRUE(node->has_parameter("plugins.arr_plugin.thresholds"));
+  auto thresholds = node->get_parameter("plugins.arr_plugin.thresholds").as_double_array();
+  ASSERT_EQ(thresholds.size(), 3u);
+  EXPECT_DOUBLE_EQ(thresholds[0], 0.5);
+  EXPECT_DOUBLE_EQ(thresholds[1], 1.5);
+  EXPECT_DOUBLE_EQ(thresholds[2], 2.5);
```
mfaferek93, line 183: Nothing in this test verifies the negative path - that an actually unsupported YAML type still falls through to the "Skipping param" branch. If a future refactor accidentally starts accepting everything in the `else`, CI would still pass.

Add one assertion with an unsupported shape (nested map is probably easiest in YAML):

```yaml
plugins.arr_plugin.weird:
  nested: value
```

Then `EXPECT_FALSE(node->has_parameter("plugins.arr_plugin.weird"));`. Pairs naturally with a log-capture check on the warning if you have that infrastructure.

tsan_suppressions.txt
```diff
@@ -8,6 +8,9 @@
 # FastDDS internal thread pool and transport layer
 race:eprosima::fastdds::*
 race:eprosima::fastrtps::*
+# FastCDR type serialization races against discovery thread during
+# rclcpp service Client creation. Upstream race in libfastcdr, not our code.
+race:eprosima::fastcdr::*
```
mfaferek93: Two concerns on this suppression:

**1. No upstream reference.** The comment says "Upstream race in libfastcdr, not our code" but there's no link to an eProsima/Fast-CDR issue, no committed TSan stack, and no reproducer. For a suppression that mutes a safety-adjacent library we depend on, linking an upstream bug (or filing one) should be standard hygiene - otherwise we have no signal for when to retire this.

**2. The wildcard is broader than needed.** `race:eprosima::fastcdr::*` silently masks any future race we introduce via our own FastCDR misuse (e.g. accidentally sharing a `Cdr` across threads in our code). Narrow to the specific symbol on the reported TSan stack, e.g. `race:eprosima::fastcdr::Cdr::serialize*`.

Also - this change is scope-unrelated to the YAML array params fix. A follow-up PR with the upstream link + a narrower pattern would be cleaner for bisect.

src/ros2_medkit_gateway/include/ros2_medkit_gateway/param_utils.hpp
bburda: Fixed. Split the catch: `ParameterAlreadyDeclaredException` stays silent (idempotent case when a base class already declared the same name), everything else goes to `RCLCPP_WARN` so YAML typos, invalid types and RCL errors surface to the operator instead of vanishing into DEBUG.

src/ros2_medkit_gateway/test/test_plugin_config.cpp
bburda: Fixed. Added a small `ScopedYamlFile` RAII helper that removes the temp file on scope exit, applied to both `YamlPluginParamsReachGateway` and the new array test. Went with RAII over a gtest fixture because the tests also need `rclcpp::init`/`shutdown` around the test body, which fixture SetUp/TearDown would make harder to read - the file cleanup is the only shared concern.

src/ros2_medkit_gateway/test/test_plugin_config.cpp
bburda: Fixed with the empty-string case: `plugins.arr_plugin.labels: ["", "mid", ""]`. This exercises the `elem != nullptr ? elem : ""` branch via `rcl_yaml_param_parser` the only way YAML can reach it - rcl stores empty strings with a non-null `data[i]` pointing at an empty C string, so technically this tests the `elem != nullptr` arm rather than the `""` fallback. The true null-pointer case needs an `rcl_variant_t` constructed by hand, which would mean refactoring the helper to accept an injection point - out of scope here. The comment in `param_utils.hpp` still documents why the fallback exists.

src/ros2_medkit_gateway/test/test_plugin_config.cpp
bburda: Added `YamlPluginPrefixScopingAndDottedKeys` which verifies two related guarantees the positive-path tests miss: (1) params outside the requested prefix are not declared (guards the `rfind(prefix, 0) == 0` filter) and (2) dotted-key YAML gets flattened into a single scalar with no intermediate parent parameter. Honest about the limitation in the test comment: the literal `else` branch in `declare_plugin_params_from_yaml` fires only when `rcl_variant_t` has all `*_value` fields null, which `rcl_yaml_param_parser` never emits for well-formed YAML. Exercising it for real would require constructing an `rcl_variant_t` by hand and injecting it, which I did not do in this PR.

tsan_suppressions.txt
bburda: Both concerns addressed.

1. Upstream reference added: https://github.com/eProsima/Fast-DDS/issues/6340 - exact same race stack (`fastcdr::Cdr::serialize_array` write vs `deserialize_array` read on `TopicPayloadPool::do_get_payload` memory), still open. Now linked from the suppression comment, gives us a retirement signal.

2. Pattern narrowed from `race:eprosima::fastcdr::*` to `race:eprosima::fastcdr::Cdr::serialize*` + `race:eprosima::fastcdr::Cdr::deserialize*`. Matches the exact symbols on the reported TSan stack, so a future race we introduce via our own FastCDR misuse (sharing a `Cdr` across threads, etc.) would still fire.

On scope: I considered splitting to a follow-up PR but kept it here intentionally. The suppression was added in direct response to TSan failing on this PR branch after I added the YAML array params change, and the preference on this repo is to fix flakies when you find them rather than defer them. A follow-up would also create an ordering constraint (suppression must land first or this PR re-reds) which is uglier for bisect than a single commit with a narrow, linked suppression. Happy to split if you feel strongly - just say the word.

mfaferek93 (approved):

PR #365, opened 2026-04-11
feat(plugins): add ros2_medkit_opcua plugin

### LINKED ISSUE #364: add OPC-UA gateway plugin (ros2_medkit_opcua)
## Summary

Proposal to add a new plugin package `ros2_medkit_opcua` that bridges OPC-UA capable PLCs into the SOVD entity tree, exposing PLC tags as standard SOVD data endpoints and mapping PLC alarms to SOVD faults via `ros2_medkit_msgs`.

Following the pattern set by `ros2_medkit_graph_provider`, the plugin lives under `src/ros2_medkit_plugins/` and implements `GatewayPlugin` + `IntrospectionProvider` against the current `get_routes()` plugin API.

## Motivation

Many industrial robotics deployments are hybrid: ROS 2 robots operating alongside PLC-controlled fixtures, conveyors, safety systems, and process equipment. Today there is no open, SOVD-native way to observe PLC state from the same diagnostic surface that exposes ROS 2 topics, nodes, and faults.

An OPC-UA plugin fills this gap:

- **Unified diagnostics** - a single SOVD API exposes both ROS 2 entities and PLC tags
- **Fault correlation across layers** - a pump failure reported by a PLC alarm becomes a first-class SOVD fault in the same database that carries ROS 2 diagnostics
- **Open standards stack** - OPC-UA is the dominant open industrial protocol, and pairing it with ros2_medkit's SOVD gateway gives users a fully open-source path for mixed ROS 2 + PLC diagnostics
- **Real-world demand** - integrators working on ROS 2 + industrial PLC deployments have asked for exactly this path, and upcoming OPC-UA-related community engagement (ROSCon 2026 talk proposal with Siemens) would benefit from an in-tree reference implementation
- **Non-trivial plugin API showcase** - exercises `get_routes()`, `IntrospectionProvider`, `PluginContext::register_capability`, fault service clients, and `rclcpp` publishers from `set_context()`. Complements `ros2_medkit_graph_provider` as a reference for future plugin authors

## Scope

A single new colcon package `ros2_medkit_opcua` under `src/ros2_medkit_plugins/`.

mfaferek93:
## Summary

Adds a new `ros2_medkit_opcua` plugin under `src/ros2_medkit_plugins/` that bridges OPC-UA capable PLCs into the SOVD entity tree, following the same pattern as `ros2_medkit_graph_provider`. The plugin implements `GatewayPlugin` + `IntrospectionProvider` against the current `get_routes()` API and works with any compliant OPC-UA server (OpenPLC, Siemens S7, Beckhoff TwinCAT, Allen-Bradley via Kepware, KUKA KR C5, etc.) by just changing the YAML node map. Closes #364.

## What the plugin does

Connects over `opc.tcp`, generates SOVD entities from a YAML node map, exposes live values via `x-plc-data`, accepts writes with type and range validation via `x-plc-operations`, reports connection status via `x-plc-status`, maps threshold PLC alarms to SOVD faults via `ros2_medkit_msgs` services, and optionally bridges numeric values to ROS 2 `std_msgs/Float32` topics. Supports all four OPC 10000-6 node identifier types (numeric, string, GUID, opaque) so one binary handles every vendor; `config/tank_demo_nodes.yaml` ships ready-to-paste examples.

## Tests and CI

79 unit tests across a Humble + Jazzy + Rolling matrix plus 16 Docker integration assertions against an OpenPLC tank demo, all green. A new `.github/workflows/opcua-plugin.yml` runs only when plugin files, gateway plugin API headers, or `ros2_medkit_msgs` change (with a nightly safety-net cron), and the base `ci.yml` skips this package via `--packages-skip` so unrelated PRs pay zero extra cost. Test domain 220-229 is carved from the `integration_tests` range.

## Open items (tracked separately)

Intentionally kept out of this PR: #366 vendor `open62541pp` before adding the package to rosdistro release (today `FetchContent` is fine because the plugin is intentionally not in `release.packages`), #367 certificate-based OPC-UA authentication (current version is Anonymous + SecurityPolicy=None, LAN-only), and #368 auto-browse CLI (deferred pending validated demand - UaExpert and `python -m asyncua.tools.uals` cover the workflow today).

### DIFF
--- a/.github/workflows/opcua-plugin.yml
+++ b/.github/workflows/opcua-plugin.yml
@@ -29,6 +29,87 @@ on:
     - cron: '0 4 * * *'
 
 jobs:
+  unit-tests:
+    name: Unit tests (${{ matrix.ros_distro }})
+    runs-on: ubuntu-latest
+    strategy:
+      fail-fast: false
+      matrix:
+        include:
+          - ros_distro: humble
+            os_image: ubuntu:jammy
+          - ros_distro: jazzy
+            os_image: ubuntu:noble
+          - ros_distro: rolling
+            os_image: ubuntu:noble
+    continue-on-error: ${{ matrix.ros_distro == 'rolling' }}
+    container:
+      image: ${{ matrix.os_image }}
+    timeout-minutes: 60
+    defaults:
+      run:
+        shell: bash
+    steps:
+      - name: Install Git
+        run: |
+          apt-get update
+          apt-get install -y git
+
+      - name: Checkout repository
+        uses: actions/checkout@v4
+
+      - name: Set up ROS 2 ${{ matrix.ros_distro }}
+        uses: ros-tooling/setup-ros@v0.7
+        with:
+          required-ros-distributions: ${{ matrix.ros_distro }}
+
+      - name: Install ccache
+        run: apt-get install -y ccache
+
+      - name: Cache ccache
+        uses: actions/cache@v4
+        with:
+          path: /root/.cache/ccache
+          key: ccache-opcua-${{ matrix.ros_distro }}-${{ github.sha }}
+          restore-keys: |
+            ccache-opcua-${{ matrix.ros_distro }}-
+
+      - name: Install dependencies
+        run: |
+          apt-get update
+          apt-get install -y ros-${{ matrix.ros_distro }}-test-msgs libyaml-cpp-dev libssl-dev
+          source /opt/ros/${{ matrix.ros_distro }}/setup.bash
+          rosdep update
+          rosdep install --from-paths src --ignore-src -y \
+            --skip-keys='ament_cmake_clang_format ament_cmake_clang_tidy test_msgs nav2_msgs'
+
+      - name: Build ros2_medkit_opcua (and upstream deps)
+        env:
+          CCACHE_DIR: /root/.cache/ccache
+          CCACHE_MAXSIZE: 500M
+          CCACHE_SLOPPINESS: pch_defines,time_macros
+        run: |
+          source /opt/ros/${{ matrix.ros_distro }}/setup.bash
+          colcon build --symlink-install \
+            --packages-up-to ros2_medkit_opcua \
+            --packages-skip vda5050_agent ros2_medkit_vda5050_msgs \
+            --cmake-args -DCMAKE_BUILD_TYPE=Release \
+            --event-handlers console_direct+
+          ccache -s
+
+      - name: Run unit tests
+        timeout-minutes: 15
+        run: |
+          source /opt/ros/${{ matrix.ros_distro }}/setup.bash
+          colcon test --return-code-on-test-failure \
+            --packages-select ros2_medkit_opcua \
+            --ctest-args -LE linter \
+            --event-handlers console_direct+
+
+      - name: Show test results
+        if: always()
+        run: colcon test-result --verbose
+
   integration:
     name: Integration (OpenPLC)
     runs-on: ubuntu-latest
--- a/.github/workflows/opcua-plugin.yml
+++ b/.github/workflows/opcua-plugin.yml
@@ -80,8 +80,13 @@ jobs:
           apt-get install -y ros-${{ matrix.ros_distro }}-test-msgs libyaml-cpp-dev libssl-dev
           source /opt/ros/${{ matrix.ros_distro }}/setup.bash
           rosdep update
+          # Only skip nav2_msgs because vda5050_agent declares it as a dep but
+          # the apt package is not available on all distros. We do NOT skip
+          # ament_cmake_clang_format or test_msgs here: upstream medkit
+          # packages (ros2_medkit_serialization, gateway) require them at
+          # configure time via find_package.
           rosdep install --from-paths src --ignore-src -y \
-            --skip-keys='ament_cmake_clang_format ament_cmake_clang_tidy test_msgs nav2_msgs'
+            --skip-keys='nav2_msgs'
 
       - name: Build ros2_medkit_opcua (and upstream deps)
         env:
@@ -90,9 +95,11 @@ jobs:
           CCACHE_SLOPPINESS: pch_defines,time_macros
         run: |
           source /opt/ros/${{ matrix.ros_distro }}/setup.bash
+          # --packages-up-to naturally excludes unrelated packages
+          # (vda5050_agent, discovery plugins, etc.) because they are not in
+          # the dependency chain of ros2_medkit_opcua.
           colcon build --symlink-install \
             --packages-up-to ros2_medkit_opcua \
-            --packages-skip vda5050_agent ros2_medkit_vda5050_msgs \
             --cmake-args -DCMAKE_BUILD_TYPE=Release \
             --event-handlers console_direct+
           ccache -s
--- a/.github/workflows/opcua-plugin.yml
+++ b/.github/workflows/opcua-plugin.yml
@@ -1,0 +1,134 @@
+name: OPC-UA Plugin Integration
+
+# Runs the OpenPLC-based Docker integration suite for ros2_medkit_opcua.
+# Triggers only when the plugin itself, the gateway plugin/provider API, or
+# ros2_medkit_msgs change. A nightly cron provides a safety net against silent
+# breaks that slip past the path filter.
+
+on:
+  pull_request:
+    branches: [main]
+    paths:
+      - 'src/ros2_medkit_plugins/ros2_medkit_opcua/**'
+      - 'src/ros2_medkit_gateway/include/ros2_medkit_gateway/plugins/**'
+      - 'src/ros2_medkit_gateway/include/ros2_medkit_gateway/providers/**'
+      - 'src/ros2_medkit_gateway/include/ros2_medkit_gateway/http/error_codes.hpp'
+      - 'src/ros2_medkit_msgs/**'
+      - '.github/workflows/opcua-plugin.yml'
+  push:
+    branches: [main]
+    paths:
+      - 'src/ros2_medkit_plugins/ros2_medkit_opcua/**'
+      - 'src/ros2_medkit_gateway/include/ros2_medkit_gateway/plugins/**'
+      - 'src/ros2_medkit_gateway/include/ros2_medkit_gateway/providers/**'
+      - 'src/ros2_medkit_gateway/include/ros2_medkit_gateway/http/error_codes.hpp'
+      - 'src/ros2_medkit_msgs/**'
+      - '.github/workflows/opcua-plugin.yml'
+  schedule:
+    # Daily at 04:00 UTC - safety net against breaks that slip past the path filter.
+    - cron: '0 4 * * *'
+
+jobs:
+  integration:
+    name: Integration (OpenPLC)
+    runs-on: ubuntu-latest
+    timeout-minutes: 30
+    steps:
+      - name: Checkout repository
+        uses: actions/checkout@v4
+
+      - name: Build OpenPLC container
+        run: |
+          docker build \
+            -t openplc-tank \
+            src/ros2_medkit_plugins/ros2_medkit_opcua/docker/openplc/
+
+      - name: Build gateway + OPC-UA plugin image
+        run: |
+          docker build \
+            -f src/ros2_medkit_plugins/ros2_medkit_opcua/docker/Dockerfile.gateway \
+            -t gateway-opcua .
+
+      - name: Start OpenPLC
+        timeout-minutes: 3
+        run: |
+          docker network create plc-demo
+          docker run -d --name openplc --network plc-demo openplc-tank
+          for i in $(seq 1 60); do
+            if docker logs openplc 2>&1 | grep -q "PLC State: RUNNING"; then
+              echo "OpenPLC running after $((i * 2))s"
+              break
+            fi
+            if [ "$i" -eq 60 ]; then
+              echo "FAIL: OpenPLC did not start"
+              docker logs openplc 2>&1 | tail -20
+              exit 1
+            fi
+            sleep 2
+          done
+
+      - name: Start gateway
+        run: |
+          docker run -d --name gateway --network plc-demo -p 8080:8080 \
+            -e ROS_DOMAIN_ID=60 \
+            -e OPCUA_ENDPOINT_URL="opc.tcp://openplc:4840/openplc/opcua" \
+            -e OPCUA_NODE_MAP_PATH="/config/tank_nodes.yaml" \
+            gateway-opcua \
+            bash -c "
+              mkdir -p /var/lib/ros2_medkit/rosbags /config
+              echo 'manifest_version: \"1.0\"' > /config/manifest.yaml
+              source /opt/ros/jazzy/setup.bash && source /root/ws/install/setup.bash
+              PLUGIN_PATH=\$(find /root/ws/install -name 'libros2_medkit_opcua_plugin.so' | head -1)
+              ros2 run ros2_medkit_gateway gateway_node \
+                --ros-args --params-file /config/gateway_params.yaml \
+                -p plugins.opcua.path:=\$PLUGIN_PATH \
+                -p discovery.mode:=hybrid \
+                -p discovery.manifest_path:=/config/manifest.yaml \
+                -p discovery.manifest_strict_validation:=false"
+
+      - name: Wait for PLC entities and live data
+        timeout-minutes: 3
+        run: |
+          echo "Waiting for entity discovery..."
+          for i in $(seq 1 60); do
+            if curl -sf http://localhost:8080/api/v1/apps 2>/dev/null | \
+               jq -e '.items | map(.id) | contains(["tank_process"])' >/dev/null 2>&1; then
+              echo "Entities discovered after $((i * 2))s"
+              break
+            fi
+            if [ "$i" -eq 60 ]; then
+              echo "Timeout waiting for entities"
+              docker logs gateway 2>&1 | tail -20
+              exit 1
+            fi
+            sleep 2
+          done
+          echo "Waiting for PLC data to flow..."
+          for i in $(seq 1 30); do
+            if curl -sf http://localhost:8080/api/v1/components/openplc_runtime/x-plc-status 2>/dev/null | \
+               jq -e '.connected == true and .poll_count > 0' >/dev/null 2>&1; then
+              echo "PLC data flowing after extra $((i * 2))s"
+              exit 0
+            fi
+            sleep 2
+          done
+          echo "WARNING: PLC not connected yet, running tests anyway"
+          curl -sf http://localhost:8080/api/v1/components/openplc_runtime/x-plc-status 2>/dev/null | jq . || true
+          docker logs gateway 2>&1 | tail -10
+
+      - name: Run integration tests
+        run: bash src/ros2_medkit_plugins/ros2_medkit_opcua/docker/scripts/run_integration_tests.sh
+
+      - name: Dump gateway logs on failure
+        if: failure()
+        run: |
+          echo "=== OpenPLC logs ==="
+          docker logs openplc 2>&1 | tail -60 || true
+          echo "=== Gateway logs ==="
+          docker logs gateway 2>&1 | tail -60 || true
+
+      - name: Cleanup
+        if: always()
+        run: |
+          docker rm -f gateway openplc 2>/dev/null || true
+          docker network rm plc-demo 2>/dev/null || true
--- a/.github/workflows/opcua-plugin.yml
+++ b/.github/workflows/opcua-plugin.yml
@@ -1,5 +1,8 @@
 name: OPC-UA Plugin Integration
 
+permissions:
+  contents: read
+
 # Runs the OpenPLC-based Docker integration suite for ros2_medkit_opcua.
 # Triggers only when the plugin itself, the gateway plugin/provider API, or
 # ros2_medkit_msgs change. A nightly cron provides a safety net against silent
--- a/src/ros2_medkit_plugins/ros2_medkit_opcua/CMakeLists.txt
+++ b/src/ros2_medkit_plugins/ros2_medkit_opcua/CMakeLists.txt
@@ -31,8 +31,13 @@ find_package(ros2_medkit_msgs REQUIRED)
 find_package(rclcpp REQUIRED)
 find_package(std_msgs REQUIRED)
 find_package(nlohmann_json REQUIRED)
-find_package(yaml-cpp REQUIRED)
 find_package(OpenSSL REQUIRED)
+
+# yaml-cpp via the medkit compat helper: jazzy exports the native
+# ``yaml-cpp::yaml-cpp`` cmake target via ``yaml_cpp_vendor``; humble may fall
+# back to the system library. The macro handles both cases and fails with a
+# clear error if ``yaml_cpp_vendor`` is missing.
+medkit_find_yaml_cpp()
 
 # ---- open62541pp via FetchContent ----
 # open62541pp v0.16.0 is MPLv2 licensed. Linking from Apache 2.0 code to MPLv2
@@ -97,12 +102,6 @@ medkit_target_dependencies(ros2_medkit_opcua_plugin
   std_msgs
 )
 
-# open62541pp headers emit -Wconversion warnings that would propagate under
-# the strict project-wide warning flags from ROS2MedkitWarnings. Relax only
-# on this target; our own code still compiles with full warnings via the
-# ROS2MedkitWarnings include above.
-target_compile_options(ros2_medkit_opcua_plugin PRIVATE -Wno-conversion)
-
 # Allow unresolved symbols - they resolve from the host process at runtime
 target_link_options(ros2_medkit_opcua_plugin PRIVATE
   -Wl,--unresolved-symbols=ignore-all
@@ -111,7 +110,7 @@ target_link_options(ros2_medkit_opcua_plugin PRIVATE
 target_link_libraries(ros2_medkit_opcua_plugin
   open62541pp::open62541pp
   nlohmann_json::nlohmann_json
-  yaml-cpp
+  yaml-cpp::yaml-cpp
   OpenSSL::SSL OpenSSL::Crypto
 )
 
@@ -153,11 +152,10 @@ if(BUILD_TESTING)
   target_include_directories(test_node_map PRIVATE
     ${CMAKE_CURRENT_SOURCE_DIR}/include
   )
-  target_compile_options(test_node_map PRIVATE -Wno-conversion)
   target_link_libraries(test_node_map
     open62541pp::open62541pp
     nlohmann_json::nlohmann_json
-    yaml-cpp
+    yaml-cpp::yaml-cpp
   )
   medkit_set_test_domain(test_node_map)
 
@@ -168,7 +166,6 @@ if(BUILD_TESTING)
   target_include_directories(test_opcua_client PRIVATE
     ${CMAKE_CURRENT_SOURCE_DIR}/include
   )
-  target_compile_options(test_opcua_client PRIVATE -Wno-conversion)
   target_link_libraries(test_opcua_client
     open62541pp::open62541pp
     nlohmann_json::nlohmann_json
--- a/src/ros2_medkit_plugins/ros2_medkit_opcua/CMakeLists.txt
+++ b/src/ros2_medkit_plugins/ros2_medkit_opcua/CMakeLists.txt
@@ -1,4 +1,4 @@
-# Copyright 2026 mfaferek
+# Copyright 2026 mfaferek93
 #
 # Licensed under the Apache License, Version 2.0 (the "License");
 # you may not use this file except in compliance with the License.
--- a/src/ros2_medkit_plugins/ros2_medkit_opcua/CMakeLists.txt
+++ b/src/ros2_medkit_plugins/ros2_medkit_opcua/CMakeLists.txt
@@ -145,6 +145,30 @@ if(BUILD_TESTING)
   include(ROS2MedkitTestDomain)
   medkit_init_test_domains(START 220 END 229)
 
+  ament_add_gtest(test_opcua_plugin
+    test/test_opcua_plugin.cpp
+    src/opcua_plugin.cpp
+    src/opcua_client.cpp
+    src/node_map.cpp
+    src/opcua_poller.cpp
+  )
+  target_include_directories(test_opcua_plugin PRIVATE
+    ${CMAKE_CURRENT_SOURCE_DIR}/include
+  )
+  medkit_target_dependencies(test_opcua_plugin
+    ros2_medkit_msgs
+    ros2_medkit_gateway
+    rclcpp
+    std_msgs
+  )
+  target_link_libraries(test_opcua_plugin
+    open62541pp::open62541pp
+    nlohmann_json::nlohmann_json
+    yaml-cpp::yaml-cpp
+    OpenSSL::SSL OpenSSL::Crypto
+  )
+  medkit_set_test_domain(test_opcua_plugin)
+
   ament_add_gtest(test_node_map
     test/test_node_map.cpp
     src/node_map.cpp
--- a/src/ros2_medkit_plugins/ros2_medkit_opcua/CMakeLists.txt
+++ b/src/ros2_medkit_plugins/ros2_medkit_opcua/CMakeLists.txt
@@ -1,0 +1,181 @@
+# Copyright 2026 mfaferek
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+cmake_minimum_required(VERSION 3.16)
+project(ros2_medkit_opcua)
+
+set(CMAKE_CXX_STANDARD 17)
+set(CMAKE_CXX_STANDARD_REQUIRED ON)
+
+# Shared cmake modules (multi-distro compat)
+find_package(ros2_medkit_cmake REQUIRED)
+include(ROS2MedkitCompat)
+include(ROS2MedkitCcache)
+include(ROS2MedkitSanitizers)
+include(ROS2MedkitWarnings)
+
+find_package(ament_cmake REQUIRED)
+find_package(ros2_medkit_gateway REQUIRED)
+find_package(ros2_medkit_msgs REQUIRED)
+find_package(rclcpp REQUIRED)
+find_package(std_msgs REQUIRED)
+find_package(nlohmann_json REQUIRED)
+find_package(yaml-cpp REQUIRED)
+find_package(OpenSSL REQUIRED)
+
+# ---- open62541pp via FetchContent ----
+# open62541pp v0.16.0 is MPLv2 licensed. Linking from Apache 2.0 code to MPLv2
+# libraries is permitted - MPLv2 is a weak per-file copyleft that only requires
+# source disclosure for modifications to the MPLv2 files themselves. We ship
+# the library unmodified via FetchContent pinned to a specific upstream tag.
+include(FetchContent)
+fetchcontent_declare(
+  open62541pp
+  GIT_REPOSITORY https://github.com/open62541pp/open62541pp.git
+  GIT_TAG v0.16.0
+  GIT_SUBMODULES_RECURSE ON
+)
+# Build open62541 as part of open62541pp, with -fPIC for shared library
+set(UAPP_INTERNAL_OPEN62541 ON CACHE BOOL "" FORCE)
+set(CMAKE_POSITION_INDEPENDENT_CODE ON CACHE BOOL "" FORCE)
+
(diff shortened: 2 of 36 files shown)


bburda (commented):
**Out-of-diff findings (files not changed in this PR):**

**[critical] `docs/design/index.rst` - missing toctree entry + symlink**
Package ships 180-line `design/index.rst` but no `ros2_medkit_opcua/index` entry in `docs/design/index.rst` toctree, and no symlink `docs/design/ros2_medkit_opcua -> ../../src/ros2_medkit_plugins/ros2_medkit_opcua/design`. Sphinx will not build the design doc.

**[critical] `docs/changelog.rst` - missing CHANGELOG aggregation**
All other packages have `.. include::` directives in `docs/changelog.rst`. The opcua CHANGELOG (20 lines, 0.4.0 entry) will not be published. Append `.. include:: ../src/ros2_medkit_plugins/ros2_medkit_opcua/CHANGELOG.rst`.

src/ros2_medkit_plugins/ros2_medkit_opcua/src/opcua_plugin_exports.cpp
```diff
@@ -0,0 +1,30 @@
...
+  return PLUGIN_API_VERSION;
+}
+
+extern "C" GATEWAY_PLUGIN_EXPORT GatewayPlugin * create_plugin() {
+  return new ros2_medkit_gateway::OpcuaPlugin();
+}
+
+extern "C" GATEWAY_PLUGIN_EXPORT IntrospectionProvider * get_introspection_provider(GatewayPlugin * plugin) {
+  return static_cast<ros2_medkit_gateway::OpcuaPlugin *>(plugin);
+}
```
bburda, line 30: **[critical] Plugin bypasses DataProvider/OperationProvider/FaultProvider framework**

The gateway on `main` has a fully functioning plugin entity framework (PR 355 + hardening):
- `DataProvider` (`list_data`, `read_data`, `write_data`) with `tl::expected`
- `OperationProvider` (`list_operations`, `get_operation`, `execute_operation`)
- `FaultProvider` (`list_faults`, `get_fault`, `clear_fault`)
- Per-entity routing in `plugin_manager.cpp` (`get_data_provider_for_entity()` etc.)
- Auto-deduced capabilities from provider presence

This plugin does NOT use any of it. Instead it registers custom `x-plc-*` routes via `get_routes()`, meaning:
1. Standard SOVD endpoints (`/data`, `/operations`) return 404 for PLC entities
2. MCP tools (`sovd_read_data`, `sovd_execute_operation`) cannot access PLC data
3. Web UI Data/Operations tabs empty for PLC entities
4. Cyclic subscriptions framework cannot sample PLC values
5. Fleet gateway aggregation skips PLC data/operations

The `x-plc-*` prefix also breaks the `x-medkit-*` convention (see `graph_provider_plugin.cpp:315,341`).

**Recommendation:** Implement `DataProvider` + `OperationProvider` (optionally `FaultProvider`), export via dlsym here. Map existing handlers to provider methods. Keep only `x-medkit-plc-status` as a vendor route via `get_routes()` (connection health has no standard SOVD equivalent).

```cpp
class OpcuaPlugin : public GatewayPlugin,
                    public IntrospectionProvider,
                    public DataProvider,
                    public OperationProvider {
```
```cpp
extern "C" DataProvider * get_data_provider(GatewayPlugin * p) {
  return dynamic_cast<DataProvider *>(p);
}
```

src/ros2_medkit_plugins/ros2_medkit_opcua/CMakeLists.txt
```diff
@@ -0,0 +1,178 @@
...
+  find_package(ament_cmake_clang_format REQUIRED)
+  file(GLOB_RECURSE _format_files
+    "include/*.hpp" "src/*.cpp" "test/*.cpp"
+  )
+  ament_clang_format(${_format_files}
+    CONFIG_FILE "${CMAKE_CURRENT_SOURCE_DIR}/../../../.clang-format")
+
+  find_package(ament_cmake_gtest REQUIRED)
+
+  include(ROS2MedkitTestDomain)
```
bburda, line 145: **[critical] No unit tests for opcua_plugin.cpp (574 lines) and opcua_poller.cpp (251 lines)**

Only `test_node_map` and `test_opcua_client` are registered. The two largest source files have zero unit coverage:

**opcua_plugin.cpp** - untested: severity-string mapping (lines 468-480), range validation, type coercion (bool/int/string/double), `is_valid_path_segment` edge cases, `publish_values` NaN/Infinity filter, `build_data_response`.

**opcua_poller.cpp** - untested: `evaluate_alarms` threshold directions (above/below), edge-only callback semantics, bool-valued alarm activation, reconnect loop, subscription fallback.

Docker integration covers only discovery + 2 writes + 3 error cases - no alarms, no range validation, no type coercion, no non-finite values.

Recommendation: add `test_opcua_plugin.cpp` and `test_opcua_poller.cpp` with test doubles for `PluginContext` and `OpcuaClient`.

src/ros2_medkit_plugins/ros2_medkit_opcua/include/ros2_medkit_opcua/opcua_plugin.hpp
```diff
@@ -0,0 +1,112 @@
...
+///
+/// Vendor REST endpoints:
+///   GET  /apps/{id}/x-plc-data          - All OPC-UA values for entity
+///   GET  /apps/{id}/x-plc-data/{node}   - Single node value
+///   POST /apps/{id}/x-plc-operations/{op} - Write value to PLC
+///   GET  /components/{id}/x-plc-status  - Connection state and stats
+class OpcuaPlugin : public ros2_medkit_gateway::GatewayPlugin, public ros2_medkit_gateway::IntrospectionProvider {
+ public:
+  OpcuaPlugin();
+  ~OpcuaPlugin() override;
```
bburda, line 71: **[important] Missing plugin shutdown guard pattern**

Plugins holding ROS resources (service clients, publishers, background threads) should implement: (1) `std::atomic<bool> shutdown_requested_{false}`, (2) idempotent `shutdown()` via `exchange(true)`, (3) callback guards, (4) destructor calling `shutdown()`.

All four violated:
- No `shutdown_requested_` atomic
- `shutdown()` (opcua_plugin.cpp:164) not idempotent
- Callbacks (`on_alarm_change`, `publish_values`, route handlers) don't check shutdown flag
- `~OpcuaPlugin() = default` does NOT call `shutdown()`

Compare `graph_provider_plugin.cpp:320-325`, `TopicBeaconPlugin`, `ParameterBeaconPlugin` which all follow this pattern. Practical risk mitigated by member declaration order (poller_ last), but the convention is consistent across all sibling plugins.

src/ros2_medkit_plugins/ros2_medkit_opcua/src/opcua_plugin.cpp
```diff
@@ -0,0 +1,574 @@
...
+  }
+
+  return result;
+}
+
+// -- Route handlers --
+
+void OpcuaPlugin::handle_plc_data(const PluginRequest & req, PluginResponse & res) {
+  auto entity_id = req.path_param(1);
+  auto entity = ctx_->validate_entity_for_route(req, res, entity_id);
```
bburda, line 282: **[important] Route handlers don't null-check `ctx_` / `poller_`**

All four handlers dereference `ctx_->validate_entity_for_route(...)` and `poller_->snapshot()` without null guards. Compare `graph_provider_plugin.cpp:362-365` which defensively checks `if (!ctx_)`. If `set_context()` throws partway, this segfaults instead of returning 503.

Recommendation: one-line guard at top of each handler:
```cpp
if (!ctx_ || !poller_) {
  res.send_error(503, ERR_SERVICE_UNAVAILABLE, "OPC-UA plugin not initialized");
  return;
}
```

src/ros2_medkit_plugins/ros2_medkit_opcua/src/opcua_poller.cpp
```diff
@@ -0,0 +1,251 @@
...
+
+      // Attempt reconnect with original config (preserves timeout, etc.)
+      if (client_.connect(client_.current_config())) {
+        if (config_.prefer_subscriptions) {
+          setup_subscriptions();
+        }
+      } else {
+        // Wait before retry using configured interval
+        auto wait_ms = config_.reconnect_interval.count();
+        for (int64_t i = 0; i < wait_ms / 100 && running_.load(); ++i) {
```
bburda: **[important] Poll/reconnect sleep loops spin on intervals < 100 ms; no backoff**

Integer division `wait_ms / 100` yields 0 for intervals under 100 ms - the loop executes zero iterations and the thread busy-retries. `poll_interval_ms: 0` = 100% CPU spin hammering the PLC. Same issue at line 155 for poll interval.

Also: reconnect uses a fixed interval (default 5s) with no exponential backoff. Many down PLCs = saturated network and log pipeline.

Also: `set_poll_callback` (line 74) uses `assert(!running_.load())` which compiles out in Release, leaving a silent data race. `set_alarm_callback` correctly uses a mutex - inconsistent.

Recommendation: Replace sleep loops with `std::condition_variable::wait_for(lock, interval, [&]{ return !running_.load(); })` so `stop()` wakes the thread immediately. Add exponential backoff for reconnect (cap at 60s). Guard `poll_callback_` with a mutex like `alarm_callback_`.

src/ros2_medkit_plugins/ros2_medkit_opcua/test/test_opcua_client.cpp
```diff
@@ -0,0 +1,79 @@
...
+#include "ros2_medkit_opcua/opcua_client.hpp"
+
+#include <gtest/gtest.h>
+
+namespace ros2_medkit_gateway {
+
+// These tests verify OpcuaClient behavior without a running OPC-UA server.
+// Integration tests with a real server are in the Docker demo.
+
+TEST(OpcuaClientTest, DefaultState) {
```
bburda, line 24: **[important] All 8 tests only exercise disconnected-state smoke paths**

411 lines of client code are "tested" by 79 lines that verify "returns false/empty when disconnected". None of `current_config()` persistence, batched `read_values` with mixed results, `add_monitored_item`, `maybe_mark_disconnected` status-code classifier, or subscription lifecycle are exercised.

Recommendation: stand up an in-process `opcua::Server` fixture (open62541pp ships one) and exercise real connect/read/write/subscribe paths; or inject a mock transport layer.

src/ros2_medkit_plugins/ros2_medkit_opcua/test/test_node_map.cpp
```diff
@@ -0,0 +1,401 @@
...
+    data_type: bool
+    writable: false
+)";
+    f.close();
+  }
+
+  std::string yaml_path_;
+};
+
+TEST_F(NodeMapTest, LoadsYaml) {
```
bburda, line 87: **[important] Missing edge cases**

Not tested:
- Malformed YAML syntax (only `/nonexistent/path.yaml` tested)
- Duplicate `node_id` (`node_id_index_[entry.node_id_str] = idx` silently overwrites at node_map.cpp:265)
- Duplicate `(entity_id, data_name)` pair
- Range fields `min_value`/`max_value` parsed but never asserted
- Alarm with missing fields (partial `alarm:` block)
- Round-trip through `load()` for GUID/Opaque node-ID types (only `parse_node_id` covers them)

Recommendation: add `MalformedYamlSyntax`, `MissingRequiredFields`, `DuplicateNodeId`, `DuplicateEntityDataName`, `RangeValidationFields`, `AlarmPartialConfig`, `LoadsAllFourNodeIdTypes`. Document duplicate-handling intent.

src/ros2_medkit_plugins/ros2_medkit_opcua/docker/scripts/run_integration_tests.sh
```diff
@@ -0,0 +1,102 @@
...
+    if [ "$ok" = "true" ]; then
+        echo -e "  ${GREEN}PASS${NC}: $desc"
+        PASS=$((PASS + 1))
+    else
+        echo -e "  ${RED}FAIL${NC}: $desc"
+        FAIL=$((FAIL + 1))
+    fi
+}
+
+echo -e "${YELLOW}=== OpenPLC Tank Demo Integration Tests ===${NC}\n"
```
bburda, line 26: **[important] Integration tests skip alarm-to-fault, topic bridge, reconnect**

The 16 assertions cover discovery, status, reads, 2 writes, 3 error cases. They do NOT verify:
- Alarm activation/clearing (the core fault-bridging feature)
- ROS 2 topic publishing (`std_msgs::msg::Float32`)
- Fault manager service integration
- Subscription vs poll mode
- Reconnection after disconnect

Most error assertions only check `has("error_code")` without validating the actual code value.

Recommendation: add assertions that (a) inject a tank-level triggering `PLC_LOW_LEVEL` and verify the fault, (b) `ros2 topic echo` the published topic, (c) replace `has(...)` with `.error_code == "RESOURCE_NOT_FOUND"` etc.

src/ros2_medkit_plugins/ros2_medkit_opcua/src/node_map.cpp
```diff
@@ -0,0 +1,383 @@
...
+      node_id_index_.clear();
+      entity_defs_.clear();
+      return false;
+    }
+
+    entries_.clear();
+    entity_index_.clear();
+    node_id_index_.clear();
+
+    for (size_t i = 0; i < nodes.size(); ++i) {
```
bburda, line 198: **[important] YAML validation gaps: charset, bounds, error reporting, partial state**

1. **No charset/length validation**: `entity_id`, `data_name`, `fault_code` read verbatim from YAML with no check against `[A-Za-z0-9_-]` / max 256 chars (the same rule enforced by `is_valid_path_segment` on URL params). Asymmetric - URL inputs are validated but YAML inputs are not.

2. **No bounds on node count**: `YAML::LoadFile` has no size limit; `entries_` uncapped. 100k entries = 100k ROS 2 publishers = DDS resource exhaustion.

3. **Silent errors**: `parse_node_id` returns empty `NodeId{}` for all malformed forms without logging. `load()` catch block (line 273) writes to `std::cerr` instead of `log_error()`. Operators see only `"Failed to load node map"` with no indication which entry/field failed.

4. **Partial state on exception**: catch block returns false but doesn't clear `entries_`/indexes - callers see `load() == false` but `entries_.size() != 0`.

Recommendation: add `validate_identifier(s, 256)` rejecting bad entries; cap `nodes.size() <= 10000`; return structured errors from `parse_node_id` (`std::optional<NodeId>`); log per-entry via `log_error()`; clear all state in catch block.

src/ros2_medkit_plugins/ros2_medkit_opcua/src/opcua_plugin.cpp
```diff
@@ -0,0 +1,574 @@
...
+  // Connect uses Anonymous auth with SecurityPolicy=None.
+
+  if (config.contains("node_map_path")) {
+    node_map_path_ = config["node_map_path"].get<std::string>();
+  }
+
+  if (config.contains("prefer_subscriptions")) {
+    poller_config_.prefer_subscriptions = config["prefer_subscriptions"].get<bool>();
+  }
+  if (config.contains("subscription_interval_ms")) {
```
bburda, line 111: **[important] Undocumented params + missing input clamping**

`subscription_interval_ms` is parsed here but missing from the README parameters table. Users enabling `prefer_subscriptions: true` cannot discover how to tune it.

Also: `poll_interval_ms` is passed straight through with no bounds. Combined with the integer-division-by-100 sleep loop in opcua_poller.cpp:154, `poll_interval_ms: 0` causes a CPU spin. Negative / INT_MIN is undefined.

Also: `ros2_topic` field is parsed in node_map.cpp:241 with regex validation but undocumented in README or example YAML.

Recommendation: add `subscription_interval_ms` and `ros2_topic` to README. Clamp: `std::clamp(poll_interval, 100ms, 60000ms)`.

src/ros2_medkit_plugins/ros2_medkit_opcua/src/opcua_plugin.cpp
```diff
@@ -0,0 +1,574 @@
...
+        log_info("PLC bridge: skipping non-numeric " + entry.node_id_str + " (data_type=" + entry.data_type + ")");
+        continue;
+      }
+      publishers_[entry.node_id_str] =
+          node->create_publisher<std_msgs::msg::Float32>(entry.ros2_topic, rclcpp::SensorDataQoS());
+      log_info("PLC bridge: " + entry.node_id_str + " -> " + entry.ros2_topic + " (Float32)");
+    }
+  }
+
+  log_warn(
```
bburda, line 163: **[important] SecurityPolicy=None warning doesn't include endpoint URL**

Operators grepping logs can't identify which connection is insecure. README documents the limitation (good), but this log line should include the target endpoint.

Recommendation: `log_warn("OPC-UA security: Anonymous auth, SecurityPolicy=None for " + client_config_.endpoint_url + ". Not suitable for untrusted networks.");`

src/ros2_medkit_plugins/ros2_medkit_opcua/CMakeLists.txt
```diff
@@ -0,0 +1,178 @@
...
+# ---- open62541pp via FetchContent ----
+# open62541pp v0.16.0 is MPLv2 licensed. Linking from Apache 2.0 code to MPLv2
+# libraries is permitted - MPLv2 is a weak per-file copyleft that only requires
+# source disclosure for modifications to the MPLv2 files themselves. We ship
+# the library unmodified via FetchContent pinned to a specific upstream tag.
+include(FetchContent)
+fetchcontent_declare(
+  open62541pp
+  GIT_REPOSITORY https://github.com/open62541pp/open62541pp.git
+  GIT_TAG v0.16.0
```
bburda: **[important] FetchContent uses mutable GIT_TAG without hash verification**

`GIT_TAG v0.16.0` is a mutable git tag - upstream can force-push it. CMake FetchContent doesn't verify content hashes for git sources. `GIT_SUBMODULES_RECURSE ON` extends this to open62541 (submodule).

Recommendation: replace `v0.16.0` with the full 40-char commit SHA of that release. Add comment `# v0.16.0 pinned by SHA`.

src/ros2_medkit_plugins/ros2_medkit_opcua/src/opcua_plugin.cpp
```diff
@@ -0,0 +1,574 @@
...
+        write_val);
+    if (v < *entry->min_value || v > *entry->max_value) {
+      res.send_error(400, ERR_INVALID_REQUEST,
+                     "Value " + std::to_string(v) + " out of range [" + std::to_string(*entry->min_value) + ", " +
+                         std::to_string(*entry->max_value) + "]");
+      return;
+    }
+  }
+
+  if (!client_->write_value(entry->node_id, write_val)) {
```
bburda: **[important] Write failure collapses all OPC-UA errors to generic 502**

`write_value` returns `bool` - connection loss, bad type, permission denied, server reject all become one 502. Client cannot distinguish "retry later" from "fix your request".

Recommendation: change `write_value` to return `tl::expected<void, std::string>` with the OPC-UA `StatusCode`. Map: connection loss -> 502, bad type -> 400 `ERR_INVALID_REQUEST`, server reject -> 400/409 with vendor code.

src/ros2_medkit_plugins/ros2_medkit_opcua/src/opcua_client.cpp
```diff
@@ -0,0 +1,411 @@
...
+      r.good = false;
+      maybe_mark_disconnected(impl_->connected, e);
+    }
+    results.push_back(std::move(r));
+  }
+  return results;
+}
+
+bool OpcuaClient::write_value(const opcua::NodeId & node_id, const OpcuaValue & value) {
+  std::lock_guard<std::mutex> lock(impl_->client_mutex);
```
bburda, line 265: **[suggestion] `write_value` holds mutex across two OPC-UA round-trips**

`readValue()` (discover type) then `writeValueScalar()` (write), all under `client_mutex_`. Poller thread blocks on every write. Since `data_type` is already known from the YAML `NodeMapEntry`, the read-probe is unnecessary.

Recommendation: pass declared `data_type` into `write_value()`, write directly - halves mutex hold time and eliminates a round-trip.

src/ros2_medkit_plugins/ros2_medkit_opcua/package.xml
```diff
@@ -0,0 +1,33 @@
+<?xml version="1.0"?>
+<?xml-model href="http://download.ros.org/schema/package_format3.xsd" schematypens="http://www.w3.org/2001/XMLSchema"?>
+<package format="3">
+  <name>ros2_medkit_opcua</name>
+  <version>0.4.0</version>
+  <description>OPC-UA gateway plugin for ros2_medkit - bridges PLC systems into the SOVD entity tree</description>
+  <maintainer email="michal.faferek@selfpatch.ai">mfaferek93</maintainer>
+  <license>Apache-2.0</license>
+  <url type="website">https://selfpatch.ai</url>
+  <url type="repository">https://github.com/selfpatch/ros2_medkit</url>
```
bburda, line 10: **[suggestion]** `open62541pp` is used throughout but has no `<depend>` entry. Rosdistro reviewers will be confused. Add an XML comment explaining it's vendored via FetchContent (linked to follow-up 366).

.github/workflows/opcua-plugin.yml
```diff
@@ -0,0 +1,222 @@
+name: OPC-UA Plugin Integration
```
bburda, line 1: **[suggestion]** No `permissions:` block. Inconsistent with `ci.yml` (`permissions: {}`) and `docs.yml`. Add `permissions: contents: read` at workflow level.

src/ros2_medkit_plugins/ros2_medkit_opcua/src/opcua_plugin.cpp
```diff
@@ -0,0 +1,574 @@
...
+    ctx_->register_entity_capability(node_map_.component_id(), "x-plc-status");
+  }
+
+  for (const auto & def : node_map_.entity_defs()) {
+    App app;
+    app.id = def.id;
+    app.name = def.name;
+    app.component_id = def.component_id;
+    app.external = true;
+    app.is_online = client_->is_connected();
```
bburda, line 255: **[suggestion] `is_online` set at introspect() time, never updated after reconnect**

An app registered during initial connect failure stays marked offline in `GET /apps/tank_process` even after poller reconnects. `x-plc-status` shows correct live state, but the entity detail response is stale.

Recommendation: either always set `is_online = true` and rely on `x-plc-status` for liveness, or update the entity cache on reconnect callback.

src/ros2_medkit_plugins/ros2_medkit_opcua/src/opcua_poller.cpp
```diff
@@ -0,0 +1,251 @@
...
+        snapshot_.connected = false;
+      }
+
+      // Attempt reconnect with original config (preserves timeout, etc.)
+      if (client_.connect(client_.current_config())) {
+        if (config_.prefer_subscriptions) {
+          setup_subscriptions();
+        }
+      } else {
+        // Wait before retry using configured interval
```
bburda: **[suggestion] Reconnect loop has no exponential backoff**

Fixed 5s retry forever against a down PLC. With many PLCs down, this saturates network and log pipelines.

Recommendation: exponential backoff capped at 60s; throttle repeated error logs (log every N failures, not every attempt).

src/ros2_medkit_plugins/ros2_medkit_opcua/src/opcua_plugin_exports.cpp
mfaferek93: Fixed in `a2f53552`. Implemented DataProvider + OperationProvider + FaultProvider (8 methods, 3 dlsym exports). Standard SOVD `/data`, `/operations`, `/faults` now work for PLC entities. Vendor `x-plc-*` routes kept alongside.

src/ros2_medkit_plugins/ros2_medkit_opcua/CMakeLists.txt
mfaferek93: Fixed in `72f4a5f7`. New `test_opcua_plugin.cpp` with FakePluginContext: 13 provider tests (DataProvider list/read/write-readonly/write-missing, OperationProvider list/execute/readonly, FaultProvider list/get). Plus 6 NodeMap edge cases + 3 client tests. Total: 105.

src/ros2_medkit_plugins/ros2_medkit_opcua/include/ros2_medkit_opcua/opcua_plugin.hpp
mfaferek93: Fixed in `a2f53552`. Added `shutdown_requested_` atomic + `exchange(true)` + destructor calls `shutdown()` + guards in `publish_values`, `on_alarm_change`, all handlers and providers.

src/ros2_medkit_plugins/ros2_medkit_opcua/src/opcua_plugin.cpp
mfaferek93: Fixed in `a2f53552`. All 4 route handlers + 8 provider methods check `!ctx_ || !poller_ || shutdown_requested_` at entry, return 503.

src/ros2_medkit_plugins/ros2_medkit_opcua/src/opcua_poller.cpp
mfaferek93: Fixed in `a2f53552` + `e14411e0`. Both sleep loops replaced with `condition_variable::wait_for`. `assert` replaced with `std::logic_error`. Poll interval clamped to [100, 60000] ms. Exponential backoff on reconnect capped at 60s.

src/ros2_medkit_plugins/ros2_medkit_opcua/test/test_opcua_client.cpp
mfaferek93: Fixed in `371f3d35`. Added `WriteValueReturnsNotConnected`, `WriteValueWithTypeHintDisconnected`, `CurrentConfigPersistence`.

src/ros2_medkit_plugins/ros2_medkit_opcua/test/test_node_map.cpp
mfaferek93: Fixed in `371f3d35`. Added `MalformedYamlSyntax`, `MissingRequiredFields`, `DuplicateNodeId`, `UnknownDataTypeDefaultsToFloat`, `RangeValidationFieldsParsed`, `AlarmPartialConfigDefaultsSeverity`.

src/ros2_medkit_plugins/ros2_medkit_opcua/docker/scripts/run_integration_tests.sh
mfaferek93: Partially fixed. Added 2 SOVD `/data` endpoint assertions proving DataProvider works e2e. Alarm trigger/clear removed - OpenPLC simulation overrides external writes on next PLC cycle (100ms). Would need cooperative PLC program. Total: 18 assertions.

src/ros2_medkit_plugins/ros2_medkit_opcua/src/node_map.cpp
mfaferek93: Fixed in `e14411e0`. Required field validation, data_type enum check, duplicate node_id detection (first wins), 10k entry cap, `std::cerr` replaced with `RCLCPP_WARN/ERROR`.

src/ros2_medkit_plugins/ros2_medkit_opcua/src/opcua_plugin.cpp
mfaferek93: Poll interval clamped in `e14411e0`. README param docs still TODO.

src/ros2_medkit_plugins/ros2_medkit_opcua/src/opcua_plugin.cpp
mfaferek93: Fixed in `a2f53552`. Warning now includes endpoint URL.

src/ros2_medkit_plugins/ros2_medkit_opcua/CMakeLists.txt
mfaferek93: Fixed in `a2f53552`. Pinned to SHA `122ea4c842193918b97fd2f2def4fbecabb33ffe`.

src/ros2_medkit_plugins/ros2_medkit_opcua/src/opcua_plugin.cpp
mfaferek93: Fixed in `e14411e0`. `write_value` returns `tl::expected<void, WriteErrorInfo>` with NotConnected/TypeMismatch/AccessDenied/NodeNotFound/TransportError. Callers map to 400/403/502.

src/ros2_medkit_plugins/ros2_medkit_opcua/src/opcua_client.cpp
mfaferek93: Fixed in `e14411e0`. Added `data_type_hint` parameter - skips `readValue` probe when hint provided. All callers pass `entry->data_type`.

src/ros2_medkit_plugins/ros2_medkit_opcua/package.xml
mfaferek93: Fixed in `a2f53552`. Added XML comment explaining FetchContent + link to #366.

.github/workflows/opcua-plugin.yml
mfaferek93: Fixed in `a2f53552`. Added `permissions: contents: read`.

src/ros2_medkit_plugins/ros2_medkit_opcua/src/opcua_plugin.cpp
mfaferek93: `introspect()` runs every 2s (`refresh_interval_ms: 2000`) and re-reads `client_->is_connected()`. Updates within one discovery cycle after reconnect. 2s staleness acceptable for diagnostic status.

src/ros2_medkit_plugins/ros2_medkit_opcua/src/opcua_poller.cpp
mfaferek93: Fixed in `e14411e0`. Reconnect wait doubles per failed attempt, capped at 60s, resets on success.

src/ros2_medkit_plugins/ros2_medkit_opcua/src/opcua_plugin.cpp
```diff
@@ -0,0 +1,941 @@
...
+    if (v < *entry->min_value || v > *entry->max_value) {
+      return tl::make_unexpected(DataProviderErrorInfo{DataProviderError::InvalidValue,
+                                                       "Value " + std::to_string(v) + " out of range [" +
+                                                           std::to_string(*entry->min_value) + ", " +
+                                                           std::to_string(*entry->max_value) + "]",
+                                                       400});
+    }
+  }
+
+  if (!client_->write_value(entry->node_id, write_val)) {
```
bburda: Bug: `write_data()` calls the 2-argument `write_value(node_id, value)`, omitting the `data_type_hint`. Both `handle_plc_operations()` and `execute_operation()` correctly pass `entry->data_type` as the third argument:

```cpp
// handle_plc_operations (line ~395) and execute_operation (line ~855):
auto write_result = client_->write_value(entry->node_id, write_val, entry->data_type);

// write_data (here):
if (!client_->write_value(entry->node_id, write_val)) {  // missing hint
```

Without the hint, `write_value` must do an extra OPC-UA round-trip (`readValue` type-probe) before every write. Should be:
```cpp
auto write_result = client_->write_value(entry->node_id, write_val, entry->data_type);
```

src/ros2_medkit_plugins/ros2_medkit_opcua/src/opcua_plugin.cpp
```diff
@@ -0,0 +1,941 @@
...
+    return tl::make_unexpected(DataProviderErrorInfo{DataProviderError::TransportError,
```
bburda: Bug (consequence of the above): Because `write_data()` uses the 2-argument `write_value()` (returns `bool`), it cannot distinguish error causes. All failures land here as `TransportError` (502).

Compare with `execute_operation()` which uses the 3-argument version (returns `tl::expected<void, WriteErrorInfo>`) and correctly maps:
- `WriteError::TypeMismatch` -> `InvalidParameters` (400)
- `WriteError::AccessDenied` -> `Rejected` (403)
- others -> `TransportError` (502)

A type mismatch write through the standard SOVD `PUT /data/{name}` endpoint will return 502 instead of 400.

src/ros2_medkit_plugins/ros2_medkit_opcua/src/opcua_plugin.cpp
```diff
@@ -0,0 +1,941 @@
...
+  if (!entry->writable) {
+    return tl::make_unexpected(
+        OperationProviderErrorInfo{OperationProviderError::Rejected, "Data point is read-only: " + data_name, 400});
+  }
+  if (!parameters.contains("value")) {
+    return tl::make_unexpected(OperationProviderErrorInfo{OperationProviderError::InvalidParameters,
+                                                          "Missing 'value' field in parameters", 400});
+  }
+
+  OpcuaValue write_val;
```
bburda: Suggestion: The value parsing (bool/int/string/float dispatch), range validation, and OPC-UA write logic is duplicated three times:
1. `handle_plc_operations()` (~line 361)
2. `write_data()` (~line 693)
3. `execute_operation()` (here, ~line 792)

The three copies have already diverged - `write_data()` uses the 2-arg `write_value()` while the other two use 3-arg (see comments above). Consider extracting a shared `parse_and_write_value()` helper to keep them in sync and prevent further drift.

src/ros2_medkit_plugins/ros2_medkit_opcua/src/opcua_plugin.cpp
```diff
@@ -0,0 +1,941 @@
...
+
+#include <algorithm>
+#include <cctype>
+#include <chrono>
+#include <cmath>
+#include <cstdlib>
+
+namespace ros2_medkit_gateway {
+
+using namespace ros2_medkit_gateway;
```
bburda: Nit: This `using namespace ros2_medkit_gateway;` is redundant - we're already inside `namespace ros2_medkit_gateway { ... }` (line 30).

src/ros2_medkit_plugins/ros2_medkit_opcua/test/test_opcua_plugin.cpp
```diff
@@ -0,0 +1,295 @@
...
+}
+const std::string & PluginRequest::path() const {
+  static const std::string empty;
+  return empty;
+}
+const std::string & PluginRequest::body() const {
+  static const std::string empty;
+  return empty;
+}
+std::string PluginRequest::query_param(const std::string &) const {
```
bburda, line 54: Note: This `query_param()` stub is not used by the plugin but is part of the `PluginRequest` interface. Looks like it was inherited from the graph_provider test template. Not a problem, but if `PluginRequest` gains new methods, this stub and the graph_provider one need to stay in sync - otherwise the test won't link. Consider whether a shared test stub header would be worth it once there are 3+ plugins with this pattern.

src/ros2_medkit_plugins/ros2_medkit_opcua/src/opcua_plugin.cpp
mfaferek93: Fixed in `39f0a640`. `write_data()` now uses 3-arg `write_value(node_id, value, entry->data_type)` matching the other two callers.

src/ros2_medkit_plugins/ros2_medkit_opcua/src/opcua_plugin.cpp
mfaferek93: Fixed in same commit. `write_data()` now maps WriteError codes: TypeMismatch->400, AccessDenied->403, Transport->502.

src/ros2_medkit_plugins/ros2_medkit_opcua/src/opcua_plugin.cpp
mfaferek93: Fixed in `39f0a640`. Removed.

src/ros2_medkit_plugins/ros2_medkit_opcua/src/opcua_plugin.cpp
mfaferek93: Fixed in `ee04d195`. Extracted `parse_coerce_validate()` helper in anonymous namespace. All 3 callers (handle_plc_operations, write_data, execute_operation) now use it. Net -36 lines.

src/ros2_medkit_plugins/ros2_medkit_opcua/test/test_opcua_plugin.cpp
mfaferek93: fixed

src/ros2_medkit_plugins/ros2_medkit_opcua/src/opcua_plugin.cpp
mfaferek93: fixed

bburda (approved):
LGTM!

PR #370, opened 2026-04-11
fix(aggregation): read app-to-component binding from SOVD is-located-on

### LINKED ISSUE #369: [BUG] Aggregation drops app->component binding when fetching apps from peers
# Bug report

### Steps to reproduce
1. Deploy two gateways (A and B), each hosting at least one component and one app where the app is bound to the component.
2. Configure gateway A to aggregate from gateway B.
3. Query gateway A: ``GET /api/v1/apps`` and ``GET /api/v1/apps/{app-id}`` for the apps owned by peer B.

### Expected behavior
Remote apps should carry their ``component_id`` so that the app-to-component binding survives aggregation. ``GET /api/v1/apps/{app-id}`` on the aggregator should still return the ``is-located-on`` relationship URI.

### Actual behavior
``parse_app()`` in ``src/ros2_medkit_gateway/src/aggregation/peer_client.cpp`` reads ``x-medkit.componentId`` (camelCase), while the gateway's own serializer emits ``x-medkit.component_id`` (snake_case) via the ``XMedkit`` builder in ``handle_list_apps`` / ``handle_get_app``. The key never matches, so the parsed ``App::component_id`` stays empty and downstream HATEOAS links (``is-located-on``) are omitted.

### Root cause
Two issues compound here:

1. The parser relied on a vendor-specific ``x-medkit`` field instead of SOVD's standard ``is-located-on`` relationship (ISO 17978-3, section 7.6). This would also have failed against any SOVD-compliant peer from another vendor.
2. Even as a fallback, the key spelling was wrong.

### Fix
- ``parse_app()`` should first read the SOVD-standard ``is-located-on`` URI (direct field or ``_links.is-located-on``), and extract the component ID from the trailing path segment. This works with any SOVD-compliant peer.
- Vendor fallback remains available for backward compatibility, but now spells the key correctly (``component_id``).

### Environment
- ros2_medkit version: main (commit 22d9b552)
- ROS 2 distro: Jazzy
- OS: Ubuntu 24.04

### Additional information
Affected file: ``src/ros2_medkit_gateway/src/aggregation/peer_client.cpp`` (``parse_app`` function).
Test coverage gap: ``test_peer_client.cpp::fetch_entities_parses_relationship_fields`` used the same snake_case spelling in its mock, masking the bug.

bburda:
## Summary

Aggregation dropped the app-to-component binding when fetching apps from peers because ``parse_app()`` read ``x-medkit.componentId`` (camelCase), while the gateway's own serializer emits ``x-medkit.component_id`` (snake_case) via the ``XMedkit`` builder in ``handle_list_apps`` / ``handle_get_app``. The key never matched and ``App::component_id`` stayed empty.

Rather than fixing the vendor key spelling, switch to SOVD's standard ``is-located-on`` relationship (ISO 17978-3, section 7.6). The gateway already emits this at the entity root and in ``_links.is-located-on``, and any SOVD-compliant peer will emit it too. ``x-medkit.component_id`` (snake_case) is kept as a fallback for backward compatibility.

``parse_app`` now:
1. Reads ``is-located-on`` (direct field) and extracts the trailing segment after ``/components/``. Handles both absolute URLs and path-only references.
2. Falls back to ``_links.is-located-on`` if the root field is absent.
3. Falls back to ``x-medkit.component_id`` (snake_case, matching the live ``XMedkit`` builder) only if the standard fields are absent.

---

## Issue

- closes #369

---

## Type

- [x] Bug fix
- [ ] New feature or tests
- [ ] Breaking change
- [ ] Documentation only

---

## Testing

- ``colcon test --packages-select ros2_medkit_gateway`` - 1995 tests, 0 failures
- ``ctest -R test_peer_client -V`` - 25/25 pass including the new case below
- New test ``PeerClientHappyPath.fetch_entities_parses_is_located_on_without_vendor_extension`` exercises the standard-only parse path (no ``x-medkit`` extension) with both the direct ``is-located-on`` field and the ``_links.is-located-on`` HATEOAS variant. It also covers the absolute-URL form of the URI.
- Updated ``PeerClientHappyPath.fetch_entities_parses_relationship_fields`` to mirror what real gateways emit (both the standard relationship and the vendor fallback, with correct snake_case spelling).

---

## Checklist

- [x] Breaking changes are clearly described (and announced in docs / changelog if needed) - no breaking changes
- [ ] Docs were updated if behavior or public API changed - no API change, purely internal parser fix

### DIFF
--- a/src/ros2_medkit_gateway/src/aggregation/peer_client.cpp
+++ b/src/ros2_medkit_gateway/src/aggregation/peer_client.cpp
@@ -198,7 +198,11 @@ Component parse_component(const nlohmann::json & j) {
  * URI (ISO 17978-3, §7.6). Peers may emit it as either an absolute URL
  * (``http://host:port/api/v1/components/{id}``) or a path-only reference
  * (``/api/v1/components/{id}``). Extract the trailing segment after the
- * ``/components/`` marker.
+ * ``/components/`` marker and validate it with ``is_valid_entity_id`` so
+ * that a malformed or hostile peer URI cannot smuggle path traversal or
+ * percent-encoded junk into the aggregator's entity cache.
+ * Returns an empty string if the URI contains no marker or if the
+ * extracted segment fails validation.
  */
 std::string component_id_from_located_on(const std::string & uri) {
   static const std::string kMarker = "/components/";
@@ -211,7 +215,8 @@ std::string component_id_from_located_on(const std::string & uri) {
     return "";
   }
   auto id_end = uri.find_first_of("/?#", id_start);
-  return uri.substr(id_start, id_end - id_start);
+  auto candidate = uri.substr(id_start, id_end - id_start);
+  return is_valid_entity_id(candidate) ? candidate : std::string{};
 }
 
 /**
@@ -243,9 +248,13 @@ App parse_app(const nlohmann::json & j) {
     const auto & xm = j["x-medkit"];
     // Vendor fallback: gateway emits x-medkit.component_id (snake_case) via
     // XMedkit builder in discovery_handlers.cpp. Only used if the SOVD
-    // standard is-located-on field is absent.
+    // standard is-located-on field is absent. Validated for the same
+    // reasons as component_id_from_located_on - the value is peer-provided.
     if (app.component_id.empty()) {
-      app.component_id = xm.value("component_id", "");
+      auto candidate = xm.value("component_id", "");
+      if (is_valid_entity_id(candidate)) {
+        app.component_id = candidate;
+      }
     }
     app.source = xm.value("source", "");
     app.is_online = xm.value("is_online", false);
--- a/src/ros2_medkit_gateway/src/aggregation/peer_client.cpp
+++ b/src/ros2_medkit_gateway/src/aggregation/peer_client.cpp
@@ -192,16 +192,61 @@ Component parse_component(const nlohmann::json & j) {
 }
 
 /**
- * @brief Parse an App from JSON
+ * @brief Extract the component ID from an ``is-located-on`` URI.
+ *
+ * SOVD exposes the app-to-component binding as a standard relationship
+ * URI (ISO 17978-3, §7.6). Peers may emit it as either an absolute URL
+ * (``http://host:port/api/v1/components/{id}``) or a path-only reference
+ * (``/api/v1/components/{id}``). Extract the trailing segment after the
+ * ``/components/`` marker.
+ */
+std::string component_id_from_located_on(const std::string & uri) {
+  static const std::string kMarker = "/components/";
+  auto pos = uri.rfind(kMarker);
+  if (pos == std::string::npos) {
+    return "";
+  }
+  auto id_start = pos + kMarker.size();
+  if (id_start >= uri.size()) {
+    return "";
+  }
+  auto id_end = uri.find_first_of("/?#", id_start);
+  return uri.substr(id_start, id_end - id_start);
+}
+
+/**
+ * @brief Parse an App from JSON.
+ *
+ * The app-to-component binding is recovered from SOVD's standard
+ * ``is-located-on`` relationship (body field or ``_links`` entry),
+ * which every SOVD-compliant peer emits. Vendor fallback:
+ * ``x-medkit.component_id`` (gateway's own extension).
  */
 App parse_app(const nlohmann::json & j) {
   App app;
   app.id = j.value("id", "");
   app.name = j.value("name", "");
   app.description = j.value("description", "");
+
+  // SOVD-standard: is-located-on relationship (ISO 17978-3, §7.6)
+  if (j.contains("is-located-on") && j["is-located-on"].is_string()) {
+    app.component_id = component_id_from_located_on(j["is-located-on"].get<std::string>());
+  }
+  if (app.component_id.empty() && j.contains("_links") && j["_links"].is_object()) {
+    const auto & links = j["_links"];
+    if (links.contains("is-located-on") && links["is-located-on"].is_string()) {
+      app.component_id = component_id_from_located_on(links["is-located-on"].get<std::string>());
+    }
+  }
+
   if (j.contains("x-medkit") && j["x-medkit"].is_object()) {
     const auto & xm = j["x-medkit"];
-    app.component_id = xm.value("component_id", "");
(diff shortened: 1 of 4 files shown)


src/ros2_medkit_gateway/src/aggregation/peer_client.cpp
```diff
@@ -192,16 +192,61 @@ Component parse_component(const nlohmann::json & j) {
...
+  if (pos == std::string::npos) {
+    return "";
+  }
+  auto id_start = pos + kMarker.size();
+  if (id_start >= uri.size()) {
+    return "";
+  }
+  auto id_end = uri.find_first_of("/?#", id_start);
+  return uri.substr(id_start, id_end - id_start);
+}
```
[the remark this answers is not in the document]
bburda: Good catch. Pushed in 5c3e8362: both `component_id_from_located_on()` and the `x-medkit.component_id` fallback now run their candidate through `is_valid_entity_id()` and leave `App::component_id` empty on failure, matching the existing `fetch_entities()` sanitization. Added `fetch_entities_rejects_malicious_component_id_in_located_on` covering percent-encoded path traversal, unencoded spaces, slash-containing vendor values, and empty vendor values.

src/ros2_medkit_gateway/src/aggregation/peer_client.cpp
```diff
@@ -192,16 +192,70 @@ Component parse_component(const nlohmann::json & j) {
...
   app.name = j.value("name", "");
   app.description = j.value("description", "");
+
+  // SOVD-standard: is-located-on relationship (ISO 17978-3, §7.6)
+  if (j.contains("is-located-on") && j["is-located-on"].is_string()) {
+    app.component_id = component_id_from_located_on(j["is-located-on"].get<std::string>());
+  }
+  if (app.component_id.empty() && j.contains("_links") && j["_links"].is_object()) {
+    const auto & links = j["_links"];
+    if (links.contains("is-located-on") && links["is-located-on"].is_string()) {
```
mfaferek93, line 242: **[minor]** `.is_string()` check here is correct for self-aggregation (gateway's own `LinksBuilder::add()` emits plain strings via `links_[rel] = href`). But per HAL+JSON spec, `_links` entries should be objects with `href` field (`{"href": "..."}`) not bare strings. A strictly-compliant third-party SOVD peer would emit the object form and this parser would miss it, falling through to vendor extension silently.

Not blocking for this PR since it's a pre-existing HAL gap in the gateway (emitter + parser both use strings). But worth noting as a future interop improvement - adding `else if (links["is-located-on"].is_object())` with `.value("href", "")` extraction would cover both forms.

src/ros2_medkit_gateway/test/test_peer_client.cpp
```diff
@@ -329,6 +333,109 @@ TEST(PeerClientHappyPath, fetch_entities_parses_relationship_fields) {
   t.join();
 }
 
+// App->component binding via SOVD-standard is-located-on relationship only
+// (no x-medkit vendor extension). Exercises the parse path used by peers
+// implementing the ISO 17978-3 standard from other vendors.
+// @verifies REQ_INTEROP_018
+TEST(PeerClientHappyPath, fetch_entities_parses_is_located_on_without_vendor_extension) {
```
mfaferek93, line 340: **[minor]** Good coverage for the two main URI forms (absolute URL + path-only). Missing edge cases that the code handles correctly (verified by tracing 11 inputs against `component_id_from_located_on`):

- Trailing slash: `/api/v1/components/ecu-a/` -> `ecu-a` (via `find_first_of` matching `/`)
- Query param: `/api/v1/components/ecu-a?foo=bar` -> `ecu-a`
- Fragment: `/api/v1/components/ecu-a#section` -> `ecu-a`

All produce correct results but none have test coverage. Quick win: add one test with a URI containing trailing delimiters.

src/ros2_medkit_gateway/test/test_peer_client.cpp
```diff
@@ -329,6 +333,109 @@ TEST(PeerClientHappyPath, fetch_entities_parses_relationship_fields) {
...
+  t.join();
+}
+
+// Malicious peer attempts to smuggle path traversal / percent-encoding /
+// oversize IDs through the app->component binding. Both the SOVD-standard
+// is-located-on parser and the x-medkit vendor fallback must reject these
+// candidates (leaving App::component_id empty) so they cannot propagate into
+// the aggregator's entity cache or response-side path construction.
+// @verifies REQ_INTEROP_018
+TEST(PeerClientHappyPath, fetch_entities_rejects_malicious_component_id_in_located_on) {
```
mfaferek93, line 412: **[minor]** This malicious-peer test correctly exercises vendor fallback **rejection** paths (`bad/id`, `bad%20id`, empty string). But there's no test for a **successful** vendor-only fallback - an app with a valid `x-medkit.component_id` and NO `is-located-on` field.

The modified `fetch_entities_parses_relationship_fields` test above has both fields present, so SOVD wins and vendor `component_id` path is effectively dead code there.

Suggestion: add one app entry with only `"x-medkit": {"component_id": "perception-ecu"}` (no `is-located-on`) and assert `component_id == "perception-ecu"` to cover the happy-path fallback.

src/ros2_medkit_gateway/src/aggregation/peer_client.cpp
bburda: Fixed in 8ee276ee. parse_app() now handles the HAL+JSON object form in addition to bare strings. Added app-hal-object entry to the is-located-on test to cover this path.

src/ros2_medkit_gateway/test/test_peer_client.cpp
bburda: Fixed in 8ee276ee. Added trailing slash, query param, and fragment URI forms to the test - all correctly extract ecu-a via the find_first_of("/?#") delimiter in component_id_from_located_on().

src/ros2_medkit_gateway/test/test_peer_client.cpp
bburda: Fixed in 8ee276ee. Added fetch_entities_parses_vendor_only_component_id_fallback test - app with only x-medkit.component_id and no is-located-on field, asserting component_id is correctly extracted through the vendor fallback path.

mfaferek93 (approved):

PR #374, opened 2026-04-13
fix(aggregation): hierarchical/leaf Component routing with provenance and warnings

### LINKED ISSUE #373: [BUG] Collision-merged peer components return empty runtime data on primary
# Bug report

### Steps to reproduce

1. Run two or more gateway instances (primary + one or more peers) with aggregation enabled on the primary.
2. Declare a shared parent Component (e.g. ``robot-alpha``) across all manifests, with each ECU declaring its own child Component via ``parent_component_id``. This is the topology used by the ``multi_ecu_aggregation`` demo.
3. After discovery, issue sub-resource requests through the primary, for example:
   - ``GET /api/v1/components/{leaf_ecu_id}/logs``
   - ``GET /api/v1/components/{leaf_ecu_id}/hosts``
   - ``GET /api/v1/components/{leaf_ecu_id}/data``

### Expected behavior

The aggregation layer treats Components with the same symmetry as Areas:

- A **leaf Component** maps to exactly one ECU. When the same ID is announced by a peer, the peer owns the runtime state (data, logs, hosts, operations, faults) and the primary should transparently forward every request - detail endpoint and all sub-resources - to that peer.
- A **hierarchical parent Component** (referenced as ``parent_component_id`` by any other Component in the merged set) has no runtime state of its own; it only groups its children. The primary should serve it locally with the merged view, exactly like an Area whose Components come from different peers.
- When more than one peer legitimately or accidentally announces the same leaf ID, the primary should flag the collision to operators without taking the gateway down.

### Actual behavior

``EntityMerger::merge_components`` previously skipped inserting a routing entry for any collision-merged Component, treating the merged entity as locally owned. That assumption is wrong for leaves: with no routing entry, sub-resource requests were handled on the primary, which has no runtime state for the peer's leaf, so ``/logs`` came back empty, ``/hosts`` missed peer apps, and ``/data`` had no resources.

The naive fix (route every collision to the peer) then broke the hierarchical case: ``GET /components/{parent_id}`` forwarded to a random peer's view and the merged cache was lost. Sub-components that arrive from different peers (typical for the multi-ECU demo) compounded the problem.

Additionally, there was no operator-visible signal when two peers declared the same leaf ID, and ``source`` on the merged entity could not express the fact that several contributors fed into a single view.

### Environment

- ros2_medkit version: main (reproduces on current tip)
- ROS 2 distro: Jazzy / Humble / Rolling
- OS: Linux (Ubuntu 22.04 / 24.04)

### Additional information

The aggregation design doc previously described "Remote-only Components get a routing table entry", which matched the implementation but did not reflect the ECU-ownership model the merge logic otherwise followed. The fix needs to classify merged Components into hierarchical parents (served locally with merged view) and leaves (routed to the owning peer), emit structured warnings on multi-peer leaf collisions, and expose provenance to clients so they can distinguish local-only, peer-only, and merged entities.

bburda:
## Summary

Make the aggregation layer treat Components with the same symmetry as Areas, so collision-merged peer Components work correctly for every deployment shape the project supports (single peer, multi-peer, hierarchical parent, daisy chain).

- ``EntityMerger::merge_components`` now records a provisional routing entry on collision. Every Component ID refers to at most one ECU when it is a leaf, and the peer owns the runtime state.
- A new pure ``classify_component_routing`` free function (``aggregation/classification.hpp``) runs after all peers have been merged. A Component is classified as a **hierarchical parent** if any other Component in the merged set references it via ``parent_component_id``; those are removed from the routing table and served locally from the merged cache. Leaves stay routed to the owning peer.
- Multi-peer leaf collisions are reported as structured ``LeafCollisionWarning`` objects. ``AggregationManager`` surfaces them on ``MergedPeerResult.leaf_warnings``, writes an RCLCPP_WARN line listing every claimant, and exposes the list via a new ``/health.warnings`` array. Routing falls back to last-writer-wins because rejection does not fix a deployment anomaly.
- Every merged entity (Area, Component, App, Function) now carries an optional ``x-medkit.contributors`` list populated during merge (``local`` seeded before the peer loop, ``peer:<name>`` appended by ``EntityMerger`` with de-duplication). Clients can distinguish local-only, peer-only, and merged views without relying on the single-valued ``source`` field. In daisy-chain topologies each hop surfaces only its direct upstream by design.
- Pre-existing doc comment referencing a third-party commercial product was stripped from ``sovd_service_interface.hpp``.
- Aggregation design doc now documents the symmetric classification, the multi-peer warning behaviour, and the ``contributors`` provenance, with an updated PlantUML merge diagram plus a new post-merge classification diagram.

### SOVD API impact

- **No changes** to SOVD-defined request paths, response schemas, or error codes.
- Additive x-medkit extensions only: new optional ``x-medkit.contributors`` field on entity responses and a new optional ``warnings`` array on ``/health``. Both are emitted only when non-empty / when aggregation is active, so existing clients ignoring unknown fields are unaffected.

---

## Issue

- closes #373

---

## Type

- [x] Bug fix
- [x] New feature or tests (classification seam, /health.warnings, contributors, daisy-chain integration test)
- [ ] Breaking change
- [x] Documentation only (design doc updates)

---

## Testing

All local runs green:

- ``./scripts/test.sh unit --packages-select ros2_medkit_gateway`` - 1976 tests, 0 failures. Includes:
  - 6 new ``AggregationClassification`` tests covering leaf vs hierarchical, single/multi-peer, subcomponents, and warning emission.
  - 4 new ``EntityMerger`` contributor tests (local/peer append, dedup, app-prefix-only-peer).
  - Updated ``merged_components_route_to_peer`` plus the synthetic-collision test from the previous iteration.
- ``./scripts/test.sh test_peer_aggregation --packages-select ros2_medkit_integration_tests`` - prior integration coverage still green, 82 tests.
- ``./scripts/test.sh test_daisy_chain_aggregation --packages-select ros2_medkit_integration_tests`` - new integration test with 3 gateways in isolated DDS domains (primary -> peer_B -> peer_C). Verifies hierarchical parent served locally, 1-hop and 2-hop leaf forwarding, contributors, and empty /health.warnings.
- ``./scripts/test.sh lint --packages-select ros2_medkit_gateway`` - clean.
- Sphinx docs build with no new warnings.

---

## Checklist

- [x] Breaking changes are clearly described (there are none; only additive x-medkit extensions)

### DIFF
--- a/docs/api/rest.rst
+++ b/docs/api/rest.rst
@@ -89,8 +89,8 @@ Server Capabilities
    Health check endpoint. Returns HTTP 200 if gateway is operational.
 
    When aggregation is enabled (``capabilities.aggregation == true`` in
-   the root response), the body includes two additional x-medkit
-   extension fields:
+   the root response), the body includes additional x-medkit extension
+   fields:
 
    - ``peers`` - array of peer status objects (URL, name, reachability,
      last-seen timestamp) for every configured or discovered peer.
@@ -99,6 +99,22 @@ Server Capabilities
      there are no active anomalies). Each warning carries ``code``,
      ``message``, ``entity_ids``, and ``peer_names``. See
      :doc:`warning_codes` for the stable list of codes.
+   - ``warning_schema_version`` - integer contract version for the
+     ``warnings`` array. Clients key on this instead of string-matching
+     codes. See :doc:`warning_codes` ``Schema versioning``.
+
+   .. note::
+
+      Security: ``/health`` is currently reachable without
+      authentication by default (``auth.enabled`` defaults to
+      ``false``), and even with auth enabled the endpoint is readable
+      by the ``viewer``, ``operator``, and ``configurator`` roles. The
+      ``peers`` array enumerates every configured peer's name and URL,
+      which reveals deployment topology. This is by design for
+      operator observability in trusted LANs, but on shared-infra or
+      multi-tenant installs you should front the endpoint with an
+      authenticating reverse proxy or restrict the peer-name field to
+      admin-gated callers at the ingress.
 
    **Example Response (aggregation enabled, one leaf collision):**
 
@@ -115,6 +131,7 @@ Server Capabilities
           {"name": "peer_b", "url": "http://peer-b:8080", "healthy": true},
           {"name": "peer_c", "url": "http://peer-c:8080", "healthy": true}
         ],
+        "warning_schema_version": 1,
         "warnings": [
           {
             "code": "leaf_id_collision",
--- a/docs/api/rest.rst
+++ b/docs/api/rest.rst
@@ -54,11 +54,16 @@ Server Capabilities
         }
       }
 
-   The ``capabilities.aggregation`` flag is ``true`` when the gateway has
-   one or more aggregation peers configured. Clients can feature-detect
-   aggregation-only response fields (``peers``, ``warnings`` on ``/health``
-   and ``x-medkit.contributors`` on entities) using this flag instead of
-   probing for field presence.
+   The ``capabilities.aggregation`` flag is ``true`` when the aggregation
+   subsystem is enabled on this gateway (i.e. ``aggregation.enabled=true``
+   in config, which wires up an ``AggregationManager``). It does NOT
+   require peers to be present - a gateway with aggregation enabled but
+   zero peers still reports ``true`` and still emits the
+   aggregation-only response fields (``peers``, which may be an empty
+   array, and ``warnings`` on ``/health``; ``x-medkit.contributors`` on
+   entities, which will contain only ``"local"`` until a peer
+   contributes). Clients can feature-detect those fields using this
+   flag instead of probing for field presence.
 
 ``GET /api/v1/version-info``
    Get gateway version and status information.
--- a/docs/api/rest.rst
+++ b/docs/api/rest.rst
@@ -49,9 +49,16 @@ Server Capabilities
           "triggers": true,
           "updates": false,
           "authentication": false,
-          "tls": false
+          "tls": false,
+          "aggregation": false
         }
       }
+
+   The ``capabilities.aggregation`` flag is ``true`` when the gateway has
+   one or more aggregation peers configured. Clients can feature-detect
+   aggregation-only response fields (``peers``, ``warnings`` on ``/health``
+   and ``x-medkit.contributors`` on entities) using this flag instead of
+   probing for field presence.
 
 ``GET /api/v1/version-info``
    Get gateway version and status information.
@@ -75,6 +82,43 @@ Server Capabilities
 
 ``GET /api/v1/health``
    Health check endpoint. Returns HTTP 200 if gateway is operational.
+
+   When aggregation is enabled (``capabilities.aggregation == true`` in
+   the root response), the body includes two additional x-medkit
+   extension fields:
+
+   - ``peers`` - array of peer status objects (URL, name, reachability,
+     last-seen timestamp) for every configured or discovered peer.
+   - ``warnings`` - array of structured operator-actionable aggregation
+     warnings (always present when aggregation is active; empty when
+     there are no active anomalies). Each warning carries ``code``,
+     ``message``, ``entity_ids``, and ``peer_names``. See
+     :doc:`warning_codes` for the stable list of codes.
+
+   **Example Response (aggregation enabled, one leaf collision):**
+
+   .. code-block:: json
+
+      {
+        "status": "healthy",
+        "timestamp": 1776185189048036615,
+        "discovery": {
+          "mode": "hybrid",
+          "strategy": "hybrid_discovery"
+        },
+        "peers": [
+          {"name": "peer_b", "url": "http://peer-b:8080", "healthy": true},
+          {"name": "peer_c", "url": "http://peer-c:8080", "healthy": true}
+        ],
+        "warnings": [
+          {
+            "code": "leaf_id_collision",
+            "message": "Component 'ecu-x' is announced by multiple peers (peer_b, peer_c); routing falls back to last-writer-wins which is non-deterministic. Resolve by renaming the Component on one side or by modelling it as a hierarchical parent (declare a child Component with parentComponentId='ecu-x' on the owning peer).",
+            "entity_ids": ["ecu-x"],
+            "peer_names": ["peer_b", "peer_c"]
+          }
+        ]
+      }
 
 Discovery Endpoints
 -------------------
--- a/src/ros2_medkit_gateway/include/ros2_medkit_gateway/discovery/models/area.hpp
+++ b/src/ros2_medkit_gateway/include/ros2_medkit_gateway/discovery/models/area.hpp
@@ -14,6 +14,8 @@
 
 #pragma once
 
+#include "ros2_medkit_gateway/discovery/models/common.hpp"
+
 #include <nlohmann/json.hpp>
 #include <string>
 #include <vector>
--- a/src/ros2_medkit_gateway/include/ros2_medkit_gateway/discovery/models/area.hpp
+++ b/src/ros2_medkit_gateway/include/ros2_medkit_gateway/discovery/models/area.hpp
@@ -29,15 +29,16 @@ using json = nlohmann::json;
  * They provide a hierarchical organization of components.
  */
 struct Area {
-  std::string id;                 ///< Unique identifier (e.g., "powertrain")
-  std::string name;               ///< Human-readable name (e.g., "Powertrain System")
-  std::string namespace_path;     ///< ROS 2 namespace path (e.g., "/powertrain")
-  std::string type = "Area";      ///< Entity type (always "Area")
-  std::string translation_id;     ///< Internationalization key
-  std::string description;        ///< Human-readable description
-  std::vector<std::string> tags;  ///< Tags for filtering
-  std::string parent_area_id;     ///< Parent area ID for sub-areas
-  std::string source;             ///< Origin of this area (e.g., "manifest", "heuristic")
+  std::string id;                         ///< Unique identifier (e.g., "powertrain")
+  std::string name;                       ///< Human-readable name (e.g., "Powertrain System")
+  std::string namespace_path;             ///< ROS 2 namespace path (e.g., "/powertrain")
+  std::string type = "Area";              ///< Entity type (always "Area")
+  std::string translation_id;             ///< Internationalization key
+  std::string description;                ///< Human-readable description
+  std::vector<std::string> tags;          ///< Tags for filtering
+  std::string parent_area_id;             ///< Parent area ID for sub-areas
+  std::string source;                     ///< Origin of this area (e.g., "manifest", "heuristic")
+  std::vector<std::string> contributors;  ///< Aggregation provenance: "local" and/or "peer:<name>"
 
   /**
    * @brief Convert to JSON representation
@@ -70,6 +71,9 @@ struct Area {
     }
     if (!source.empty()) {
       x_medkit["source"] = source;
+    }
+    if (!contributors.empty()) {
+      x_medkit["contributors"] = contributors;
     }
     j["x-medkit"] = x_medkit;
 
--- a/src/ros2_medkit_gateway/include/ros2_medkit_gateway/discovery/models/area.hpp
+++ b/src/ros2_medkit_gateway/include/ros2_medkit_gateway/discovery/models/area.hpp
@@ -73,7 +73,7 @@ struct Area {
       x_medkit["source"] = source;
     }
     if (!contributors.empty()) {
-      x_medkit["contributors"] = contributors;
+      x_medkit["contributors"] = sorted_contributors(contributors);
     }
     j["x-medkit"] = x_medkit;
 
--- a/src/ros2_medkit_gateway/include/ros2_medkit_gateway/http/warning_codes.hpp
+++ b/src/ros2_medkit_gateway/include/ros2_medkit_gateway/http/warning_codes.hpp
@@ -1,0 +1,37 @@
+// Copyright 2026 bburda
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+#pragma once
+
+namespace ros2_medkit_gateway {
+
+/// Warning codes surfaced on /health.warnings.
+///
+/// These describe deployment anomalies the gateway chooses to flag without
+/// taking itself offline (contrast with error codes in error_codes.hpp which
+/// fail individual requests). Operators can key on these strings to drive
+/// dashboards, alerts, or automated remediation.
+///
+/// Keep this list in sync with docs/api/warning_codes.rst. All codes MUST be
+/// stable strings; rename = breaking change for downstream consumers.
+
+/// A leaf (non-hierarchical) Component ID was announced by more than one peer
+/// during aggregation merge. Routing falls back to last-writer-wins, so
+/// requests for the affected Component reach one peer non-deterministically.
+/// Resolve by renaming the Component on one side or by modelling it as a
+/// hierarchical parent (declare a child Component with
+/// ``parentComponentId`` pointing at the colliding ID on the owning peer).
+inline constexpr const char * WARN_LEAF_ID_COLLISION = "leaf_id_collision";
+
+}  // namespace ros2_medkit_gateway
--- a/src/ros2_medkit_gateway/include/ros2_medkit_gateway/http/warning_codes.hpp
+++ b/src/ros2_medkit_gateway/include/ros2_medkit_gateway/http/warning_codes.hpp
@@ -34,4 +34,11 @@ namespace ros2_medkit_gateway {
 /// ``parentComponentId`` pointing at the colliding ID on the owning peer).
 inline constexpr const char * WARN_LEAF_ID_COLLISION = "leaf_id_collision";
 
+/// Schema version for the ``warnings`` array on ``GET /health``. Clients can
+/// key on this integer to detect supported warning codes without
+/// string-matching on individual codes. Increment whenever a code is added,
+/// removed, or the shape of an individual warning object changes. Keep in
+/// sync with docs/api/warning_codes.rst.
+inline constexpr int kWarningSchemaVersion = 1;
+
 }  // namespace ros2_medkit_gateway
--- a/src/ros2_medkit_gateway/src/aggregation/aggregation_manager.cpp
+++ b/src/ros2_medkit_gateway/src/aggregation/aggregation_manager.cpp
@@ -17,6 +17,7 @@
 #include <algorithm>
 #include <future>
 #include <string>
+#include <unordered_set>
 #include <utility>
 #include <vector>
 
@@ -397,6 +398,11 @@ AggregationManager::MergedPeerResult AggregationManager::fetch_and_merge_peer_en
     }));
   }
 
+  // Per-peer Component IDs each peer claimed, fed into classify_component_routing
+  // after the loop so that hierarchical-vs-leaf classification sees the full
+  // merged Component set (including subcomponents from other peers).
+  std::vector<PeerClaim> peer_component_claims;
+
   // Collect results and merge sequentially (merge order must be deterministic)
   for (auto & f : futures) {
     auto pfr = f.get();
@@ -406,6 +412,13 @@ AggregationManager::MergedPeerResult AggregationManager::fetch_and_merge_peer_en
       }
       continue;
     }
+
+    PeerClaim claim;
+    claim.peer_name = pfr.peer_name;
+    for (const auto & c : pfr.entities.components) {
+      claim.claimed_entity_ids.insert(c.id);
+    }
+    peer_component_claims.push_back(std::move(claim));
 
     EntityMerger merger(pfr.peer_name);
     merged.areas = merger.merge_areas(merged.areas, pfr.entities.areas);
@@ -449,6 +462,43 @@ AggregationManager::MergedPeerResult AggregationManager::fetch_and_merge_peer_en
           RCLCPP_WARN(*logger, "Entity ID collision: '%s' prefixed for peer '%s'", id.c_str(), name.c_str());
         }
       }
+    }
+  }
+
+  // Reclassify Component routing entries against the fully-merged Component
+  // set. Hierarchical parents (Components referenced as parent_component_id by
+  // another Component) are removed from the routing table so they are served
+  // locally with the merged view, mirroring Areas and Functions. Leaves keep
+  // routing to the owning peer. Multi-peer leaf collisions surface as warnings.
+  std::unordered_set<std::string> all_component_ids;
+  all_component_ids.reserve(merged.components.size());
+  for (const auto & c : merged.components) {
+    all_component_ids.insert(c.id);
+  }
+  for (auto it = merged.routing_table.begin(); it != merged.routing_table.end();) {
+    if (all_component_ids.count(it->first) > 0u) {
+      it = merged.routing_table.erase(it);
+    } else {
+      ++it;
+    }
+  }
+  auto classified = classify_component_routing(merged.components, peer_component_claims);
+  for (auto & [id, name] : classified.routing_table) {
+    merged.routing_table[id] = std::move(name);
+  }
+  merged.leaf_warnings = std::move(classified.leaf_warnings);
+
+  for (const auto & w : merged.leaf_warnings) {
+    if (logger) {
+      std::string peers_str;
+      for (size_t i = 0; i < w.peer_names.size(); ++i) {
+        if (i > 0u) {
+          peers_str += ", ";
+        }
+        peers_str += w.peer_names[i];
+      }
+      RCLCPP_WARN(*logger, "Leaf Component '%s' announced by multiple peers: %s. Routing uses last-writer-wins.",
+                  w.entity_id.c_str(), peers_str.c_str());
     }
   }
 
--- a/src/ros2_medkit_gateway/src/aggregation/aggregation_manager.cpp
+++ b/src/ros2_medkit_gateway/src/aggregation/aggregation_manager.cpp
@@ -347,6 +347,21 @@ AggregationManager::MergedPeerResult AggregationManager::fetch_and_merge_peer_en
   merged.apps = local_apps;
   merged.functions = local_functions;
 
+  // Seed aggregation provenance: every locally-sourced entity contributes
+  // "local" to its own contributors list. EntityMerger will append
+  // "peer:<name>" entries as peer contributions are folded in.
+  auto seed_local = [](auto & entities) {
+    for (auto & e : entities) {
+      if (e.contributors.empty()) {
+        e.contributors.emplace_back("local");
+      }
+    }
+  };
+  seed_local(merged.areas);
+  seed_local(merged.components);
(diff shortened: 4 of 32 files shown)


src/ros2_medkit_gateway/include/ros2_medkit_gateway/discovery/models/area.hpp
```diff
@@ -71,6 +72,9 @@ struct Area {
     if (!source.empty()) {
       x_medkit["source"] = source;
     }
+    if (!contributors.empty()) {
+      x_medkit["contributors"] = sorted_contributors(contributors);
+    }
```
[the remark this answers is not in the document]
bburda: Fixed in 73d1433b: added `#include "ros2_medkit_gateway/discovery/models/common.hpp"` to area.hpp so the header is self-contained regardless of include order (component.hpp and function.hpp already had it).

src/ros2_medkit_integration_tests/test/features/test_daisy_chain_aggregation.test.py
```diff
@@ -0,0 +1,328 @@
...
+PEER_C_PORT = get_test_port(2)
+PRIMARY_URL = f'http://localhost:{PRIMARY_PORT}{API_BASE_PATH}'
+PEER_B_URL = f'http://localhost:{PEER_B_PORT}{API_BASE_PATH}'
+PEER_C_URL = f'http://localhost:{PEER_C_PORT}{API_BASE_PATH}'
+
+# Secondary DDS domains: avoid collision with test_peer_aggregation (offset=1
+# -> domain 230). Daisy chain uses offsets 2 and 3 -> domains 231 and 232.
+PRIMARY_DOMAIN = get_test_domain_id(0)
+PEER_B_DOMAIN = get_test_domain_id(2)
+PEER_C_DOMAIN = get_test_domain_id(3)
```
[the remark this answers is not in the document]
bburda: Good catch, thanks - and you were right about the daisy chain not actually getting distinct DDS domains. Fixed in 73d1433b: `get_test_domain_id(offset)` now maps offsets 1..3 one-to-one to domains 230..232 and rejects anything outside that range with a `ValueError` (DDS max is 232 so silent overflow was the wrong failure mode). Paired that with a CTest `RESOURCE_LOCK medkit_secondary_dds_domains` on `test_peer_aggregation` and `test_daisy_chain_aggregation` in the integration-tests CMakeLists so the two multi-gateway tests never share a secondary domain concurrently.

src/ros2_medkit_integration_tests/test/features/test_daisy_chain_aggregation.test.py
```diff
@@ -0,0 +1,328 @@
...
+        # A leaf ECU that came entirely from peer_b must have contributors
+        # ["peer:peer_b"]. This used to pass silently even if the field
+        # was dropped (because the test looked at a forwarded response from
+        # peer_b which emits the field itself). We only assert the shape
+        # here - forwarding origin is covered by the 1-hop test above.
+        r = requests.get(f'{PRIMARY_URL}/components/ecu-b', timeout=5)
+        self.assertEqual(r.status_code, 200)
+        contributors = r.json().get('x-medkit', {}).get('contributors', [])
+        # Stable ordering: "local" first when present, then peer:* sorted.
+        self.assertEqual(contributors[:1], ['local'])
```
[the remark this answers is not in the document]
bburda: Fair point - the comment described the primary's merged view, but the assertion pins the forwarded response from peer_b. Rewrote the docstring in 73d1433b so it matches reality: the request is transparently forwarded, peer_b serialises the response locally, and the assertion guards that peer_b's `x-medkit.contributors` is present and proxied intact - a regression dropping the field on Area/App/Function detail handlers would flip it. The forwarding itself is covered by `test_primary_forwards_ecu_b_detail_one_hop`.

src/ros2_medkit_gateway/src/openapi/schema_builder.cpp
```diff
@@ -199,10 +217,44 @@ nlohmann::json SchemaBuilder::health_schema() {
...
-          {"properties",
-           {{"status", {{"type", "string"}}}, {"timestamp", {{"type", "integer"}}}, {"discovery", discovery_schema}}},
-          {"required", {"status"}}};
+  nlohmann::json peer_status_schema = {{"type", "object"}, {"additionalProperties", true}};
+
+  nlohmann::json aggregation_warning_schema = {
+      {"type", "object"},
+      {"description",
+       "Operator-actionable aggregation warning. Codes are documented in "
+       "docs/api/warning_codes.rst and stable across releases."},
```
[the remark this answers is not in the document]
bburda: Chose to add the missing doc rather than drop the reference: `docs/api/warning_codes.rst` now exists in 73d1433b with the stable code table (currently just `leaf_id_collision`) plus remediation guidance, and it is linked from `docs/api/index.rst`. The schema description stays pointing at that canonical location.

src/ros2_medkit_gateway/include/ros2_medkit_gateway/http/warning_codes.hpp
```diff
@@ -0,0 +1,37 @@
...
+
+/// Warning codes surfaced on /health.warnings.
+///
+/// These describe deployment anomalies the gateway chooses to flag without
+/// taking itself offline (contrast with error codes in error_codes.hpp which
+/// fail individual requests). Operators can key on these strings to drive
+/// dashboards, alerts, or automated remediation.
+///
+/// Keep this list in sync with docs/api/warning_codes.rst. All codes MUST be
+/// stable strings; rename = breaking change for downstream consumers.
```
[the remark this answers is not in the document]
bburda: Same fix as the schema_builder thread: `docs/api/warning_codes.rst` now exists in 73d1433b with the stable code table and is wired into `docs/api/index.rst`, so this header comment points at a real canonical location.

src/ros2_medkit_gateway/src/http/handlers/health_handlers.cpp
```diff
@@ -69,6 +70,33 @@ void HealthHandlers::handle_health(const httplib::Request & req, httplib::Respon
     // Add peer status when aggregation is active
     if (auto * agg = ctx_.aggregation_manager()) {
       response["peers"] = agg->get_peer_status();
+
+      // Surface operator-actionable aggregation warnings (x-medkit extension).
+      // Always an array when aggregation is active; empty means no active
+      // warnings. Clients can feature-detect via /.capabilities.aggregation
+      // in the root response.
```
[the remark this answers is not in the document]
bburda: Added in 73d1433b: `docs/api/rest.rst` now has a dedicated `GET /api/v1/health` block describing both `peers` and `warnings` arrays, with an example response showing a realistic `leaf_id_collision` payload and a cross-link to the new `warning_codes` page. Always-array semantics and the feature-detection path via `capabilities.aggregation` are documented alongside.

src/ros2_medkit_gateway/src/http/handlers/health_handlers.cpp
```diff
@@ -130,6 +158,7 @@ void HealthHandlers::handle_root(const httplib::Request & req, httplib::Response
         {"tls", tls_config.enabled},
         {"scripts", ctx_.node() && ctx_.node()->get_script_manager() != nullptr &&
                         ctx_.node()->get_script_manager()->has_backend()},
+        {"aggregation", ctx_.aggregation_manager() != nullptr},
         {"vendor_extensions",
```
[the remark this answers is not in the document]
bburda: Done in 73d1433b: the root-endpoint example in `docs/api/rest.rst` now includes `"aggregation": false`, followed by a paragraph explaining that the flag gates the aggregation-only response fields (`peers` and `warnings` on `/health`, `x-medkit.contributors` on entity responses), so clients can feature-detect without probing for field presence.

docs/api/rest.rst
```diff
@@ -49,10 +49,17 @@ Server Capabilities
...
+          "tls": false,
+          "aggregation": false
         }
       }
 
+   The ``capabilities.aggregation`` flag is ``true`` when the gateway has
+   one or more aggregation peers configured. Clients can feature-detect
+   aggregation-only response fields (``peers``, ``warnings`` on ``/health``
+   and ``x-medkit.contributors`` on entities) using this flag instead of
+   probing for field presence.
```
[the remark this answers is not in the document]
bburda: Good catch - the doc wording did not match the code. Fixed in 3999013f: the paragraph now says the flag reflects the aggregation subsystem being enabled (`AggregationManager` wired in), notes that `peers` may be an empty array and `x-medkit.contributors` may contain only `"local"` when no peers are active, and reiterates that clients should feature-detect the extension fields via this flag rather than probing for presence.

src/ros2_medkit_integration_tests/CMakeLists.txt
```diff
@@ -133,6 +133,16 @@ if(BUILD_TESTING)
...
     message(FATAL_ERROR "Integration test domain range exhausted (${_MEDKIT_DOMAIN_START}-${_MEDKIT_DOMAIN_END})")
   endif()
+
+  # Serialise multi-gateway tests that share the secondary DDS domain pool
+  # (230-232, see ros2_medkit_test_utils.constants.get_test_domain_id). CTest
+  # will not run two tests holding this resource lock concurrently, so they
+  # cannot collide on those shared domains.
+  set_tests_properties(
+    test_peer_aggregation
+    test_daisy_chain_aggregation
```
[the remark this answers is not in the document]
bburda: Right - I missed `test_cross_ecu_fanout`, which also calls `get_test_domain_id(1)` and therefore claims domain 230 just like `test_peer_aggregation`. Added it to the same `RESOURCE_LOCK medkit_secondary_dds_domains` in 3999013f so the three multi-gateway tests are now serialised against each other. Extended the comment to spell out the rule: any future test calling `get_test_domain_id` with a non-zero offset must be added to this lock list.

src/ros2_medkit_gateway/src/aggregation/classification.cpp
```diff
@@ -0,0 +1,74 @@
...
+
+ClassifiedRouting classify_component_routing(const std::vector<Component> & merged_components,
+                                             const std::vector<PeerClaim> & peer_claims) {
+  ClassifiedRouting result;
+
+  // Identify hierarchical parents: any Component referenced as parent_component_id.
+  std::unordered_set<std::string> hierarchical_parents;
+  for (const auto & comp : merged_components) {
+    if (!comp.parent_component_id.empty()) {
+      hierarchical_parents.insert(comp.parent_component_id);
```
mfaferek93: **Major - no validation on `parent_component_id`.** This `insert` fires unconditionally for any non-empty parent ID. Three malformed inputs produce silent misbehavior (round-2 traced end-to-end):

1. **Nonexistent parent** (`child.parent_component_id = "ghost"`): "ghost" lands in `hierarchical_parents` set, gets excluded from `routing_table` (line 45-47). No Component object exists. Client `GET /components/ghost` → 404 via `validate_entity_for_route` (handler_context.cpp:254-267). The child itself is correctly routed, but the parent reference is a dangling pointer in the response.

2. **Cycle A↔B** (A.parent = B, B.parent = A): both IDs enter `hierarchical_parents`, neither enters `routing_table`. Both served from local merged cache (both exist there), never forwarded to their owning peer. Semantically wrong: if peer_b owns B, client reading `/components/B` hits the local stub instead of the real peer data.

3. **Self-parent** (A.parent = A): A inserted as its own hierarchical parent, excluded from routing, served locally. Breaks the invariant that hierarchical parents aggregate their children.

Pre-classification sanitization also absent at `peer_client.cpp:180` (JSON deserialization) and `entity_merger.cpp`. Zero guard anywhere.

Suggested fix:
```cpp
std::unordered_set<std::string> known_ids;
for (const auto & comp : merged_components) known_ids.insert(comp.id);

for (const auto & comp : merged_components) {
  const auto & p = comp.parent_component_id;
  if (p.empty()) continue;
  if (p == comp.id) {
    RCLCPP_WARN(..., "Component '%s' declares itself as parent, ignoring", p.c_str());
    continue;
  }
  if (known_ids.count(p) == 0u) {
    RCLCPP_WARN(..., "Component '%s' parent_component_id='%s' not in merged set", comp.id.c_str(), p.c_str());
    continue;
  }
  hierarchical_parents.insert(p);
}
// BFS/DFS cycle detection pass here.
```

Plus unit tests: `parent_references_nonexistent_component`, `two_way_parent_cycle`, `self_parent`.

src/ros2_medkit_gateway/src/aggregation/aggregation_manager.cpp
```diff
@@ -397,6 +413,11 @@ AggregationManager::MergedPeerResult AggregationManager::fetch_and_merge_peer_en
     }));
   }
 
+  // Per-peer Component IDs each peer claimed, fed into classify_component_routing
+  // after the loop so that hierarchical-vs-leaf classification sees the full
+  // merged Component set (including subcomponents from other peers).
+  std::vector<PeerClaim> peer_component_claims;
```
mfaferek93, line 419: **Major - routing destination flips silently across merge snapshots under peer health churn.**

`peer_component_claims` is built by iterating `futures` (line 422 context below) in insertion order of the healthy-peer snapshot at line 371-374. `is_healthy()` filter silently drops failing peers. The "last writer" in `classification.cpp:46` (`result.routing_table[id] = pc.peer_name;`) is whichever peer sits last in this vector.

**Concrete scenario (round-2 traced):**
- Merge N: peers `[peer_b, peer_c]` both claim `ecu-shared`. Last writer = peer_c. `/health.warnings` surfaces both claimants. Routing destination = peer_c.
- Merge N+1: peer_c fails health check, dropped from snapshot. `peer_component_claims = [peer_b]`. Last writer = peer_b. Only 1 claimant = **no collision = no warning**. Routing destination silently flipped from peer_c to peer_b.
- Operator sees /health go from "collision warning present" to "warnings: []" with no signal that routing destination changed.

This is worse than pure nondeterminism: the warning mechanism itself can hide the flip.

Three options:
1. **Sticky routing**: remember last-chosen peer across merges, only flip if that peer becomes unhealthy.
2. **Deterministic ordering**: `std::sort(peer_component_claims.begin(), ..., by peer_name)` before classification so the "last writer" is reproducible.
3. **Emit routing-change event**: new warning code `WARN_ROUTING_DESTINATION_CHANGED` whenever a routing entry for a given `entity_id` differs from the previous merge's entry.

At minimum, update `design/aggregation.rst:291` to call out cross-snapshot instability and the warning-disappearance edge case so operators can plan around it.

src/ros2_medkit_gateway/src/http/handlers/health_handlers.cpp
```diff
@@ -69,6 +70,33 @@ void HealthHandlers::handle_health(const httplib::Request & req, httplib::Respon
...
+        std::string message = "Component '" + w.entity_id + "' is announced by multiple peers (" + peers_list +
+                              "); routing falls back to last-writer-wins which is non-deterministic. Resolve by "
+                              "renaming the Component on one side or by modelling it as a hierarchical parent "
+                              "(declare a child Component with parentComponentId='" +
+                              w.entity_id + "' on the owning peer).";
+        warnings.push_back({
+            {"code", WARN_LEAF_ID_COLLISION},
+            {"message", std::move(message)},
+            {"entity_ids", json::array({w.entity_id})},
+            {"peer_names", w.peer_names},
```
mfaferek93, line 102: **Minor - peer names leak on unauthenticated endpoint.** Round-2 verified: `auth.enabled` defaults to `false` at `gateway_node.cpp:169`. Anyone reachable to `/health` enumerates every peer name in the deployment. Even with auth enabled, `/health` is in VIEWER/OPERATOR/CONFIGURATOR roles so any authenticated user sees the names.

Fine in controlled LAN deployments, leaks topology in shared-infra / multi-tenant setups. Since design intent is "operator-visible", that's OK - but please document explicitly in `docs/api/rest.rst` so operators know peer names on this endpoint are public-by-default.

If you want to gate it: wrap the `warnings` array in the same auth config that gates admin-adjacent info, while leaving the top-level `status`/`uptime` unauthenticated for classical health probes.

src/ros2_medkit_gateway/include/ros2_medkit_gateway/http/warning_codes.hpp
```diff
@@ -0,0 +1,37 @@
...
+
+/// A leaf (non-hierarchical) Component ID was announced by more than one peer
+/// during aggregation merge. Routing falls back to last-writer-wins, so
+/// requests for the affected Component reach one peer non-deterministically.
+/// Resolve by renaming the Component on one side or by modelling it as a
+/// hierarchical parent (declare a child Component with
+/// ``parentComponentId`` pointing at the colliding ID on the owning peer).
+inline constexpr const char * WARN_LEAF_ID_COLLISION = "leaf_id_collision";
```
mfaferek93, line 35: **Minor - no schema version for the `warnings` array.** Codes are stable strings (good), but when the second code ships, clients have no contract to feature-detect "supports new_code_X" without string-matching. `capabilities.aggregation` is a boolean — too coarse.

Suggested: add `response["warning_schema_version"] = 1;` in `health_handlers.cpp`, bump on every new code added, document in warning_codes.rst. Typed clients (MCP, Web UI, Foxglove) can then make the "known codes" set explicit and log+degrade on unknown codes per version.

src/ros2_medkit_gateway/test/test_aggregation_classification.cpp
```diff
@@ -0,0 +1,202 @@
...
+  std::vector<PeerClaim> claims = {
+      make_claim("peer_b", {"ecu-shared"}),
+      make_claim("peer_c", {"ecu-shared"}),
+  };
+
+  auto result = classify_component_routing(merged, claims);
+
+  // Routing exists - last-writer-wins (peer_c processed last here).
+  ASSERT_EQ(result.routing_table.count("ecu-shared"), 1u);
+  EXPECT_EQ(result.routing_table.at("ecu-shared"), "peer_c");
```
mfaferek93, line 192: **Minor - collision test is unidirectional.** Input is hardcoded `[peer_b, peer_c]` at line 184-185, asserted routing destination is `peer_c` (last). Flipping input to `[peer_c, peer_b]` should route to `peer_b`. That flip case is not tested, so a regression that breaks last-writer semantics (e.g., accidentally sorting claims by peer_name before classification) wouldn't be caught.

Add a parametrized or second test:
```cpp
TEST(AggregationClassification, leaf_collision_routing_respects_input_order) {
  // [peer_c, peer_b] → routing must go to peer_b
  ...
}
```

Also missing from this file (round-2 confirmed all 6 test names): `two_way_parent_cycle`, `self_parent_ignored`, `parent_references_nonexistent_component`, `empty_merged_set`, `four_level_nesting`. See companion Major comment on classification.cpp:29.

src/ros2_medkit_integration_tests/test/features/test_daisy_chain_aggregation.test.py
```diff
@@ -0,0 +1,335 @@
...
+
+        for label, url in (('primary', PRIMARY_URL),
+                           ('peer_b', PEER_B_URL),
+                           ('peer_c', PEER_C_URL)):
+            cls._wait_for_health(url, label)
+        # Primary must see ecu-b (1 hop) AND ecu-c (2 hops) via aggregation.
+        # Sub-components (with parent_component_id) appear under the parent's
+        # /subcomponents endpoint, not the top-level /components list.
+        cls._wait_for_subcomponents(
+            PRIMARY_URL, 'robot-x', {'ecu-a', 'ecu-b', 'ecu-c'}, 'primary',
```
mfaferek93, line 198: **Minor - no end-to-end negative collision scenario in daisy-chain tests.** All 10 test methods are positive paths. `robot-x` collision is intentional hierarchical merge (not an error case), and line 275 asserts `body['warnings'] == []` for healthy chain.

Recommend a dedicated test: configure peer_b AND peer_c to both claim a new leaf `ecu-shared` (no parent_component_id), verify from primary:
1. `/components/ecu-shared` returns 200 and routes somewhere (either peer_b or peer_c).
2. `/health.warnings` contains a `leaf_id_collision` entry listing both `peer_b` and `peer_c` under `peer_names`.
3. `x-medkit.contributors` for the routed response reflects the resolving peer (not both).

This exercises the actual wire-level warning path under the real daisy-chain topology that unit tests can't simulate. Also covers the nondeterminism concern on `aggregation_manager.cpp:419` via a follow-up test that kills one of the colliding peers and re-asserts warnings disappearance + routing flip (per the Major on that line).

src/ros2_medkit_gateway/test/test_entity_merger.cpp
```diff
@@ -408,6 +431,81 @@ TEST(EntityMerger, remote_only_area_gets_routing_entry) {
...
+  // Local keeps "local"
+  ASSERT_EQ(result[0].contributors.size(), 1u);
+  EXPECT_EQ(result[0].contributors[0], "local");
+  // Prefixed remote gets only peer contributor
+  ASSERT_EQ(result[1].contributors.size(), 1u);
+  EXPECT_EQ(result[1].contributors[0], "peer:peer_b");
+}
+
+// @verifies REQ_INTEROP_003
+TEST(EntityMerger, contributors_no_duplicate_on_repeat_merge) {
```
mfaferek93, line 487: **Minor - contributor presentation order not asserted on the user-visible path.** This test verifies dedup in the internal `contributors_` vector, which is correct. But clients actually see the output of `XMedkit::contributors()` in `x_medkit.cpp:110-128`, which sorts `["local", peers alphabetical]`.

Suggest one additional test that exercises the sorted output: inputs arriving in reverse order (e.g., `["peer:zulu", "peer:alpha", "local"]`) should serialize as `["local", "peer:alpha", "peer:zulu"]`. Otherwise a regression flipping the sort direction wouldn't be caught by unit tests — only by an integration test that happens to inspect the exact contributor order.

src/ros2_medkit_gateway/src/aggregation/classification.cpp
bburda: Validation added in `classify_component_routing`: self-parent, unknown parent IDs, and cycle members are dropped from `hierarchical_parents` and surfaced as `malformed_parent_warnings`, which the aggregation manager logs via `RCLCPP_WARN`. Affected Components fall back to leaf routing. Three unit tests cover the cases (ghost parent, self parent, two-way cycle) in `test_aggregation_classification.cpp`. Commit 95bac930.

src/ros2_medkit_gateway/src/aggregation/aggregation_manager.cpp
bburda: Went with the documentation-minimum option you listed. `design/aggregation.rst` now spells out the cross-snapshot flip mechanics and the warning-disappearance edge case, with the recommendation to alert on the empty→warning and warning→empty transitions rather than on warning presence alone. Sticky routing was rejected because it masks the anomaly instead of surfacing it; deterministic sort-by-peer-name directly conflicts with the input-order contract asserted by the companion test-level comment. Commit 95bac930.

src/ros2_medkit_gateway/src/http/handlers/health_handlers.cpp
bburda: Documented in `docs/api/rest.rst` with an explicit security note on `/health`: peer names are exposed without auth by default, and even with auth enabled the field is readable by `viewer`/`operator`/`configurator`. The note recommends an authenticating reverse proxy or ingress-level restriction on shared-infra deployments. Commit 95bac930.

src/ros2_medkit_gateway/include/ros2_medkit_gateway/http/warning_codes.hpp
bburda: Added `warning_schema_version: 1` on the `/health` response alongside `warnings`, backed by a `kWarningSchemaVersion` constant in `warning_codes.hpp` and a `Schema versioning` section in `docs/api/warning_codes.rst` that documents the bump policy and the log-and-ignore contract for unknown codes. The new integration test asserts the field is present on both healthy-chain and collision paths. Commit 95bac930.

src/ros2_medkit_gateway/test/test_aggregation_classification.cpp
bburda: Added `leaf_collision_routing_respects_input_order` with flipped input `[peer_c, peer_b]` asserting routing resolves to `peer_b`. The three malformed-parent tests from the companion major comment (`parent_references_nonexistent_component_falls_back_to_leaf`, `self_parent_is_ignored_and_warned`, `two_way_parent_cycle_falls_back_to_leaves_with_warning`) are in the same file. Commit 95bac930.

src/ros2_medkit_integration_tests/test/features/test_daisy_chain_aggregation.test.py
bburda: Added a dedicated three-gateway scenario in `test_leaf_collision_aggregation.test.py`: `peer_b` and `peer_c` both claim a leaf `ecu-shared`, primary aggregates both directly. Assertions cover `warning_schema_version == 1`, a `leaf_id_collision` entry listing both peer names, and `GET /components/ecu-shared` returning 200 without a merged-view response. The peer-churn follow-up (kill a colliding peer, re-assert warning disappearance + routing flip) was intentionally deferred to keep this PR focused; it belongs with the sticky-routing-vs-warning tradeoff discussion. Commit 95bac930.

src/ros2_medkit_gateway/test/test_entity_merger.cpp
bburda: Added three tests in `test_x_medkit.cpp` exercising the user-visible path through `XMedkit::contributors()`: reverse-order input `[peer:zulu, peer:alpha, local]` serialises as `[local, peer:alpha, peer:zulu]`, plus the empty-input and no-local variants. Also refactored `XMedkit::contributors()` to delegate to `sorted_contributors()` in `common.hpp` so list and detail responses share a single sort implementation (two copies silently drift, one helper cannot). Commit 95bac930.

src/ros2_medkit_integration_tests/CMakeLists.txt
```diff
@@ -133,6 +133,18 @@ if(BUILD_TESTING)
...
+  # (230-232, see ros2_medkit_test_utils.constants.get_test_domain_id). CTest
+  # will not run two tests holding this resource lock concurrently, so they
+  # cannot collide on those shared domains. Every integration test that
+  # calls get_test_domain_id(offset>0) must be listed here.
+  set_tests_properties(
+    test_peer_aggregation
+    test_cross_ecu_fanout
+    test_daisy_chain_aggregation
+    PROPERTIES RESOURCE_LOCK medkit_secondary_dds_domains
+  )
```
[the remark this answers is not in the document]
bburda: CMake `NAME_WE` returns the file name without its **longest** extension, so `test_daisy_chain_aggregation.test.py` maps to target `test_daisy_chain_aggregation` (not `test_daisy_chain_aggregation.test`). Confirmed via `ctest -N` on the built package — `test_peer_aggregation`, `test_cross_ecu_fanout`, `test_daisy_chain_aggregation`, and `test_leaf_collision_aggregation` all match their `set_tests_properties` entries, and the `RESOURCE_LOCK` is enforced correctly at runtime. No change needed.

src/ros2_medkit_gateway/src/http/x_medkit.cpp
```diff
@@ -104,6 +107,26 @@ XMedkit & XMedkit::last_feedback(const nlohmann::json & feedback) {
...
+  // consistent order regardless of how peers were merged internally.
+  std::vector<std::string> sorted(contributors.begin(), contributors.end());
+  std::sort(sorted.begin(), sorted.end(), [](const std::string & a, const std::string & b) {
+    const bool a_local = (a == "local");
+    const bool b_local = (b == "local");
+    if (a_local != b_local) {
+      return a_local;
+    }
+    return a < b;
+  });
```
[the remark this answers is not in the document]
bburda: Refactored: `XMedkit::contributors()` now delegates to `sorted_contributors()` in `discovery/models/common.hpp`. Both list responses (direct `to_json()`) and detail responses (via `XMedkit`) now share one sort implementation. Commit 95bac930.

mfaferek93 (approved):

PR #377, opened 2026-04-15
feat(plugins): notify_entities_changed + manifest fragments

### LINKED ISSUE #376: Plugin SDK: notify gateway of entity-surface changes + manifest reload
## Problem

Plugins that implement `UpdateProvider` (or any other plugin that mutates the entity surface at runtime) currently have no way to tell the gateway "my entities changed, please refresh your tree". The gateway's entity tree is populated from:

1. The manifest file, loaded once at startup.
2. Runtime discovery, which picks up ROS 2 graph changes on a polling interval.

When a plugin deploys a new app (e.g. drops a Python node into `/opt/app-root/addons/<id>/main.py` and starts it) the runtime discovery catches the new node within a few seconds, but:

- It shows up as an **orphan** (not in manifest), so it can't be attached to a parent component.
- There's no way for the plugin to ship a matching manifest fragment at deploy time and have the gateway pick it up.
- After the plugin deletes the app (e.g. rollback), the manifest entry — if it ever existed — is never cleared.

This is a gap for any plugin whose job is to mutate the entity surface: OTA installers, rosbag injectors, dynamic config deployers, hot-reloadable adapters.

## Proposal

Two parts:

### 1. `PluginContext::notify_entities_changed(scope)`

Add to `ros2_medkit_gateway::PluginContext` a method plugins call when they have finished mutating entities:

```cpp
struct EntityChangeScope {
  std::optional<std::string> area_id;
  std::optional<std::string> component_id;
  // empty/null = full refresh
};

virtual void notify_entities_changed(const EntityChangeScope & scope) = 0;
```

The gateway responds by:
- Re-reading the manifest file from disk (cheap, just a single yaml parse).
- Scanning a well-known fragments directory (new, see part 2).
- Re-running runtime discovery for the affected scope.
- Merging the fresh result into the entity cache via the existing `merge_pipeline`.

Existing `ResourceChangeNotifier` stays as-is — it's for resource-item changes (faults, data, config), not entity-surface changes. The two concerns are separate.

### 2. Manifest fragments directory

Introduce a gateway parameter `manifest.fragments_dir` (default: empty = disabled). When set, the manifest loader scans that directory for `*.yaml` / `*.yml` files at every manifest load and merges them on top of the base manifest. Fragment schema is a strict subset of the main manifest (apps / functions / subcomponents only; no new areas or top-level components without `allow_manifest_override: true`).

This gives plugins a place to drop per-deploy manifest chunks alongside their deployed files. Removing the fragment (during plugin's cleanup / rollback) + calling `notify_entities_changed` makes the entry disappear.

## Non-goals

- No hot-reload of plugin binaries.
- No change to SSE / trigger / websocket APIs (use existing `ResourceChangeNotifier` for resource-level events).
- No new HTTP endpoint (the gateway's admin surface is and should remain SOVD-only).

## Compatibility

- `PluginContext` is a versioned interface (`PLUGIN_API_VERSION`). This change bumps the API version and gets a default no-op implementation for backwards compatibility with plugins built against the previous header.
- `manifest.fragments_dir` defaults to empty, so no behavior change for users who don't opt in.

## Acceptance criteria

- [ ] New `EntityChangeScope` type and `PluginContext::notify_entities_changed` method.
- [ ] Gateway implementation re-parses manifest + runs discovery for the given scope.
- [ ] New `manifest.fragments_dir` parameter with fragment loader.
- [ ] Unit tests covering: full-surface refresh, scoped refresh (by area, by component), fragment merge, fragment removal after refresh.
- [ ] Integration test: a test plugin drops a fragment file, calls `notify_entities_changed`, the new app appears in `GET /apps`; removes the fragment + notifies, the app disappears.
- [ ] Docs: `docs/design/` page describing the lifecycle, extended CHANGELOG, PLUGIN_API_VERSION bump.

bburda:
## Summary

Implements issue #376.

Plugin SDK gains two new pieces that together let plugins mutate the entity surface at runtime and have the gateway reflect those changes immediately:

1. **`PluginContext::notify_entities_changed(EntityChangeScope)`** (plugin API v7) - a lifecycle hook a plugin calls after it finishes adding or removing entities on disk or in the running runtime. Default implementation is a no-op so v6 plugins keep working unchanged; a v7 gateway responds by reloading the manifest source and re-running discovery for the entity cache.
2. **`discovery.manifest.fragments_dir`** - a directory of drop-in `*.yaml` / `*.yml` manifest chunks that the gateway scans on every manifest load / reload and merges on top of the base manifest. Fragments contribute `apps`, `components`, `functions`; reserved top-level fields (`areas`, `metadata`, `scripts`, `capabilities`, `lock_overrides`) are rejected with a `FRAGMENT_FORBIDDEN_FIELD` validation error. Files merged in alphabetical order so duplicate-id errors are deterministic.

The two are intentionally decoupled from `ResourceChangeNotifier`, which stays for resource-item change events (faults, data, configurations). Entity-surface refresh and resource-item streaming are different concerns - plugins that do both simply call both APIs.

No SOVD admin endpoint was added: the surface-refresh trigger is a plugin-side call against `PluginContext`, not an HTTP route.

---

## Issue

- closes #376

---

## Type

- [ ] Bug fix
- [x] New feature or tests
- [ ] Breaking change
- [ ] Documentation only

---

## Testing

- 10 unit tests covering `ManifestManager::apply_fragments` merge semantics (no fragments, missing dir, single merge, deterministic multi-file order, non-yaml skipped, forbidden areas, duplicate ids, reload picks up added + removed files, setter round-trip).
- 6 unit tests covering `EntityChangeScope` + the default no-op `PluginContext::notify_entities_changed` backwards-compat guarantee.
- 3 integration tests driving a real `GatewayNode` through the full lifecycle: fragment dropped after startup + notify -> visible; fragment removed + notify -> gone; notify-with-no-fragments is a safe no-op.
- Full gateway unit + integration test suite (77 tests) green.

## Notes for reviewers

- This PR targets `fix/aggregation-proxy-logs-and-hosts` (PR #374). Rebasing onto `main` once #374 merges is trivial - no overlap with its changes.
- Scope hint inside `EntityChangeScope` is informational in v7 (gateway always does a full `refresh_cache`); the interface is in place so a future scoped-rediscovery optimisation lands behind it without an API change.
- Design walkthrough in `src/ros2_medkit_gateway/design/plugin_entity_notifications.rst`; `CHANGELOG.rst` has an Unreleased section.


src/ros2_medkit_gateway/include/ros2_medkit_gateway/discovery/manifest/manifest_manager.hpp
```diff
@@ -73,10 +73,50 @@ class ManifestManager {
...
+   * the base manifest.
+   *
+   * Merge rules:
+   *   * apps / components / functions are appended; duplicate IDs across
+   *     the merged manifest cause validation to fail the same way they
+   *     would in a single file.
+   *   * fragments may not declare top-level `areas`, `metadata`, `config`,
+   *     `scripts`, `capabilities`, or `lock_overrides` - those must stay in
+   *     the base manifest. A fragment that declares any of them is skipped
+   *     with a warning.
```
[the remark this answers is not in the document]
bburda: Doc aligned with behavior in 2cdd5aed: forbidden fields list now includes `discovery`, and the text says each violation is a `FRAGMENT_FORBIDDEN_FIELD` validation error that fails the load / reload.

src/ros2_medkit_gateway/include/ros2_medkit_gateway/discovery/manifest/manifest_parser.hpp
```diff
@@ -51,6 +51,19 @@ class ManifestParser {
...
+   *
+   * Fragments declare a subset of a manifest (apps, components, functions)
+   * intended to be merged on top of a base manifest at load time. They do
+   * NOT carry a manifest_version - that field is owned by the base.
+   *
+   * @param file_path Path to fragment YAML
+   * @return Parsed manifest with only the fragment-allowed sections populated
+   * @throws std::runtime_error if file cannot be read or parsed
+   */
+  Manifest parse_fragment_file(const std::string & file_path) const;
```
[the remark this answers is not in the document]
bburda: Doc updated in 2cdd5aed to describe the actual tolerant behavior: fragments may omit `manifest_version` (a synthetic `"1.0"` is injected before parsing) or declare it explicitly, and forbidden fields are still parsed so `apply_fragments` can reject them with a specific validation error.

src/ros2_medkit_gateway/src/gateway_node.cpp
```diff
@@ -1497,6 +1504,32 @@ AggregationManager * GatewayNode::get_aggregation_manager() const {
...
+  // since the last load stay invisible to the pipeline.
+  if (discovery_mgr_) {
+    if (auto * manifest = discovery_mgr_->get_manifest_manager()) {
+      if (!manifest->get_manifest_path().empty()) {
+        manifest->reload_manifest();
+      }
+    }
+  }
+  refresh_cache();
+}
```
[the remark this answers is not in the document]
bburda: Real bug, fixed in 2cdd5aed. Added a `std::recursive_mutex refresh_mutex_` on `GatewayNode`. `refresh_cache()` locks it at the top, and `handle_entity_change_notification()` locks it around both the `reload_manifest()` step and the subsequent `refresh_cache()` call (recursive so the re-entry is a no-op). The timer callback and any plugin thread now serialize for the full refresh pipeline, not just the cache write.

src/ros2_medkit_gateway/src/plugins/plugin_context.cpp
```diff
@@ -236,6 +236,16 @@ class GatewayPluginContext : public PluginContext {
     return node_->get_condition_registry();
   }
 
+  // ---- Entity surface notifications (plugin API v7) ----
+
+  void notify_entities_changed(const EntityChangeScope & scope) override {
+    // Delegate to GatewayNode::handle_entity_change_notification which
+    // serializes concurrent plugin notifications onto the discovery thread.
+    // Pass the scope by value because the caller's copy may disappear
+    // before the worker thread dequeues the request.
```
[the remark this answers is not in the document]
bburda: Comment rewritten in 2cdd5aed to match the actual synchronous behavior: forwards directly to `GatewayNode::handle_entity_change_notification`, caller thread runs the full refresh before returning, and serialization across concurrent notifications is handled by the gateway-side refresh mutex.

src/ros2_medkit_gateway/include/ros2_medkit_gateway/gateway_node.hpp
```diff
@@ -186,6 +187,17 @@ class GatewayNode : public rclcpp::Node {
...
   AggregationManager * get_aggregation_manager() const;
 
+  /**
+   * @brief Handle a plugin's `PluginContext::notify_entities_changed` request.
+   *
+   * Runs a single `refresh_cache()` pass synchronously. The scope hint is
+   * accepted and logged but the current implementation ignores it and always
+   * does a full refresh - future work may limit the rediscovery to the
+   * indicated area / component. The entry point is thread-safe: concurrent
+   * callers serialize on the cache's internal mutex.
```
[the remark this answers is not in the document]
bburda: Header comment rewritten in 2cdd5aed. It now says the entry point is safe from any thread because `refresh_cache()` serializes on an internal refresh mutex (not the cache mutex), and calls out that the mutex is needed because the refresh pipeline reaches beyond the cache into discovery state that is not itself thread-safe.

src/ros2_medkit_gateway/src/discovery/manifest/manifest_manager.cpp
```diff
@@ -444,6 +464,88 @@ void ManifestManager::build_lookup_maps() {
...
+      validation_result_.add_error("FRAGMENT_FORBIDDEN_FIELD",
+                                   "fragment may not declare '" + field + "'", path.string());
+      log_error("Fragment " + path.string() + " declares forbidden field '" + field + "'");
+      ok = false;
+    };
+    if (!fragment.areas.empty()) reject("areas");
+    // manifest_version is auto-injected by parse_fragment_file when the
+    // fragment omits it, so we don't check that field here - fragments are
+    // free to declare or omit it without penalty.
+    if (!fragment.metadata.name.empty() || !fragment.metadata.version.empty()) reject("metadata");
```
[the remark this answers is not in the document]
bburda: Fixed in 2cdd5aed. `apply_fragments` now rejects `metadata` when any of `name`, `description`, `version`, or `created_at` is set. The `discovery` top-level key (which maps to `Manifest::config` and has non-empty defaults, so it is not detectable on the parsed struct) is caught by re-loading the raw YAML and checking for the `discovery` top-level key. Added three gtest cases covering `metadata.description`, `metadata.created_at`, and `discovery`.

src/ros2_medkit_gateway/src/discovery/manifest/manifest_parser.cpp
```diff
@@ -36,6 +36,52 @@ Manifest ManifestParser::parse_file(const std::string & file_path) const {
...
+  // reuse the main parse_string pipeline so every field is parsed with the
+  // exact same logic as a full manifest. Anything forbidden in fragments
+  // (areas, metadata, scripts, ...) is still parsed, which lets the caller
+  // detect and reject it with a specific error message.
+  std::ifstream file(file_path);
+  if (!file.is_open()) {
+    throw std::runtime_error("Cannot open manifest fragment: " + file_path);
+  }
+  std::stringstream buffer;
+  buffer << file.rdbuf();
```
mfaferek93, line 66: **Major - no file size limit before reading fragment into memory.** `buffer << file.rdbuf()` reads the entire file with no cap (same pattern at line 34-35 for base manifest, pre-existing). `YAML::Load()` downstream (line 88) has no depth/entity-expansion limit either.

`fragments_dir` is user-configurable via ROS param `discovery.manifest.fragments_dir` (gateway_node.cpp:200). A 1GB file (misconfiguration or symlink to log) causes OOM. A YAML billion-laughs file (`&a [*a, *a]` repeated) causes CPU exhaustion via yaml-cpp's default anchor/alias expansion.

Suggested fix before the `ifstream` open:
```cpp
const auto fsize = std::filesystem::file_size(path, ec);
if (fsize > kMaxFragmentBytes) {
  throw std::runtime_error("fragment exceeds " + std::to_string(kMaxFragmentBytes) + " byte limit");
}
```
with e.g. `constexpr std::uintmax_t kMaxFragmentBytes = 1'048'576;` (1 MB). For YAML depth, consider post-parse tree-depth validation or yaml-cpp `LoadFile()` with a custom handler if available.

src/ros2_medkit_gateway/design/plugin_entity_notifications.rst
```diff
@@ -0,0 +1,138 @@
...
+error message.
+
+A missing fragments directory is *not* an error - the plugin can
+create the directory lazily on first install.
+
+Compatibility
+-------------
+
+* ``PLUGIN_API_VERSION`` bumped from v6 to v7.
+* ``PluginContext::notify_entities_changed`` has a default no-op
```
mfaferek93, line 165: **Minor - "continue to load unchanged" is misleading (round-2 verified).** All in-tree plugins export `return PLUGIN_API_VERSION;` from the header constant (opcua:20, graph_provider:20, sovd_service_interface:20), not a hardcoded number. Recompilation against v7 headers auto-bumps to 7 - no code changes needed. So SOURCE compatibility is correct.

But this sentence says "plugins compiled against v6 continue to load unchanged", which implies binary compatibility (no recompilation). A v6-compiled .so exporting `6` WILL be rejected by `plugin_loader.cpp:161` (`version != PLUGIN_API_VERSION`).

Suggest reword: "Plugin source code written against v6 compiles and runs unchanged against v7 headers; recompilation is required but no code changes are needed."

src/ros2_medkit_gateway/src/discovery/manifest/manifest_manager.cpp
```diff
@@ -444,6 +466,117 @@ void ManifestManager::build_lookup_maps() {
...
+  }
+  std::sort(files.begin(), files.end());
+
+  bool ok = true;
+  for (const auto & path : files) {
+    log_info("Merging manifest fragment: " + path.string());
+    Manifest fragment;
+    try {
+      fragment = parser_.parse_fragment_file(path.string());
+    } catch (const std::exception & e) {
```
mfaferek93, line 542: **Minor - one bad fragment blocks all fragments from taking effect (round-2 traced).** Parse error here records an error and `continue`s to the next file (good - valid fragments are still parsed). But `apply_fragments()` returns `false` when ANY fragment errored (line 577), which causes `load_manifest()` to fail at line 42-45, restoring the old manifest.

Result: a single typo in one fragment file prevents ALL fragments (including valid ones) from activating. The valid fragment entities were merged in-place (lines 567-575) but rolled back at the load gate.

Is this the intended contract? If yes, add a test + document in design doc. If not, consider `ok = false` only for non-recoverable errors, and log a warning for individual fragment parse failures while keeping the rest.

src/ros2_medkit_gateway/src/discovery/manifest/manifest_manager.cpp
```diff
@@ -444,6 +466,117 @@ void ManifestManager::build_lookup_maps() {
...
+    // A missing directory is "no fragments", not an error. This lets the
+    // deploy side create the directory lazily on first install.
+    return true;
+  }
+
+  // Collect + sort file paths so the merge order is deterministic. Without
+  // this the validator's "duplicate id" error could point at a different
+  // fragment run to run, which is nightmarish to diagnose.
+  std::vector<std::filesystem::path> files;
+  for (const auto & entry : std::filesystem::directory_iterator(dir, ec)) {
```
mfaferek93, line 507: **Minor - symlinks followed without validation.** `directory_iterator` defaults follow symlinks. If `fragments_dir` is world-writable or NFS-mounted, a symlink `evil.yaml -> /etc/shadow` causes the gateway to read and parse an unintended file. Currently read-only (no writes to followed path) and parse errors are caught, so impact is information leakage via error messages that may include partial file content.

Note: k8s ConfigMap mounts use symlinks internally, so a blanket `is_symlink()` reject would break that use case. Consider: validate that the resolved path is under `fragments_dir` via `std::filesystem::canonical()` comparison.

src/ros2_medkit_gateway/src/discovery/manifest/manifest_manager.cpp
```diff
@@ -444,6 +466,117 @@ void ManifestManager::build_lookup_maps() {
...
+    } catch (const std::exception & e) {
+      validation_result_.add_error("FRAGMENT_PARSE", e.what(), path.string());
+      log_error("Failed to parse fragment " + path.string() + ": " + e.what());
+      ok = false;
+      continue;
+    }
+
+    // Fragments may only contribute apps, components, functions. Reject any
+    // attempt to redeclare top-level properties owned by the base manifest.
+    auto reject = [&](const std::string & field) {
```
mfaferek93, line 551: **Minor - unknown top-level fragment fields silently ignored.** Only `apps`, `components`, `functions` (and the forbidden set) are recognized. A fragment with typo `app:` (singular) instead of `apps:` parses successfully with zero entities and zero feedback. Operator gets no signal that their fragment did nothing.

Suggest: after parsing, if the fragment contributed zero entities AND had non-empty YAML content, emit `RCLCPP_WARN("Fragment '%s' parsed but contributed no entities - check top-level field names (expected: apps, components, functions)", path)`. Catches the most common mistake without rejecting intentionally-empty fragments.

src/ros2_medkit_gateway/test/test_manifest_fragments.cpp
```diff
@@ -0,0 +1,368 @@
...
+  for (const auto & err : vr.errors) {
+    if (err.rule_id == "FRAGMENT_FORBIDDEN_FIELD" && err.message.find("discovery") != std::string::npos) {
+      found = true;
+      break;
+    }
+  }
+  EXPECT_TRUE(found) << "expected FRAGMENT_FORBIDDEN_FIELD error naming 'discovery'";
+}
+
+TEST_F(FragmentsFixture, DuplicateIdsAcrossFragmentsFailValidation) {
```
mfaferek93, line 258: **Minor - missing test: base manifest + fragment declaring the same entity ID.** This test covers cross-fragment duplicate IDs. The validator (manifest_validator.cpp:45-87 `validate_unique_ids`) runs on the merged manifest (base + fragments appended), so it SHOULD catch base+fragment collisions too - but no test proves it.

Add: `BaseAndFragmentDuplicateIdFailsValidation` - base manifest declares `app: sensor_monitor`, fragment also declares `app: sensor_monitor` -> expect validation error R002-R005.

src/ros2_medkit_gateway/include/ros2_medkit_gateway/plugins/plugin_types.hpp
```diff
@@ -28,7 +28,14 @@
...
 
 /// Current plugin API version. Plugins must export this value from plugin_api_version().
-constexpr int PLUGIN_API_VERSION = 6;
+///
+/// Version history:
+/// - v6: ScriptProvider, locking API, extended PluginContext (entity snapshot,
+///       fault listing, sampler registration).
+/// - v7: PluginContext::notify_entities_changed(EntityChangeScope) for plugins
+///       that mutate the entity surface at runtime. Default implementation is
+///       a no-op so plugins built against v6 continue to load.
```
mfaferek93: **Minor - comment says "continue to load" but means "continue to compile".** Same issue as design doc (see comment there). All plugins in the tree use `return PLUGIN_API_VERSION;` which is v7 after recompilation. A pre-compiled v6 .so binary would be rejected by the strict `!=` check at `plugin_loader.cpp:161`. This comment should clarify the recompilation requirement.

src/ros2_medkit_gateway/src/discovery/manifest/manifest_manager.cpp
```diff
@@ -444,6 +466,117 @@ void ManifestManager::build_lookup_maps() {
...
+    }
+  }
+  std::sort(files.begin(), files.end());
+
+  bool ok = true;
+  for (const auto & path : files) {
+    log_info("Merging manifest fragment: " + path.string());
+    Manifest fragment;
+    try {
+      fragment = parser_.parse_fragment_file(path.string());
```
mfaferek93, line 541: **Minor - TOCTOU race on fragment files.** Directory listing at line 496-507 is captured, then files are read sequentially here. If a plugin writes a fragment concurrently (e.g., via `notify_entities_changed` from another thread), the gateway may read a partially-written file → YAML parse error → all fragments rejected (per the all-or-nothing behavior at line 516+577).

Suggested contract for plugins: write to a temp file first, then atomic `rename()` into `fragments_dir`. Worth documenting in `plugin_context.hpp` docstring for `notify_entities_changed` and in the design doc.

src/ros2_medkit_gateway/include/ros2_medkit_gateway/discovery/manifest/manifest_manager.hpp
```diff
@@ -216,8 +257,16 @@ class ManifestManager {
   rclcpp::Node * node_;
   mutable std::mutex mutex_;
 
+  /// Merge the fragments found in `fragments_dir_` on top of `base`.
+  /// Returns true on success (or no fragments); false if any fragment
+  /// rejected the merge (conflicting IDs, disallowed top-level fields).
+  /// Validation errors from individual fragments are appended to
```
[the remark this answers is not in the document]
bburda: Docstring rewritten in 2befc477 to say apply_fragments only reports per-fragment parse / forbidden-field failures and that duplicate IDs are surfaced by the subsequent ManifestValidator::validate() pass over the merged manifest.

src/ros2_medkit_gateway/src/discovery/manifest/manifest_manager.cpp
```diff
@@ -444,6 +466,117 @@ void ManifestManager::build_lookup_maps() {
...
+    // allow_manifest_override=true) are indistinguishable from "not declared" on the
+    // parsed struct, so detect a `discovery:` top-level key by re-reading the raw YAML.
+    try {
+      YAML::Node raw = YAML::LoadFile(path.string());
+      if (raw && raw["discovery"]) {
+        reject("discovery");
+      }
+    } catch (const YAML::Exception & e) {
+      validation_result_.add_error("FRAGMENT_PARSE", e.what(), path.string());
+      log_error("Failed to re-parse fragment " + path.string() + " for discovery check: " + e.what());
```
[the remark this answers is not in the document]
bburda: Switched to raw-YAML key-presence in 2befc477. apply_fragments now loads the yaml once and rejects any of areas / metadata / scripts / capabilities / lock_overrides / discovery regardless of whether the value is empty. Same raw walk also warns on unknown top-level keys (the `app:` vs `apps:` typo case). Backed by new EmptyReservedKeysAreRejected and UnknownTopLevelKeyIsIgnoredWithWarning tests in d08de44e.

src/ros2_medkit_gateway/test/test_plugin_notify_integration.cpp
```diff
@@ -0,0 +1,309 @@
...
+  }
+
+  void SetUp() override {
+    auto tmp = std::filesystem::temp_directory_path();
+    work_dir = tmp / ("medkit-notify-integ-" + std::to_string(::testing::UnitTest::GetInstance()->random_seed()));
+    std::filesystem::create_directories(work_dir);
+    fragments_dir = work_dir / "fragments";
+    std::filesystem::create_directories(fragments_dir);
+    manifest_path = work_dir / "manifest.yaml";
+    std::ofstream(manifest_path) << kBaseManifest;
```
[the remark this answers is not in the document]
bburda: Fixed in 3ba837ed. work_dir now derives from `current_test_info()->name()` + pid, and is `remove_all`d up-front so a leftover from an aborted run cannot leak into a new case.

src/ros2_medkit_gateway/src/discovery/manifest/manifest_parser.cpp
bburda: Size cap added in 33b37f2d. `parse_fragment_file` now rejects any file larger than `ManifestParser::kMaxFragmentBytes` (1 MiB) via `std::filesystem::file_size` before opening, so yaml-cpp never sees oversized input. Test FragmentOverSizeLimitIsRejected in d08de44e locks it in. yaml-cpp's internal billion-laughs handling is a library-side concern tracked separately - the size cap at least bounds total memory even in that pathological case.

src/ros2_medkit_gateway/design/plugin_entity_notifications.rst
bburda: Reworded in 4129e73f. The doc now says plugin source written against v6 compiles unchanged against v7 headers (source-compatible) and explicitly calls out that binary compatibility is not provided - strict equality against `plugin_api_version()` means a pre-compiled v6 .so is rejected and recompilation is required.

src/ros2_medkit_gateway/src/discovery/manifest/manifest_manager.cpp
bburda: Intended behaviour, now documented. Added an "All-or-nothing fragment contract" section in the design doc (4129e73f) explaining the rationale: a partial merge produces a client-visible entity tree that matches neither the old nor the new deployment, which is worse than keeping the previous manifest active. Plugins that want independent failure domains should use separate fragment directories. Locked in with OneBrokenFragmentFailsEntireLoad in d08de44e.

src/ros2_medkit_gateway/src/discovery/manifest/manifest_manager.cpp
bburda: Fixed in 2befc477. The directory scan now canonicalizes each entry path and rejects any path whose real location is not a descendant of `canonical(fragments_dir)`. Symlinks pointing inside the directory (the k8s ConfigMap case) keep working. Backed by SymlinkEscapingFragmentsDirIsSkipped in d08de44e.

src/ros2_medkit_gateway/src/discovery/manifest/manifest_manager.cpp
bburda: Covered by the same raw-YAML walk added in 2befc477: after checking forbidden keys, any unrecognised top-level key (anything outside apps / components / functions / manifest_version / the forbidden set) is logged with `RCLCPP_WARN` naming the bad key and the allowed alternatives. UnknownTopLevelKeyIsIgnoredWithWarning in d08de44e asserts the fragment merges clean and contributes zero entities.

src/ros2_medkit_gateway/test/test_manifest_fragments.cpp
bburda: Added in d08de44e as BaseAndFragmentDuplicateIdFailsValidation. Base declares `ecu-primary` as a component, fragment declares an app with the same id; load_manifest returns false and the validation result has errors, proving the existing validate_unique_ids pass runs over the merged manifest.

src/ros2_medkit_gateway/include/ros2_medkit_gateway/plugins/plugin_types.hpp
bburda: Reworded in 4129e73f to match the design doc change (source-compatible, not binary-compatible; strict equality loader check; recompilation required).

src/ros2_medkit_gateway/src/discovery/manifest/manifest_manager.cpp
bburda: Contract now documented. In plugin_context.hpp `notify_entities_changed` docstring and in the design doc "Plugin write contract" section (both 4129e73f), plugins that write into fragments_dir are required to write to `fragments_dir/.tmp-<id>.yaml`, fsync, then `rename()` atomically to `<id>.yaml`. The rename is the commit point visible to the next reload. Reader-side, the new 1 MiB size cap and canonical-path symlink scoping make partial / unbounded reads bounded.

mfaferek93 (approved):

PR #379, opened 2026-04-15
fix(updates): seed Pending state on register so /updates/{id}/status returns 200

### LINKED ISSUE #378: UpdateManager::register_update does not seed an initial Pending state, causing GET /updates/{id}/status to 404
## Summary

After a successful `POST /updates`, a follow-up `GET /updates/{id}/status` returns `404` with `vendor_code: x-medkit-update-not-found` until something else (e.g. `/updates/{id}/prepare`) runs. The backend registered the package, but `UpdateManager::register_update` never touched `states_`, so `get_status` can't find the id.

## Impact

Any client that decides what to do with an update based on a non-null status response ends up stuck. In particular, the web UI's `UpdatesDashboard` (and its `UpdateCard`) hides the Prepare / Execute / Delete action buttons while the status fetch returns 404 and shows "Status unavailable" instead. There is no path forward from the UI for a freshly registered update.

## Repro

```bash
curl -X POST http://localhost:8080/api/v1/updates \
  -H 'Content-Type: application/json' \
  -d '{"id":"demo","update_name":"demo","origins":["uptane-server"]}'
# 200 OK -> {"id":"demo"}

curl http://localhost:8080/api/v1/updates/demo/status
# 404 -> {"error_code":"vendor-error","vendor_code":"x-medkit-update-not-found",
#         "message":"No status available for package 'demo'"}
```

## Suggested fix

After a successful `backend_->register_update(...)`, seed `states_[id]` with `{phase: None, status: Pending}` so the first status query returns `{"status":"pending"}`. Idempotent - preserves any already-seeded state. Matches the pattern already used inside the prepare / execute flows.

bburda:
## Summary

Closes #378. After a successful `backend_->register_update(...)`, seed `states_[id]` with `phase=None` / `status=Pending` so the first `GET /updates/{id}/status` returns `200 {"status":"pending"}` instead of `404 vendor-error`. Clients that gate UI on a non-null status (web UI `UpdatesDashboard`) can now render Prepare / Execute / Delete buttons immediately after registering a new update.

Idempotent - preserves an already-seeded state if a concurrent caller got there first (matches the defensive pattern already used in `start_prepare` / `run_prepare`).

## Test plan

- [x] New unit test `UpdateManagerTest.StatusPendingRightAfterRegister` covers the happy path.
- [x] Existing `StatusNotFoundForUnknown` still covers the NotFound path for truly unregistered IDs.
- [x] Full `test_update_manager` suite passes locally (20/20).

### DIFF
--- a/docs/api/rest.rst
+++ b/docs/api/rest.rst
@@ -1143,6 +1143,10 @@ Without such a plugin, all endpoints return ``501 Not Implemented``.
 
    **Status values:** ``pending``, ``inProgress``, ``completed``, ``failed``
 
+   A successful ``POST /api/v1/updates`` seeds a ``pending`` status for the package,
+   so this endpoint returns ``200`` with ``{"status": "pending"}`` immediately after
+   registration, before any ``prepare`` or ``execute`` call.
+
    When ``status`` is ``failed``, an ``error`` object is included:
 
    .. code-block:: json
@@ -1155,7 +1159,7 @@ Without such a plugin, all endpoints return ``501 Not Implemented``.
         }
       }
 
-   - **404 Not Found:** No status available (package not found or no operation started)
+   - **404 Not Found:** Package is not registered
 
 Cyclic Subscriptions
 --------------------
--- a/src/ros2_medkit_gateway/src/updates/update_manager.cpp
+++ b/src/ros2_medkit_gateway/src/updates/update_manager.cpp
@@ -104,18 +104,17 @@ tl::expected<void, UpdateError> UpdateManager::register_update(const nlohmann::j
   // consumers (UpdatesDashboard) gate action buttons on the status field
   // and won't render "Prepare" when the status query fails, so a newly-
   // registered update would otherwise be stuck from the frontend's view.
-  // Preserve an already-seeded state if a concurrent caller raced us
-  // through the backend (shouldn't happen with current backends, but the
-  // defensive check keeps the method idempotent).
-  if (metadata.contains("id") && metadata["id"].is_string()) {
-    const auto id = metadata["id"].get<std::string>();
-    std::lock_guard<std::mutex> lock(mutex_);
-    auto & state_ptr = states_[id];
-    if (!state_ptr) {
-      state_ptr = std::make_unique<PackageState>();
-      state_ptr->phase = UpdatePhase::None;
-      state_ptr->status = UpdateStatusInfo{UpdateStatus::Pending, std::nullopt, std::nullopt, std::nullopt};
-    }
+  // Callers (update_handlers.cpp) validate id presence + non-empty string
+  // before invoking register_update, so metadata.at("id") is guaranteed to
+  // hold a string at this point. The !state_ptr check matches the
+  // idempotent-seed pattern used in start_prepare / run_prepare.
+  const auto id = metadata.at("id").get<std::string>();
+  std::lock_guard<std::mutex> lock(mutex_);
+  auto & state_ptr = states_[id];
+  if (!state_ptr) {
+    state_ptr = std::make_unique<PackageState>();
+    state_ptr->phase = UpdatePhase::None;
+    state_ptr->status = UpdateStatusInfo{UpdateStatus::Pending, std::nullopt, std::nullopt, std::nullopt};
   }
   return {};
 }
--- a/src/ros2_medkit_gateway/src/updates/update_manager.cpp
+++ b/src/ros2_medkit_gateway/src/updates/update_manager.cpp
@@ -99,6 +99,24 @@ tl::expected<void, UpdateError> UpdateManager::register_update(const nlohmann::j
         return tl::make_unexpected(UpdateError{UpdateErrorCode::Internal, err.message});
     }
   }
+  // Seed an initial Pending state so GET /updates/<id>/status returns
+  // {"status":"pending"} immediately after register, rather than 404. UI
+  // consumers (UpdatesDashboard) gate action buttons on the status field
+  // and won't render "Prepare" when the status query fails, so a newly-
+  // registered update would otherwise be stuck from the frontend's view.
+  // Preserve an already-seeded state if a concurrent caller raced us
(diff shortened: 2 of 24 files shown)


src/ros2_medkit_gateway/src/updates/update_manager.cpp
```diff
@@ -99,6 +99,24 @@ tl::expected<void, UpdateError> UpdateManager::register_update(const nlohmann::j
...
+  if (metadata.contains("id") && metadata["id"].is_string()) {
+    const auto id = metadata["id"].get<std::string>();
+    std::lock_guard<std::mutex> lock(mutex_);
+    auto & state_ptr = states_[id];
+    if (!state_ptr) {
+      state_ptr = std::make_unique<PackageState>();
+      state_ptr->phase = UpdatePhase::None;
+      state_ptr->status = UpdateStatusInfo{UpdateStatus::Pending, std::nullopt, std::nullopt, std::nullopt};
+    }
+  }
```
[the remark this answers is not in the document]
bburda: Updated `docs/api/rest.rst` to describe the new contract: successful `POST /api/v1/updates` seeds a `pending` status, so `GET /updates/{id}/status` returns `200 {"status":"pending"}` immediately after registration. 404 is reserved for packages that are not registered.

src/ros2_medkit_gateway/src/updates/update_manager.cpp
```diff
@@ -99,6 +99,24 @@ tl::expected<void, UpdateError> UpdateManager::register_update(const nlohmann::j
...
+  // defensive check keeps the method idempotent).
+  if (metadata.contains("id") && metadata["id"].is_string()) {
+    const auto id = metadata["id"].get<std::string>();
+    std::lock_guard<std::mutex> lock(mutex_);
+    auto & state_ptr = states_[id];
+    if (!state_ptr) {
+      state_ptr = std::make_unique<PackageState>();
+      state_ptr->phase = UpdatePhase::None;
+      state_ptr->status = UpdateStatusInfo{UpdateStatus::Pending, std::nullopt, std::nullopt, std::nullopt};
+    }
```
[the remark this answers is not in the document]
bburda: Added `TestUpdatesPrepareExecute.test_00_status_pending_right_after_register` in `src/ros2_medkit_integration_tests/test/features/test_updates.test.py`. It registers a package and immediately fetches `/updates/{id}/status`, asserting 200 with `status=pending` and no progress/error keys. Existing `test_06_status_not_found_for_unknown` continues to cover the unknown-ID 404 path.

docs/api/rest.rst
```diff
@@ -1155,7 +1159,7 @@ Without such a plugin, all endpoints return ``501 Not Implemented``.
         }
       }
 
-   - **404 Not Found:** No status available (package not found or no operation started)
+   - **404 Not Found:** Package is not registered
```
mfaferek93, line 1162: **Minor - API contract change worth noting in CHANGELOG.** Previously 404 meant "package not found OR no operation started" (registered-but-idle was 404). Now registered-but-idle returns `200 pending`. Any downstream client (MCP `sovd_check_update_status`, Web UI `UpdatesDashboard`) that used 404 as a signal for "registered but nothing started yet" will see different behavior.

This IS the intended fix - just worth a one-liner in CHANGELOG.rst so consumers know the semantics shifted:
> `GET /updates/{id}/status` no longer returns 404 for registered packages; returns `200 {"status": "pending"}` instead.

src/ros2_medkit_gateway/src/updates/update_manager.cpp
```diff
@@ -99,6 +99,24 @@ tl::expected<void, UpdateError> UpdateManager::register_update(const nlohmann::j
...
     }
   }
+  // Seed an initial Pending state so GET /updates/<id>/status returns
+  // {"status":"pending"} immediately after register, rather than 404. UI
+  // consumers (UpdatesDashboard) gate action buttons on the status field
+  // and won't render "Prepare" when the status query fails, so a newly-
+  // registered update would otherwise be stuck from the frontend's view.
+  // Preserve an already-seeded state if a concurrent caller raced us
+  // through the backend (shouldn't happen with current backends, but the
+  // defensive check keeps the method idempotent).
```
mfaferek93: **Minor - guard is double-defensive (round-2 verified).** `update_handlers.cpp:108` already validates `id` is present, non-empty string before calling `register_update()`. This guard can never fail with current callers. Only fires if a future code path bypasses the HTTP handler layer.

Fine as defensive code. Consider adding a `RCLCPP_WARN` on the `else` branch so a future non-handler caller gets a breadcrumb when seeding silently skips:
```cpp
} else {
  RCLCPP_WARN(logger_, "register_update succeeded but metadata has no 'id' - status not seeded");
}
```

docs/api/rest.rst
bburda: Added a `**Fixes:**` entry under Unreleased in `src/ros2_medkit_gateway/CHANGELOG.rst` documenting the semantic shift (404 for registered-but-idle -> 200 pending, 404 reserved for unregistered) so downstream consumers like MCP `sovd_check_update_status` and the Web UI `UpdatesDashboard` have a clear heads-up.

src/ros2_medkit_gateway/src/updates/update_manager.cpp
bburda: Good observation. Rather than log-guarding a branch that can't fire, dropped the outer `if (metadata.contains("id") && metadata["id"].is_string())` entirely and replaced it with `metadata.at("id").get<std::string>()`. The handler (`update_handlers.cpp:108`) already enforces the precondition, so expressing it via `.at()` (which would throw clearly if the contract were ever broken) is cleaner than defensive validation for an unreachable path. Also avoids plumbing a logger into `UpdateManager` just for a WARN that can't fire. The inner `!state_ptr` idempotency check is kept (matches `start_prepare`/`run_prepare`).

mfaferek93 (dismissed):

mfaferek93 (approved):

PR #384, opened 2026-04-21
feat(data): pool-backed TopicDataProvider architecture

### LINKED ISSUE #375: Rewrite NativeTopicSampler to subscription-pool design
## Context

``NativeTopicSampler`` (``src/ros2_medkit_gateway/src/native_topic_sampler.cpp``) provides the data-sampling backend for ``GET /components/{id}/data/{topic}`` and ``GET /apps/{id}/data/{topic}``. It was added as a fast replacement for shelling out to ``ros2 topic echo``. The current implementation creates a fresh ``GenericSubscription`` on every sample request, serialized through a single global mutex.

## Observed pain points

- **Per-request subscription lifecycle.** Every call to ``sample_topic`` does ``create_generic_subscription`` + destructor. This touches rmw internals that are not safe against concurrent callers on the same node, and the current ``sampling_mutex_`` papers over that rather than removing the stress. Each cycle is a candidate for the intermittent Rolling SIGSEGV we see on PR #374 CI: ``test_component_topic_rpm`` crashed the gateway with exit code ``-11`` inside ``sample_topic``.
- **``sample_topics_parallel`` is not parallel.** It spawns ``std::async`` tasks but each task re-acquires ``sampling_mutex_``. Latency for N topics is N × timeout, not max(timeout).
- **Redundant graph queries per sample.** ``sample_topic`` calls ``get_topic_info``, ``get_topic_publishers``, ``get_topic_subscribers``, ``get_topic_type`` - 3-4 round-trips into rmw just for metadata.
- **httplib thread blocking.** The handler thread holds ``sampling_mutex_`` for up to ``timeout_sec`` (default 1 s). Under concurrent load this serialises behind a single mutex and can starve the httplib thread pool - the aggregation design doc already warns about this pattern.
- **No dedup / back-pressure.** Ten concurrent requests for the same topic create ten separate subscriptions in sequence, not one subscription serving ten callers.
- **Existing kludges.** ``sampling_mutex_``, the dedicated ``sampling_cb_group_``, and the heap-allocated ``SampleState`` all exist to dodge races that a better design would simply not create.

## Proposed design (subscription pool)

Keep the public ``NativeTopicSampler`` interface intact; rewrite internals.

1. **Long-lived subscription pool.**
   - One ``GenericSubscription`` per actively-sampled topic, created lazily on first request.
   - Subscriptions live on a dedicated sampler thread (or a bounded TaskQueue). Subscriptions are never created or destroyed on an httplib handler thread.
   - TTL-based eviction: drop subscription after a configurable idle window with no sample requests.
   - Re-create subscription if the message type changes for a topic.
2. **Per-topic ring buffer.** Callback pushes the latest serialised message into a bounded ``std::deque<SerializedMessage>`` (depth 1 is sufficient for current UX; tunable). Callback is the only writer; a condition variable wakes waiters.
3. **``sample_topic`` becomes non-blocking for hot topics.**
   - If the ring buffer has a message: serialise and return immediately.
   - If the ring buffer is empty: wait on condvar up to ``timeout_sec``. Multiple concurrent callers share the wait - one arriving message wakes all of them.
4. **``sample_topics_parallel`` becomes truly parallel.** Just ring-buffer reads; no global mutex.
5. **Metadata cache.** Cache topic type and QoS by topic name, invalidated on graph change events if available, otherwise TTL.

## Benefits

- Removes the entire "per-request create_generic_subscription" class of races - likely fixes the Rolling ``test_data_read`` SIGSEGV and any future rmw-timing regressions on that path.
- Unblocks httplib threads; frees the gateway under concurrent data-reads.
- True parallel sampling for multi-topic requests.
- Lower jitter: no subscription startup cost once a topic is warm.
- Observability win: pool state (active subscriptions, eviction count, message rates) is easy to expose on ``/health`` as an x-medkit extension.

## Costs / risks

- Additional memory: N active topics × message_size × buffer_depth.
- Needs new tests: soak test around create/evict cycles, thread-stress test with many concurrent samplers on the same topic, type-change regression.
- Has to preserve exact existing behaviour for the "no publishers" fast path (metadata-only response) so ``test_app_no_topics`` and similar keep passing.

## Out of scope

- Changing the REST contract (``GET /components/{id}/data/{topic}``).
- Switching off of ``dynmsg``-based deserialisation (``ros2_medkit_serialization::JsonSerializer`` stays).
- Long-term subscription persistence across gateway restart.

## Acceptance criteria

- [ ] ``NativeTopicSampler`` public header unchanged (same methods, same signatures, same return types).
- [ ] Existing unit tests (``test_native_topic_sampler``) pass without modification.
- [ ] New unit tests cover pool lifecycle (create, evict, re-create on type change, concurrent first-request race).
- [ ] ``test_data_read`` is stable across 50 consecutive runs on Jazzy, Humble, and Rolling.
- [ ] Benchmark shows p50 / p99 latency for hot-topic ``GET /components/{id}/data/{topic}`` improved vs baseline, and concurrent-topic parallelism verified.
- [ ] Design doc (either extending ``src/ros2_medkit_gateway/design/`` or a new page) documents the pool, TTL, and sampler thread ownership.

## Related

- Issue #344 - similar SIGSEGV pattern traced to a subscription-destructor race.
- PR #374 - Rolling CI failure that motivated this proposal (``test_component_topic_rpm``, gateway exit code -11).

bburda:
## Summary

- Fix the rcl hash-map race behind the sample-time SIGSEGV in `test_data_read` (issue #375) by serialising every subscription / callback-group creation on a single worker thread (`Ros2SubscriptionExecutor`) and routing subscription lifetime through a RAII handle (`Ros2SubscriptionSlot`).
- Introduce a transport-neutral `TopicDataProvider` interface and a pool-backed default implementation (`Ros2TopicDataProvider`) that subscribes once per topic and serves many samples from the cached latest message, with LRU + idle-safety-net + graph-change eviction, publisher-matching QoS, and a cold-wait cap that preserves cpp-httplib liveness under sample load.
- Delete the legacy `NativeTopicSampler` entirely (class + tests + all production fallbacks) so the new pool is the single sampling path. Surface the new machinery via `x-medkit-*` stats on `GET /health`, add a drill-in endpoint `GET /health/subscription-pool`, document the architecture under `design/ros2_subscription_architecture.rst`, and gate regressions in CI with `scripts/check_no_naked_subscriptions.sh`.

---

## Issue

- closes #375

---

## Type

- [ ] Bug fix
- [x] New feature or tests
- [ ] Breaking change
- [ ] Documentation only

Primarily a new architecture that also fixes a latent crash. The SOVD REST contract is unchanged; changes are entirely below the handler layer.

---

## Testing

- Unit tests: 2110 / 2110 green on Jazzy. New tests cover the pool (`test_ros2_topic_data_provider.cpp`, 19 cases) including publisher-QoS matching, LRU promote/evict, idle sweep, cold-wait cap, timeout-without-publisher resilience, rapid-timeout with active publisher, and a concurrent-sample regression test (`ConcurrentSampleFromMultipleThreadsDoesNotCrash`) that reliably reproduced the original race before the fix.
- Executor tests cover `run_sync` deadlines, queue backpressure, shutdown-drain, watchdog, graph-change delivery.
- Integration: existing data/discovery/health integration tests continue to pass against the new provider wiring.
- Local Jazzy build verified with pre-commit + pre-push hooks. The new grep gate blocks any `create_generic_subscription` / `create_subscription<` / `create_callback_group` outside `ros2_common/` and `test/`. `NativeTopicSampler` grep returns zero matches in production code.

---

## Checklist

Follow-ups (outside this PR's scope):
- Migrate remaining legacy subscriber call-sites listed in `check_no_naked_subscriptions.sh` (fault subscribers, operation manager, log manager, trigger_topic_subscriber) onto `Ros2SubscriptionSlot`.


src/ros2_medkit_gateway/src/ros2_common/ros2_subscription_slot.cpp
```diff
@@ -0,0 +1,81 @@
...
+    sub_.reset();
+    return;
+  }
+
+  // Normal path: post destroy to the worker so we do not race with
+  // concurrent rcl mutations. Bounded deadline to avoid hanging teardown.
+  auto result = exec_.run_sync<void>(std::function<void()>([this] {
+                                       sub_.reset();
+                                     }),
+                                     kDestroyDeadline);
```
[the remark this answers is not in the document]
bburda: Fixed in 3d3db541. The destroy task now moves the subscription out of the slot into its own capture list, so the handle's lifetime is decoupled from this surviving run_sync's deadline. A shutdown-flag fast path drops the sub synchronously when the executor is already winding down, avoiding the queued-task-after-free window.

src/ros2_medkit_gateway/src/http/handlers/data_handlers.cpp
```diff
@@ -198,10 +199,21 @@ void DataHandlers::handle_get_data_item(const httplib::Request & req, httplib::R
...
-    auto sample = native_sampler->sample_topic(full_topic_path, data_access_mgr->get_topic_sample_timeout());
+    TopicSampleResult sample;
+    const auto timeout_sec = data_access_mgr->get_topic_sample_timeout();
+    if (auto * provider = data_access_mgr->get_topic_data_provider()) {
+      const auto timeout_ms = std::chrono::milliseconds{static_cast<std::int64_t>(std::max(timeout_sec, 0.0) * 1000.0)};
+      auto r = provider->sample(full_topic_path, timeout_ms);
+      if (r) {
+        sample = *r;
+      }
+    } else if (auto * native_sampler = data_access_mgr->get_native_sampler()) {
```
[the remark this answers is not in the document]
bburda: Fixed in b02c0259. The handler now inspects the tl::expected return from provider->sample() and forwards the error through ctx_.send_error with the provider's code, http_status, and params instead of falling through to the metadata-only response.

src/ros2_medkit_gateway/src/data_access_manager.cpp
```diff
@@ -178,7 +179,26 @@ json DataAccessManager::sample_result_to_json(const TopicSampleResult & sample)
...
+  const auto timeout_ms = std::chrono::milliseconds{static_cast<std::int64_t>(std::max(timeout_sec, 0.0) * 1000.0)};
+
+  TopicSampleResult sample;
+  if (topic_data_provider_) {
+    auto r = topic_data_provider_->sample(topic_name, timeout_ms);
+    if (!r) {
+      RCLCPP_WARN(node_->get_logger(), "TopicDataProvider::sample('%s') failed: %s [%d]", topic_name.c_str(),
+                  r.error().message.c_str(), r.error().http_status);
+      throw TopicNotAvailableException(topic_name);
+    }
```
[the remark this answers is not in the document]
bburda: Fixed in 3d3db541. DataAccessManager now preserves err.http_status: 404 keeps the original TopicNotAvailableException mapping, every other status raises the new ProviderErrorException which handlers translate back to the original code + HTTP status instead of collapsing to 404.

src/ros2_medkit_gateway/src/data/ros2_topic_data_provider.cpp
```diff
@@ -0,0 +1,769 @@
...
+      if (r) {
+        return *r;
+      }
+      TopicSampleResult fallback;
+      fallback.topic_name = t;
+      fallback.has_data = false;
+      fallback.timestamp_ns = now_ns_epoch();
+      return fallback;
+    }));
+  }
```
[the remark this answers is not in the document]
bburda: Fixed in ab558e0d. sample_parallel now chunks at data_provider.max_parallel_samples (default 8, configurable in gateway_params.yaml, range 1-256) instead of fanning out one std::async per topic.

src/ros2_medkit_gateway/src/data/ros2_topic_data_provider.cpp
```diff
@@ -0,0 +1,769 @@
...
+  if (!exec_->node()) {
+    throw std::invalid_argument{"Ros2SubscriptionExecutor has no subscription node"};
+  }
+  // Serializer is optional for discovery-only use; sampling path will check.
+
+  graph_listener_token_ = exec_->on_graph_change([this] {
+    on_graph_change();
+  });
+
+  if (cfg_.idle_sweep_tick.count() > 0 && cfg_.idle_safety_net.count() > 0) {
```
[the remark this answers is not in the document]
bburda: Fixed in 3d3db541. The graph callback captures a shared alive flag; ~Ros2TopicDataProvider clears that flag before remove_graph_change(). Already-queued callbacks check the flag and no-op instead of dereferencing a dangling this.

src/ros2_medkit_gateway/src/data/ros2_topic_data_provider.cpp
```diff
@@ -0,0 +1,769 @@
...
+    result.has_data = false;
+  }
+  return result;
+}
+
+tl::expected<std::vector<TopicSampleResult>, ErrorInfo>
+Ros2TopicDataProvider::sample_parallel(const std::vector<std::string> & topics, std::chrono::milliseconds timeout) {
+  if (shutdown_.load(std::memory_order_acquire)) {
+    return tl::unexpected(ErrorInfo{"ERR_GATEWAY_SHUTDOWN", "gateway shutting down", 503, nlohmann::json::object()});
+  }
```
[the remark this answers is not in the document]
bburda: Fixed in 3d3db541. Added the vendor code ERR_X_MEDKIT_GATEWAY_SHUTDOWN (value x-medkit-gateway-shutdown) in error_codes.hpp and switched both shutdown return sites in ros2_topic_data_provider.cpp to use it.

src/ros2_medkit_gateway/src/http/rest_server.cpp
```diff
@@ -323,6 +323,18 @@ void RESTServer::setup_routes() {
...
+          [this](auto & req, auto & res) {
+            health_handlers_->handle_subscription_pool(req, res);
+          })
+      .tag("Server")
+      .summary("Subscription pool drill-in")
+      .description(
+          "Per-topic snapshot of the ROS 2 subscription pool. x-medkit vendor extension; "
+          "intended for diagnostics, not for tight polling.")
+      .response(200, "Pool snapshot")
+      .operation_id("getSubscriptionPool");
```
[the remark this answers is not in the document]
bburda: The GET /health/subscription-pool endpoint was removed in 38c1e2eb. Subscription pool stats now ride on the main /health payload under the x-medkit-data-provider and x-medkit-subscription-executor namespaces, so there is no separate endpoint to document.

src/ros2_medkit_gateway/src/http/handlers/health_handlers.cpp
```diff
@@ -112,6 +126,27 @@ void HealthHandlers::handle_health(const httplib::Request & req, httplib::Respon
...
+      if (auto * tdp = ctx_.node()->get_topic_data_provider()) {
+        response["x-medkit-subscription-pool"] = tdp->x_medkit_pool_snapshot();
+        response["x-medkit-data-provider"] = tdp->x_medkit_stats().value("x-medkit-data-provider", json::object());
+        response["x-medkit-subscription-executor"] =
+            tdp->x_medkit_stats().value("x-medkit-subscription-executor", json::object());
+      } else {
+        response["x-medkit-subscription-pool"] = json::array();
+      }
+    }
+    HandlerContext::send_json(res, response);
```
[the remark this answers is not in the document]
bburda: handle_subscription_pool was removed alongside the endpoint in 38c1e2eb. The stats are now emitted by the existing /health handler, which already constructs its JSON body as an object.

src/ros2_medkit_gateway/include/ros2_medkit_gateway/models/error_info.hpp
```diff
@@ -0,0 +1,47 @@
...
+ *   Must be serializable and must not contain sensitive information.
+ *
+ * @par Thread safety
+ * Plain value type; safe to copy and move across threads without synchronization.
+ */
+struct ErrorInfo {
+  std::string code;                                  ///< SOVD error code (e.g. ERR_TOPIC_NOT_FOUND)
+  std::string message;                               ///< Human-readable message (truncated to 512 chars by handlers)
+  int http_status{500};                              ///< HTTP status in SOVD range 400-599
+  nlohmann::json params = nlohmann::json::object();  ///< Optional structured parameters
```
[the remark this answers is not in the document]
bburda: Fixed in 3d3db541. The comment now references ERR_RESOURCE_NOT_FOUND and ERR_X_MEDKIT_ROS2_TOPIC_UNAVAILABLE, both of which exist in error_codes.hpp.

src/ros2_medkit_gateway/include/ros2_medkit_gateway/ros2_common/ros2_subscription_executor.hpp
```diff
@@ -0,0 +1,290 @@
...
+   * @brief Construct and start the worker thread.
+   *
+   * @param gateway_node Owning gateway node. Used only to derive the subscription node
+   *                     name and namespace; no references retained after construction.
+   * @param main_executor Executor that will spin both the gateway node and the
+   *                      newly-created subscription node. Must outlive this executor.
+   * @param cfg           Bounded resource configuration.
+   */
+  Ros2SubscriptionExecutor(const std::shared_ptr<rclcpp::Node> & gateway_node, rclcpp::Executor & main_executor,
+                           Config cfg = Config());
```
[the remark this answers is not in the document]
bburda: Fixed in ab558e0d by removing the unused main_executor parameter entirely. The ctor now matches the actual ownership model: the subscription node is owned by an internal SingleThreadedExecutor that is spun from the worker thread via spin_some(), and the main gateway executor never sees it. Doc block updated to describe that contract.

src/ros2_medkit_gateway/include/ros2_medkit_gateway/discovery/runtime_discovery.hpp
```diff
@@ -175,7 +175,7 @@ class RuntimeDiscoveryStrategy : public DiscoveryStrategy {
 
   rclcpp::Node * node_;
   RuntimeConfig config_;
-  NativeTopicSampler * topic_sampler_{nullptr};
+  TopicDataProvider * topic_sampler_{nullptr};
```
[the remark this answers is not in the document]
bburda: Renamed to topic_data_provider_ in ab558e0d (both header and runtime_discovery.cpp).

src/ros2_medkit_gateway/src/data/ros2_topic_data_provider.cpp
```diff
@@ -0,0 +1,807 @@
...
+      if (r) {
+        return *r;
+      }
+      TopicSampleResult fallback;
+      fallback.topic_name = t;
+      fallback.has_data = false;
+      fallback.timestamp_ns = now_ns_epoch();
+      return fallback;
+    }));
+  }
```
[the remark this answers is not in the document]
bburda: Fixed in ab558e0d. sample_parallel now chunks topics at data_provider.max_parallel_samples (default 8, configurable in gateway_params.yaml, range 1-256) instead of launching one std::async per topic.

src/ros2_medkit_gateway/src/main.cpp
```diff
@@ -12,24 +12,96 @@
...
 
-  // MultiThreadedExecutor is required because cpp-httplib handler threads call
-  // Node::create_generic_subscription() for topic sampling. SingleThreadedExecutor
-  // does not synchronize external subscription creation with its internal iteration,
-  // causing non-deterministic SIGSEGV on rolling. All gateway callbacks are already
-  // protected by mutexes (EntityCache, LogManager, TriggerManager, etc.).
+  // MultiThreadedExecutor is required: the Ros2SubscriptionExecutor's
+  // subscription_node lives on the same executor as the gateway node, and
+  // parallel dispatch prevents the serial worker and the main executor from
+  // starving each other.
```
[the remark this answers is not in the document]
bburda: Fixed in ab558e0d. The misleading comment was rewritten to explain that the subscription node lives on an internal SingleThreadedExecutor owned by Ros2SubscriptionExecutor, not the main gateway executor. Also removed the dead main_executor ctor parameter it was describing.

scripts/check_no_naked_subscriptions.sh
```diff
@@ -0,0 +1,106 @@
...
+  violations=$(echo "${violations}" | grep -vE "${LEGACY_PATTERN}" || true)
+fi
+
+if [[ -n "${violations}" ]]; then
+  echo "FAIL: naked rclcpp subscription / callback-group creation outside ros2_common/:"
+  echo "${violations}"
+  echo
+  echo "Route these through ros2_medkit_gateway::ros2_common::Ros2SubscriptionSlot"
+  echo "so the rcl hash-map race fixed in issue #375 does not come back."
+  echo "See docs/design/ros2_subscription_architecture.rst for the allowed pattern."
```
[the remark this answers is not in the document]
bburda: Fixed in ab558e0d.

src/ros2_medkit_gateway/src/data/ros2_topic_data_provider.cpp
```diff
@@ -0,0 +1,807 @@
...
+      auto r = sample(t, timeout);
+      if (r) {
+        return *r;
+      }
+      TopicSampleResult fallback;
+      fallback.topic_name = t;
+      fallback.has_data = false;
+      fallback.timestamp_ns = now_ns_epoch();
+      return fallback;
+    }));
```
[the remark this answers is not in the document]
bburda: Fixed in ab558e0d. sample_parallel now returns tl::unexpected(err) on the first provider failure instead of converting it into a metadata-only TopicSampleResult. Missing topics still return has_data=false (not an error), matching sample()'s contract.

src/ros2_medkit_gateway/src/data/ros2_topic_data_provider.cpp
```diff
@@ -0,0 +1,807 @@
...
+    return std::nullopt;
+  }
+  TopicInfo info;
+  info.name = topic;
+  if (!it->second.empty()) {
+    info.type = it->second[0];
+  }
+  info.publisher_count = exec_->node()->count_publishers(topic);
+  info.subscriber_count = exec_->node()->count_subscribers(topic);
+  return info;
```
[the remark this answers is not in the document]
bburda: Fixed in ab558e0d. Every exec_->node() graph introspection call in the provider (get_topic_info, discover_all, has_publishers, get_topic_publishers/subscribers, get_topic_connection, build_component_topic_map, discover_topics_by_namespace, get_topics_for_namespace, qos_for, update_topic_type_if_needed) is now wrapped in try/catch(std::runtime_error), logged at DEBUG, and returns empty results so callers continue gracefully during shutdown.

src/ros2_medkit_gateway/src/data/ros2_topic_data_provider.cpp
```diff
@@ -0,0 +1,807 @@
...
+  s.pool_hits = pool_hits_.load();
+  s.pool_misses = pool_misses_.load();
+  s.metadata_cache_size = 0;
+  s.metadata_cache_cap = cfg_.metadata_cache_cap;
+  s.evictions_total = evictions_total_.load();
+  s.type_change_events = type_change_events_.load();
+  s.graph_events_received = graph_events_received_.load();
+  s.concurrent_cold_waits = concurrent_cold_waits_.load();
+  s.cold_wait_cap = cfg_.cold_wait_cap;
+  return s;
```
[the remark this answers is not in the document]
bburda: Fixed in ab558e0d by dropping the unimplemented cache entirely: removed metadata_cache_cap from Config, metadata_cache_size/cap from PoolStats, the x_medkit_stats keys, the ROS 2 parameter declaration in main.cpp, and the gateway_params.yaml block. If a metadata cache is added later we will reintroduce the config/stats alongside the implementation.

src/ros2_medkit_gateway/src/main.cpp
```diff
@@ -12,24 +12,96 @@
...
+  // (rolling; recent jazzy patch releases) asserts 'node needs to be
+  // associated with an executor' and aborts with exit -6 when the managers'
+  // shutdown paths touch service clients. Explicit teardown avoids that:
+  //   1. drop the provider (clears pool entries via the subscription worker)
+  //   2. reset sub_exec (joins worker, removes subscription_node from executor)
+  //   3. remove the gateway node from the executor and drop our ref so
+  //      ~GatewayNode runs with the executor still alive.
+  data_provider.reset();
+  sub_exec.reset();
+  executor.remove_node(node);
```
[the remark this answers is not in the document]
bburda: Fixed in ab558e0d. main() now calls node->set_topic_data_provider(nullptr) before the local shared_ptr resets, so GatewayNode drops its own shared_ptr first and data_provider.reset() actually tears down the provider. Teardown comment updated to reflect the new step.

src/ros2_medkit_gateway/src/ros2_common/ros2_subscription_executor.cpp
```diff
@@ -0,0 +1,302 @@
...
+    degraded_.store(false, std::memory_order_release);
+    return;
+  }
+  const std::uint64_t now_ns = steady_now_ns();
+  const std::uint64_t age_ms = (now_ns > started_ns) ? (now_ns - started_ns) / 1'000'000U : 0U;
+  if (age_ms > static_cast<std::uint64_t>(cfg_.watchdog_threshold.count())) {
+    if (!degraded_.exchange(true, std::memory_order_acq_rel)) {
+      watchdog_trips_.fetch_add(1, std::memory_order_relaxed);
+      RCLCPP_ERROR(rclcpp::get_logger("ros2_subscription_executor"),
+                   "Watchdog tripped: task age=%lums threshold=%lldms", static_cast<unsigned long>(age_ms),
```
mfaferek93: **Major.** Watchdog is **observational only** - on trip it sets `degraded_=true` + increments counter + logs once. Stuck task continues to hold the worker. No timeout enforcement, no thread interrupt, no exec recreation. Plus: `degraded_` flaps to `false` (line 261) on the same tick when the threshold is briefly not exceeded - even when the task is still genuinely stuck.

Fix: either rename to `stuck_task_detector` and document as observational-only, or add an action (forced-kill flag callers can poll, or recreate the worker on threshold breach).

src/ros2_medkit_gateway/src/ros2_common/ros2_subscription_executor.cpp
```diff
@@ -0,0 +1,302 @@
...
+    for (std::size_t i = 0; i < kMaxGraphListeners; ++i) {
+      if (graph_slot_used_[i] && graph_callbacks_[i]) {
+        snapshot[count++] = graph_callbacks_[i];
+      }
+    }
+  }
+  for (std::size_t i = 0; i < count; ++i) {
+    // If the queue is full we accept losing this graph fan-out for some
+    // listeners; the idle-safety-net eviction in upstream providers compensates.
+    (void)post([cb = std::move(snapshot[i])] {
```
mfaferek93: **Major.** Graph callback fan-out **drops callbacks under backpressure** with no recovery. Each registered callback is `post()`ed individually; if queue fills mid-fan-out, half listeners get the event, half don't. Single graph event, partial delivery, permanent miss until next graph change. Comment at line 287 admits this and relies on "idle-safety-net eviction in upstream providers" which can take 15 min by default to recover.

Fix: post one wrapper that fires the entire snapshot under `graph_mtx_` so backpressure is all-or-nothing.

src/ros2_medkit_gateway/src/ros2_common/ros2_subscription_executor.cpp
```diff
@@ -0,0 +1,302 @@
...
+rclcpp::Node * Ros2SubscriptionExecutor::node() const noexcept {
+  return subscription_node_.get();
+}
+
+std::size_t Ros2SubscriptionExecutor::on_graph_change(GraphCallback cb) {
+  std::lock_guard<std::mutex> lk(graph_mtx_);
+  for (std::size_t i = 0; i < kMaxGraphListeners; ++i) {
+    if (!graph_slot_used_[i]) {
+      graph_callbacks_[i] = std::move(cb);
+      graph_slot_used_[i] = true;
```
mfaferek93, line 127: **Major.** Lock-ordering hazard on `graph_mtx_`: `on_graph_change()` and `remove_graph_change()` both take this mutex, and the API allows a registrant to call `remove_graph_change()` from inside their own callback (running on the worker, post-fan-out) → recursive non-recursive mutex → deadlock. The fan-out itself is fine because it snapshots before firing, but nothing prevents the registrant from doing the wrong thing.

Fix: document "do not call `remove_graph_change` from your callback" in the header, or copy the slot before invoking.

mfaferek93 (commented):
Great change, I added a few comments :)

src/ros2_medkit_gateway/src/ros2_common/ros2_subscription_executor.cpp
```diff
@@ -0,0 +1,302 @@
...
+void Ros2SubscriptionExecutor::aux_loop() {
+  // Independent thread driving watchdog + graph-event polling so they keep
+  // running even when the worker thread is blocked inside a long-running
+  // run_sync task. Both tick functions only touch this object's atomics and
+  // the shared graph_event_; neither creates, destroys, or dispatches
+  // callbacks on subscription_node_, so concurrency with the worker is safe.
+  const auto tick = std::min(cfg_.watchdog_tick, cfg_.graph_poll_tick);
+  auto next_watchdog = std::chrono::steady_clock::now() + cfg_.watchdog_tick;
+  auto next_graph = std::chrono::steady_clock::now() + cfg_.graph_poll_tick;
+  while (!shutdown_flag_->load(std::memory_order_acquire)) {
```
mfaferek93, line 238: **Major.** `aux_loop` polls every 100 ms (`graph_poll_tick`) regardless of `watchdog_threshold` - thread wakes 10×/sec for a watchdog that may fire every 1000 ms. Plus `sleep_for(tick)` is not shutdown-aware, so shutdown latency is bounded by `graph_poll_tick=100ms`.

Fix: switch to `std::condition_variable::wait_until(min(next_watchdog, next_graph))` with shutdown predicate. Cleaner shutdown + lower idle CPU on quiet gateways.

src/ros2_medkit_gateway/include/ros2_medkit_gateway/ros2_common/ros2_subscription_executor.hpp
```diff
@@ -0,0 +1,293 @@
...
+   * is at capacity.
+   */
+  bool post(std::function<void()> task);
+
+  /**
+   * @brief Accessor for the subscription node. Use only from within a task posted to
+   *        this executor; concurrent rcl mutations on other threads violate the
+   *        single-writer invariant.
+   */
+  rclcpp::Node * node() const noexcept;
```
mfaferek93, line 183: **Major.** `node()` is documented as worker-only but exposed as a free `const noexcept` getter returning raw `rclcpp::Node *`. Nothing prevents new code from doing `exec.node()->create_subscription(...)` straight on an HTTP thread - the exact bypass the executor exists to eliminate. The CI gate (`check_no_naked_subscriptions.sh`) catches `create_*` patterns but not arbitrary node-pointer operations.

Fix: rename to `node_unsafe_for_external_use()`, or return a guard token that asserts `std::this_thread::get_id() == worker_id_` for any rcl-mutating op.

src/ros2_medkit_gateway/src/ros2_common/ros2_subscription_slot.cpp
```diff
@@ -0,0 +1,107 @@
...
+    return;
+  }
+
+  // Normal path: post destroy to the worker so we do not race with
+  // concurrent rcl mutations. Bounded deadline to avoid hanging teardown.
+  // The wrapper owns sub_to_drop by move; even if run_sync's deadline fires
+  // before the worker drains the task, the destructor returns safely while
+  // the queued task still holds the last reference and releases it when it
+  // finally runs (normal path or shutdown queue drain).
+  auto result = exec_->run_sync<void>(std::function<void()>([sub = std::move(sub_to_drop)]() mutable {
```
mfaferek93: **Major (×2 issues at this site).**

**(1) Recursive deadlock when slot dtor runs on the worker thread.** This `run_sync` posts to the worker and waits up to `kDestroyDeadline=30s`. If a callback running ON THE WORKER drops its own slot (e.g. graph callback evicts a pool entry whose slot it owns), the worker is the one waiting → 30s hang → deadline expires → leaked rcl handle deferred to executor shutdown drain. The graph callback path in `Ros2TopicDataProvider::on_graph_change` is the realistic trigger.

*Fix:* check `std::this_thread::get_id() == exec_->worker_thread_id()` and run `sub_to_drop.reset()` inline in that case; or assert/document "a callback must not drop its own slot."

**(2) Destructor can throw.** `run_sync` builds a `std::function` + promise/future + `tl::expected` move - any of those can throw on OOM and escape this non-`noexcept` dtor. If another exception is in flight at the time, `std::terminate`.

*Fix:* wrap the body in `try { ... } catch (...) { /* log, swallow */ }` and mark `~Ros2SubscriptionSlot() noexcept;`.

src/ros2_medkit_gateway/src/data/ros2_topic_data_provider.cpp
```diff
@@ -0,0 +1,894 @@
...
+    }
+  }
+  for (auto & entry : to_drop) {
+    entry->shutdown.store(true, std::memory_order_release);
+    std::lock_guard<std::mutex> bl(entry->buf_mtx);
+    entry->buf_cv.notify_all();
+  }
+}
+
+TopicSampleResult Ros2TopicDataProvider::build_metadata_only_sample(const std::string & topic) {
```
mfaferek93: **Major.** **Dead code.** `build_metadata_only_sample` is declared in the header (`ros2_topic_data_provider.hpp:157`) and defined here, but **zero callers** in the entire repo. The header comment for `topic_data_provider.hpp:69-72` describes a metadata-only fallback path that this helper used to serve, but `sample()` now returns 503 (`ERR_X_MEDKIT_COLD_WAIT_CAP_EXCEEDED`) on cap exhaustion instead.

This matches the doc/code drift in `design/ros2_subscription_architecture.rst:149-150` ("return metadata-only immediately") - the design intent moved on, the code did not.

Fix: delete the helper (and the doc claim), or re-wire it as the cold-wait-cap fallback if metadata-only is still the desired behavior. `-Wunused-private-method` will eventually fire.

src/ros2_medkit_gateway/src/data/ros2_topic_data_provider.cpp
```diff
@@ -0,0 +1,894 @@
...
+  const auto threshold = cfg_.idle_safety_net;
+
+  std::vector<std::shared_ptr<PoolEntry>> to_drop;
+  {
+    std::lock_guard<std::mutex> lk(pool_mtx_);
+    for (auto it = pool_.begin(); it != pool_.end();) {
+      const std::string topic = it->first;
+      std::chrono::steady_clock::time_point last_sample;
+      {
+        std::lock_guard<std::mutex> bl(it->second->buf_mtx);
```
mfaferek93, line 641: **Major.** Lock-ordering inconsistency. Idle sweep takes `pool_mtx_` and then `buf_mtx_` together (this site). `sample()` takes `pool_mtx_` first, drops it, then takes `buf_mtx_` (lines 271-281, then 351-352). Currently safe because `sample()` does not reacquire `pool_mtx_` while holding `buf_mtx_` - but **fragile**: a future edit that reads pool state under `buf_mtx_` deadlocks against this sweep.

Fix: document the lock order at the top of the file (`pool_mtx_` always before `buf_mtx_`, never reversed), or switch `last_sample_time` to `std::atomic<rep>` so the sweep does not need `buf_mtx_` at all.

src/ros2_medkit_gateway/src/data/ros2_topic_data_provider.cpp
```diff
@@ -0,0 +1,894 @@
...
+    for (std::size_t i = base; i < end; ++i) {
+      const auto & t = topics[i];
+      futures.push_back(std::async(std::launch::async, [this, t, timeout] {
+        return sample(t, timeout);
+      }));
+    }
+    for (auto & f : futures) {
+      auto r = f.get();
+      if (!r) {
+        return tl::unexpected(r.error());
```
mfaferek93, line 551: **Major.** `sample_parallel` returns the first error and aborts the whole batch. Chunk of 8 topics, 4 cold + 4 hot, `cold_wait_cap=4` → 4 cold callers get 503, batch fails on the first error. Plus: `f.get()` joins sequentially but on early `return tl::unexpected(r.error())`, subsequent futures' `~future()` blocks until tasks complete - caller is forced to wait for all in-flight before getting the error.

This silently regresses `GET /components/{id}/data` under load (one cold-cap hit fails the entire response).

Fix: decide and document - should `COLD_WAIT_CAP_EXCEEDED` for one topic be treated as metadata-only-for-that-topic (partial success), or fail-batch? Either way, current behavior is a runtime trap.

src/ros2_medkit_gateway/src/data/ros2_topic_data_provider.cpp
```diff
@@ -0,0 +1,894 @@
...
+  }
+  for (const auto & pub : pubs) {
+    const auto & q = pub.qos_profile();
+    if (q.reliability() == rclcpp::ReliabilityPolicy::Reliable) {
+      any_reliable = true;
+    }
+    if (q.durability() == rclcpp::DurabilityPolicy::TransientLocal) {
+      any_transient_local = true;
+    }
+  }
```
mfaferek93, line 293: **Major.** `qos_for()` only matches reliability + transient_local. Publisher-side `Deadline`, `Lifespan`, and `Liveliness=ManualByTopic` are ignored. On Cyclone DDS this means the connection is refused (DDS enforces deadline contracts); on FastDDS the subscriber stays connected but never receives - manifesting as permanent cold-wait timeout with no obvious cause.

Fix: at minimum, log a WARN when `qos_to_profile()` shows publishers using non-default deadline/lifespan/liveliness so operators can spot the mismatch. Better: copy those fields into the subscriber QoS too.

src/ros2_medkit_gateway/test/test_ros2_topic_data_provider.cpp
```diff
@@ -0,0 +1,642 @@
...
+    std_msgs::msg::Int32 msg;
+    int counter = 0;
+    while (!stop.load(std::memory_order_acquire)) {
+      msg.data = counter++;
+      pub->publish(msg);
+      std::this_thread::sleep_for(5ms);
+    }
+  });
+
+  constexpr int kThreads = 4;
```
mfaferek93: **Major.** `ConcurrentSampleFromMultipleThreadsDoesNotCrash` is the regression test for issue #375, but it **passes even when the fix is reverted** in ~99% of runs without TSan. The original race window is in `create_generic_subscription` racing under `pool_mtx_` Phase 2 - i.e. cold-create only. With 4 threads × 10 iterations on a HOT topic, the first miss wins the race almost immediately and all subsequent iterations hit the fast path.

Fix: (1) add comment marking this test as "TSan/ASan required to catch original bug"; (2) bump to ~16 threads × 50 iterations across rotating topics to widen the cold-create window; (3) variant that calls `pool_->clear()` between iterations to force each call into the cold path; (4) gate CI with `ctest --repeat-until-fail=10` for this file specifically.

scripts/check_no_naked_subscriptions.sh
```diff
@@ -0,0 +1,106 @@
...
+    exit 2
+  fi
+done
+
+# Stage 1: raw matches. grep exit 1 (no match) is legitimate "no violations";
+# exit >=2 (bad pattern, unreadable file, IO error) must NOT be masked - that
+# is precisely the fail-open we want to avoid. Disable set -e around the call
+# so we can inspect the exit status explicitly.
+set +e
+raw_matches=$(grep -rEn "${FORBIDDEN_PATTERN}" "${scan_dirs[@]}" --include='*.cpp' --include='*.hpp' 2>&1)
```
mfaferek93: **Major.** `grep -rEn` runs **line-by-line**; a constructor call split across lines slips past the gate:

```cpp
auto sub = node->create_subscription
    <std_msgs::msg::String>(...);
```

No single line contains `create_subscription<` so the FORBIDDEN_PATTERN doesn't match. clang-format may not even reformat this for callers using deep template args. Synthetic verified.

Fix: switch to `pcre2grep -M` (multi-line), or assemble file-as-string with awk, or widen the regex to `create_subscription[[:space:]<]` (accepts some false positives that the gate's allowlist already handles).

scripts/check_no_naked_subscriptions.sh
```diff
@@ -0,0 +1,106 @@
...
+# means removing it from ALLOWED_LEGACY_FILES and routing through
+# Ros2SubscriptionSlot instead.
+#
+# Exit code:
+#   0  no violations outside ros2_common/, test/, or the allowlist
+#   1  at least one forbidden call outside allowed locations
+
+set -euo pipefail
+
+GATEWAY_ROOT="src/ros2_medkit_gateway"
```
mfaferek93, line 36: **Major.** Gate scope is hardcoded to `${GATEWAY_ROOT}` only. PR #384 also lands a sibling fix in `src/ros2_medkit_fault_manager/src/snapshot_capture.cpp` whose own comment (line 36-37) reads:

> *Locking the destroy-side too closes the window (same race class as issue #375 in the gateway).*

Fault manager has the same race, but is not under this gate. Same goes for `rosbag_capture.cpp`, `param_beacon_plugin.cpp`, etc. The gate misrepresents itself as a project-wide regression gate.

Fix: extend `scan_dirs` to `src/ros2_medkit_*/{src,include}` minus an explicit allowlist, **or** rename the script + comment block to make "gateway-only" explicit and document where the same rule applies elsewhere (and how it's enforced - or admit it isn't).

src/ros2_medkit_gateway/design/ros2_subscription_architecture.rst
```diff
@@ -0,0 +1,256 @@
...
+1. Graph query for publisher count / type. If no publishers, return
+   metadata-only immediately.
+2. Pool lookup under ``pool_mtx_``. Hit: promote the topic to the MRU end of
+   the LRU list and return the shared entry.
+3. Miss: build the entry shell, create the slot via
+   ``Ros2SubscriptionSlot::create_generic`` (this goes through the executor's
+   ``run_sync``), then insert under ``pool_mtx_``. If the pool is at cap,
+   evict the LRU entry first.
+4. Lock ``entry->buf_mtx``. If ``latest`` present, deserialize and return. If
+   cold and ``concurrent_cold_waits_ >= cold_wait_cap``, return metadata-only
```
mfaferek93: **Major.** Doc/code drift. Lines 149-150 say:

> *cold and `concurrent_cold_waits_ >= cold_wait_cap`, return metadata-only immediately*

Lines 184-186 repeat this ("further cold callers bounce back metadata-only immediately"). But the actual code at `src/data/ros2_topic_data_provider.cpp:374-381` returns `ErrorInfo{ERR_X_MEDKIT_COLD_WAIT_CAP_EXCEEDED, "...retry with backoff", 503, ...}` - a behavior change explicitly justified inline in the comment but never propagated to this doc.

Integration consequence: the dead `build_metadata_only_sample` helper (`ros2_topic_data_provider.cpp:570`) is the relic of the original design.

Fix: rewrite both passages to describe 503 + retry semantics. The helper should be deleted or re-wired in the same PR.

src/ros2_medkit_gateway/include/ros2_medkit_gateway/http/error_codes.hpp
```diff
@@ -99,6 +99,15 @@ constexpr const char * ERR_X_MEDKIT_ROS2_TOPIC_UNAVAILABLE = "x-medkit-ros2-topi
 /// ROS 2 action server not available
 constexpr const char * ERR_X_MEDKIT_ROS2_ACTION_UNAVAILABLE = "x-medkit-ros2-action-unavailable";
 
+/// Gateway is shutting down; request cannot be served
+constexpr const char * ERR_X_MEDKIT_GATEWAY_SHUTDOWN = "x-medkit-gateway-shutdown";
```
mfaferek93, line 103: **Major.** Three new public-facing `x-medkit-*` error codes (`gateway-shutdown`, `subscribe-failed`, `cold-wait-cap-exceeded` at lines 103/106/109) - **none of them are documented in `docs/api/rest.rst`**. Clients (MCP server, Web UI, anyone with retry-on-5xx logic) get a behavior change with zero docs:

- `cold-wait-cap-exceeded` is a 503 that should be retried with backoff (semantic distinct from `gateway-shutdown` 503).
- `subscribe-failed` is a 500 that indicates a transient rcl error.
- `gateway-shutdown` is a 503 that should NOT be retried by the same client (gateway is going down).

Fix: add the three codes to `docs/api/rest.rst` (or `docs/api/error_codes.rst` if that's the destination) with semantics + retry guidance. Without that, MCP and Web UI will conflate them all into "503 transient, retry" and hammer a shutting-down gateway.

src/ros2_medkit_gateway/include/ros2_medkit_gateway/data_access_manager.hpp
```diff
@@ -123,7 +131,7 @@ class DataAccessManager {
   mutable std::shared_mutex publishers_mutex_;
 
   std::unique_ptr<TypeIntrospection> type_introspection_;
-  std::unique_ptr<NativeTopicSampler> native_sampler_;
+  TopicDataProvider * topic_data_provider_{nullptr};  ///< Non-owning; set at wiring time.
   int max_parallel_samples_;
```
mfaferek93: **Major.** `max_parallel_samples_` is declared, validated (line 39 in cpp), and logged at init - but **never actually used** anywhere else. Verified with grep across the repo: zero read sites in production code. The chunked parallel dispatch moved to `Ros2TopicDataProvider::sample_parallel` which uses `cfg_.max_parallel_samples` (provider config, default 8), set via the YAML key `data_provider.max_parallel_samples`.

Result: two parameters look like sister knobs, only one is alive:
- `max_parallel_topic_samples: 10` (DAM, default 10) - **accepted but ignored**
- `data_provider.max_parallel_samples: 8` (provider, default 8) - **actually used**

The init log at `data_access_manager.cpp:55-56` reports `max_parallel_samples=10` even though sampling actually uses 8. Operators tuning the wrong knob get no signal that nothing changed.

Fix: delete the field + parameter declaration + README references at `README.md:291,1160,1167`, or flag deprecated and stop logging.

src/ros2_medkit_gateway/src/main.cpp
```diff
@@ -12,24 +12,100 @@
...
+  // previously killed /data on Rolling when concurrent HTTP handler threads
+  // created subscriptions on the same node.
+  const auto exec_cfg = declare_executor_config(*node);
+  const auto dp_cfg = declare_data_provider_config(*node);
+  auto sub_exec = std::make_shared<ros2_medkit_gateway::ros2_common::Ros2SubscriptionExecutor>(node, exec_cfg);
+  auto serializer = std::make_shared<ros2_medkit_serialization::JsonSerializer>();
+  auto data_provider = std::make_shared<ros2_medkit_gateway::Ros2TopicDataProvider>(sub_exec, serializer, dp_cfg);
+  node->set_topic_data_provider(data_provider);
+
   executor.spin();
```
mfaferek93: **Major.** `executor.spin()` is called bare, no `try/catch`. If a handler throws an unrecoverable exception that escapes to spin(), the explicit teardown block below (lines 102-110) is bypassed - which is exactly the rclcpp abort scenario the comment block above warns about ("~GatewayNode running against a dead executor... aborts with exit -6").

Fix: wrap in `try { executor.spin(); } catch (...) { /* log, fall through to teardown */ }` or use an RAII guard so the teardown runs unconditionally. Bonus: this also makes the unhappy path testable.

src/ros2_medkit_gateway/src/ros2_common/ros2_subscription_executor.cpp
bburda: Fixed in 28c89a01. Renamed the log line to mark the watchdog as observational only and made the degraded_ state sticky for the in-flight task: it now clears only when the worker is observed idle (started_ns==0), not on every sub-threshold tick of the same long-running task. Recovery action (kicking the worker / surfacing 503) is intentionally still out of scope - that is its own design discussion.

src/ros2_medkit_gateway/src/ros2_common/ros2_subscription_executor.cpp
bburda: Fixed in 28c89a01. fire_graph_callbacks now snapshots all listeners under graph_mtx_ and posts a SINGLE wrapper that fires the entire snapshot, so a queue overflow mid fan-out either delivers the event to everyone or to no one. graph_poll_tick refires on the next graph change instead of waiting for the 15-min idle-safety-net.

src/ros2_medkit_gateway/src/ros2_common/ros2_subscription_executor.cpp
bburda: Fixed in 28c89a01. Added an explicit @warning to the on_graph_change header doc telling registrants not to call remove_graph_change() from inside their own callback, with the recommended pattern (drop the token via a separate posted task).

src/ros2_medkit_gateway/src/ros2_common/ros2_subscription_executor.cpp
bburda: Fixed in 28c89a01. aux_loop now uses condition_variable::wait_until(min(next_watchdog, next_graph)) with a shutdown predicate. The destructor signals aux_cv_ alongside queue_cv_, so destruction returns within microseconds rather than the next 100ms graph_poll_tick boundary.

src/ros2_medkit_gateway/include/ros2_medkit_gateway/ros2_common/ros2_subscription_executor.hpp
bburda: Fixed in 28c89a01. Header doc on node() now lists allowed read-only graph queries (get_topic_names_and_types, count_publishers, etc) vs forbidden mutations (create_subscription, create_publisher, ...) with an explicit pointer to scripts/check_no_naked_subscriptions.sh as the structural enforcement. Did not rename the method since 41 callers are legitimate read-only graph queries; the gate (now multi-line aware, see e3cdeda3) catches the actual create_* mutations.

src/ros2_medkit_gateway/src/data/ros2_topic_data_provider.cpp
bburda: Fixed in 28c89a01. Deleted both the header declaration and the impl - it was a relic of the original cold-wait fallback that moved to a 503 + retry contract. Design doc passages that described the old metadata-only behavior were rewritten in e3cdeda3 to match the actual code.

src/ros2_medkit_gateway/src/data/ros2_topic_data_provider.cpp
bburda: Fixed in 28c89a01. Documented the canonical lock order at the top of ros2_topic_data_provider.cpp: pool_mtx_ before buf_mtx_, never reversed. sweep_idle_entries() is the only nested-lock site; any future edit that flips the order will conflict with that comment immediately.

src/ros2_medkit_gateway/src/data/ros2_topic_data_provider.cpp
bburda: Fixed in 28c89a01 by switching to partial-success per your suggestion. Per-topic recoverable errors (cold-wait cap, subscribe-failed) now embed into the corresponding TopicSampleResult via new error_code / error_message / error_http_status fields. Only ERR_X_MEDKIT_GATEWAY_SHUTDOWN remains batch-fatal. The interface contract was updated and a regression test (SampleParallelEmbedsPerTopicErrorInsteadOfFailingBatch in 2fe6723f) pins the new behavior.

src/ros2_medkit_gateway/src/data/ros2_topic_data_provider.cpp
bburda: Fixed in 28c89a01. qos_for now copies publisher Deadline / Lifespan / Liveliness into the subscriber QoS (using the first non-default value observed across publishers). When publishers declare conflicting contracts on the same topic, the provider logs a WARN naming the topic so operators can spot the misconfiguration.

src/ros2_medkit_gateway/test/test_ros2_topic_data_provider.cpp
bburda: Addressed in 2fe6723f. Bumped the test from 4x10 to 16x50 with a high cold_wait_cap so it actually exercises the cold-create race window rather than being limited by back-pressure shedding. Added a sibling test ConcurrentSampleStressColdCreatePath that rotates topic names per iteration so every sample goes through the cold-create path (the actual rcl_subscription_init race site). Both tests carry a comment noting they are TSan/ASan-required to catch the original bug; CI sanitizer-asan and sanitizer-tsan run all unit tests on every push.

scripts/check_no_naked_subscriptions.sh
bburda: Fixed in e3cdeda3. The inner scan now slurps each .cpp/.hpp via perl with -0777 (single-string read) so a constructor split across lines (auto sub = node->create_subscription
    <T>(...);) is detected. Verified locally with a synthetic split-call .cpp: the previous grep -rEn passed, the new perl scan flags it. Failure path also hardened to fail-closed: missing scan_dirs OR no source files OR perl missing all exit 2 instead of silently passing.

scripts/check_no_naked_subscriptions.sh
bburda: Fixed in e3cdeda3. scan_dirs extended to src/ros2_medkit_fault_manager/{src,include}. Allowlisted snapshot_capture.cpp/.hpp (uses LockedSubscriptionGuard for in-place serialisation) and rosbag_capture.cpp (uses its own node + executor with no shared rcl hash map) with explicit rationale comments. Failure message updated to call out the new project-wide scope.

src/ros2_medkit_gateway/design/ros2_subscription_architecture.rst
bburda: Fixed in e3cdeda3. Both passages in design/ros2_subscription_architecture.rst rewritten to match the actual 503 + retry contract that the code returns (lines 149-150 and 184-186). The dead build_metadata_only_sample helper was removed in 28c89a01.

src/ros2_medkit_gateway/include/ros2_medkit_gateway/http/error_codes.hpp
bburda: Fixed in e3cdeda3. docs/api/rest.rst Common Error Codes table now lists all three new vendor codes with semantics and per-code retry guidance: ERR_X_MEDKIT_GATEWAY_SHUTDOWN (do NOT retry against the same gateway), ERR_X_MEDKIT_SUBSCRIBE_FAILED (retry once with short backoff), ERR_X_MEDKIT_COLD_WAIT_CAP_EXCEEDED (retry with exponential backoff; params.cold_wait_cap carries the cap). Clients with retry-on-5xx logic can now distinguish the cases instead of conflating them.

src/ros2_medkit_gateway/include/ros2_medkit_gateway/data_access_manager.hpp
bburda: Fixed in 28c89a01. Deleted the field, parameter declaration, validation block, and init log reference. Live parallel-sample concurrency lives at data_provider.max_parallel_samples on the TopicDataProvider. Updated README + design/index.rst + gateway_params.yaml to point operators at the live knob (e3cdeda3).

src/ros2_medkit_gateway/src/main.cpp
bburda: Fixed in 28c89a01. main.cpp now wraps executor.spin() in try/catch and falls through to the explicit teardown block on any uncaught throw, with an RCLCPP_ERROR log line so the failure is visible. The exact rclcpp abort scenario the comment block warned about (~GatewayNode running against a dead executor, exit -6) no longer applies.

src/ros2_medkit_gateway/src/ros2_common/ros2_subscription_slot.cpp
bburda: Both issues fixed in 28c89a01.

(1) Worker self-deadlock: ~Ros2SubscriptionSlot now compares std::this_thread::get_id() to exec_->worker_thread_id() and drops the subscription inline (single-writer-safe because we ARE the single writer) when the dtor runs on the worker thread. The graph-callback-evicts-its-own-slot path no longer deadlocks on run_sync.

(2) Throwing dtor: dtor is now noexcept and the body is wrapped in try/catch logging at ERROR level, so OOM during run_sync's promise/future construction cannot escape and trip std::terminate when another exception is in flight.

mfaferek93 (approved):

Issue #338, opened 2026-04-02
[BUG] OpenAPI spec: entity list responses lack proper typing for generated clients


bburda:
# Bug report

### Steps to reproduce

1. Generate a typed client from the OpenAPI spec
2. Use the client to fetch entity lists (areas, components, apps, functions)
3. Try to access entity properties through the generated types

### Expected behavior

Entity list responses should be properly typed so generated clients can access all fields without casts.

### Actual behavior

Entity list endpoints return `EntityList` schema which is just `{items: Entity[]}` where `Entity` has `{id, name, type, href}`. The actual gateway responses include `x-medkit` extensions with rich metadata (ROS 2 node info, capabilities, etc.) that are invisible to the generated types.

This forces the web UI to use 15 `as unknown as SovdEntity` casts in store.ts to work around the type system. Every entity list response is cast through `unknown`, bypassing the type safety that the generated client should provide.

### Suggested fix

1. Add proper `x-medkit` sub-schemas to entity types in the OpenAPI spec (DataItem, OperationItem, FaultListItem, ConfigurationMetaData, BulkDataDescriptor)
2. Include entity metadata fields that the gateway actually returns
3. Regenerate clients after spec update

Related: selfpatch/ros2_medkit#328 (tracks other spec gaps including lock/trigger schemas)

### Environment
- ros2_medkit version: 0.4.0
- Discovered during: web UI migration to generated client (ros2_medkit_web_ui#53)

PR #390, opened 2026-04-26
Nest x-medkit vendor fields on SOVD endpoint payloads

### LINKED ISSUE #385: Standardize vendor fields on SOVD-endpoint payloads as nested x-medkit object
## Summary

We want one consistent convention for vendor fields on SOVD-standard endpoint payloads: **nested `"x-medkit": { ... }`**. Today that rule is broken in at least one place: `GET /updates/{id}/status` emits the vendor phase as a flat top-level key `"x-medkit-phase"` instead of `"x-medkit": { "phase": "..." }`. Surfaced during review of https://github.com/selfpatch/ros2_medkit_web_ui/pull/72 (new typed client-side support for that field).

Other SOVD-endpoint payloads already follow the nested convention (examples from existing handlers and transforms):
- Faults collections: `"x-medkit": { "count": N }`
- Data / operations detail: `"x-medkit": { "ros2": { "kind": "...", "topic": "...", "type": "..." }, "access": "read|write", "entity_id": "..." }`
- Bulkdata: `"x-medkit": { "fault_code": "...", "duration_sec": 5.0, "format": "mcap" }`

## Proposed solution

1. On SOVD-standard endpoint responses, move all `x-medkit-*` flat top-level vendor keys into a nested `x-medkit` object. Concrete case today: `GET /updates/{id}/status` emits `{ "x-medkit": { "phase": "prepared" } }`.
2. Reflect the change in schema + docs (`docs/api/rest.rst`, OpenAPI spec).
3. Downstream clients update type + readers in lockstep (web UI, MCP server, foxglove extension, generated clients).

## Out of scope

- Vendor-extension endpoints like `/apps/{id}/x-medkit-param-beacon`. That route lives under the OpenAPI `x-<vendor>-<name>` namespace and its payload is fully vendor-owned, so the nesting rule does not apply there.
- Capability / route naming - not affected.

## Additional context

Existing nested examples for reference:
- `src/ros2_medkit_gateway/src/http/handlers/fault_handlers.cpp` (faults `x-medkit.count`)
- `src/ros2_medkit_gateway/src/http/handlers/data_handlers.cpp`, `operations_handlers.cpp` (`x-medkit.ros2.*`, `x-medkit.access`, `x-medkit.entity_id`)
- `src/ros2_medkit_gateway/src/http/handlers/bulkdata_handlers.cpp`

Flat cases to migrate (non-exhaustive):
- `src/ros2_medkit_gateway/include/ros2_medkit_gateway/updates/update_types.hpp` - `x-medkit-phase` on `GET /updates/{id}/status`

bburda:
## Summary

Migrates the two remaining SOVD-standard endpoint payloads that emit flat
`x-medkit-*` top-level vendor keys to the nested `x-medkit: {...}`
convention that every other endpoint already follows.

- `GET /updates/{id}/status` -> `x-medkit.phase` (was flat `x-medkit-phase`)
- `GET /apps|components/{id}/configurations` -> nested `x-medkit.source`
  on items, `x-medkit.{source,node}` on per-parameter entries (was flat
  `x-medkit-source` / `x-medkit-node`)

The OpenAPI schemas for `UpdateStatus` and `ConfigurationMetaData` now
declare the nested object so generated clients pick up accurate typing.

Adds `test_openapi_response_drift` - a runtime drift validator that
walks every GET endpoint declared in `GET /docs`, fetches a real
response, and validates it against the declared response schema. The
original `x-medkit-phase` divergence slipped past every typed client
because the schema never declared the field; this test catches that
class of regression. The compile-time `emitter -> schema` contract is
deferred to #338.

---

## Issue

- closes #385

---

## Type

- [x] Bug fix
- [x] New feature or tests
- [x] Breaking change
- [ ] Documentation only

Breaking on the wire format of two endpoints. Cross-repo grep showed
zero current consumers reading the legacy flat keys (web UI, MCP,
foxglove, demos, generated clients, commercial). 

---

## Testing

- `colcon build --packages-up-to ros2_medkit_gateway ros2_medkit_integration_tests` (Jazzy, Release)
- `colcon test --packages-select ros2_medkit_gateway --ctest-args -L linter` -> green
- `ctest -R test_update_manager` -> 24/24 passing (UpdateStatusToJson + UpdateManagerFailureTest assertions updated)
- Smoke: `GET /api/v1/docs` returns `UpdateStatus.properties.x-medkit` schema with `phase` enum
- Drift integration test (`test_openapi_response_drift`) runs in CI - requires `python3-jsonschema` (added to `ros2_medkit_integration_tests/package.xml`)

---

## Checklist

- [x] Breaking changes are clearly described
- [x] Docs were updated (`docs/api/rest.rst` example payload + vendor extension prose)

### DIFF
--- a/src/ros2_medkit_gateway/include/ros2_medkit_gateway/ros2_common/ros2_subscription_executor.hpp
+++ b/src/ros2_medkit_gateway/include/ros2_medkit_gateway/ros2_common/ros2_subscription_executor.hpp
@@ -189,12 +189,19 @@ class Ros2SubscriptionExecutor final {
    * disappearing, types changing). Use to invalidate per-topic caches or evict stale
    * pool entries.
    *
-   * @warning A registered callback must NOT call remove_graph_change() on its own
-   *          token from inside the callback - graph_mtx_ is non-recursive and the
-   *          synchronous in-flight drain in remove_graph_change() would wait
-   *          indefinitely on the very call holding the slot's in_flight counter.
-   *          Drop the token through a separate task posted to the worker if
-   *          dynamic deregistration is needed.
+   * @warning DO NOT call remove_graph_change() on this token from inside the
+   *          callback body. The deadlock is concrete: remove_graph_change()
+   *          acquires graph_mtx_ and waits for graph_in_flight_[token] to
+   *          drain to zero, but the wrapper that decrements is the very call
+   *          frame that ran your callback. The worker thread is now blocked
+   *          on cv.wait, the decrement never runs, the wait never returns.
+   *          Symptom in production: hung worker thread, /health stops
+   *          updating, no log line. There is no runtime detection (would
+   *          need thread_local TLS, banned in code linked into the gateway
+   *          plugin MODULE; see CLAUDE.md). Workaround for dynamic
+   *          deregistration: have the callback hand the token to a
+   *          non-worker thread via a queue/promise, and call
+   *          remove_graph_change() from there.
    *
    * @return Opaque token in range [0, kMaxGraphListeners). `kMaxGraphListeners` if
    *         all slots are taken.
--- a/src/ros2_medkit_gateway/include/ros2_medkit_gateway/ros2_common/ros2_subscription_executor.hpp
+++ b/src/ros2_medkit_gateway/include/ros2_medkit_gateway/ros2_common/ros2_subscription_executor.hpp
@@ -190,16 +190,29 @@ class Ros2SubscriptionExecutor final {
    * pool entries.
    *
    * @warning A registered callback must NOT call remove_graph_change() on its own
-   *          token from inside the callback - graph_mtx_ is non-recursive and a
-   *          self-removal attempt deadlocks. Drop the token through a separate task
-   *          posted to the worker if dynamic deregistration is needed.
+   *          token from inside the callback - graph_mtx_ is non-recursive and the
+   *          synchronous in-flight drain in remove_graph_change() would wait
+   *          indefinitely on the very call holding the slot's in_flight counter.
+   *          Drop the token through a separate task posted to the worker if
+   *          dynamic deregistration is needed.
    *
    * @return Opaque token in range [0, kMaxGraphListeners). `kMaxGraphListeners` if
    *         all slots are taken.
    */
   [[nodiscard]] std::size_t on_graph_change(GraphCallback cb);
 
-  /// Remove a previously-registered graph callback. Idempotent. See warning on on_graph_change.
+  /**
+   * @brief Remove a previously-registered graph callback. Idempotent.
+   *
+   * Synchronously drains in-flight execution: returns only after any worker
+   * thread that already entered this slot's callback has exited it. After
+   * remove_graph_change() returns, the caller can safely tear down state that
+   * the callback's captured `this` pointer references; the executor guarantees
+   * no further dispatch and no in-flight call.
+   *
+   * See warning on on_graph_change(): self-removal from inside the callback
+   * deadlocks.
+   */
   void remove_graph_change(std::size_t token);
 
   /// True after shutdown has started. Monotonic. Use to skip re-posting work on teardown.
@@ -269,10 +282,19 @@ class Ros2SubscriptionExecutor final {
   std::atomic<std::size_t> watchdog_trips_{0};
   std::atomic<std::size_t> graph_events_received_{0};
 
-  // Graph callbacks (pre-allocated, Tier 1)
+  // Graph callbacks (pre-allocated, Tier 1).
+  // graph_in_flight_[i] tracks how many worker invocations of slot i's
+  // callback are currently mid-execution. fire_graph_callbacks() increments
+  // it when snapshotting under graph_mtx_, the snapshot wrapper decrements
+  // it after the callback returns and notifies graph_in_flight_cv_. This
+  // lets remove_graph_change() block until in-flight execution drains, so a
+  // consumer's destructor can safely tear down state that the callback's
+  // captured `this` references.
   mutable std::mutex graph_mtx_;
   std::array<GraphCallback, kMaxGraphListeners> graph_callbacks_{};
   std::array<bool, kMaxGraphListeners> graph_slot_used_{};
+  std::array<int, kMaxGraphListeners> graph_in_flight_{};
+  std::condition_variable graph_in_flight_cv_;
 
   // Public rclcpp graph API - stable across Humble / Jazzy / Rolling
   rclcpp::Event::SharedPtr graph_event_;
--- a/src/ros2_medkit_gateway/src/openapi/schema_builder.cpp
+++ b/src/ros2_medkit_gateway/src/openapi/schema_builder.cpp
@@ -160,12 +160,20 @@ nlohmann::json SchemaBuilder::items_wrapper(const nlohmann::json & item_schema) 
 }
 
 nlohmann::json SchemaBuilder::configuration_metadata_schema() {
-  return {{"type", "object"},
+  return {
+      {"type", "object"},
+      {"properties",
+       {{"id", {{"type", "string"}, {"description", "Configuration parameter ID"}}},
+        {"name", {{"type", "string"}, {"description", "Parameter name"}}},
+        {"type", {{"type", "string"}, {"description", "Parameter type (e.g. 'parameter')"}}},
+        {"x-medkit",
+         {{"type", "object"},
+          {"description", "Vendor extensions (medkit)"},
           {"properties",
-           {{"id", {{"type", "string"}, {"description", "Configuration parameter ID"}}},
-            {"name", {{"type", "string"}, {"description", "Parameter name"}}},
-            {"type", {{"type", "string"}, {"description", "Parameter type (e.g. 'parameter')"}}}}},
-          {"required", {"id", "name", "type"}}};
+           {{"source",
+             {{"type", "string"},
+              {"description", "App ID that owns this parameter (only present in aggregated configurations)"}}}}}}}}},
+      {"required", {"id", "name", "type"}}};
 }
 
 nlohmann::json SchemaBuilder::configuration_read_value_schema() {
@@ -538,13 +546,24 @@ nlohmann::json SchemaBuilder::update_status_schema() {
       {"properties", {{"name", {{"type", "string"}}}, {"progress", {{"type", "number"}}}}},
       {"required", {"name", "progress"}}};
 
+  nlohmann::json x_medkit_schema = {
+      {"type", "object"},
+      {"description", "Vendor extensions (medkit)"},
+      {"properties",
+       {{"phase",
+         {{"type", "string"},
+          {"enum", {"none", "preparing", "prepared", "executing", "executed", "failed", "deleting"}},
+          {"description", "Internal lifecycle phase, distinguishes prepare-completed from execute-completed"}}}}},
+      {"required", {"phase"}}};
+
   return {{"type", "object"},
           {"properties",
            {{"status", {{"type", "string"}, {"enum", {"pending", "inProgress", "completed", "failed"}}}},
             {"progress", {{"type", "number"}}},
             {"sub_progress", {{"type", "array"}, {"items", sub_progress_schema}}},
-            {"error", {{"type", "string"}}}}},
-          {"required", {"status"}}};
+            {"error", {{"type", "string"}}},
+            {"x-medkit", x_medkit_schema}}},
+          {"required", {"status", "x-medkit"}}};
 }
 
 nlohmann::json SchemaBuilder::log_configuration_schema() {
--- a/src/ros2_medkit_gateway/src/openapi/schema_builder.cpp
+++ b/src/ros2_medkit_gateway/src/openapi/schema_builder.cpp
@@ -172,7 +172,10 @@ nlohmann::json SchemaBuilder::configuration_metadata_schema() {
           {"properties",
            {{"source",
              {{"type", "string"},
-              {"description", "App ID that owns this parameter (only present in aggregated configurations)"}}}}}}}}},
+              {"description", "App ID that owns this parameter (only present in aggregated configurations)"}}},
+            {"node",
+             {{"type", "string"},
+              {"description", "Node FQN providing this parameter (only present in aggregated configurations)"}}}}}}}}},
       {"required", {"id", "name", "type"}}};
 }
 
@@ -556,11 +559,14 @@ nlohmann::json SchemaBuilder::update_status_schema() {
           {"description", "Internal lifecycle phase, distinguishes prepare-completed from execute-completed"}}}}},
       {"required", {"phase"}}};
 
-  // x-medkit is optional in the schema (matches the convention used by
-  // fault_detail_schema and entity_detail_schema): SOVD does not require
-  // vendor extensions, so a SOVD-compliant client must be able to ignore
-  // it. The gateway always emits it; the explicit handler guard in
-  // test_openapi_response_drift covers regression risk.
+  // x-medkit is optional in the SOVD payload (clients may ignore vendor
+  // extensions; matches the convention in fault_detail_schema and
+  // entity_detail_schema). When the gateway DOES emit the x-medkit object,
+  // however, ``phase`` is mandatory inside it - that scope is enforced by
+  // the inner ``required: {phase}`` above, NOT by listing x-medkit in the
+  // parent's required list. The drift test in test_openapi_response_drift
+  // covers regression on the emit side. If x-medkit is ever dropped from
+  // the parent properties, the inner required must be revisited too.
   return {{"type", "object"},
           {"properties",
            {{"status", {{"type", "string"}, {"enum", {"pending", "inProgress", "completed", "failed"}}}},
--- a/src/ros2_medkit_gateway/src/openapi/schema_builder.cpp
+++ b/src/ros2_medkit_gateway/src/openapi/schema_builder.cpp
@@ -556,6 +556,11 @@ nlohmann::json SchemaBuilder::update_status_schema() {
           {"description", "Internal lifecycle phase, distinguishes prepare-completed from execute-completed"}}}}},
       {"required", {"phase"}}};
 
+  // x-medkit is optional in the schema (matches the convention used by
+  // fault_detail_schema and entity_detail_schema): SOVD does not require
+  // vendor extensions, so a SOVD-compliant client must be able to ignore
+  // it. The gateway always emits it; the explicit handler guard in
+  // test_openapi_response_drift covers regression risk.
   return {{"type", "object"},
           {"properties",
            {{"status", {{"type", "string"}, {"enum", {"pending", "inProgress", "completed", "failed"}}}},
@@ -563,7 +568,7 @@ nlohmann::json SchemaBuilder::update_status_schema() {
             {"sub_progress", {{"type", "array"}, {"items", sub_progress_schema}}},
             {"error", {{"type", "string"}}},
             {"x-medkit", x_medkit_schema}}},
-          {"required", {"status", "x-medkit"}}};
+          {"required", {"status"}}};
 }
 
 nlohmann::json SchemaBuilder::log_configuration_schema() {
--- a/src/ros2_medkit_gateway/src/ros2_common/ros2_subscription_executor.cpp
+++ b/src/ros2_medkit_gateway/src/ros2_common/ros2_subscription_executor.cpp
@@ -135,9 +135,17 @@ void Ros2SubscriptionExecutor::remove_graph_change(std::size_t token) {
   if (token >= kMaxGraphListeners) {
     return;
   }
-  std::lock_guard<std::mutex> lk(graph_mtx_);
+  std::unique_lock<std::mutex> lk(graph_mtx_);
   graph_slot_used_[token] = false;
   graph_callbacks_[token] = nullptr;
+  // Wait for any worker-thread invocation that already snapshotted this
+  // slot to finish executing the callback. Without this drain, a wrapper
+  // task posted by fire_graph_callbacks() could call cb() against a
+  // captured `this` after the consumer's destructor has freed members,
+  // producing the heap-use-after-free in the snapshotted callback body.
+  graph_in_flight_cv_.wait(lk, [this, token] {
+    return graph_in_flight_[token] == 0;
+  });
 }
 
 Ros2SubscriptionExecutor::Stats Ros2SubscriptionExecutor::stats() const {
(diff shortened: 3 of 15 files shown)


src/ros2_medkit_integration_tests/test/features/test_openapi_response_drift.test.py
```diff
@@ -0,0 +1,291 @@
...
+        if '$ref' in schema:
+            ref = schema['$ref']
+            if ref in seen:
+                return {}
+            if ref.startswith('#/components/schemas/'):
+                name = ref.rsplit('/', 1)[-1]
+                target = schemas.get(name)
+                if target is None:
+                    return {}
+                return _inline_refs(target, schemas, seen | {ref})
```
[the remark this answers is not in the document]
bburda: Fixed in c492f9e0: `_inline_refs` now raises `_RefResolutionError` on cycles, unknown refs, or unsupported `$ref` forms instead of returning `{}`. The validation loop catches the error and records it as a violation so multiple spec issues surface in one run.

src/ros2_medkit_integration_tests/test/features/test_openapi_response_drift.test.py
```diff
@@ -0,0 +1,291 @@
...
+
+For every GET endpoint declared in the runtime OpenAPI spec served at
+GET /docs, this test fetches a real response from the live gateway and
+validates it against the response schema declared in the same spec for the
+returned status code. A mismatch means handler output and declared schema
+have drifted apart - either the handler emits a field the schema does not
+declare (the failure mode that caused issue #385's `x-medkit-phase` to
+silently slip past every typed client), or the schema declares a required
+field the handler does not emit.
+
```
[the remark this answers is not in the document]
bburda: Fixed in c492f9e0: docstring rewritten to spell out the catch surface (required, type, enum, nested shape) and explicitly call out that extra undeclared top-level fields are NOT caught because most schemas leave `additionalProperties` open. Closing that gap is the issue #338 rework, not this PR.

src/ros2_medkit_gateway/test/test_update_manager.cpp
```diff
@@ -656,7 +656,7 @@ TEST(UpdateManagerFailureTest, DeleteRollbackUpdatesStatusPhaseAndErrorMessage)
   // Serialized payload reflects the rollback end-to-end.
   auto j = update_status_to_json(*status);
   EXPECT_EQ(j["status"], "failed");
-  EXPECT_EQ(j["x-medkit-phase"], "failed");
+  EXPECT_EQ(j["x-medkit"]["phase"], "failed");
   EXPECT_EQ(j["error"], "backend delete exploded");
```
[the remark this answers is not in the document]
bburda: Fixed in c492f9e0: switched all four sites to `contains()` + `at()`. `nlohmann::json::operator[]` on a non-const json inserts missing keys, which would let the test pass silently if `update_status_to_json` stopped emitting the vendor extension.

src/ros2_medkit_gateway/test/test_update_manager.cpp
```diff
@@ -667,20 +667,20 @@ TEST(UpdateStatusToJson, SerializesPhaseAsVendorExtension) {
   UpdateStatusInfo status{UpdateStatus::Completed, UpdatePhase::Prepared, std::nullopt, std::nullopt, std::nullopt};
   auto j = update_status_to_json(status);
   EXPECT_EQ(j["status"], "completed");
-  EXPECT_EQ(j["x-medkit-phase"], "prepared");
+  EXPECT_EQ(j["x-medkit"]["phase"], "prepared");
 }
```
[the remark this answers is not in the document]
bburda: Fixed in c492f9e0 (same change as the line 660 thread).

src/ros2_medkit_gateway/test/test_update_manager.cpp
```diff
@@ -667,20 +667,20 @@ TEST(UpdateStatusToJson, SerializesPhaseAsVendorExtension) {
...
 
 TEST(UpdateStatusToJson, ExposesExecutedPhaseForTerminalCompletion) {
   UpdateStatusInfo status{UpdateStatus::Completed, UpdatePhase::Executed, 100, std::nullopt, std::nullopt};
   auto j = update_status_to_json(status);
   EXPECT_EQ(j["status"], "completed");
-  EXPECT_EQ(j["x-medkit-phase"], "executed");
+  EXPECT_EQ(j["x-medkit"]["phase"], "executed");
   EXPECT_EQ(j["progress"], 100);
```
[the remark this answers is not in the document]
bburda: Fixed in c492f9e0 (same change as the line 660 thread).

src/ros2_medkit_gateway/test/test_update_manager.cpp
```diff
@@ -667,20 +667,20 @@ TEST(UpdateStatusToJson, SerializesPhaseAsVendorExtension) {
...
 }
 
 TEST(UpdateStatusToJson, EmitsNonePhaseForFreshlyRegistered) {
   UpdateStatusInfo status;
   auto j = update_status_to_json(status);
   EXPECT_EQ(j["status"], "pending");
-  EXPECT_EQ(j["x-medkit-phase"], "none");
+  EXPECT_EQ(j["x-medkit"]["phase"], "none");
 }
```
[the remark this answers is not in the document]
bburda: Fixed in c492f9e0 (same change as the line 660 thread).

src/ros2_medkit_integration_tests/test/features/test_openapi_response_drift.test.py
```diff
@@ -0,0 +1,279 @@
...
+            if ref.startswith('#/components/schemas/'):
+                name = ref.rsplit('/', 1)[-1]
+                target = schemas.get(name)
+                if target is None:
+                    return {}
+                return _inline_refs(target, schemas, seen | {ref})
+        return {k: _inline_refs(v, schemas, seen) for k, v in schema.items()}
+    if isinstance(schema, list):
+        return [_inline_refs(item, schemas, seen) for item in schema]
+    return schema
```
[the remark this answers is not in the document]
bburda: Fixed in c492f9e0 (same change as the later thread on this file): `_inline_refs` now raises `_RefResolutionError` instead of returning `{}` on unresolved or cyclic refs.

src/ros2_medkit_integration_tests/test/features/test_openapi_response_drift.test.py
```diff
@@ -0,0 +1,279 @@
...
+        return resource_id
+
+    return re.sub(r'\{([^}]+)\}', replacer, path)
+
+
+class TestOpenApiResponseDrift(GatewayTestCase):
+    """Catches handler-vs-spec drift on GET response bodies."""
+
+    MIN_EXPECTED_APPS = 1
+
```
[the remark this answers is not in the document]
bburda: Fixed in c492f9e0: `REQUIRED_APPS = {"temp_sensor", "calibration"}` plus `MIN_EXPECTED_APPS = 2`, so `setUpClass` now waits until both demo nodes the fixture relies on are actually discovered.

src/ros2_medkit_integration_tests/test/features/test_openapi_response_drift.test.py
```diff
@@ -0,0 +1,316 @@
...
+        """
+        resp = requests.get(
+            f'{self.BASE_URL}/updates/{self._update_pkg_id}/status', timeout=5
+        )
+        self.assertEqual(resp.status_code, 200)
+        body = resp.json()
+        self.assertNotIn('x-medkit-phase', body, 'Legacy flat key still emitted')
+        self.assertIn('x-medkit', body)
+        self.assertIn('phase', body['x-medkit'])
+
```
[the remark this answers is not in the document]
bburda: Fixed in 53fddac8: added `test_configurations_payload_uses_nested_x_medkit`. It exercises `temp_sensor`'s four declared ROS 2 parameters and asserts no flat `x-medkit-source` / `x-medkit-node` keys on items or on parameters inside `x-medkit.parameters[]`, plus a positive check that the nested vendor attribution is present.

src/ros2_medkit_gateway/src/openapi/schema_builder.cpp
```diff
@@ -160,12 +160,20 @@ nlohmann::json SchemaBuilder::items_wrapper(const nlohmann::json & item_schema)
...
+         {{"type", "object"},
+          {"description", "Vendor extensions (medkit)"},
           {"properties",
-           {{"id", {{"type", "string"}, {"description", "Configuration parameter ID"}}},
-            {"name", {{"type", "string"}, {"description", "Parameter name"}}},
-            {"type", {{"type", "string"}, {"description", "Parameter type (e.g. 'parameter')"}}}}},
-          {"required", {"id", "name", "type"}}};
+           {{"source",
+             {{"type", "string"},
+              {"description", "App ID that owns this parameter (only present in aggregated configurations)"}}}}}}}}},
```
mfaferek93: **[Blocker] Schema gap reintroduces the class of bug this PR fixes.**

`config_handlers.cpp:321` emits `param_with_source["x-medkit"] = {{"source", node_info.app_id}, {"node", node_info.node_fqn}}` on every per-parameter entry, but `configuration_metadata_schema()` here only declares `x-medkit.properties.source`. `node` lands on the wire undeclared - same drift signature that motivated #385.

Generated clients (typed) silently drop or fail-type the field. The drift test added in this PR catches divergence in *declared* schemas but not extra fields (most schemas don't set `additionalProperties: false`), so this slips through.

Fix: add `node` to the x-medkit properties block:
```cpp
{"node", {{"type", "string"},
          {"description", "Node FQN providing this parameter "
                          "(only present in aggregated configs)"}}}
```

src/ros2_medkit_gateway/src/ros2_common/ros2_subscription_executor.cpp
```diff
@@ -135,9 +135,17 @@ void Ros2SubscriptionExecutor::remove_graph_change(std::size_t token) {
...
   graph_slot_used_[token] = false;
   graph_callbacks_[token] = nullptr;
+  // Wait for any worker-thread invocation that already snapshotted this
+  // slot to finish executing the callback. Without this drain, a wrapper
+  // task posted by fire_graph_callbacks() could call cb() against a
+  // captured `this` after the consumer's destructor has freed members,
+  // producing the heap-use-after-free in the snapshotted callback body.
+  graph_in_flight_cv_.wait(lk, [this, token] {
+    return graph_in_flight_[token] == 0;
+  });
```
mfaferek93, line 148: **[Blocker] `graph_in_flight_[token]` is never reset on remove - stale count on slot reuse.**

After `cv.wait` returns here, the function ends. `graph_slot_used_[token]=false` (line 139) and `graph_callbacks_[token]=nullptr` (line 140) are reset, but `graph_in_flight_[token]` retains the old count.

Slot reuse path (`on_graph_change` at cpp:125 finds first `!graph_slot_used_[i]`) re-assigns the same token to a new callback. The new registration starts with a stale non-zero `graph_in_flight_[token]`. A subsequent `remove_graph_change` on this new registration blocks on `cv.wait` waiting for a count that was already wrong at registration time - either deadlocks until the slot fires enough graph events to decrement to zero, or hangs forever if events never come.

Round 1 missed this; round 2 verification caught it. One-line fix:
```cpp
graph_in_flight_cv_.wait(lk, [this, token] {
  return graph_in_flight_[token] == 0;
});
graph_in_flight_[token] = 0;  // reset so slot can be safely reused
```

src/ros2_medkit_integration_tests/test/features/test_openapi_response_drift.test.py
```diff
@@ -0,0 +1,363 @@
...
+    @classmethod
+    def tearDownClass(cls):
+        requests.delete(
+            f'{cls.BASE_URL}/updates/{cls._update_pkg_id}', timeout=5
+        )
+        super().tearDownClass()
+
+    def _fetch_spec(self):
+        return self.poll_endpoint_until(
```
mfaferek93, line 233: **[Major] `_fetch_spec` poll completeness - validates against partial spec.**

The lambda checks only that `paths` is non-empty/exists. If the spec endpoint populates lazily (slow CI runner, dynamic plugin registration), the test reads back the spec the moment `paths` becomes truthy - which may be just one entry. Subsequent endpoints registered later are silently un-validated.

This is exactly the failure mode the drift test is supposed to catch.

Suggestion: poll until either `len(spec['paths']) >= MIN_EXPECTED_PATHS` or until a known-late endpoint (e.g. a plugin-registered route like `/apps/.../x-plc-data`) is present.
```python
return self.poll_endpoint_until(
    '/docs',
    lambda d: d if (d.get('paths') and len(d['paths']) >= MIN_EXPECTED_PATHS) else None
)
```

src/ros2_medkit_integration_tests/test/features/test_openapi_response_drift.test.py
```diff
@@ -0,0 +1,363 @@
...
+
+        @verifies REQ_INTEROP_002
+        """
+        spec = self._fetch_spec()
+        schemas = spec.get('components', {}).get('schemas', {})
+        violations = []
+        validated = 0
+
+        for path, path_item in spec.get('paths', {}).items():
+            operation = path_item.get('get')
```
mfaferek93, line 280: **[Major] Class docstring claims "every endpoint" but loop walks GET only.**

POST/PUT/DELETE endpoints are declared in the spec - confirmed via `route_registry.cpp:128, 132, 136`. The drift validator skips them entirely (`path_item.get('get')` here) so request-body schemas and 4xx/5xx response shapes aren't covered.

Two paths forward:
1. (Min) Update the class docstring to spell out the exclusions: "Validates GET responses only. POST/PUT/DELETE request bodies and error response shapes are out of scope." Spawn follow-up issue.
2. (Better) Extend the loop to cover all verbs - request schemas validate against `requestBody.content.*.schema`, response schemas same as today.

Either way, leaving the docstring as-is misleads future readers.

src/ros2_medkit_integration_tests/test/features/test_openapi_response_drift.test.py
```diff
@@ -0,0 +1,363 @@
...
+    """Detect SSE streaming endpoints that would hang on a plain GET."""
+    summary = (operation.get('summary') or '') + (operation.get('description') or '')
+    return 'SSE' in summary or 'stream' in summary.lower()
+
+
+def _substitute_path_params(path, entity_map, resource_id='test_id'):
+    def replacer(match):
+        param = match.group(1)
+        if param in _ENTITY_PARAM_MAP:
+            return entity_map.get(_ENTITY_PARAM_MAP[param], 'test_entity')
```
mfaferek93: **[Major] Hardcoded `'test_entity'` fallback substitutes against bogus paths.**

When entity discovery (lines 161-167) doesn't populate an entity type (e.g. `functions` returns empty), `cls._entity_map` lacks that key. `_substitute_path_params` here returns `'test_entity'` as the default. The test then constructs URLs like `/functions/test_entity/operations` and silently skips the resulting 404 at line 258-259 (`continue` on non-200).

Result: drift test reports green for entity types it never actually validated. The operator sees "all GET endpoints validated" when in fact the function-scoped ones were skipped silently.

Fix options:
1. Fail loudly if a critical entity type returns no items in setUpClass (rather than substituting).
2. Skip endpoints that need a missing entity type rather than substitute.
3. At minimum, log a `WARNING` per substitution so the test output makes the gap visible.

src/ros2_medkit_gateway/include/ros2_medkit_gateway/ros2_common/ros2_subscription_executor.hpp
```diff
@@ -190,16 +190,29 @@ class Ros2SubscriptionExecutor final {
    * pool entries.
    *
    * @warning A registered callback must NOT call remove_graph_change() on its own
-   *          token from inside the callback - graph_mtx_ is non-recursive and a
-   *          self-removal attempt deadlocks. Drop the token through a separate task
-   *          posted to the worker if dynamic deregistration is needed.
+   *          token from inside the callback - graph_mtx_ is non-recursive and the
+   *          synchronous in-flight drain in remove_graph_change() would wait
```
mfaferek93: **[Major] Self-removal deadlock is documented but not detected at runtime.**

`graph_mtx_` is non-recursive, so a callback that calls `remove_graph_change()` on its own token from inside the callback deadlocks: `remove_graph_change` acquires `graph_mtx_` then waits on `graph_in_flight_[token] > 0` to drain, but the decrement that would unblock it lives in the wrapper that's currently blocked on the very lock `remove_graph_change` holds.

A documented precondition is fine for senior engineers, but production code paths evolve. A junior adding ad-hoc reconfiguration ("clean up myself if I see a stale config") trips this with no diagnostic - just a hung daemon.

Cheap defensive fix: stash a `thread_local` token in the wrapper before invoking the user callback, check it in `remove_graph_change` and assert with an actionable message rather than deadlock:
```cpp
if (current_callback_token == token) {
  RCLCPP_FATAL(logger, "remove_graph_change(%zu) called from inside its own callback - "
                       "deferring removal would loop forever; use post() to schedule it instead", token);
  std::abort();
}
```

src/ros2_medkit_gateway/src/openapi/schema_builder.cpp
```diff
@@ -538,12 +546,28 @@ nlohmann::json SchemaBuilder::update_status_schema() {
...
+       {{"phase",
+         {{"type", "string"},
+          {"enum", {"none", "preparing", "prepared", "executing", "executed", "failed", "deleting"}},
+          {"description", "Internal lifecycle phase, distinguishes prepare-completed from execute-completed"}}}}},
+      {"required", {"phase"}}};
+
+  // x-medkit is optional in the schema (matches the convention used by
+  // fault_detail_schema and entity_detail_schema): SOVD does not require
+  // vendor extensions, so a SOVD-compliant client must be able to ignore
+  // it. The gateway always emits it; the explicit handler guard in
```
mfaferek93: **[Major] Comment is subtle - reader may misread the requirement scope.**

The x-medkit sub-schema declares `"required": {"phase"}` at line 557 while the parent at line 571 only requires `status`. So `phase` is required *if* the x-medkit extension object is present - SOVD-correct but easy to misread as "phase is always required".

A future cleanup that drops `x-medkit` from the parent properties would leave the `required: phase` orphan as silent dead code. Tightening the comment costs nothing:

```
// x-medkit is optional in the SOVD payload (clients may ignore vendor extensions);
// however, when the gateway DOES emit it, ``phase`` is mandatory inside it. The
// drift test in test_openapi_response_drift covers regression on the emit side.
```

mfaferek93 (commented):
A few comments :)

src/ros2_medkit_gateway/src/openapi/schema_builder.cpp
bburda: Fixed in dbcd0109: added `x-medkit.node` property next to `source`. Plus a static SchemaBuilder test (`ConfigurationMetaDataXMedkitDeclaresAllEmittedFields`) that fails if either property goes missing - the drift integration test cannot detect this since `additionalProperties` is open here.

src/ros2_medkit_gateway/src/ros2_common/ros2_subscription_executor.cpp
bburda: Pushback on the bug claim: predicate-form `cv.wait(lk, pred)` is equivalent to `while (!pred()) cv.wait(lk);` - when wait returns, the predicate is true, so `graph_in_flight_[token] == 0` is guaranteed. We hold `graph_mtx_` throughout, so no concurrent path can flip the value before we release. Slot reuse therefore starts from 0 deterministically.

The suggested `graph_in_flight_[token] = 0;` line would be a no-op (assigns 0 to 0). Happy to add it as documentation of the post-condition, but I don't see a path where wait would return with a non-zero value - would appreciate a walk-through if I'm missing something.

src/ros2_medkit_integration_tests/test/features/test_openapi_response_drift.test.py
bburda: Fixed in dbcd0109: `_fetch_spec` now polls until `len(paths) >= MIN_EXPECTED_PATHS` (=30, ~50 in the live spec). A slow plugin registration can no longer let the test read back a partial spec.

src/ros2_medkit_integration_tests/test/features/test_openapi_response_drift.test.py
bburda: Fixed in dbcd0109: class docstring + `test_get_responses_match_declared_schema` docstring updated to spell out GET-only scope and reference the follow-up issue for the all-verbs extension.

src/ros2_medkit_integration_tests/test/features/test_openapi_response_drift.test.py
bburda: Fixed in dbcd0109: dropped the `test_entity` fallback. `_substitute_path_params` now raises `_MissingEntityType` when a path needs an unknown type; the loop catches it, skips the endpoint, and prints the gap in test output (currently 26 area-scoped paths surface in this fixture - the runtime_only discovery mode does not synthesise areas). Added unit tests for the helper covering happy path, resource_id substitution, and the raise-on-missing-type contract.

src/ros2_medkit_gateway/include/ros2_medkit_gateway/ros2_common/ros2_subscription_executor.hpp
bburda: Doc only - dbcd0109: rewrote the warning to describe the exact deadlock surface (worker blocks on `cv.wait`, decrement never runs because the wrapper is the same call frame) and the symptom (hung worker, no log line).

Runtime detection would need `thread_local` TLS, which is banned in code linked into the gateway plugin MODULE per CLAUDE.md (TPOFF32 / initial-exec TLS incompatible with shared objects). The detection has to be *precise* - only triggers on self-removal of the currently-running token, not any `remove_graph_change` call from the worker thread - because subscription callbacks and posted tasks legitimately run on the worker too. Without TLS I don't see how to distinguish those cases. Open to alternatives if you see one.

src/ros2_medkit_gateway/src/openapi/schema_builder.cpp
bburda: Fixed in dbcd0109: rewrote the comment to spell out that `phase` is required *only when `x-medkit` is present*, not unconditionally; flagged that dropping `x-medkit` from the parent properties would orphan the inner `required` and need a revisit.

mfaferek93 (approved):
+1

PR #392, opened 2026-04-27
refactor(gateway): introduce core/ layer scaffolding for provider injection

### LINKED ISSUE #391: Split gateway sources into a middleware-neutral core layer for independent testing and structural enforcement
## Summary

Today the gateway package compiles all of its source files into a single `gateway_lib` static library that links `rclcpp`, `rcl_interfaces`, action messages, `ros2_medkit_msgs`, and `ros2_medkit_serialization`. Most of that source — HTTP routing, request handlers, JWT authentication, fault model (debounce, SQLite storage, correlation), peer aggregation (mDNS discovery, entity merger, SSE proxy), the manifest parser, the entity cache, the lock/bulk-data/subscription/script/update managers, the OpenAPI builder, and the provider interface contracts — has no inherent dependency on ROS. Mixing the two in one library produces several friction points:

- **Testing requires a full ROS environment.** Every gateway unit test brings in `ament_target_dependencies` and pulls `rclcpp` onto the link line, even when the code under test (e.g. `auth_manager`, `fault_storage`, `entity_merger`, `peer_client`) doesn't touch ROS APIs.
- **Slower iteration on business logic.** The precompiled-header set bundles `rclcpp/rclcpp.hpp`, so any change in a neutral header invalidates a PCH that includes ROS headers and triggers larger rebuilds than necessary.
- **No structural enforcement** that new neutral code stays neutral — there is no CI guard preventing somebody from quietly adding `#include <rclcpp/rclcpp.hpp>` inside the JWT manager or the SQLite trigger store.
- **Provider abstractions are blurred.** The `DataProvider` / `OperationProvider` / `FaultProvider` / `LogProvider` / `IntrospectionProvider` interfaces are pure C++17, but they sit alongside their ROS-specific default implementations (e.g. `Ros2TopicDataProvider`) under the same include root, making the interface/implementation boundary harder to reason about.

---

Split the package's sources into two named layers, both inside the existing `ros2_medkit_gateway` package, without splitting into a separate colcon package yet:

- `core/` — middleware-neutral business logic. Compiles into a new `gateway_core` static library that links only header-only and C-level externals (cpp-httplib, nlohmann/json, yaml-cpp, tl::expected, jwt-cpp, OpenSSL, SQLite, dl). No `rclcpp`, no `rcl_interfaces`, no message-package transitive includes. Enforced by:
  - a grep-based linter (`scripts/check_core_purity.sh` wired as `gateway_core_purity` CTest);
  - a link-time smoke test (`test_gateway_core_smoke`) that compiles a translation unit including a sampling of `core/` headers and links exclusively against `gateway_core` + GTest, with no `ament_target_dependencies`. A regression in either guard fails the build.
- The remainder of the source tree — `gateway_node`, the ROS-coupled managers, `runtime_discovery`, `ros2_topic_data_provider`, the trigger subscribers, etc. — stays at its current path and forms a `gateway_ros2` static library that publicly links `gateway_core`.

The `gateway_node` executable and existing tests link `gateway_ros2` so they transitively get both layers; nothing changes for downstream consumers of the gateway. Headers that move acquire a one-line forwarding shim at their old paths so out-of-tree plugins continue to compile without modification.

This issue tracks only the structural relocation. Subsequent issues will cover the per-manager refactors (data access, operation, configuration, fault facade, log, trigger) that introduce provider injection so the managers themselves move into `core/`.

---

Outcome metrics expected after the relocation:
- 35 source files compile via `gateway_core` without any ROS link dependency.
- Two structural CI guards: `gateway_core_purity` (grep) and `gateway_core_smoke` (link-time).
- PCH split: `gateway_core` PCH contains std + json + httplib + tl::expected; `gateway_ros2` PCH adds rclcpp/rclcpp.hpp on top.
- Backwards-compat shim headers at every legacy path so out-of-tree consumers (graph_provider, opcua, sovd_service_interface, linux_introspection, param_beacon, topic_beacon) compile unchanged.
- Existing test suite (~2500 unit, ~3200 integration, ~2600 clang-tidy targets) stays green.

bburda:
# Pull Request

## Summary

Splits `src/ros2_medkit_gateway/` into a middleware-neutral `core/` layer and the existing top-level layout (ROS adapters), introducing two named build targets and two structural CI guards as outlined in the linked issue. Pure structural change — no business logic edits, no signature changes, no public API changes for consumers. Subsequent issues will cover the per-manager refactors (data access, operation, configuration, fault facade, log, trigger) that introduce provider injection so the managers themselves move into `core/`.

What changed:

- 35 source files relocated into `src/core/` (`include/ros2_medkit_gateway/core/` for headers). New `gateway_core` STATIC library compiles them with header-only and C-level externals (cpp-httplib, nlohmann/json, yaml-cpp, tl::expected, jwt-cpp, OpenSSL, SQLite, dl) and zero ROS dependencies.
- The remaining 56 ROS-coupled sources stay at their current paths and form a new `gateway_ros2` STATIC library that publicly links `gateway_core`. `gateway_node` and the existing test targets link `gateway_ros2`, so they transitively get both layers.
- Every `#include` path inside the package rewritten via sed against the rename history of the relocation commits — pure mechanical edit, no other source changes.
- `compat.hpp` is reserved for the eventual `namespace ros2_medkit_gateway = medkit;` alias once production code has been file-by-file converted to `namespace medkit::`. Adding the alias today would be ill-formed in any TU that subsequently opens `namespace ros2_medkit_gateway { ... }` (which every gateway header currently does), so until the migration completes the header is a documented no-op.
- Backwards-compat shim headers at every legacy path (`providers/`, `plugins/`, `discovery/models/`, `http/error_codes.hpp`, `discovery/introspection_provider.hpp`) so out-of-tree consumers (graph_provider, opcua, sovd_service_interface, linux_introspection, param_beacon, topic_beacon) compile without modification.
- `scripts/check_core_purity.sh` greps `core/` for any direct ROS include and is wired as the `gateway_core_purity` CTest under the `linter` label. Smoke-tested by planting a `#include <rclcpp/rclcpp.hpp>` — correctly rejected.
- `test_gateway_core_smoke` compiles a translation unit that includes a sampling of `core/` headers and links exclusively against `gateway_core` + GTest, with no `ament_target_dependencies`. Catches transitive ROS coupling that the grep guard might miss.
- PCH split: `gateway_core` PCH covers std + `nlohmann/json.hpp` + `httplib.h` + `tl/expected.hpp`; `gateway_ros2` PCH adds `rclcpp/rclcpp.hpp` on top.
- Pre-emptive clang-tidy fixes (`cppcoreguidelines-special-member-functions` Rule of Five, `bugprone-exception-escape` on `main`, `performance-unnecessary-value-param`) surfaced by the new build setup.
- Bonus: removes a timing race in `test_combined_introspection.test_01_procfs_returns_200_on_host` where a freshly forked process could momentarily report `RssAnon=0` before any memory was touched. The poll helper now waits for `rss_bytes > 0` before asserting.

---

## Issue

- closes #391

---

## Type

- [ ] Bug fix
- [x] New feature or tests
- [ ] Breaking change
- [ ] Documentation only

---

## Testing

Local validation on Jazzy:

| Stage | Result |
|---|---|
| `colcon build --symlink-install` (15 packages) | clean (~4 min) |
| `./scripts/test.sh unit` | 2561 tests, 0 errors, 0 failures |
| `./scripts/test.sh lint --packages-select ros2_medkit_gateway` | 3494 tests, 0 errors, 0 failures (includes the new `gateway_core_purity` linter) |
| `./scripts/test.sh integ --packages-select ros2_medkit_integration_tests` | 3211 tests, 0 errors, 0 failures |
| `./scripts/test.sh tidy --packages-select ros2_medkit_gateway` | 2636 tests, 0 errors, 0 failures (clang-tidy run, ~32 min) |
| `colcon test --ctest-args -R gateway_core_smoke` | passes — compiles and links against `gateway_core` alone with no `ament_target_dependencies` |
| `gateway_core_purity` planted-violation smoke test | a planted `#include <rclcpp/rclcpp.hpp>` in `core/` was correctly rejected |

Reviewers should verify: (1) `gateway_core_purity` runs as part of the linter job, (2) `test_gateway_core_smoke` builds and runs without ament dependencies, (3) downstream packages that consume the gateway's headers still compile via the shim layer.

---

## Checklist

- [x] Breaking changes are clearly described (and announced in docs / changelog if needed) — none, public API and namespaces preserved via shim headers; the `medkit::` namespace migration is deferred to a follow-up so `compat.hpp` ships as a documented no-op
- [x] Tests were added or updated if needed — `gateway_core_purity` linter, `test_gateway_core_smoke` link test, and a polling fix in `test_combined_introspection`
- [ ] Docs were updated if behavior or public API changed — no public API change; CHANGELOG entries are added at release time per project convention


src/ros2_medkit_gateway/include/ros2_medkit_gateway/compat.hpp
```diff
@@ -0,0 +1,29 @@
...
+//   namespace medkit {}
+//   namespace ros2_medkit_gateway = medkit;
+// would be ill-formed in any translation unit that subsequently opens
+// `namespace ros2_medkit_gateway { ... }` (which every gateway header
+// currently does). The migration to namespace medkit:: must therefore
+// happen file-by-file in a dedicated follow-up; this header will then
+// be repurposed to host the alias once all production code has been
+// converted.
+//
+// Until then, including this header is a no-op.
```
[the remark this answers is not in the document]
bburda: Out of scope for this PR. The linked issue scopes this change to the structural relocation only and explicitly defers the namespace migration: `compat.hpp` is reserved for the eventual `namespace ros2_medkit_gateway = medkit;` alias once production code has been file-by-file converted, and is intentionally a documented no-op until then (see commit 7b0a134e). The PR description has been updated to reflect this.

src/ros2_medkit_gateway/test/test_gateway_core_smoke.cpp
```diff
@@ -0,0 +1,79 @@
...
+#include "ros2_medkit_gateway/core/discovery/models/function.hpp"
+#include "ros2_medkit_gateway/core/providers/data_provider.hpp"
+#include "ros2_medkit_gateway/core/providers/fault_provider.hpp"
+#include "ros2_medkit_gateway/core/providers/host_info_provider.hpp"
+#include "ros2_medkit_gateway/core/providers/introspection_provider.hpp"
+#include "ros2_medkit_gateway/core/providers/log_provider.hpp"
+#include "ros2_medkit_gateway/core/providers/operation_provider.hpp"
+#include "ros2_medkit_gateway/core/providers/script_provider.hpp"
+#include "ros2_medkit_gateway/core/providers/update_provider.hpp"
+
```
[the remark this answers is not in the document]
bburda: Out of scope for this PR. The linked issue scopes the smoke test as `a sampling of core/ headers` and explicitly defers the per-manager and provider-injection refactors to subsequent issues. Adding `core/http/rest_server.hpp` or `core/http/handlers/handlers.hpp` to this TU would not strengthen it the way you describe - both transitively pull in `HandlerContext`, which holds an `rclcpp::Node *` and lives outside `core/` for that reason. Splitting `HandlerContext` into a neutral interface plus a ROS-only implementation is part of the follow-up work, after which those headers can join this smoke TU.

docs/tutorials/plugin-system.rst
```diff
@@ -148,7 +148,7 @@ implements the provider interface, even if the class inherits from it.
 .. code-block:: cmake
 
    add_library(my_plugin MODULE src/my_plugin.cpp)
-   target_link_libraries(my_plugin gateway_lib)
+   target_link_libraries(my_plugin gateway_ros2)
```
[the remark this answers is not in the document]
bburda: Fixed in commit 4a9c1d52. The tutorial now uses `medkit_target_dependencies(my_plugin ros2_medkit_gateway)` for headers and `-Wl,--unresolved-symbols=ignore-all` for runtime symbol resolution, with an explicit note not to link `gateway_ros2` directly. Matches the in-tree pattern in `ros2_medkit_graph_provider/CMakeLists.txt`.

src/ros2_medkit_gateway/include/ros2_medkit_gateway/core/plugins/plugin_context.hpp
```diff
@@ -0,0 +1,282 @@
...
+ * the gateway's thread-safe caches internally.
+ */
+class PluginContext {
+ public:
+  virtual ~PluginContext() = default;
+
+  // ---- ROS 2 access (replaces set_node) ----
+
+  /// Get ROS 2 node pointer for subscriptions, service clients, etc.
+  virtual rclcpp::Node * node() const = 0;
```
mfaferek93: Exposing `rclcpp::Node *` here (with the `namespace rclcpp { class Node; }` forward decl at line 33) leaks ROS coupling into core/ at the API signature level. `check_core_purity.sh` regex only matches `#include` directives, not forward decls or signatures, so neither CI guard catches it. Cleanest fix: move `node()` into a `RosPluginContext : PluginContext` subclass living in gateway_ros2.

src/ros2_medkit_gateway/scripts/check_core_purity.sh
```diff
@@ -0,0 +1,48 @@
...
+# Run from the package root directory.
+
+set -euo pipefail
+
+CORE_DIRS=(
+  "include/ros2_medkit_gateway/core"
+  "src/core"
+)
+
+FORBIDDEN_PATTERN='#include[[:space:]]*[<"](rclcpp|rcl_interfaces|rosidl|action_msgs|std_msgs|sensor_msgs|geometry_msgs|builtin_interfaces|ros2_medkit_msgs|rcutils|rmw)[/_]'
```
mfaferek93: Regex misses several ROS 2 prefixes likely to drift into core/ later: `lifecycle_msgs`, `tf2`/`tf2_ros`, `*_msgs` siblings (`nav_msgs`, `diagnostic_msgs`, `trajectory_msgs`, `visualization_msgs`), `rcpputils`, `pluginlib`, `class_loader`, `rosbag2_*`. None used today, so preventive only - but an incomplete denylist defeats the linter's purpose.

src/ros2_medkit_gateway/test/test_gateway_core_smoke.cpp
```diff
@@ -0,0 +1,79 @@
...
+#include "ros2_medkit_gateway/core/discovery/models/component.hpp"
+#include "ros2_medkit_gateway/core/discovery/models/function.hpp"
+#include "ros2_medkit_gateway/core/providers/data_provider.hpp"
+#include "ros2_medkit_gateway/core/providers/fault_provider.hpp"
+#include "ros2_medkit_gateway/core/providers/host_info_provider.hpp"
+#include "ros2_medkit_gateway/core/providers/introspection_provider.hpp"
+#include "ros2_medkit_gateway/core/providers/log_provider.hpp"
+#include "ros2_medkit_gateway/core/providers/operation_provider.hpp"
+#include "ros2_medkit_gateway/core/providers/script_provider.hpp"
+#include "ros2_medkit_gateway/core/providers/update_provider.hpp"
```
mfaferek93, line 37: Smoke test excludes the core header with the highest signature-level coupling risk: `core/plugins/plugin_context.hpp` (forward-declares `rclcpp::Node` and exposes it via virtual method). Also missing: all `core/managers/*` and `core/aggregation/*` headers. Adding 4-5 more includes here would surface forward-decl-style leaks at link time.

src/ros2_medkit_gateway/include/ros2_medkit_gateway/core/plugins/plugin_context.hpp
bburda: Addressed in commit ce1f4503. `node()` and the `namespace rclcpp { class Node; }` forward decl are gone from `core/plugins/plugin_context.hpp`. The new `include/ros2_medkit_gateway/plugins/ros_plugin_context.hpp` (in gateway_ros2, not core) defines `class RosPluginContext : public PluginContext` with `virtual rclcpp::Node * node() const = 0;` plus the moved `make_gateway_plugin_context` factory and an `as_ros_plugin_context` helper. The gateway always passes a RosPluginContext through `GatewayPlugin::set_context`, so in-tree plugins keep the existing override signature and only change their member type from `PluginContext *` to `RosPluginContext *` plus a one-line `as_ros_plugin_context(context)` cast in set_context. The legacy `plugins/plugin_context.hpp` shim re-exports both, so out-of-tree consumers compile unchanged. Smoke test now also pins `PluginContext` so the regression cannot land unnoticed.

src/ros2_medkit_gateway/scripts/check_core_purity.sh
bburda: Addressed in commit ce1f4503. The denylist now covers all the prefixes you listed plus generic `*_msgs` / `*_interfaces` patterns so any future sibling drops in unannounced (`nav_msgs`, `diagnostic_msgs`, `trajectory_msgs`, `visualization_msgs`, `lifecycle_msgs`), and we also catch `tf2`/`tf2_ros`/`tf2_msgs`/`tf2_geometry_msgs`, `rosbag2_*`, `pluginlib`, `class_loader`, `rcpputils`, `ament_index_cpp`/`ament_index_python`. Sanity-tested by planting one violating include for each new prefix - all five rejected with FORBIDDEN ROS INCLUDE.

src/ros2_medkit_gateway/test/test_gateway_core_smoke.cpp
bburda: Addressed in commit ce1f4503. `core/plugins/plugin_context.hpp` is now in the smoke TU together with `core/managers/{bulk_data_store,lock_manager}.hpp` and `core/aggregation/{entity_merger,network_utils}.hpp`, with `static_assert` references on `PluginContext`, `BulkDataStore`, `LockManager`, and `EntityMerger` so the includes are exercised at compile time. The TU still links against `gateway_core` alone, so any forward-decl-style or signature-level rclcpp creep into these headers fails at link time. `rest_server.hpp` and `http/handlers/handlers.hpp` are intentionally left out for now: both transitively pull in HandlerContext, which carries an `rclcpp::Node *` and is part of the next-stage decoupling tracked in the follow-up provider-injection issues per the linked issue's scope.

mfaferek93 (approved):

Issue #380, opened 2026-04-17
Global /faults/stream SSE events lack entity context for downstream consumers


mfaferek93:
## Problem

Consumers subscribing to the gateway-wide `/api/v1/faults/stream` receive a fault payload (`fault_code`, `severity`, `reporting_sources`, `description`, ...) but no explicit SOVD entity identifier. `reporting_sources` contains ROS node names which follow a different convention than SOVD entity IDs (for example `/bridge/diagnostic_bridge` vs `diagnostic-bridge`), so a consumer that wants to pull the associated `bulk-data/rosbags/{fault_code}` has to enumerate apps + components and HEAD-probe each one until a 200 comes back.

Surfaced while building the first external integration (bridge to an Apache Arrow Flight data platform): see `resolve_entity_for_download()` in the demo bridge, which does O(N) HEAD probes on apps + components as the pragmatic workaround. That works fine at ~10 entities but is wasteful latency-wise and conceptually wrong - the gateway already knows which entity owns the bag, it just doesn't tell the SSE consumer.

### A. x-medkit extension in SSE payload

Add `x-medkit-entity-type` and `x-medkit-entity-id` fields inside the SSE fault event JSON. The consumer reads them directly and issues one request to `bulk-data`.

- Pros: zero extra round-trips, backward compatible (vendor extension, existing SOVD consumers ignore it).
- Cons: introduces a new `x-medkit-*` extension just as #140 proposes moving away from them to dedicated REST resources.

### B. Per-entity fault streams

Consumers subscribe to `/api/v1/{entity_type}/{id}/faults/stream` instead of the gateway-wide stream. Entity context is implicit in the URL path.

- Pros: SOVD-native shape, aligns with the direction of #140.
- Cons: consumer needs to enumerate entities upfront (but that's a one-time discovery, not per-fault). For a global 'catch everything across the gateway' use case the consumer has to fan out into N subscriptions.

## Related

- #140 - Replace x-medkit response extensions with dedicated REST resources (tension with option A)
- #248 / #270 - x-medkit-graph extension precedent (shows option A is workable if we accept more x-medkit)
- #325 - Rethink runtime discovery entity mapping to align with SOVD entity model

## Ask

Confirm a direction (A, B, or a third option) so the demo bridge can replace the HEAD-probe workaround with something the gateway explicitly supports.

bburda:
Worth surfacing a related symptom of the same root cause we hit during a synthetic-component audit:

[...]

PR #394, opened 2026-04-28
fix(gateway): aggregate component /logs and /bulk-data from hosted apps' fqns

### LINKED ISSUE #393: [BUG] Synthetic components return empty results from /logs and /bulk-data resource collections
# Bug report

### Steps to reproduce

1. Launch the gateway with at least one runtime-discovered ROS 2 node grouped into a synthetic / runtime-discovered component (any node whose component is not declared in a manifest):
   ```
   ros2 launch ros2_medkit_gateway gateway.launch.py
   ros2 launch ros2_medkit_integration_tests demo_nodes.launch.py
   ```
2. Identify the synthetic component that hosts a demo app:
   ```
   curl -s localhost:8080/api/v1/components | jq '.items[].id'
   COMP_ID=...  # the one whose /hosts contains a demo app
   curl -s localhost:8080/api/v1/components/$COMP_ID/hosts
   ```
3. Query log entries for that component:
   ```
   curl -s "localhost:8080/api/v1/components/$COMP_ID/logs?severity=debug" | jq
   ```
4. Query bulk-data rosbag descriptors for the same component (after a fault triggers a capture):
   ```
   curl -s "localhost:8080/api/v1/components/$COMP_ID/bulk-data/rosbags" | jq
   ```

### Expected behavior

Both endpoints return entries aggregated from the apps hosted by the component, with `x-medkit.aggregation_level=component` and a non-empty `items` list (mirroring how `/areas/{id}/logs` and `/functions/{id}/logs` already aggregate).

### Actual behavior

Both endpoints return `{"items": []}` even when the component groups N apps that have logs / rosbags. No error, no warning - just a quiet empty response.

Root cause is the same in both handlers: synthetic / runtime-discovered components have an empty `fqn` *and* an empty `namespace_path`, while manifest-defined components have non-empty values.

- `LogHandlers::handle_get_logs` calls `LogManager::get_logs({entity.fqn}, prefix_match=true, ...)` for the COMPONENT branch. With an empty fqn, the prefix-match against per-node logger name buffers selects nothing and returns zero items. AREA and FUNCTION already aggregate via hosted apps' `effective_fqn()`, so they work correctly; COMPONENT is the only entity type still on the legacy path.
- `BulkDataHandlers::get_source_filters` has a special case for FUNCTION (`cache.get_apps_for_function()` + `effective_fqn()`) but falls through to `entity.fqn.empty() ? entity.namespace_path : entity.fqn` for COMPONENT. Both fields empty → returns `{}` source filters, so `handle_list_descriptors` (rosbag listing) and the ownership check in `handle_download` get nothing to work with.

### Environment

- ros2_medkit version: main (reproduces on every recent release; introduced when synthetic components became normal in #325)
- ROS 2 distro: any (Jazzy / Humble / Rolling)
- OS: any

### Additional information

Fix mirrors the AREA / FUNCTION pattern already in both handlers - resolve hosted apps via `cache.get_apps_for_component()`, build their `effective_fqn()` list, fall through to the namespace-prefix path only when the component has no hosted apps but declares a non-empty namespace (manifest deployments where a component groups topics rather than nodes).

A related but distinct symptom on the per-entity faults handler is tracked separately - see [comment on #380](https://github.com/selfpatch/ros2_medkit/issues/380#issuecomment-4336673964) for details. That one is a scope-leak rather than an empty-result bug and needs a different fix (changing the GetFault service contract), so it is not in scope here.

bburda:
## Summary

Fix two handlers that silently returned empty results for synthetic / runtime-discovered components, by mirroring the AREA / FUNCTION aggregation pattern already present in the same files.

- ``LogHandlers::handle_get_logs`` - the COMPONENT branch fed ``entity.fqn`` to ``LogManager::get_logs(..., prefix_match=true)``. Synthetic components have an empty ``fqn``, so the prefix match selected nothing and the response was an empty ``items`` list even when the component grouped N active apps. Resolve hosted apps via ``cache.get_apps_for_component()`` and call ``get_logs(host_fqns, prefix_match=false)``. Manifest deployments with a non-empty ``fqn`` and no hosted apps fall through to the original namespace-prefix path. Response now also carries ``x-medkit.aggregation_level=component``, ``app_count``, and ``aggregation_sources`` for parity with AREA and FUNCTION.
- ``BulkDataHandlers::get_source_filters`` - identical pattern. The handler had a FUNCTION-only aggregation arm and fell through to ``entity.fqn`` / ``entity.namespace_path`` for COMPONENT. Both empty for synthetic components, so source filters were ``{}``, descriptor lists came back empty, and ownership checks in ``handle_download`` short-circuited. Extended the aggregation arm to handle COMPONENT identically; manifest-only components still fall through to the namespace prefix path.
- ``docs/api/rest.rst`` updated to describe the actual aggregation behavior for ``/components/{id}/logs`` instead of the stale ``namespace prefix match`` wording.

### Tests

- New integration test ``test_component_get_logs_aggregates_child_apps`` in ``test_logging_api.test.py`` verifies a synthetic component returns non-empty ``items`` plus the new aggregation metadata. Asserts ``len(items) > 0`` directly because the original bug returned 0 items with otherwise valid ``x-medkit`` metadata, so checking ``app_count`` alone is not sufficient. Matches ``aggregation_sources`` by FQN substring instead of bare app id.
- New integration test ``test_bulk_data_component_aggregates_child_apps`` in ``test_bulk_data_api.test.py`` verifies the equivalent fix for the bulk-data endpoint.

### SOVD API impact

- No changes to SOVD-defined request paths, response schemas, or error codes.
- Additive ``x-medkit`` extensions on ``/components/{id}/logs`` only - new optional ``aggregation_level``, ``aggregated``, ``app_count``, ``aggregation_sources`` fields, emitted only when local aggregation is active. Existing clients that ignore unknown ``x-medkit.*`` fields are unaffected. ``/components/{id}/bulk-data/rosbags`` payload schema is unchanged.

---

## Issue

- closes #393

A related but distinct symptom on ``handle_get_fault`` for synthetic components is a scope-leak that needs a different fix (changing the GetFault service contract), tracked separately - see comment on #380.

---

## Type

- [x] Bug fix
- [x] New feature or tests (two new integration tests + ``x-medkit`` aggregation metadata on ``/components/{id}/logs``)
- [ ] Breaking change
- [x] Documentation only (``docs/api/rest.rst`` log endpoints description)

---

## Testing

All local runs green:

- ``./scripts/test.sh unit --packages-select ros2_medkit_gateway`` - 2099 assertions across 81 tests, 0 failures.
- ``./scripts/test.sh test_logging_api --packages-select ros2_medkit_integration_tests`` - 22 logging tests + shutdown, 0 failures. Includes the new ``test_component_get_logs_aggregates_child_apps``.
- ``./scripts/test.sh test_bulk_data_api --packages-select ros2_medkit_integration_tests`` - 10 tests, 0 failures. Includes the new ``test_bulk_data_component_aggregates_child_apps``.
- ``./scripts/test.sh lint --packages-select ros2_medkit_gateway ros2_medkit_integration_tests`` - clean (clang-format, cmake-lint, ament-copyright, flake8, pep257, cppcheck, xmllint).
- pre-push clang-tidy on changed files - clean.

---

## Checklist

- [x] Breaking changes are clearly described (there are none; only additive ``x-medkit`` extensions on ``/components/{id}/logs``)

### DIFF
--- a/docs/api/rest.rst
+++ b/docs/api/rest.rst
@@ -737,7 +737,8 @@ Logs Endpoints
 --------------
 
 Query and configure the /rosout ring buffer for an entity. Supported entity types:
-**areas** (namespace prefix match), **components** (namespace prefix match), **apps** (exact FQN match),
+**areas** (aggregated from hosted apps, namespace prefix fallback), **components** (aggregated from
+hosted apps, namespace prefix fallback for manifest-only deployments), **apps** (exact FQN match),
 and **functions** (aggregated from hosted apps).
 
 .. note::
@@ -748,7 +749,11 @@ and **functions** (aggregated from hosted apps).
    storage backend or take full ownership of the log pipeline (see plugin development docs).
 
 ``GET /api/v1/components/{id}/logs``
-   Query log entries for all nodes in the component namespace (prefix match).
+   Query log entries aggregated from the component's hosted apps. Resolves child apps via
+   the entity cache and queries each by exact FQN. Falls back to namespace prefix match only
+   when the component has no hosted apps but declares a non-empty namespace (manifest-only
+   deployments where the component groups topics rather than nodes). The response carries
+   ``x-medkit.aggregation_level=component``, ``app_count``, and ``aggregation_sources``.
 
 ``GET /api/v1/apps/{id}/logs``
    Query log entries for the specific app node (exact match).
--- a/docs/api/rest.rst
+++ b/docs/api/rest.rst
@@ -752,8 +752,10 @@ and **functions** (aggregated from hosted apps).
    Query log entries aggregated from the component's hosted apps. Resolves child apps via
    the entity cache and queries each by exact FQN. Falls back to namespace prefix match only
    when the component has no hosted apps but declares a non-empty namespace (manifest-only
-   deployments where the component groups topics rather than nodes). The response carries
-   ``x-medkit.aggregation_level=component``, ``app_count``, and ``aggregation_sources``.
+   deployments where the component groups topics rather than nodes). The response always
+   carries ``x-medkit.aggregation_level=component`` and ``aggregated=true``; the
+   ``app_count`` and ``aggregation_sources`` fields are populated only when hosted-app
+   aggregation is active and are omitted under the namespace-prefix fallback.
 
 ``GET /api/v1/apps/{id}/logs``
    Query log entries for the specific app node (exact match).
--- a/src/ros2_medkit_gateway/include/ros2_medkit_gateway/http/handlers/bulkdata_handlers.hpp
+++ b/src/ros2_medkit_gateway/include/ros2_medkit_gateway/http/handlers/bulkdata_handlers.hpp
@@ -112,26 +112,18 @@ class BulkDataHandlers {
    */
   static std::string get_rosbag_mimetype(const std::string & format);
 
+ private:
+  HandlerContext & ctx_;
+
   /**
    * @brief Get source filters for rosbag queries based on entity type.
    *
-   * - APP / AREA: returns the entity's FQN or namespace path (single filter).
-   * - FUNCTION: aggregates non-empty effective_fqn() values across all hosted apps
-   *   (no fallback - functions are pure aggregated views).
-   * - COMPONENT: aggregates from hosted apps; falls back to FQN/namespace_path only
-   *   when the component has no hosted apps (manifest deployments where the component
-   *   groups topics rather than nodes). This avoids the synthetic-component bug where
-   *   empty fqn + empty namespace_path produced zero source filters.
-   *
-   * Public to enable unit testing the entity-type branching directly.
-   *
-   * @param entity Entity information
-   * @return Vector of source filter strings (empty if no valid filters)
+   * Thin instance wrapper that fetches the cache from ctx_ and delegates to
+   * detail::compute_bulkdata_source_filters. The pure logic (entity-type
+   * branching) is unit-tested via the free function instead of the member
+   * to keep the handler's public surface unchanged.
    */
   std::vector<std::string> get_source_filters(const EntityInfo & entity) const;
-
- private:
-  HandlerContext & ctx_;
 
   /**
    * @brief Stream file contents to HTTP response.
@@ -154,6 +146,34 @@ class BulkDataHandlers {
    */
   static std::string resolve_rosbag_file_path(const std::string & path);
 };
+
+namespace detail {
+
+/**
+ * @brief Compute rosbag source filters for an entity based on its type.
+ *
+ * Pure helper that drives ``BulkDataHandlers::get_source_filters``. Lives in
+ * a ``detail`` namespace to signal "not part of the public API" while still
+ * being directly unit-testable without spinning up a ``GatewayNode``.
+ *
+ * - APP / AREA: returns the entity's FQN or namespace path (single filter).
+ * - FUNCTION: aggregates non-empty ``effective_fqn()`` values across all
+ *   hosted apps (no fallback - functions are pure aggregated views).
+ * - COMPONENT: aggregates from hosted apps; falls back to FQN/namespace_path
+ *   only when the component has no hosted apps (manifest deployments where
+ *   the component groups topics rather than nodes). This avoids the
+ *   synthetic-component bug where empty fqn + empty namespace_path produced
+ *   zero source filters.
+ *
+ * @param cache Entity cache to resolve hosted apps in (used for FUNCTION /
+ *              COMPONENT only)
+ * @param entity Entity information
+ * @return Vector of source filter strings (empty if no valid filters)
+ */
+std::vector<std::string> compute_bulkdata_source_filters(const ThreadSafeEntityCache & cache,
+                                                         const EntityInfo & entity);
+
+}  // namespace detail
 
 }  // namespace handlers
 }  // namespace ros2_medkit_gateway
--- a/src/ros2_medkit_gateway/include/ros2_medkit_gateway/http/handlers/bulkdata_handlers.hpp
+++ b/src/ros2_medkit_gateway/include/ros2_medkit_gateway/http/handlers/bulkdata_handlers.hpp
@@ -17,6 +17,7 @@
 #include <httplib.h>
 
 #include <string>
+#include <vector>
 
 #include "ros2_medkit_gateway/http/handlers/handler_context.hpp"
 
@@ -111,20 +112,26 @@ class BulkDataHandlers {
    */
   static std::string get_rosbag_mimetype(const std::string & format);
 
- private:
-  HandlerContext & ctx_;
-
   /**
    * @brief Get source filters for rosbag queries based on entity type.
    *
-   * For apps/components/areas: returns the entity's FQN or namespace path.
-   * For functions: aggregates FQNs from all hosting apps (read-only
-   * aggregated view - upload/delete are blocked at the route level).
+   * - APP / AREA: returns the entity's FQN or namespace path (single filter).
+   * - FUNCTION: aggregates non-empty effective_fqn() values across all hosted apps
+   *   (no fallback - functions are pure aggregated views).
+   * - COMPONENT: aggregates from hosted apps; falls back to FQN/namespace_path only
+   *   when the component has no hosted apps (manifest deployments where the component
+   *   groups topics rather than nodes). This avoids the synthetic-component bug where
+   *   empty fqn + empty namespace_path produced zero source filters.
+   *
+   * Public to enable unit testing the entity-type branching directly.
    *
    * @param entity Entity information
    * @return Vector of source filter strings (empty if no valid filters)
    */
   std::vector<std::string> get_source_filters(const EntityInfo & entity) const;
+
+ private:
+  HandlerContext & ctx_;
 
   /**
    * @brief Stream file contents to HTTP response.
--- a/src/ros2_medkit_gateway/src/http/handlers/bulkdata_handlers.cpp
+++ b/src/ros2_medkit_gateway/src/http/handlers/bulkdata_handlers.cpp
@@ -539,10 +539,16 @@ std::string BulkDataHandlers::get_rosbag_mimetype(const std::string & format) {
 }
 
 std::vector<std::string> BulkDataHandlers::get_source_filters(const EntityInfo & entity) const {
+  return detail::compute_bulkdata_source_filters(ctx_.node()->get_thread_safe_cache(), entity);
+}
+
+namespace detail {
+
+std::vector<std::string> compute_bulkdata_source_filters(const ThreadSafeEntityCache & cache,
+                                                         const EntityInfo & entity) {
   if (entity.type == EntityType::FUNCTION) {
     // Functions are pure aggregated views over hosted apps - if no apps host the function,
     // there is nothing to query. No fall-through to fqn/namespace_path.
-    const auto & cache = ctx_.node()->get_thread_safe_cache();
     return HandlerContext::resolve_app_host_fqns(cache, cache.get_apps_for_function(entity.id));
   }
 
@@ -552,7 +558,6 @@ std::vector<std::string> BulkDataHandlers::get_source_filters(const EntityInfo &
     // empty descriptor lists plus failed ownership checks on download. Resolve hosted
     // apps first; manifest deployments where the component groups topics rather than
     // nodes still need the namespace prefix path, so fall through if no apps host it.
-    const auto & cache = ctx_.node()->get_thread_safe_cache();
     auto filters = HandlerContext::resolve_app_host_fqns(cache, cache.get_apps_for_component(entity.id));
     if (!filters.empty()) {
       return filters;
@@ -567,6 +572,8 @@ std::vector<std::string> BulkDataHandlers::get_source_filters(const EntityInfo &
   }
   return {filter};
 }
+
+}  // namespace detail
 
 }  // namespace handlers
 }  // namespace ros2_medkit_gateway
--- a/src/ros2_medkit_gateway/src/http/handlers/bulkdata_handlers.cpp
+++ b/src/ros2_medkit_gateway/src/http/handlers/bulkdata_handlers.cpp
@@ -539,33 +539,28 @@ std::string BulkDataHandlers::get_rosbag_mimetype(const std::string & format) {
 }
 
 std::vector<std::string> BulkDataHandlers::get_source_filters(const EntityInfo & entity) const {
-  // FUNCTION and COMPONENT both aggregate rosbags from their hosting apps.
-  // Synthetic / runtime-discovered components have an empty fqn / namespace_path,
-  // so falling through to the bare-fqn path silently returned zero source filters
-  // and produced empty descriptor lists plus failed ownership checks on download.
-  if (entity.type == EntityType::FUNCTION || entity.type == EntityType::COMPONENT) {
+  if (entity.type == EntityType::FUNCTION) {
+    // Functions are pure aggregated views over hosted apps - if no apps host the function,
+    // there is nothing to query. No fall-through to fqn/namespace_path.
     const auto & cache = ctx_.node()->get_thread_safe_cache();
-    auto host_app_ids = (entity.type == EntityType::FUNCTION) ? cache.get_apps_for_function(entity.id)
-                                                              : cache.get_apps_for_component(entity.id);
-    std::vector<std::string> filters;
-    filters.reserve(host_app_ids.size());
-    for (const auto & app_id : host_app_ids) {
-      auto app = cache.get_app(app_id);
-      if (app) {
-        auto fqn = app->effective_fqn();
-        if (!fqn.empty()) {
-          filters.push_back(fqn);
-        }
-      }
-    }
+    return HandlerContext::resolve_app_host_fqns(cache, cache.get_apps_for_function(entity.id));
+  }
+
+  if (entity.type == EntityType::COMPONENT) {
+    // Synthetic / runtime-discovered components have an empty fqn / namespace_path,
+    // so the bare-fqn path used to silently return zero source filters and produce
+    // empty descriptor lists plus failed ownership checks on download. Resolve hosted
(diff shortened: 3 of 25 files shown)


src/ros2_medkit_gateway/src/http/handlers/log_handlers.cpp
```diff
@@ -177,11 +177,71 @@ void LogHandlers::handle_get_logs(const httplib::Request & req, httplib::Respons
...
+  // when the component grouped 20+ active nodes.
   // -----------------------------------------------------------------------
-  const bool prefix_match = (entity.type == EntityType::COMPONENT);
+  if (entity.type == EntityType::COMPONENT) {
+    const auto & cache = ctx_.node()->get_thread_safe_cache();
+    const auto app_ids = cache.get_apps_for_component(entity_id);
+
+    std::vector<std::string> host_fqns;
+    host_fqns.reserve(app_ids.size());
+    for (const auto & app_id : app_ids) {
```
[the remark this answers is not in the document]
bburda: Done. Extracted the host-FQN resolution loop into a shared helper `HandlerContext::resolve_app_host_fqns` (now used by FUNCTION/AREA/COMPONENT branches in `log_handlers` and FUNCTION/COMPONENT in `bulkdata_handlers`) and added 7 `ResolveAppHostFqnsTest` cases pinning the silent-skip semantics (missing apps, empty effective_fqn, bound_fqn precedence). Also added 3 `ComponentLogs*` cases on the existing `AreaAggregationTest` fixture that exercise `handle_get_logs` end-to-end via the REST server: synthetic component aggregates from hosted apps with x-medkit metadata, manifest-only component falls through, empty component returns empty items without crashing.

src/ros2_medkit_integration_tests/test/features/test_logging_api.test.py
```diff
@@ -229,6 +229,43 @@ def test_component_get_logs_returns_200(self):
...
+                comp_id = c['id']
+                break
+        self.assertIsNotNone(comp_id, 'No component hosts temp_sensor')
+
+        data = self.get_json(f'/components/{comp_id}/logs?severity=debug')
+        self.assertIn('items', data)
+        # Items must be non-empty - the original bug was a silent empty list
+        # for synthetic components, so app_count alone is not sufficient.
+        self.assertGreater(len(data['items']), 0,
+                           'Component logs aggregation returned zero items')
```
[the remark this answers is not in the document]
bburda: Done. Wrapped the assertion in `poll_endpoint_until` with the same `lambda d: d if d.get("items") else None` predicate that `test_app_get_logs_has_items_after_startup` uses, plus a 15s timeout.

docs/api/rest.rst
```diff
@@ -748,7 +749,11 @@ and **functions** (aggregated from hosted apps).
    storage backend or take full ownership of the log pipeline (see plugin development docs).
 
 ``GET /api/v1/components/{id}/logs``
-   Query log entries for all nodes in the component namespace (prefix match).
+   Query log entries aggregated from the component's hosted apps. Resolves child apps via
+   the entity cache and queries each by exact FQN. Falls back to namespace prefix match only
+   when the component has no hosted apps but declares a non-empty namespace (manifest-only
+   deployments where the component groups topics rather than nodes). The response carries
+   ``x-medkit.aggregation_level=component``, ``app_count``, and ``aggregation_sources``.
```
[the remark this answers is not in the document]
bburda: Done. Updated to make the field semantics explicit: the response always carries `aggregation_level=component` and `aggregated=true`; `app_count` and `aggregation_sources` are populated only when hosted-app aggregation is active and are omitted under the namespace-prefix fallback.

src/ros2_medkit_gateway/src/http/handlers/bulkdata_handlers.cpp
```diff
@@ -554,7 +558,11 @@ std::vector<std::string> BulkDataHandlers::get_source_filters(const EntityInfo &
         }
       }
     }
-    return filters;
+    if (!filters.empty()) {
+      return filters;
+    }
+    // Manifest deployments where a component groups topics rather than nodes
+    // still need the namespace prefix path to work, so fall through.
   }
```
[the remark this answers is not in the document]
bburda: Good catch. Split FUNCTION and COMPONENT into separate arms - FUNCTION now always returns the host-derived filters (no fall-through to fqn/namespace_path, since functions are pure aggregated views), COMPONENT keeps the fall-through for manifest-only deployments. Added a `BulkDataSourceFiltersTest.FunctionWithoutHostsReturnsEmptyEvenIfFqnSet` regression guard that pins this exact behavior - constructs a function with no hosts but with fqn/namespace_path set on the entity and asserts the returned filter list is empty.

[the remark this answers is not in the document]
bburda: Done. Made `get_source_filters` public (calling it through the REST surface forces a real FaultManager service which is not viable in unit tests) and added 9 `BulkDataSourceFiltersTest` cases covering APP / AREA / FUNCTION / COMPONENT branching with both happy paths and edge cases: synthetic component with hosted apps, manifest-only fall-through, empty component, function with hosts, function without hosts (regression guard), app fqn fall-through, app+namespace fall-through, app with neither, area fqn passthrough.

src/ros2_medkit_gateway/include/ros2_medkit_gateway/http/handlers/bulkdata_handlers.hpp
```diff
@@ -111,21 +112,27 @@ class BulkDataHandlers {
...
+   *   when the component has no hosted apps (manifest deployments where the component
+   *   groups topics rather than nodes). This avoids the synthetic-component bug where
+   *   empty fqn + empty namespace_path produced zero source filters.
+   *
+   * Public to enable unit testing the entity-type branching directly.
    *
    * @param entity Entity information
    * @return Vector of source filter strings (empty if no valid filters)
    */
   std::vector<std::string> get_source_filters(const EntityInfo & entity) const;
```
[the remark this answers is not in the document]
bburda: Done. Moved `get_source_filters` back to private and extracted the entity-type branching to `handlers::detail::compute_bulkdata_source_filters` (a free function in a `detail` namespace, declared at the bottom of `bulkdata_handlers.hpp`). The instance method is now a one-line wrapper that fetches the cache from `ctx_` and delegates. Tests target the free function directly with a bare `ThreadSafeEntityCache` - no `BulkDataHandlers` instance, no `GatewayNode`, no DDS context required.

src/ros2_medkit_gateway/src/http/handlers/log_handlers.cpp
```diff
@@ -128,17 +118,9 @@ void LogHandlers::handle_get_logs(const httplib::Request & req, httplib::Respons
...
-        }
-        auto fqn = app->effective_fqn();
-        if (!fqn.empty()) {
-          host_fqns.push_back(std::move(fqn));
-        }
-      }
+      auto comp_fqns = HandlerContext::resolve_app_host_fqns(cache, cache.get_apps_for_component(comp_id));
+      host_fqns.insert(host_fqns.end(), std::make_move_iterator(comp_fqns.begin()),
+                       std::make_move_iterator(comp_fqns.end()));
     }
```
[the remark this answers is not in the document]
bburda: Done. Added `#include <iterator>` to `log_handlers.cpp`.

mfaferek93 (commented):
A few comments

src/ros2_medkit_gateway/src/http/handlers/log_handlers.cpp
```diff
@@ -177,11 +160,58 @@ void LogHandlers::handle_get_logs(const httplib::Request & req, httplib::Respons
...
-  const bool prefix_match = (entity.type == EntityType::COMPONENT);
+  if (entity.type == EntityType::COMPONENT) {
+    const auto & cache = ctx_.node()->get_thread_safe_cache();
+    auto host_fqns = HandlerContext::resolve_app_host_fqns(cache, cache.get_apps_for_component(entity_id));
+
+    json result;
+    result["items"] = json::array();
+    XMedkit ext;
+    ext.entity_id(entity_id);
+    ext.add("aggregation_level", "component");
```
mfaferek93, line 178: **[Blocker] x-medkit aggregation metadata emitted here is not declared in the LogList schema - same drift class as PR #390.**

This block (and the AREA / FUNCTION arms above) emits `x-medkit.aggregation_level`, `aggregated`, `app_count`, `aggregation_sources` on the response. The corresponding response schema (`schema_builder.cpp:660` registers `LogEntryList` as `items_wrapper_ref("LogEntry")` which produces `{items: array}` only - no x-medkit field). Generated/typed clients silently drop these fields. Drift test added in #390 won't catch it because most schemas don't set `additionalProperties: false`.

Same issue downstream: `BulkDataDescriptorList` (`schema_builder.cpp:700`) - `bulkdata_handlers.cpp` also emits aggregation metadata in this PR.

Fix: dedicated `log_list_schema()` (and `bulk_data_list_schema()`) that wraps `items_wrapper(...)` and adds a top-level x-medkit subschema with the four fields. Mirror the pattern used by `fault_detail_schema:68-117`. The drift test you ran on #390 should then catch this kind of gap going forward IF the affected response schemas set `additionalProperties: false`.

src/ros2_medkit_plugins/ros2_medkit_opcua/docker/Dockerfile.gateway
```diff
@@ -20,10 +20,21 @@ WORKDIR ${COLCON_WS}
...
+# apt-get update has to be re-run here: the previous RUN cleared
+# /var/lib/apt/lists, so without a fresh fetch rosdep's internal
+# apt-get install calls fail with "Unable to locate package" for anything
+# that wasn't pulled in by the first RUN's apt-get install (notably
+# python3-jsonschema, which the integration_tests package declares as a
+# test_depend and which rosdep tries to install even with BUILD_TESTING=OFF).
+# This was previously masked by Docker layer cache hits on CI; cold builds
+# always failed.
 RUN bash -c "source /opt/ros/jazzy/setup.bash && \
+    apt-get update && \
```
mfaferek93, line 33: **[Blocker] Same anti-pattern still present in root `Dockerfile:71-75`** (not modified by this PR, but the rationale you wrote in the comment block above applies there verbatim).

Root `Dockerfile` lines 35-51 clear `/var/lib/apt/lists/*`, then line 71-75 runs `rosdep install` in a fresh RUN with **no** preceding `apt-get update`. Same `python3-jsonschema` test_depend would fail on cold Docker builds the moment BuildKit evicts the warm layer. Currently masked by warm cache.

Apply the identical pattern from this commit to the root Dockerfile RUN block:
```
RUN bash -c "source /opt/ros/${ROS_DISTRO}/setup.bash && \
    apt-get update && \
    rosdep update && \
    rosdep install --from-paths src --ignore-src -r -y --skip-keys='...' && \
    rm -rf /var/lib/apt/lists/* && \
    colcon build ..."
```

src/ros2_medkit_gateway/src/http/handlers/bulkdata_handlers.cpp
```diff
@@ -539,31 +539,41 @@ std::string BulkDataHandlers::get_rosbag_mimetype(const std::string & format) {
...
-      auto app = cache.get_app(app_id);
-      if (app) {
-        auto fqn = app->effective_fqn();
-        if (!fqn.empty()) {
-          filters.push_back(fqn);
-        }
-      }
+    // Functions are pure aggregated views over hosted apps - if no apps host the function,
+    // there is nothing to query. No fall-through to fqn/namespace_path.
+    return HandlerContext::resolve_app_host_fqns(cache, cache.get_apps_for_function(entity.id));
```
mfaferek93, line 552: **[Major] Silent breaking change for FUNCTION endpoint.**

Previous behavior: when a FUNCTION had no hosted apps but had a non-empty `entity.fqn`, `get_source_filters` fell through to the fqn/namespace_path path and returned a single-element filter list. New behavior here: returns the empty result of `resolve_app_host_fqns(...)` unconditionally - no fall-through.

Locked by `FunctionWithoutHostsReturnsEmptyEvenIfFqnSet` in test_bulkdata_handlers.cpp:317-320, which is the right regression guard - **but the change is not in the PR description and not in any CHANGELOG**. Downstream consumers reading `/functions/{id}/bulk-data/rosbags` for a function-without-hosts will see empty descriptor lists where before they saw at least the namespace-prefix match.

Either: (a) add a CHANGELOG entry + PR description note calling out the FUNCTION semantics change, or (b) preserve the previous fall-through (`if (filters.empty() && !entity.fqn.empty()) filters.push_back(entity.fqn);`) and update the test.

src/ros2_medkit_gateway/src/http/handlers/handler_context.cpp
```diff
@@ -346,5 +346,22 @@ void HandlerContext::send_json(httplib::Response & res, const json & data) {
...
+  std::vector<std::string> fqns;
+  fqns.reserve(app_ids.size());
+  for (const auto & app_id : app_ids) {
+    auto app = cache.get_app(app_id);
+    if (!app) {
+      continue;
+    }
+    auto fqn = app->effective_fqn();
+    if (!fqn.empty()) {
+      fqns.push_back(std::move(fqn));
```
mfaferek93, line 370: **[Major] Duplicate `effective_fqn`s are pushed without deduplication; no test pins the contract.**

Bare `push_back` here means if two app_ids resolve to the same effective_fqn (e.g. two app entries pointing at the same ROS node, or a manifest+runtime double-bind), the returned vector contains duplicates. Downstream `log_mgr->get_logs(host_fqns, false, ...)` semantics on duplicates is unspecified - either dedups internally and the duplicates are wasted CPU, or queries twice and merges.

Existing `ResolveAppHostFqnsTest::ResolvesMultipleAppsPreservesInputOrder` (test:1158-1170) uses *distinct* fqns (`temp_sensor` vs `rpm_sensor`), so the duplicate scenario is not pinned.

Pick one and lock it:
1. Add a `DuplicateEffectiveFqnsAreNotDeduplicated` test that asserts the current behavior (passes today, regression guard for callers).
2. OR add `std::sort` + `std::unique` here with deterministic order, plus a `DuplicateEffectiveFqnsAreDeduplicated` test.

src/ros2_medkit_gateway/src/http/handlers/bulkdata_handlers.cpp
bburda: The previous behavior on `main` did not fall through. The FUNCTION arm in `bulkdata_handlers.cpp` on `origin/main` ends with an unconditional `return filters;`:

```cpp
std::vector<std::string> BulkDataHandlers::get_source_filters(const EntityInfo & entity) const {
  if (entity.type == EntityType::FUNCTION) {
    // ... build filters from hosting apps' effective_fqns ...
    return filters;   // unconditional, even when filters is empty
  }
  // namespace_path fallback only ran for non-FUNCTION entity types
  ...
}
```

A FUNCTION-without-hosts on `main` returned `{}` regardless of `entity.fqn`, and the same case after this PR still returns `{}`. There is no behavior change for FUNCTION here, so nothing to call out in the description. `FunctionWithoutHostsReturnsEmptyEvenIfFqnSet` pins the existing semantics, not a regression introduced here.

Suggestion (b) (`if (filters.empty() && !entity.fqn.empty()) filters.push_back(entity.fqn);`) would actually introduce a *new* behavior that never existed: FUNCTION-no-hosts-with-fqn would start returning `{entity.fqn}`. That also conflicts with the entity model where FUNCTION is a capability grouping over hosted apps rather than a ROS node with its own fqn.

Leaving the code unchanged. Happy to revisit if there is a commit/path on `main` where the FUNCTION arm did fall through that I might have missed.

src/ros2_medkit_gateway/src/http/handlers/log_handlers.cpp
bburda: Done for `LogEntryList`. Added a dedicated `log_entry_list_schema()` that wraps `items` and declares an `x-medkit` subschema with all eight fields the handler emits (`entity_id`, `aggregation_level` enum, `aggregated`, `host_count`, `component_count`, `app_count`, `aggregation_sources`, `contributors`). Wired into `component_schemas()` under `LogEntryList`. Added a `LogEntryListXMedkitDeclaresAggregationFields` static unit test mirroring the existing `ConfigurationMetaDataXMedkitDeclaresAllEmittedFields`.

`BulkDataDescriptorList` was left as-is. `bulkdata_handlers.cpp` does not emit response-level aggregation metadata in this PR - the only `x-medkit` it sets is per-descriptor (`desc["x-medkit"] = item.metadata`), which is already covered by the existing `bulk_data_descriptor_schema()` declaring `x-medkit: {type: object, additionalProperties: true}`. Happy to revisit if you spotted aggregation metadata being set on the wrapper that I missed.

src/ros2_medkit_plugins/ros2_medkit_opcua/docker/Dockerfile.gateway
bburda: Done. Applied the same pattern to the root `Dockerfile` rosdep RUN: re-run `apt-get update` at the start, and move the `rm -rf /var/lib/apt/lists/*` cleanup to the end of the same layer so the builder image still inherits no apt cache. Comment block in the diff explains why.

src/ros2_medkit_gateway/src/http/handlers/handler_context.cpp
bburda: Picked option 2 (dedup + test). Switched `resolve_app_host_fqns` to an order-preserving dedup using an `unordered_set` seen guard - first occurrence wins, so the existing `ResolvesMultipleAppsPreservesInputOrder` invariant still holds. Updated the docstring to spell out the new contract.

Added `DuplicateEffectiveFqnsAreDeduplicated` that pins three failure modes: two distinct app_ids resolving to the same effective_fqn collapse to one entry, the same app_id requested twice also collapses, and an intervening distinct fqn keeps its original first-seen position.

mfaferek93 (approved):

PR #397, opened 2026-04-29
refactor(gateway): pure-C++ managers via Transport adapters + discovery split + graph-event refresh

### LINKED ISSUE #396: Decouple gateway managers from rclcpp via provider injection for testability and reduced flakiness
## Summary

Today every gateway test pulls `rclcpp` and the ament environment because the manager classes (`DataAccessManager`, `OperationManager`, `ConfigurationManager`, `FaultManager` facade, `LogManager`, `TriggerManager`) take `rclcpp::Node *` directly and route I/O inline. That coupling makes unit tests heavy (multi-second startup), forces every test fixture through `MultiThreadedExecutor` lifecycle, and makes test failures timing-sensitive - flaky `process spawned but graph not yet seen` and `subscription destroyed mid-callback` patterns appear repeatedly in CI history.

This issue tracks the next pass after #391: each manager keeps its routing and business logic but gains a `register_provider(...)` injection point, with the ROS-specific I/O extracted into `Ros2*Provider` classes living in the adapter layer. After the change, manager unit tests compose with mock providers and link only against `gateway_core` (the neutral library introduced in #391), which removes `rclcpp` from the unit-test link line and eliminates the executor-lifecycle setup that is the root cause of intermittent failures.

---

Concretely:

* Six managers gain provider injection: `Data`, `Operation`, `Configuration`, `Fault`, `Log`, `Trigger`. Each manager moves to `core/managers/`. The ROS-specific default behaviour is extracted into `Ros2*Provider` implementations under `src/ros2/providers/` (or equivalent path in the adapter layer), registered statically by `gateway_node` at startup. Plugins continue to register per-entity providers via `PluginManager` exactly as today.
* `RuntimeDiscoveryStrategy` becomes `Ros2RuntimeIntrospection : IntrospectionProvider`. The discovery framework in `core/` then routes through the existing `IntrospectionProvider` chain and stops referencing `rclcpp::Node`. `HybridDiscoveryStrategy` collapses into the standard `MergePipeline` configuration.
* `TypeIntrospection` moves to `ros2_medkit_serialization`, where `rosidl_typesupport_cpp` and `rosidl_typesupport_introspection_cpp` already live. The gateway then depends on the serializer for type schemas instead of duplicating the rosidl bridge inside the gateway tree.
* Runtime discovery trigger switches from cyclic `wall_timer(refresh_interval_ms_)` polling to event-driven `graph_event->check_and_clear()`. The polling timer remains as a low-frequency safety backstop. Reduces idle CPU when the graph is stable and improves new-node detection latency.

Out of scope (intentionally deferred): folding `core/` into a separate colcon package - that is a layout change and does not affect the test-link-line goal.

---

Outcome metrics expected after the work:

* Manager unit tests link only `gateway_core` + GTest (no `ament_target_dependencies`), proven by extending `test_gateway_core_smoke` to include a manager headers and have a representative manager test compile against the mock-provider variant.
* Idle gateway process CPU usage drops measurably once the discovery refresh stops firing on a stable graph (sample with `top` before/after on the demo nodes scenario).
* Existing test suite (~2500 unit, ~3200 integration, ~2600 clang-tidy) stays green without timing-sensitive sleeps in tests we touch.
* New-node detection latency (time from `ros2 run` to gateway `/apps` listing the new entity) drops below 500 ms in steady-state operation.

bburda:
# Pull Request

## Summary

This PR is the second pass after #391: each SOVD manager keeps its routing and
business logic, but the ROS 2 binding is extracted behind neutral Transport
adapters. Manager unit tests can now compose with mock transports and link only
against `gateway_core`, which removes `rclcpp` from the unit-test link line and
eliminates the executor-lifecycle setup that was the root cause of intermittent
test failures (`process spawned but graph not yet seen`,
`subscription destroyed mid-callback`).

The change also folds `RuntimeDiscoveryStrategy` into the existing
`IntrospectionProvider` chain (built-in graph queries become just another
provider), relocates `TypeIntrospection` next to the rest of the
`rosidl_typesupport` glue in `ros2_medkit_serialization`, and switches the
runtime-discovery refresh from cyclic polling to rclcpp graph-event driven, with
the timer kept only as a low-frequency safety backstop.

This PR stacks on #392 (`feat/medkit-core-scaffold`). When #392 merges, GitHub
auto-rebases this branch onto `main`.

---

## Issue

- closes #396

---

## Type

- [x] New feature or tests
- [ ] Bug fix
- [ ] Breaking change
- [x] Documentation only

---

## Testing

The changes were exercised through the full local test suite. Reviewers can
reproduce by running, from the gateway worktree:

- [x] `colcon build --symlink-install` - whole workspace compiles cleanly
      (15 packages, no errors)
- [x] `./scripts/test.sh unit` - 2727 unit tests, 0 failures, 0 errors. The
      new mock-transport tests link only against `gateway_core` + GTest with
      no `ament_target_dependencies`.
- [x] `./scripts/test.sh lint --packages-select ros2_medkit_gateway` - linters
      pass (clang-format, copyright, cmake-lint, flake8, pep257, xmllint)
- [x] `./scripts/test.sh integ --packages-select ros2_medkit_integration_tests`
      - 4501 tests, 0 failures, including the `test_graph_event_discovery`
      and `test_triggers_persistent` features
- [x] `colcon build --packages-select ros2_medkit_gateway --cmake-args -DENABLE_CLANG_TIDY=ON`
      then `./scripts/test.sh tidy --packages-select ros2_medkit_gateway` -
      clang-tidy clean on every changed file
- [x] `colcon test --packages-select ros2_medkit_gateway --ctest-args -R "gateway_core_smoke"`
      - the smoke test links only `gateway_core` + GTest and proves all eight
      provider interfaces, seven Transport ports, and six manager class names
      are reachable without rclcpp

To verify the discovery latency improvement, launch the gateway, then
`ros2 run` a demo node and observe `/apps` listing the new entity below 500 ms.

---

## Checklist

---

## Commits

13 commits build up the main change in self-contained, reviewable steps. Each
step keeps the suite green and ends with the project compiling cleanly.

1. `refactor(gateway): relocate operation result types into core/operations`
2. `refactor(gateway): relocate parameter result types into core/configuration`
3. `refactor(gateway): relocate FaultResult into core/faults`
4. `feat(gateway): introduce neutral Transport interfaces under core/transports`
5. `refactor(gateway): route DataAccessManager through TopicTransport adapter`
6. `refactor(gateway): route OperationManager through Service/Action transports`
7. `refactor(gateway): route ConfigurationManager through ParameterTransport`
8. `refactor(gateway): route FaultManager through FaultServiceTransport`
9. `refactor(gateway): route LogManager through LogSource adapter`
10. `refactor(gateway): route TriggerManager through TopicSubscriptionTransport`
11. `refactor(gateway): convert runtime discovery to IntrospectionProvider`
12. `refactor: relocate TypeIntrospection to ros2_medkit_serialization`
13. `refactor(gateway): drive discovery refresh from rclcpp graph events`

Two functional fixes, one doc sweep, and a small clang-tidy / formatting
follow-up were folded into the branch during verification:

- `fix(gateway): only sweep orphan triggers on the backstop refresh` - the
  graph-event refresh path was running the orphan-trigger sweep during the
  cold-start window, before DDS discovery had populated the entity cache.
  Restored persistent triggers were then treated as orphans and removed from
  the SQLite store. Sweep now runs only from the backstop tick, where the
  cache is guaranteed to reflect a settled DDS view.
- `fix(opcua): bump test_opcua_plugin TIMEOUT to 240s` - the suite legitimately
  needs ~90 s wallclock for its 13 OPC UA connection-failure tests; the
  default 60 s timeout truncated the run mid-suite under heavy parallel
  colcon test load.
- `docs(gateway): refresh design index for managers-in-core layout` updates
  the layered-library note, the TriggerTopicSubscriber class summary, and
  the TriggerManager arrow on the class diagram so the design document
  matches the new manager / transport split.
- `fix(gateway): resolve clang-tidy findings on touched files` and a small
  follow-up commit address every clang-tidy warning the incremental PR
  job emits on the cpp files this branch modifies (named-parameter,
  ``std::bind`` -> lambda, single-character ``find()``, copied
  ``shared_ptr`` parameters, ``vector::reserve`` before ``emplace_back``,
  and one ``operator+=`` rewrite in ``fault_handlers``).


src/ros2_medkit_gateway/src/http/handlers/fault_handlers.cpp
```diff
@@ -163,69 +163,94 @@ json extract_primary_value(const std::string & message_type, const json & full_d
...
+  if (fault_json.contains("reporting_sources") && fault_json["reporting_sources"].is_array()) {
+    reporting_sources = fault_json["reporting_sources"];
   }
 
-  response["x-medkit"] = {{"occurrence_count", fault.occurrence_count},
+  response["x-medkit"] = {{"occurrence_count", fault_json.value("occurrence_count", static_cast<uint64_t>(0))},
                           {"reporting_sources", reporting_sources},
-                          {"severity_label", severity_to_label(fault.severity)},
-                          {"status_raw", fault.status}};
+                          {"severity_label", severity_to_label(severity)},
```
mfaferek93, line 271: `severity_to_label()` defined at lines 119-134 is off-by-one vs `ros2_medkit_msgs/msg/Fault.msg`. Map is `{0:DEBUG,1:INFO,2:WARN,3:ERROR,4:FATAL}` but Fault.msg defines `SEVERITY_INFO=0..SEVERITY_CRITICAL=3`. Every SOVD response here gets a wrong `severity_label`: `severity:0` (INFO) labeled `DEBUG`, `severity:3` (CRITICAL) labeled `ERROR`. Fix the map to `{0:INFO,1:WARN,2:ERROR,3:CRITICAL}` and add a test asserting label-vs-int consistency.

src/ros2_medkit_gateway/src/ros2/transports/ros2_fault_service_transport.cpp
```diff
@@ -378,10 +344,9 @@ FaultResult FaultManager::get_snapshots(const std::string & fault_code, const st
   result.success = response->success;
 
   if (response->success) {
-    // Parse the JSON data from the service response
     try {
       result.data = json::parse(response->data);
-    } catch (const json::exception & e) {
+    } catch (const json::exception & /*e*/) {
```
mfaferek93: `catch (const json::exception & /*e*/)` silently swallows the parse failure - no log, no counter. Operators can't tell whether the `raw_data` fallback fires once or constantly. Log `e.what()` at WARN before the fallback. Sibling methods in this file repeat the same pattern; apply consistently.

src/ros2_medkit_gateway/src/core/managers/log_manager.cpp
```diff
@@ -112,61 +123,64 @@ json LogManager::entry_to_json(const LogEntry & e) {
...
-  // crashing the gateway's ROS 2 subscription callback.
+  // crashing the source's delivery callback.
   bool suppress_buffer = false;
-  if (plugin_mgr_) {
-    for (auto * observer : plugin_mgr_->get_log_observers()) {
+  if (provider_registry_) {
+    for (auto * observer : provider_registry_->log_observers()) {
       try {
         if (observer->on_log_entry(entry)) {
           suppress_buffer = true;
```
mfaferek93, line 178: At `buffers_.size() >= max_buffer_size_ * 10` (default cap = 2000), logs from a new node name are dropped silently. Same pattern at lines 222-223. No WARN, no counter - operators see missing logs with no diagnostic. Add a one-shot WARN per newly-dropped node (track via `std::unordered_set<std::string> dropped_nodes_`) and expose a counter via the manager config endpoint.

src/ros2_medkit_gateway/src/http/handlers/sse_fault_handler.cpp
```diff
@@ -220,7 +220,7 @@ std::string SSEFaultHandler::format_sse_event(const ros2_medkit_msgs::msg::Fault
 
   nlohmann::json json_event;
   json_event["event_type"] = sanitized_event_type;
```
mfaferek93, line 237: Outside this hunk but visible while reviewing: `on_fault_event()` at lines 96-98 silently `pop_front()`s events when `event_queue_.size() > kMaxBufferedEvents` (=100). Same diagnostic blind spot as the log manager - clients lose fault notifications under backpressure with zero log/counter signal. Worth a rate-limited WARN + dropped-events counter in the same SSE rework.

src/ros2_medkit_gateway/test/test_data_access_manager_routing.cpp
```diff
@@ -0,0 +1,127 @@
...
+class MockTopicTransport : public TopicTransport {
+ public:
+  MockTopicTransport() = default;
+  ~MockTopicTransport() override = default;
+  MockTopicTransport(const MockTopicTransport &) = delete;
+  MockTopicTransport & operator=(const MockTopicTransport &) = delete;
+  MockTopicTransport(MockTopicTransport &&) = delete;
+  MockTopicTransport & operator=(MockTopicTransport &&) = delete;
+
+  json publish(const std::string & topic_path, const std::string & msg_type, const json & data,
```
mfaferek93, line 39: `MockTopicTransport::publish()` always returns `{"status":"published",...}`; `sample()` only branches on `sample_publishers_ > 0`. No failure-path coverage - DataAccessManager's error propagation (timeout, serialization failure, transport error) is unexercised. Add `publish_success_` / `sample_success_` / `sample_error_msg_` knobs and test cases asserting transport errors propagate to the caller. Same concern applies to the other mock transports in this PR.

src/ros2_medkit_gateway/src/ros2/conversions/fault_msg_conversions.cpp
```diff
@@ -0,0 +1,100 @@
...
+nlohmann::json environment_data_to_json(const ros2_medkit_msgs::msg::EnvironmentData & env_data) {
+  using nlohmann::json;
+
+  json snapshots = json::array();
+  for (const auto & s : env_data.snapshots) {
+    json snap;
+    snap["type"] = s.type;
+    snap["name"] = s.name;
+
+    if (s.type == "freeze_frame") {
```
mfaferek93, line 72: Rosbag and freeze_frame snapshots have disjoint JSON shapes - freeze_frame has `data/topic/message_type/captured_at_ns` (lines 72-80), rosbag has `fault_code/size_bytes/duration_sec/format` (lines 81-86). The handler at `fault_handlers.cpp:215` does `s.value("fault_code", fault_code)` defaulting to the parent fault when key missing. If a future schema change drops `fault_code` from rosbag snapshots, every rosbag silently inherits the parent fault. Add an explicit `snapshot_type` discriminator and validate shape per type, or drop the silent fallback in the handler.

src/ros2_medkit_gateway/launch/gateway.launch.py
```diff
@@ -53,7 +53,10 @@ def generate_launch_description():
 
     declare_refresh_arg = DeclareLaunchArgument(
         'refresh_interval_ms', default_value='2000',
```
mfaferek93: Default `2000` (ms) here conflicts with `config/gateway_params.yaml:61` which defaults to `30000`. With graph-event-driven refresh + safety-backstop semantics, 30s is the documented intent (per README + design docs). Operators using `ros2 launch` get an aggressive 2s backstop; YAML-driven deployments get 30s. Align both to one value.

src/ros2_medkit_gateway/src/core/managers/operation_manager.cpp
```diff
@@ -0,0 +1,491 @@
...
+void OperationManager::update_goal_feedback(const std::string & goal_id, const json & feedback) {
+  std::lock_guard<std::mutex> lock(goals_mutex_);
+  auto it = tracked_goals_.find(goal_id);
+  if (it != tracked_goals_.end()) {
+    it->second.last_feedback = feedback;
+    it->second.last_update = std::chrono::system_clock::now();
+  }
+}
+
+void OperationManager::cleanup_old_goals(std::chrono::seconds max_age) {
```
mfaferek93, line 415: `cleanup_old_goals()` only evicts SUCCEEDED/CANCELED/ABORTED. Goals in ACCEPTED/EXECUTING that never receive a terminal status update (action-server crash, lost status) keep `last_update` frozen and live forever in `tracked_goals_` (unbounded `std::map`). Add an age-based force-evict for non-terminal states (`now - last_update > N * max_age`, e.g. N=2) with a WARN log so operators see when it triggers.

src/ros2_medkit_gateway/src/ros2/transports/ros2_log_source.cpp
```diff
@@ -0,0 +1,84 @@
...
+
+        cb(entry);
+      });
+}
+
+void Ros2LogSource::stop() {
+  // Mark shutdown so any in-flight callback short-circuits before touching
+  // members that are about to destruct.
+  shutdown_requested_.store(true, std::memory_order_release);
+
```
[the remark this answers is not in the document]
bburda: Fixed in 1b19baec - `Ros2LogSource::start()` now clears `shutdown_requested_` to false (under the same mutex) before installing the callback. Added regression test `Ros2LogSourceTest.RestartResetsShutdownLatch` that drives a stop()/start()/publish cycle and asserts the callback fires.

src/ros2_medkit_gateway/src/ros2/transports/ros2_service_transport.cpp
```diff
@@ -0,0 +1,160 @@
...
+  using ros2_medkit_serialization::ServiceActionTypes;
+  ServiceCallResult result;
+
+  try {
+    auto client = get_or_create_client(service_path, service_type);
+
+    if (!client->wait_for_service(std::chrono::seconds(5))) {
+      result.success = false;
+      result.error_message = "Service not available: " + service_path;
+      return result;
```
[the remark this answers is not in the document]
bburda: Fixed in 1b19baec - `wait_for_service()` now uses the caller-provided timeout instead of a hardcoded 5 s. The async call below still bounds itself via `future.wait_for()`.

src/ros2_medkit_gateway/src/ros2/transports/ros2_service_transport.cpp
```diff
@@ -0,0 +1,160 @@
...
+#include "ros2_medkit_serialization/serialization_error.hpp"
+#include "ros2_medkit_serialization/service_action_types.hpp"
+#include "ros2_medkit_serialization/type_cache.hpp"
+
+namespace ros2_medkit_gateway::ros2 {
+
+Ros2ServiceTransport::Ros2ServiceTransport(rclcpp::Node * node)
+  : node_(node), serializer_(std::make_shared<ros2_medkit_serialization::JsonSerializer>()) {
+  RCLCPP_INFO(node_->get_logger(), "Ros2ServiceTransport initialised (native serialization)");
+}
```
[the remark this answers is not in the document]
bburda: Declining: `Ros2ServiceTransport` is constructed by `GatewayNode` which always passes `this`. The null-node case is unreachable in production wiring. Adding a defensive throw here would catch only programmer error in tests, which already cover the wiring chain. Keeping the precondition implicit to match the other transport adapters (`Ros2ActionTransport`, `Ros2ParameterTransport`, `Ros2FaultServiceTransport`, `Ros2LogSource` - none of them validate).

src/ros2_medkit_gateway/src/ros2/transports/ros2_action_transport.cpp
```diff
@@ -0,0 +1,469 @@
...
+
+  try {
+    using ros2_medkit_serialization::ServiceActionTypes;
+
+    auto & clients = get_or_create_clients(action_path, action_type);
+
+    if (!clients.send_goal_client->wait_for_service(std::chrono::seconds(5))) {
+      result.error_message = "Action server not available: " + action_path;
+      return result;
+    }
```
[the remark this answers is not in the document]
bburda: Fixed in 1b19baec - `send_goal()` now uses the caller-provided timeout for `wait_for_service()`. The cancel/get_result paths keep a 2 s wait which is acceptable since the server has already been confirmed alive in send_goal; can revisit as follow-up if it bites.

src/ros2_medkit_gateway/include/ros2_medkit_gateway/data_access_manager.hpp
```diff
@@ -14,132 +14,7 @@
...
-   * @brief Get default timeout for topic sampling (from parameter)
-   */
-  double get_default_topic_timeout() const {
-    return topic_sample_timeout_sec_;
-  }
-};
-
-}  // namespace ros2_medkit_gateway
+// Backwards-compatibility shim - header relocated to core/managers/.
+#include "ros2_medkit_gateway/core/managers/data_access_manager.hpp"
```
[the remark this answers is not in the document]
bburda: Resolved differently in 77cd6f42 - the backwards-compat shim headers were deleted entirely. They guarded a scenario (out-of-tree consumers) that doesn't exist in our monorepo, so the `#pragma message` noise was overengineering on top of overengineering. All in-tree consumers now use the canonical `core/managers/` paths directly.

src/ros2_medkit_gateway/include/ros2_medkit_gateway/fault_manager.hpp
```diff
@@ -14,141 +14,7 @@
...
-  mutable std::mutex get_mutex_;
-  mutable std::mutex clear_mutex_;
-  mutable std::mutex snapshots_mutex_;
-  mutable std::mutex rosbag_mutex_;
-  mutable std::mutex list_rosbags_mutex_;
-};
-
-}  // namespace ros2_medkit_gateway
+// Backwards-compatibility shim - header relocated to core/managers/.
+#include "ros2_medkit_gateway/core/managers/fault_manager.hpp"
```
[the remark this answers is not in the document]
bburda: Resolved differently in 77cd6f42 - shim header deleted entirely (no out-of-tree consumers existed). Canonical path `core/managers/fault_manager.hpp` is now the only path.

src/ros2_medkit_gateway/src/core/managers/data_access_manager.cpp
```diff
@@ -0,0 +1,82 @@
...
+    if (!sample.type_info.empty() && !sample.type_info.is_null()) {
+      out["type_info"] = sample.type_info;
+    }
+  }
+  if (sample.status == "data" && !sample.data.is_null()) {
+    out["data"] = sample.data;
+    out["timestamp"] =
+        std::chrono::duration_cast<std::chrono::milliseconds>(std::chrono::system_clock::now().time_since_epoch())
+            .count();
+  }
```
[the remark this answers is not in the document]
bburda: Fixed in f26107ba - `get_topic_sample_native()` now emits a `timestamp` in nanoseconds since epoch for both `data` and `metadata_only` responses, matching the SOVD contract used elsewhere in the API.

src/ros2_medkit_gateway/src/ros2/transports/ros2_service_transport.cpp
```diff
@@ -0,0 +1,165 @@
...
+        std::chrono::milliseconds{static_cast<std::int64_t>(std::max(timeout.count(), 0.0) * 1000.0)};
+    auto future_status = future_and_id.wait_for(timeout_secs);
+
+    if (future_status != std::future_status::ready) {
+      client->remove_pending_request(future_and_id.request_id);
+      ros2_medkit_serialization::destroy_ros_message(&ros_request);
+      result.success = false;
+      result.error_message =
+          "Service call timed out (" + std::to_string(static_cast<int>(timeout.count())) + "s): " + service_path;
+      return result;
```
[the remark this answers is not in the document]
bburda: Fixed in f26107ba - the timeout error message now formats the budget in milliseconds via `chrono::duration_cast<milliseconds>(timeout).count()`, so sub-second timeouts (e.g., 0.1 s) no longer collapse to `0s`.

src/ros2_medkit_gateway/include/ros2_medkit_gateway/operation_manager.hpp
```diff
@@ -14,287 +14,7 @@
...
-  std::string uuid_bytes_to_hex(const std::array<uint8_t, 16> & uuid) const;
-
-  /// Map of action_path -> status subscription
-  mutable std::mutex subscriptions_mutex_;
-  std::map<std::string, rclcpp::Subscription<action_msgs::msg::GoalStatusArray>::SharedPtr> status_subscriptions_;
-};
-
-}  // namespace ros2_medkit_gateway
+// Backwards-compatibility shim - header relocated to core/managers/.
+#include "ros2_medkit_gateway/core/managers/operation_manager.hpp"
```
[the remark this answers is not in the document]
bburda: Resolved differently in 77cd6f42 - shim header deleted entirely. Canonical `core/managers/operation_manager.hpp` is now the only path.

src/ros2_medkit_gateway/include/ros2_medkit_gateway/log_manager.hpp
```diff
@@ -14,185 +14,7 @@
...
-  std::unordered_map<std::string, LogConfig> configs_;
-  mutable std::mutex configs_mutex_;
-
-  // Monotonically increasing entry ID (never resets; starts at 1)
-  std::atomic<int64_t> next_id_{1};
-};
-
-}  // namespace ros2_medkit_gateway
+// Backwards-compatibility shim - header relocated to core/managers/.
+#include "ros2_medkit_gateway/core/managers/log_manager.hpp"
```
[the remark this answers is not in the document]
bburda: Resolved differently in 77cd6f42 - shim header deleted entirely. Canonical `core/managers/log_manager.hpp` is now the only path.

src/ros2_medkit_gateway/src/http/handlers/fault_handlers.cpp
bburda: Fixed in f26107ba - `severity_to_label()` now matches `ros2_medkit_msgs/msg/Fault.msg` constants: {0:INFO, 1:WARN, 2:ERROR, 3:CRITICAL}. Tests in `test_fault_handlers.cpp` updated to use the `Fault::SEVERITY_*` constants and assert label-vs-int consistency.

src/ros2_medkit_gateway/src/ros2/transports/ros2_fault_service_transport.cpp
bburda: Fixed in f26107ba - the catch site now logs `e.what()` at WARN with fault_code/topic context before falling back to `raw_data`. Sibling methods reviewed; this was the only silent catch in the file.

src/ros2_medkit_gateway/src/core/managers/log_manager.cpp
bburda: Fixed in f26107ba - `LogManager` now tracks dropped entries via `dropped_nodes_` set and `dropped_total_` counter, emits one WARN per newly-dropped node through the injected `log_sink`, and exposes `dropped_entries_count()` publicly. Wiring the count into `/server-info` or `/diagnostics` is left as follow-up.

src/ros2_medkit_gateway/src/http/handlers/sse_fault_handler.cpp
bburda: Fixed in f26107ba - `on_fault_event()` now tracks `dropped_events_` atomically and emits a rate-limited WARN every Nth drop with the running total.

src/ros2_medkit_gateway/test/test_data_access_manager_routing.cpp
bburda: Partial fix in f26107ba - `MockTopicTransport` gained `publish_success_`, `sample_success_`, and `sample_error_msg_` knobs plus two new tests asserting failure propagation. The same pattern should be extended to the service/action/fault mocks; flagged as follow-up since it is a wider mechanical change than the immediate gap.

src/ros2_medkit_gateway/src/ros2/conversions/fault_msg_conversions.cpp
bburda: Fixed in f26107ba - the conversion layer now emits an explicit `snapshot_type` discriminator (`'freeze_frame'` or `'rosbag'`). `fault_handlers.cpp` reads it preferentially and logs a WARN if a rosbag snapshot is missing `fault_code`, instead of the silent inherit-from-parent fallback.

src/ros2_medkit_gateway/launch/gateway.launch.py
bburda: Fixed in f26107ba - both `gateway.launch.py` and `gateway_https.launch.py` now default `refresh_interval_ms` to `30000`, aligning with `gateway_params.yaml`, README, and design docs.

src/ros2_medkit_gateway/src/core/managers/operation_manager.cpp
bburda: Fixed in f26107ba - `cleanup_old_goals()` now force-evicts stuck non-terminal goals (`ACCEPTED`/`EXECUTING`) when `now - last_update > 2 * max_age` with a WARN log. `OperationManager` gained an optional `LogSink` constructor argument (same pattern as `LogManager`) to preserve observability while keeping the manager middleware-neutral.

mfaferek93 (approved):

PR #398, opened 2026-05-13
feat(gateway): add GET /apps/{id}/belongs-to discovery endpoint

### LINKED ISSUE #196: Implement `GET /apps/{app-id}/belongs-to` endpoint
## Summary

Add a new discovery endpoint that returns the Area to which a given App belongs. This is determined by finding the App's parent Component and then looking up which Area that Component belongs to.

---

### `GET /api/v1/apps/{app-id}/belongs-to`

Returns the Area that contains this App (via its parent Component's area assignment).

**Path parameters:**

| `app-id` | `string` | Yes | App identifier (alphanumeric, underscores, hyphens, max 256 chars) |

**Request body:** None

**Response `200 OK`:**

```json
{
  "items": [
    {
      "id": "powertrain",
      "name": "Powertrain",
      "href": "/api/v1/areas/powertrain"
    }
  ],
  "_links": {
    "self": "/api/v1/apps/{app-id}/belongs-to",
    "app": "/api/v1/apps/{app-id}"
  }
}
```

The `items` array will contain exactly **0 or 1** area. It may be empty if:
- The App has no parent Component (`component_id` is empty)
- The parent Component has no assigned Area

**Response `400 Bad Request`:**

Returned when the `app-id` format is invalid.

```json
{
  "error_code": "invalid-parameter",
  "message": "Invalid app ID",
  "parameters": {
    "app_id": "bad id!"
  }
}
```

**Response `404 Not Found`:**

Returned when no App with the given ID exists.

```json
{
  "error_code": "entity-not-found",
  "message": "App not found",
  "parameters": {
    "app_id": "nonexistent_app"
  }
}
```

---

### Logic flow

1. Validate `app-id` format
2. Look up the App via `DiscoveryManager::get_app(app_id)` → 404 if not found
3. Get `app.component_id` → if empty, return empty `items` array
4. Look up the Component via `DiscoveryManager::get_component(component_id)`
5. Get `component.area` → if empty, return empty `items` array
6. Look up the Area via `DiscoveryManager::get_area(area_id)` to get its name
7. Return the Area as an `EntityReference`

### Existing patterns to follow

- Follow the same handler structure as `handle_app_depends_on` and `handle_get_hosts` in `discovery_handlers.cpp`
- Use the standard relationship response envelope

### Route registration

Register in `rest_server.cpp::setup_routes()`:

```cpp
srv->Get((api_path("/apps") + R"(/([^/]+)/belongs-to$)"), handler);
```

### Handler method

Add to `DiscoveryHandlers`:
```cpp
void handle_app_belongs_to(const httplib::Request& req, httplib::Response& res);
```

### Tests

- Unit test: verify correct area returned for app with known component→area chain
- Unit test: verify empty items when app has no component
- Unit test: verify empty items when app's component has no area
- Unit test: 404 for nonexistent app
- Integration test: launch demo nodes, query the endpoint

bburda:
## Summary

Implements `GET /api/v1/apps/{app-id}/belongs-to`. Returns the area that
contains an app, resolved via the app's parent component and that
component's area assignment.

The `belongs-to` resource was already listed in app capabilities; this PR
wires it up with a real handler, route, capability URI on `GET /apps/{id}`,
unit + integration tests, and the matching `REQ_INTEROP_106` spec entry.

Pattern mirrors the existing relationship endpoints (`is-located-on`,
`depends-on`): collection wrapper with 0 or 1 element, `x-medkit.missing=true`
on the item when an area id cannot be resolved.

---

## Issue

- closes #196

---

## Type

- [ ] Bug fix
- [x] New feature or tests
- [ ] Breaking change
- [ ] Documentation only

---

## Testing

- Unit: 6 new cases in `test_discovery_handlers.cpp` covering happy path,
  no-component, no-area, missing-area, invalid id, unknown app. Plus
  `belongs-to` URI assertion in the existing `GetAppReturnsLinksAndCapabilities`
  and `supports_resource("belongs-to")` for the APP capability set.
- Integration: 2 new cases in `test_hateoas.test.py` (happy path against
  demo manifest, 404 for unknown app).
- Full local unit suite (2130 tests): green.
- Linters (clang-format, ament_copyright, cpplint, etc.) + clang-tidy on
  changed files: clean.
- Sphinx docs build: clean apart from the unrelated auto-generated
  `verification.rst` path.

Also bumps the `TriggerManagerTest.LifetimeExpiry` sleep margin in a
separate commit. The test created a 1 s trigger and slept 1500 ms; under
parallel CTest load the 500 ms margin past expiry was occasionally short
and the trigger still reported active. Bumped to 2500 ms.

---

## Checklist

### DIFF
--- a/src/ros2_medkit_gateway/src/http/handlers/discovery_handlers.cpp
+++ b/src/ros2_medkit_gateway/src/http/handlers/discovery_handlers.cpp
@@ -982,6 +982,16 @@ void DiscoveryHandlers::handle_get_app(const httplib::Request & req, httplib::Re
     using Cap = CapabilityBuilder::Capability;
     std::vector<Cap> caps = {Cap::DATA, Cap::OPERATIONS, Cap::CONFIGURATIONS,       Cap::FAULTS,
                              Cap::LOGS, Cap::BULK_DATA,  Cap::CYCLIC_SUBSCRIPTIONS, Cap::TRIGGERS};
+    // Relationship endpoints are gated the same way as the top-level URI keys
+    // above so the three advertising surfaces (top-level URIs, `_links`,
+    // `capabilities` array) describe the same set of available collections.
+    if (!app.component_id.empty()) {
+      caps.push_back(Cap::IS_LOCATED_ON);
+      caps.push_back(Cap::BELONGS_TO);
+    }
+    if (!app.depends_on.empty()) {
+      caps.push_back(Cap::DEPENDS_ON);
+    }
     if (ctx_.node()->get_script_manager() && ctx_.node()->get_script_manager()->has_backend()) {
       caps.push_back(Cap::SCRIPTS);
     }
@@ -1034,14 +1044,15 @@ void DiscoveryHandlers::handle_app_depends_on(const httplib::Request & req, http
 
     std::string app_id = req.matches[1];
 
-    auto validation_result = ctx_.validate_entity_id(app_id);
-    if (!validation_result) {
-      HandlerContext::send_error(res, 400, ERR_INVALID_PARAMETER, "Invalid app ID",
-                                 {{"details", validation_result.error()}, {"app_id", app_id}});
+    // validate_entity_for_route forwards the request to the owning peer when
+    // the app is remote (aggregation setup) - validate_entity_id alone would
+    // resolve locally only and return 404 for remote apps.
+    auto entity_opt = ctx_.validate_entity_for_route(req, res, app_id);
+    if (!entity_opt) {
       return;
     }
 
-    // Cache-first lookup: EntityCache has merged entities from peers
+    // Local entity: fetch full App model for relationship data.
     const auto & cache = ctx_.node()->get_thread_safe_cache();
     auto app_opt = cache.get_app(app_id);
     if (!app_opt) {
@@ -1109,49 +1120,62 @@ void DiscoveryHandlers::handle_app_belongs_to(const httplib::Request & req, http
 
     std::string app_id = req.matches[1];
 
-    auto validation_result = ctx_.validate_entity_id(app_id);
-    if (!validation_result) {
-      HandlerContext::send_error(res, 400, ERR_INVALID_PARAMETER, "Invalid app ID",
-                                 {{"details", validation_result.error()}, {"app_id", app_id}});
+    // Forward to peer if app is remote (aggregation setup); locally
+    // resolved entity falls through to the cache/discovery lookup below.
+    auto entity_opt = ctx_.validate_entity_for_route(req, res, app_id);
+    if (!entity_opt) {
       return;
     }
 
-    // Cache-first lookup: EntityCache has merged entities from peers
+    // Atomic snapshot avoids a mixed-generation view of App -> Component ->
+    // Area across concurrent cache refreshes. Fall back to the discovery
+    // manager (no atomic guarantee, but rare path) when the app is not in
+    // the cache.
     const auto & cache = ctx_.node()->get_thread_safe_cache();
-    auto app_opt = cache.get_app(app_id);
-    if (!app_opt) {
+    auto snapshot = cache.get_app_with_links(app_id);
+    if (!snapshot.app) {
       auto discovery = ctx_.node()->get_discovery_manager();
-      app_opt = discovery->get_app(app_id);
+      snapshot.app = discovery->get_app(app_id);
+      if (snapshot.app && !snapshot.app->component_id.empty()) {
+        snapshot.component = discovery->get_component(snapshot.app->component_id);
+        if (snapshot.component && !snapshot.component->area.empty()) {
+          snapshot.area = discovery->get_area(snapshot.component->area);
+        }
+      }
     }
 
-    if (!app_opt) {
+    if (!snapshot.app) {
       HandlerContext::send_error(res, 404, ERR_ENTITY_NOT_FOUND, "App not found", {{"app_id", app_id}});
       return;
     }
 
     json items = json::array();
-    const auto & app = *app_opt;
+    const auto & app = *snapshot.app;
 
     if (!app.component_id.empty()) {
-      auto component_opt = cache.get_component(app.component_id);
-      if (!component_opt) {
-        auto discovery = ctx_.node()->get_discovery_manager();
-        component_opt = discovery->get_component(app.component_id);
-      }
-
-      if (component_opt && !component_opt->area.empty()) {
-        const auto & area_id = component_opt->area;
+      if (!snapshot.component) {
+        // Mirror handle_app_is_located_on: surface broken parent reference
+        // with x-medkit.missing=true so HATEOAS clients can distinguish
+        // 'no parent area' from 'manifest broken, component gone'.
+        json item;
+        item["id"] = "";
+        item["name"] = "<unknown area>";
+        item["href"] = "";
+        XMedkit ext;
+        ext.add("missing", true);
+        ext.add("unresolved_component", app.component_id);
+        item["x-medkit"] = ext.build();
+        items.push_back(item);
+        RCLCPP_WARN(HandlerContext::logger(), "App '%s' belongs-to unresolvable: parent component '%s' is unknown",
+                    app_id.c_str(), app.component_id.c_str());
+      } else if (!snapshot.component->area.empty()) {
+        const auto & area_id = snapshot.component->area;
         json item;
         item["id"] = area_id;
         item["href"] = "/api/v1/areas/" + area_id;
 
-        auto area_opt = cache.get_area(area_id);
-        if (!area_opt) {
-          auto discovery = ctx_.node()->get_discovery_manager();
-          area_opt = discovery->get_area(area_id);
-        }
-        if (area_opt) {
-          item["name"] = area_opt->name.empty() ? area_id : area_opt->name;
+        if (snapshot.area) {
+          item["name"] = snapshot.area->name.empty() ? area_id : snapshot.area->name;
         } else {
           item["name"] = area_id;
           XMedkit ext;
@@ -1162,6 +1186,9 @@ void DiscoveryHandlers::handle_app_belongs_to(const httplib::Request & req, http
         }
         items.push_back(item);
       }
+      // If component resolves but has no area_id assigned, items stays
+      // empty - that is a legitimate manifest configuration (component
+      // without parent area), not a broken reference.
     }
 
     json response;
@@ -1192,40 +1219,39 @@ void DiscoveryHandlers::handle_app_is_located_on(const httplib::Request & req, h
 
     std::string app_id = req.matches[1];
 
-    auto validation_result = ctx_.validate_entity_id(app_id);
-    if (!validation_result) {
-      HandlerContext::send_error(res, 400, ERR_INVALID_PARAMETER, "Invalid app ID",
-                                 {{"details", validation_result.error()}, {"app_id", app_id}});
+    // Forward to peer if app is remote (aggregation setup).
+    auto entity_opt = ctx_.validate_entity_for_route(req, res, app_id);
+    if (!entity_opt) {
       return;
     }
 
-    // Cache-first lookup: EntityCache has merged entities from peers
+    // Atomic snapshot keeps app -> component consistent across a concurrent
+    // cache refresh. Fall back to discovery manager when the app is not in
+    // the cache.
     const auto & cache = ctx_.node()->get_thread_safe_cache();
-    auto app_opt = cache.get_app(app_id);
-    if (!app_opt) {
+    auto snapshot = cache.get_app_with_links(app_id);
+    if (!snapshot.app) {
       auto discovery = ctx_.node()->get_discovery_manager();
-      app_opt = discovery->get_app(app_id);
+      snapshot.app = discovery->get_app(app_id);
+      if (snapshot.app && !snapshot.app->component_id.empty()) {
+        snapshot.component = discovery->get_component(snapshot.app->component_id);
+      }
     }
 
-    if (!app_opt) {
+    if (!snapshot.app) {
       HandlerContext::send_error(res, 404, ERR_ENTITY_NOT_FOUND, "App not found", {{"app_id", app_id}});
       return;
     }
 
     json items = json::array();
-    const auto & app = *app_opt;
+    const auto & app = *snapshot.app;
 
     if (!app.component_id.empty()) {
-      auto component_opt = cache.get_component(app.component_id);
-      if (!component_opt) {
-        auto discovery = ctx_.node()->get_discovery_manager();
-        component_opt = discovery->get_component(app.component_id);
-      }
-      if (component_opt) {
+      if (snapshot.component) {
         json item;
-        item["id"] = component_opt->id;
-        item["name"] = component_opt->name.empty() ? component_opt->id : component_opt->name;
-        item["href"] = "/api/v1/components/" + component_opt->id;
+        item["id"] = snapshot.component->id;
+        item["name"] = snapshot.component->name.empty() ? snapshot.component->id : snapshot.component->name;
+        item["href"] = "/api/v1/components/" + snapshot.component->id;
         items.push_back(item);
       } else {
         json item;
--- a/src/ros2_medkit_gateway/src/http/handlers/discovery_handlers.cpp
+++ b/src/ros2_medkit_gateway/src/http/handlers/discovery_handlers.cpp
@@ -972,6 +972,7 @@ void DiscoveryHandlers::handle_get_app(const httplib::Request & req, httplib::Re
 
     if (!app.component_id.empty()) {
       response["is-located-on"] = "/api/v1/components/" + app.component_id;
+      response["belongs-to"] = base_uri + "/belongs-to";
     }
 
     if (!app.depends_on.empty()) {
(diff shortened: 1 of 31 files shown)


src/ros2_medkit_gateway/test/test_trigger_manager.cpp
```diff
@@ -309,15 +309,16 @@ TEST_F(TriggerManagerTest, Multishot_NotifyTwice) {
...
   auto created = manager_->create(req);
   ASSERT_TRUE(created.has_value());
 
   EXPECT_TRUE(manager_->is_active(created->id));
 
-  // Sleep past expiry with margin for CPU contention during parallel testing
-  std::this_thread::sleep_for(std::chrono::milliseconds(1500));
+  // Sleep past expiry with a generous margin: parallel test execution
+  // and CI load can stall the scheduler well past 500 ms after a 1 s lifetime.
+  std::this_thread::sleep_for(std::chrono::milliseconds(2500));
```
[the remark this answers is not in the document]
bburda: Switched to a bounded polling loop with a 5 s deadline and 50 ms tick. On a fast machine the test exits as soon as `is_active` flips (~1 s); the deadline only kicks in if the scheduler stalls under parallel test load. Fixed in edcd41df.

src/ros2_medkit_gateway/test/test_entity_resource_model.cpp
```diff
@@ -171,6 +171,7 @@ TEST(EntityCapabilities, ComponentSupportsOperations) {
 TEST(EntityCapabilities, AppSupportsIsLocatedOn) {
   auto caps = EntityCapabilities::for_type(SovdEntityType::APP);
   EXPECT_TRUE(caps.supports_resource("is-located-on"));
+  EXPECT_TRUE(caps.supports_resource("belongs-to"));
   EXPECT_TRUE(caps.supports_collection(ResourceCollection::OPERATIONS));
```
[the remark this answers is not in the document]
bburda: Split into two tests: `AppSupportsIsLocatedOn` keeps the original assertions and a new `AppSupportsBelongsTo` covers the new capability. Fixed in edcd41df.

src/ros2_medkit_gateway/src/http/handlers/discovery_handlers.cpp
```diff
@@ -972,6 +972,7 @@ void DiscoveryHandlers::handle_get_app(const httplib::Request & req, httplib::Re
 
     if (!app.component_id.empty()) {
       response["is-located-on"] = "/api/v1/components/" + app.component_id;
+      response["belongs-to"] = base_uri + "/belongs-to";
```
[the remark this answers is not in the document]
bburda: Good catch. Added `links.add("belongs-to", base_uri + "/belongs-to")` next to `is-located-on`, gated on the same `component_id` condition. Extended `GetAppReturnsLinksAndCapabilities` to assert it. Fixed in 592fce94.

mfaferek93 (commented):
A few comments :)

src/ros2_medkit_gateway/src/http/rest_server.cpp
```diff
@@ -1131,6 +1131,18 @@ void RESTServer::setup_routes() {
...
+              [this](auto & req, auto & res) {
+                discovery_handlers_->handle_app_belongs_to(req, res);
+              })
+          .tag("Discovery")
+          .summary("Get app parent area")
+          .description(
+              "Returns the area this app belongs to via its parent component, as a 0-or-1 element "
+              "collection.")
+          .response(200, "Parent area", SB::ref("EntityList"))
+          .operation_id("getAppArea");
```
mfaferek93, line 1144: **Blocker - RBAC missing.** `auth_config.cpp` has no `GET:/api/v1/apps/*/belongs-to` entry for viewer (L49), operator (L188), or configurator (L388). The `apps/*` wildcard at L39 is single-segment (`auth_requirement_policy.cpp:111` maps `*` to `[^/]+`), so it doesn't cover `apps/X/belongs-to`. Only ADMIN has the `apps/**` fallback at L600. Result: every role except ADMIN gets 403 on this endpoint. Add 3 entries + an integration test for an authenticated viewer.

src/ros2_medkit_gateway/src/http/handlers/discovery_handlers.cpp
```diff
@@ -1098,6 +1100,89 @@ void DiscoveryHandlers::handle_app_depends_on(const httplib::Request & req, http
...
+        } else {
+          item["name"] = area_id;
+          XMedkit ext;
+          ext.add("missing", true);
+          item["x-medkit"] = ext.build();
+          RCLCPP_WARN(HandlerContext::logger(), "App '%s' belongs to unknown area '%s' (via component '%s')",
+                      app_id.c_str(), area_id.c_str(), app.component_id.c_str());
+        }
+        items.push_back(item);
+      }
```
mfaferek93, line 1193: **Blocker - asymmetry hides broken refs.** `handle_app_is_located_on:1235-1244` emits a `x-medkit.missing=true` item + `RCLCPP_WARN` when `component_opt` is empty. Here, when `app.component_id` is non-empty but the component can't be resolved (or `component->area` is empty), we silently return `items=[]`. Client can't distinguish 'app has no parent area' from 'manifest broken, component gone'. Add the symmetric else-branch + a unit test mirroring `AppIsLocatedOnReturnsMissingItemWhenHostComponentUnresolved`.

src/ros2_medkit_gateway/src/http/handlers/discovery_handlers.cpp
mfaferek93, line 975: **Blocker - `capabilities` array missing `belongs-to`.** Added to `_links` (L998) and top-level response (here), but the `caps` vector at L983 doesn't include it - `CapabilityBuilder::Capability` enum (`capability_builder.hpp:43`) has no `BELONGS_TO`/`IS_LOCATED_ON` variant. Clients reading `response.capabilities` (the canonical SOVD discovery surface) won't see the new endpoint. Three parallel capability surfaces (`capabilities`, `_links`, `EntityCapabilities`) disagree. Add `BELONGS_TO` to the enum + include it in `caps` when `!app.component_id.empty()`.

src/ros2_medkit_gateway/src/http/handlers/discovery_handlers.cpp
```diff
@@ -994,6 +995,7 @@ void DiscoveryHandlers::handle_get_app(const httplib::Request & req, httplib::Re
     links.self("/api/v1/apps/" + app.id).collection("/api/v1/apps");
     if (!app.component_id.empty()) {
       links.add("is-located-on", "/api/v1/components/" + app.component_id);
+      links.add("belongs-to", base_uri + "/belongs-to");
```
mfaferek93, line 1008: **Major - duplicated link logic vs `App::to_json`.** `models/app.cpp:127,130` serializes `is-located-on` and `depends-on` but not `belongs-to`. Any caller of `to_json` (notifications, snapshots, change events) will see a different shape than this handler. Either add `belongs-to` to `App::to_json` (same `!component_id.empty()` guard) or extract link emission into a shared helper called from both sites.

src/ros2_medkit_gateway/test/test_discovery_handlers.cpp
```diff
@@ -689,6 +691,112 @@ TEST_F(DiscoveryHandlersFixtureTest, AppIsLocatedOnUnknownAppReturns404) {
...
+  EXPECT_EQ(body["items"][0]["id"], "sensors");
+  EXPECT_EQ(body["items"][0]["name"], "Sensors");
+  EXPECT_EQ(body["items"][0]["href"], "/api/v1/areas/sensors");
+  EXPECT_EQ(body["x-medkit"]["total_count"], 1);
+  EXPECT_EQ(body["_links"]["self"], "/api/v1/apps/mapper/belongs-to");
+  EXPECT_EQ(body["_links"]["app"], "/api/v1/apps/mapper");
+}
+
+// @verifies REQ_INTEROP_106
+TEST_F(DiscoveryHandlersFixtureTest, AppBelongsToReturnsEmptyWhenAppHasNoComponent) {
```
mfaferek93, line 712: **Major - HATEOAS discovery gap for standalone apps.** This test proves the endpoint returns 200 + empty items for an app without a component. But `handle_get_app:996` only adds the `belongs-to` link when `!app.component_id.empty()`. HATEOAS-only clients hitting `GET /apps/standalone` won't see a link to follow. Either always emit the link (content can be empty) or document that standalone apps don't expose the subresource.

src/ros2_medkit_gateway/src/http/rest_server.cpp
```diff
@@ -1131,6 +1131,18 @@ void RESTServer::setup_routes() {
           .response(200, "Host component(s)", SB::ref("EntityList"))
           .operation_id("getAppHost");
 
+      reg.get(entity_path + "/belongs-to",
```
mfaferek93, line 1134: **Major - app-only route, but capability claim covers more types.** `entity_capabilities.cpp:33` (SERVER) and `:59` (COMPONENT) advertise `belongs-to` in `resources_`, but no `/api/v1/servers/{id}/belongs-to` or `/api/v1/components/{id}/belongs-to` route exists. `supports_resource("belongs-to")` on those types is a lie - client follows it, hits 404. Either add the routes (mirror this PR's app handler) or drop the entry from `resources_` for SERVER/COMPONENT.

src/ros2_medkit_gateway/src/http/handlers/discovery_handlers.cpp
```diff
@@ -1098,6 +1100,89 @@ void DiscoveryHandlers::handle_app_depends_on(const httplib::Request & req, http
...
+    auto validation_result = ctx_.validate_entity_id(app_id);
+    if (!validation_result) {
+      HandlerContext::send_error(res, 400, ERR_INVALID_PARAMETER, "Invalid app ID",
+                                 {{"details", validation_result.error()}, {"app_id", app_id}});
+      return;
+    }
+
+    // Cache-first lookup: EntityCache has merged entities from peers
+    const auto & cache = ctx_.node()->get_thread_safe_cache();
+    auto app_opt = cache.get_app(app_id);
```
mfaferek93: **Major - mixed-generation snapshot.** Three sequential `cache.get_*` lookups here (L1121), L1136, L1148 take independent locks (`thread_safe_entity_cache.cpp:135/144/153`), plus per-call fallback to `discovery_manager`. Concurrent cache refresh can yield app from gen N, component from N+1, area from N - and the discovery fallback can resurrect stale manifest data the cache deliberately dropped. Same shape exists in `is-located-on` and `depends-on`. Worth a follow-up issue: atomic `get_app_with_relationships()` or generation-token retry.

src/ros2_medkit_gateway/src/http/handlers/discovery_handlers.cpp
```diff
@@ -1098,6 +1100,89 @@ void DiscoveryHandlers::handle_app_depends_on(const httplib::Request & req, http
...
+void DiscoveryHandlers::handle_app_belongs_to(const httplib::Request & req, httplib::Response & res) {
+  try {
+    if (req.matches.size() < 2) {
+      HandlerContext::send_error(res, 400, ERR_INVALID_REQUEST, "Invalid request");
+      return;
+    }
+
+    std::string app_id = req.matches[1];
+
+    auto validation_result = ctx_.validate_entity_id(app_id);
```
mfaferek93: **Major - no peer forwarding for remote entities.** `handle_get_app:925` uses `validate_entity_for_route`, which routes through `handler_context.cpp:273-277` aggregation forwarding when `entity_info.is_remote && aggregation_mgr_`. This handler uses only `validate_entity_id` (format check). In a multi-gateway/aggregation setup, `/apps/<remote-app>/belongs-to` returns 404 locally instead of forwarding to the owning peer. `handle_app_is_located_on:1195` and `handle_app_depends_on:1037` have the same gap - systemic, but this PR inherits it. Switch to `validate_entity_for_route` (or a variant that allows the forwarding decision without type-coercion errors).

src/ros2_medkit_gateway/test/test_trigger_manager.cpp
```diff
@@ -309,15 +309,23 @@ TEST_F(TriggerManagerTest, Multishot_NotifyTwice) {
...
   ASSERT_TRUE(created.has_value());
 
   EXPECT_TRUE(manager_->is_active(created->id));
 
-  // Sleep past expiry with margin for CPU contention during parallel testing
-  std::this_thread::sleep_for(std::chrono::milliseconds(1500));
+  // Poll for expiry: finishes quickly on a fast machine but tolerates
+  // scheduler stalls under parallel test load. Deadline is well past the
+  // 1 s lifetime so a busy CI runner does not race the check.
+  const auto deadline = std::chrono::steady_clock::now() + std::chrono::milliseconds(5000);
```
mfaferek93, line 322: **Major - redundant churn with earlier commit on same hunk.** `94f905fc` bumped the fixed sleep 1500ms -> 2500ms; this commit replaces the fixed sleep with the polling loop. Two commits moving the same mechanism in opposite directions clutter the history and make `git blame` misleading. Squash `94f905fc` out (the polling version stands on its own).

src/ros2_medkit_integration_tests/test/features/test_hateoas.test.py
```diff
@@ -461,6 +461,44 @@ def test_is_located_on_apps_nonexistent(self):
...
+        self.assertLessEqual(len(data['items']), 1)
+
+        # engine-temp-sensor -> temp-sensor-hw (component) -> engine (area)
+        self.assertEqual(len(data['items']), 1)
+        area = data['items'][0]
+        self.assertEqual(area['id'], 'engine')
+        self.assertIn('name', area)
+        self.assertEqual(area['href'], '/api/v1/areas/engine')
+
+    def test_belongs_to_apps_nonexistent(self):
```
mfaferek93, line 482: **Major - thin integration coverage.** Only happy path + 404. Missing: standalone app (200+empty), component without area (200+empty), unresolved component reference (200+missing item; ties into the unit-test gap flagged on the handler), unresolved area (covered in unit tests but not integration). Also: both new tests use raw `requests.get()` with `timeout=10` instead of `self.poll_endpoint()` from the base class - the helper adds retry, which matters under CI sanitizer load given the gateway-shutdown timing changes elsewhere in this PR.

src/ros2_medkit_gateway/test/test_discovery_handlers.cpp
bburda: Documented in 97c654c8 (docs/api/rest.rst). Per SOVD ISO 17978-3 §7.6 "Each URI reference is only present if the entity supports that collection" - standalone apps deliberately do not expose the belongs-to URI, the absent link IS the spec-conformant HATEOAS signal. Added an explicit note in the API docs so HATEOAS-only clients know to expect absence rather than an empty collection.

src/ros2_medkit_gateway/src/http/rest_server.cpp
bburda: Fixed in 4727fe08 - added `GET:/api/v1/apps/*/belongs-to` to viewer, operator and configurator sections in auth_config.cpp. Also seeded the matching `apps/*/is-located-on` entries in the same three sections while I was there - they were missing as a pre-existing gap.

src/ros2_medkit_gateway/src/http/handlers/discovery_handlers.cpp
bburda: Fixed in 97c654c8. Added `BELONGS_TO` and `IS_LOCATED_ON` to `CapabilityBuilder::Capability` (the latter was also missing as a pre-existing gap). The `caps` vector in handle_get_app pushes both under the same `component_id` gate as the existing top-level URI / `_links` emission, so all three advertising surfaces describe the same set of collections.

src/ros2_medkit_gateway/src/http/handlers/discovery_handlers.cpp
bburda: Fixed in 97c654c8. Linked function is actually `App::to_capabilities(base_url)` (not `to_json()` which is the no-arg generic serializer) - added `belongs-to` there under the same `!component_id.empty()` guard as the existing `is-located-on`. Notification / snapshot / change-event consumers that go through the model now see the same shape as the handler. Tests `ToCapabilities_ContainsRelationshipLinks` and `ToCapabilities_OmitsRelationshipLinksWhenAppIsStandalone` cover happy and standalone paths.

src/ros2_medkit_gateway/src/http/handlers/discovery_handlers.cpp
bburda: Fixed in aaef97c5. handle_app_belongs_to now emits an item with `x-medkit.missing=true` plus `x-medkit.unresolved_component=<id>` and a `RCLCPP_WARN` when the parent component does not resolve, mirroring handle_app_is_located_on. A component that resolves but has no `area_id` assigned still returns `items=[]` - that is a legit manifest configuration, not a broken reference. Unit test `AppBelongsToReturnsMissingItemWhenParentComponentUnresolved` added.

src/ros2_medkit_gateway/src/http/handlers/discovery_handlers.cpp
bburda: Fixed in aaef97c5. `handle_app_belongs_to`, `handle_app_is_located_on` and `handle_app_depends_on` all use `validate_entity_for_route` now, so a remote app in an aggregation setup gets forwarded via `aggregation_mgr_` instead of returning 404 locally. The 404 payload shape changes to the generic `Entity not found` / `parameters.entity_id` form that `validate_entity_for_route` emits; `test_hateoas` was updated to match.

src/ros2_medkit_gateway/src/http/handlers/discovery_handlers.cpp
bburda: Fixed in aaef97c5 and dfc4dae2. Added `ThreadSafeEntityCache::get_app_with_links()` (app → component → area, single shared_lock) and `get_app_with_dependencies()` (app + N dependency apps, single shared_lock). `handle_app_belongs_to`, `handle_app_is_located_on` and `handle_app_depends_on` all switched over - all three relationship paths now resolve in a single cache generation. Discovery-manager fallback path preserves the old shape for the rare case where the entity is not yet cached.

src/ros2_medkit_integration_tests/test/features/test_hateoas.test.py
bburda: Extended in dfc4dae2 and c220212a. The 200-path tests now use `self.poll_endpoint()` (retry under sanitizer load) and the 404 shape was updated to the generic `validate_entity_for_route` payload. Four new integration tests added for the standalone-app and component-without-area branches on both `/belongs-to` and `/is-located-on`, backed by new manifest fixtures (`hateoas-edge-area`, `hateoas-component-no-area`, `hateoas-app-standalone`, `hateoas-app-on-area-less-component`). The unresolved-component and dangling-area scenarios stay as unit tests (`AppBelongsToReturnsMissingItemWhenParentComponentUnresolved`, `AppBelongsToReturnsMissingItemWhenAreaUnresolved`) - manifest validators R007 and R006 reject dangling references and fall back to runtime_only discovery, which would disable the rest of the suite if seeded as integration fixtures.

src/ros2_medkit_gateway/test/test_trigger_manager.cpp
bburda: Squashed. `94f905fc` is gone from the branch history - `git rebase --onto 94f905fc^ 94f905fc` with a manual merge against `edcd41df` so the polling loop is the only state visible. Branch is now 10 commits, no 1500ms-then-2500ms intermediate version.

src/ros2_medkit_gateway/src/http/rest_server.cpp
bburda: Fixed in 97c654c8 by dropping `"belongs-to"` from `EntityCapabilities::resources_` for both SERVER (entity_capabilities.cpp:33) and COMPONENT (line 59). SOVD ISO 17978-3 §7.6 defines `/belongs-to` only for `/apps`; advertising it on the other types was a 404 promise. Added inline comments citing the spec on both rows.

mfaferek93 (approved):

PR #400, opened 2026-05-14
feat(gateway): expose entity context on /faults/stream SSE events

### LINKED ISSUE #380: Global /faults/stream SSE events lack entity context for downstream consumers

bburda:
# Pull Request

## Summary

`GET /api/v1/faults/stream` event payloads now carry an optional `x-medkit` SOVD payload-extension object with `entity_type` and `entity_id` fields when the gateway can resolve the fault's first `reporting_sources` entry back to a SOVD entity. Consumers can then hit `/{entity_type}/{entity_id}/bulk-data/rosbags/{fault_code}` directly, replacing the O(N) HEAD-probe workaround that external integrations need today.

Nested under `x-medkit` per the SOVD payload-extension convention (matches existing `x-medkit.aggregation_level`, `x-medkit.phase`, etc.). Flat `x-medkit-*` names are reserved for endpoint paths (`/x-medkit-graph`) and error codes, not payload fields.

Resolution mirrors the existing `TriggerFaultSubscriber` / `LogManager` pattern:
- Manifest / hybrid mode: cache's node-to-app linking index (with and without leading `/`).
- Runtime-only fallback: ROS FQN's last segment, only emitted when an App with that id actually exists in the cache - we never point consumers at a 404.
- No match: `x-medkit` object omitted entirely (backward-compatible for existing SOVD consumers).

Resolution is snapshotted in `on_fault_event` and stored alongside the buffered event (new `QueuedEvent` struct), so a discovery refresh between enqueue and stream-out cannot retroactively flip the entity reported to consumers, and the format path stays lock-free wrt the entity cache.

Example payload:

```json
{
  "event_type": "fault_confirmed",
  "fault": {"fault_code": "MOTOR_OVERHEAT", "...": "..."},
  "timestamp": 1735830000.123,
  "x-medkit": {"entity_type": "apps", "entity_id": "motor_controller"}
}
```

---

## Issue

- closes #380

---

## Type

- [ ] Bug fix
- [x] New feature or tests
- [ ] Breaking change
- [ ] Documentation only

---

## Testing

- 4 new unit tests in `test_sse_fault_handler.cpp` covering: manifest-app match, runtime-fallback match, no-match (entity absent from cache), and empty `reporting_sources`.
- Full SSE fault handler suite (15 tests) green.
- Full gateway unit suite (2213 tests, 0 failures) green when run serially.
- clang-format / clang-tidy / ament-copyright clean on touched files.

---

## Checklist


src/ros2_medkit_gateway/src/http/handlers/sse_fault_handler.cpp
```diff
@@ -230,24 +235,89 @@ size_t SSEFaultHandler::connected_clients() const {
...
 
   return sse.str();
 }
 
+std::optional<SSEFaultHandler::EntityContext>
+SSEFaultHandler::resolve_entity_context(const ros2_medkit_msgs::msg::Fault & fault) const {
+  if (fault.reporting_sources.empty()) {
+    return std::nullopt;
+  }
+  const auto & raw_fqn = fault.reporting_sources.front();
```
mfaferek93, line 277: `reporting_sources` is a set and a debounced fault can carry several reporters (e.g. node_a + node_b on the same fault_code). `.front()` is the lexicographically-first one, not a defined owner, so x-medkit surfaces one of N valid apps by sort order. Probably fine since any co-reporter's bag is fetchable, but worth a one-line comment - "first reporting source" reads like it implies ownership.

src/ros2_medkit_gateway/src/http/handlers/sse_fault_handler.cpp
```diff
@@ -230,24 +235,89 @@ size_t SSEFaultHandler::connected_clients() const {
...
+
+  if (entity_id.empty()) {
+    RCLCPP_DEBUG(HandlerContext::logger(),
+                 "SSE fault event: no entity match for reporting source '%s' (fault_code='%s'); "
+                 "omitting x-medkit extension",
+                 raw_fqn.c_str(), fault.fault_code.c_str());
+    return std::nullopt;
+  }
+
+  return EntityContext{"apps", std::move(entity_id)};
```
mfaferek93, line 329: entity_type hardcoded to "apps" is correct (apps are the leaf reporter; components own faults transitively via their hosted apps), but it isn't obvious here. One line of comment would stop someone "fixing" it later, and note that manifest-only components (no node) have no match and fall back to discovery by design.

src/ros2_medkit_gateway/test/test_sse_fault_handler.cpp
```diff
@@ -415,6 +419,199 @@ TEST_F(SSEFaultHandlerTest, NonPositiveKeepaliveOverrideLogsWarning) {
...
+  auto payload = parse_sse_payload(output);
+
+  ASSERT_TRUE(payload.contains("x-medkit")) << payload.dump();
+  EXPECT_EQ(payload["x-medkit"]["entity_type"], "apps");
+  EXPECT_EQ(payload["x-medkit"]["entity_id"], "temp_sensor");
+
+  release_stream(res);
+}
+
+TEST_F(SSEFaultHandlerTest, StreamEmitsXMedkitForManifestApp) {
```
mfaferek93, line 476: Two suite notes: (1) no test pins manifest-over-runtime precedence - seed both a node_to_app entry and a colliding runtime app and assert the manifest id wins, otherwise that ordering contract is unguarded. (2) the suite shares the default DDS domain and a deterministic fault topic (only the publisher node name is randomized), which is why it has to run serially - a per-test ROS_DOMAIN_ID or unique topic would make it parallel-safe.

src/ros2_medkit_gateway/src/http/handlers/sse_fault_handler.cpp
bburda: Added a four-line comment at the front() call clarifying that reporting_sources is a set and the picked FQN is lexicographically first, not authoritative (co-reporter rosbag is still fetchable). Pushed in 30141281.

src/ros2_medkit_gateway/src/http/handlers/sse_fault_handler.cpp
bburda: Added a six-line comment at the EntityContext{"apps", ...} return explaining why the type is fixed (apps are the leaf reporters; components own faults transitively via their hosted apps) and noting the manifest-only-without-bound-node fallback is by design. Pushed in 30141281.

src/ros2_medkit_gateway/test/test_sse_fault_handler.cpp
bburda: Both addressed:

1. Added StreamPrefersManifestMatchOverRuntimeFallback - seeds a node_to_app entry for /bridge/diagnostic_bridge -> manifest-bridge and a colliding runtime App with id diagnostic_bridge, asserts the manifest id wins.
2. Set a per-test fault_manager.namespace ("test<N>") in SetUp so each TEST_F resolves to a unique /test<N>/fault_manager/events topic. The publisher node name uses the same counter, so the fixture no longer shares the deterministic events topic across tests.

Pushed in 30141281.

mfaferek93 (approved):
LGTM!

PR #402, opened 2026-05-15
fix(gateway): drive fault transport with its own executor

### LINKED ISSUE #399: Refactor Ros2FaultServiceTransport to use spin_until_future_complete to remove client/spin-thread race
## Summary

`Ros2FaultServiceTransport::get_snapshots()` and `get_rosbag()` use a blocking
`future.wait_for()` pattern that requires the rclcpp Client to be spun on a
separate executor thread. Under ThreadSanitizer this produces flaky data race
warnings between the executor thread cleaning up the client's pending request
and the main thread destroying the local `response` shared_ptr at the end of
the transport function. See https://github.com/selfpatch/ros2_medkit/actions/runs/25823593340/job/75870897762
for the failing run on PR #398; the stack hits `rclcpp::Executor::execute_client`
in the spin thread and `~shared_ptr<Response>` at `ros2_fault_service_transport.cpp:361`
in the main thread.

PR #398 mitigates the flake test-side by joining the spin thread before reading
the result (see [`test_fault_manager.cpp`](https://github.com/selfpatch/ros2_medkit/blob/feat/issue-196-app-belongs-to/src/ros2_medkit_gateway/test/test_fault_manager.cpp)
after that PR merges), but the underlying transport design still requires every
caller to drive an external spin thread, which is fragile.

## Proposed solution

Change `Ros2FaultServiceTransport` to drive the client itself instead of relying
on an external executor:

- Accept an executor (or build a single-threaded one internally for the call)
- Replace `auto future = client_->async_send_request(request); future.wait_for(timeout);`
  with `rclcpp::spin_until_future_complete(executor, future, timeout)`
- Remove the requirement that callers `start_spinning()` a separate thread for
  these RPC paths
- Drop the per-method `std::lock_guard` if the executor itself serializes calls

This eliminates the race by ensuring the client's pending-request cleanup
(`execute_client`) and the response destruction happen on the same thread, in a
well-defined order. It also removes the matching footgun from
`FaultManagerTest.*` and from anywhere else in the gateway that calls into
these transport methods.

## Additional context

Files involved:

- `src/ros2_medkit_gateway/src/ros2/transports/ros2_fault_service_transport.cpp`
- `src/ros2_medkit_gateway/include/ros2_medkit_gateway/ros2/transports/ros2_fault_service_transport.hpp`
- `src/ros2_medkit_gateway/test/test_fault_manager.cpp` (callers can drop
  `start_spinning()` / `stop_spinning()` around `get_*` calls once this lands)

Related: PR #398 ships a test-side workaround for the same race.

bburda:
## Summary

Refactors `Ros2FaultServiceTransport` to drive its rclcpp service clients with
a private `SingleThreadedExecutor` and `spin_until_future_complete()` instead
of blocking on a `std::future::wait_for()` that requires some external thread
to spin the host node.

The seven service clients are now created against a dedicated
`MutuallyExclusive` callback group that is explicitly **not** auto-attached to
the host node's executor. The transport owns the executor that drives this
group, so each RPC's client callback runs synchronously on the caller's
thread inside `spin_until_future_complete`. The pending-request cleanup and
the response destruction therefore happen sequentially on the same thread,
which eliminates the TSan-reported race between the gateway's spin thread
calling `rclcpp::Executor::execute_client` and the caller destroying the
local `response` shared_ptr at the end of the function.

The seven per-method mutexes collapse into one `executor_mutex_` -
`SingleThreadedExecutor::spin_until_future_complete()` is not safe for
concurrent callers on the same executor, so all RPCs serialise through it.
This is the explicit tradeoff the issue accepts in exchange for losing the
external-spin-thread dependency.

## Issue

- closes #399

## Type

- [x] Bug fix
- [ ] New feature or tests
- [ ] Breaking change
- [ ] Documentation only

## Testing

- `colcon build --packages-select ros2_medkit_gateway` clean
- `colcon test --packages-select ros2_medkit_gateway --ctest-args -LE "linter|integration"`:
  2044 tests / 0 failures
- `test_fault_manager` (18 cases) passes - covers the previously racy
  `get_snapshots` / `get_rosbag` paths and the namespaced variants
- pre-commit hooks pass (clang-format, ament-copyright, no-naked-subscriptions,
  clang-tidy on changed files)
- TSan was not run locally (WSL2 ASLR cannot be relaxed for the TSan runtime);
  the CI sanitiser jobs will validate the race is gone

The callback group is created once at transport construction (single-threaded)
and stays on the transport's private executor. The `rcl` hash-map race that
`check_no_naked_subscriptions.sh` guards against does not apply, so the file
is added to `ALLOWED_LEGACY_FILES` with a note referencing this issue.

## Checklist

### DIFF
--- a/scripts/check_no_naked_subscriptions.sh
+++ b/scripts/check_no_naked_subscriptions.sh
@@ -52,6 +52,7 @@ ALLOWED_LEGACY_FILES=(
   "${FAULT_MANAGER_ROOT}/src/snapshot_capture.cpp"                 # uses LockedSubscriptionGuard (in-place serialisation)
   "${FAULT_MANAGER_ROOT}/include/ros2_medkit_fault_manager/snapshot_capture.hpp"  # comment references the guarded API
   "${FAULT_MANAGER_ROOT}/src/rosbag_capture.cpp"                   # bag-recorder spawns its own node + executor, no shared rcl hash map
+  "${GATEWAY_ROOT}/src/ros2/transports/ros2_fault_service_transport.cpp"  # issue #399: single-threaded callback group at construction for synchronous service-client spin
 )
 
 LEGACY_PATTERN=""
--- a/scripts/check_no_naked_subscriptions.sh
+++ b/scripts/check_no_naked_subscriptions.sh
@@ -52,7 +52,6 @@ ALLOWED_LEGACY_FILES=(
   "${FAULT_MANAGER_ROOT}/src/snapshot_capture.cpp"                 # uses LockedSubscriptionGuard (in-place serialisation)
   "${FAULT_MANAGER_ROOT}/include/ros2_medkit_fault_manager/snapshot_capture.hpp"  # comment references the guarded API
   "${FAULT_MANAGER_ROOT}/src/rosbag_capture.cpp"                   # bag-recorder spawns its own node + executor, no shared rcl hash map
-  "${GATEWAY_ROOT}/src/ros2/transports/ros2_fault_service_transport.cpp"  # issue #399: single-threaded callback group at construction for synchronous service-client spin
 )
 
 LEGACY_PATTERN=""
--- a/src/ros2_medkit_gateway/include/ros2_medkit_gateway/ros2/transports/ros2_fault_service_transport.hpp
+++ b/src/ros2_medkit_gateway/include/ros2_medkit_gateway/ros2/transports/ros2_fault_service_transport.hpp
@@ -16,6 +16,7 @@
 
 #include <chrono>
 #include <cstdint>
+#include <memory>
 #include <mutex>
 #include <rclcpp/rclcpp.hpp>
 #include <string>
@@ -34,11 +35,12 @@ namespace ros2_medkit_gateway::ros2 {
 /**
  * @brief rclcpp adapter implementing FaultServiceTransport.
  *
- * Owns the seven `rclcpp::Client<ros2_medkit_msgs::srv::*>` instances and the
- * seven per-client mutexes that previously lived inside FaultManager. Each
- * mutex serialises one service direction so read operations (list, get) are
- * not blocked by slow write operations (report_fault triggers snapshot
- * capture inside the fault manager node).
+ * Owns seven `rclcpp::Client<ros2_medkit_msgs::srv::*>` instances scoped to a
+ * dedicated `MutuallyExclusive` callback group, plus a private
+ * `SingleThreadedExecutor` that drives that group. Each RPC blocks on
+ * `executor_.spin_until_future_complete()` so the client's pending-request
+ * cleanup and the response destruction happen on the calling thread instead
+ * of racing with whichever external executor spins the host node.
  *
  * Performs the ros2_medkit_msgs <-> JSON translation internally, returning
  * neutral FaultResult and FaultWithEnvJsonResult structures so the FaultManager
@@ -85,6 +87,15 @@ class Ros2FaultServiceTransport : public FaultServiceTransport {
  private:
   rclcpp::Node * node_;
 
+  /// Dedicated callback group for all fault service clients. Created with
+  /// `automatically_add_to_executor_with_node = false` so the host node's
+  /// executor does not spin these clients - we drive them ourselves.
+  rclcpp::CallbackGroup::SharedPtr callback_group_;
+
+  /// Private single-threaded executor that owns `callback_group_` and is
+  /// driven inline via `spin_until_future_complete()` on each RPC.
+  std::shared_ptr<rclcpp::executors::SingleThreadedExecutor> executor_;
+
   rclcpp::Client<ros2_medkit_msgs::srv::ReportFault>::SharedPtr report_fault_client_;
   rclcpp::Client<ros2_medkit_msgs::srv::GetFault>::SharedPtr get_fault_client_;
   rclcpp::Client<ros2_medkit_msgs::srv::ListFaults>::SharedPtr list_faults_client_;
@@ -96,16 +107,11 @@ class Ros2FaultServiceTransport : public FaultServiceTransport {
   double service_timeout_sec_{5.0};
   std::string fault_manager_base_path_{"/fault_manager"};
 
-  /// Per-client mutexes for thread-safe service calls.
-  /// Split by service client so that read operations (list, get) are not blocked
-  /// by slow write operations (report_fault with snapshot capture).
-  mutable std::mutex report_mutex_;
-  mutable std::mutex list_mutex_;
-  mutable std::mutex get_mutex_;
-  mutable std::mutex clear_mutex_;
-  mutable std::mutex snapshots_mutex_;
-  mutable std::mutex rosbag_mutex_;
-  mutable std::mutex list_rosbags_mutex_;
+  /// Serialises access to `executor_`. `SingleThreadedExecutor::spin_until_future_complete()`
+  /// is not safe for concurrent callers on the same executor, so all RPCs take
+  /// this mutex for the duration of the synchronous spin. Replaces the seven
+  /// per-client mutexes that previously gated each method.
+  mutable std::mutex executor_mutex_;
 };
 
 }  // namespace ros2_medkit_gateway::ros2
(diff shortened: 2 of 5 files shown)


src/ros2_medkit_gateway/include/ros2_medkit_gateway/ros2/transports/ros2_fault_service_transport.hpp
```diff
@@ -34,11 +35,12 @@ namespace ros2_medkit_gateway::ros2 {
...
  *
- * Owns the seven `rclcpp::Client<ros2_medkit_msgs::srv::*>` instances and the
- * seven per-client mutexes that previously lived inside FaultManager. Each
- * mutex serialises one service direction so read operations (list, get) are
- * not blocked by slow write operations (report_fault triggers snapshot
- * capture inside the fault manager node).
+ * Owns seven `rclcpp::Client<ros2_medkit_msgs::srv::*>` instances scoped to a
+ * dedicated `MutuallyExclusive` callback group, plus a private
+ * `SingleThreadedExecutor` that drives that group. Each RPC blocks on
+ * `executor_.spin_until_future_complete()` so the client's pending-request
```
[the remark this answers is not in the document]
bburda: The header documentation has been rewritten and now refers to executor_->spin_until_future_complete(). The design also changed since this comment: the service clients are now hosted on a private rclcpp::Node owned by the transport rather than a callback group, and the doc reflects that.

src/ros2_medkit_gateway/src/ros2/transports/ros2_fault_service_transport.cpp
```diff
@@ -93,9 +112,10 @@ FaultResult Ros2FaultServiceTransport::report_fault(const std::string & fault_co
...
   request->source_id = source_id;
 
+  std::lock_guard<std::mutex> lock(executor_mutex_);
   auto future = report_fault_client_->async_send_request(request);
 
-  if (future.wait_for(timeout) != std::future_status::ready) {
+  if (executor_->spin_until_future_complete(future, timeout) != rclcpp::FutureReturnCode::SUCCESS) {
     result.success = false;
     result.error_message = "ReportFault service call timed out";
     return result;
```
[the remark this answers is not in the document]
bburda: Added GetSnapshotsDrivesPrivateClientWithoutSpinningHostNode in test_fault_manager.cpp. The mock service runs on a separate node with its own spin thread, and the transport's host node is deliberately left unspun, so the RPC can only complete if the transport drives its private client executor itself.

scripts/check_no_naked_subscriptions.sh
```diff
@@ -52,6 +52,7 @@ ALLOWED_LEGACY_FILES=(
   "${FAULT_MANAGER_ROOT}/src/snapshot_capture.cpp"                 # uses LockedSubscriptionGuard (in-place serialisation)
   "${FAULT_MANAGER_ROOT}/include/ros2_medkit_fault_manager/snapshot_capture.hpp"  # comment references the guarded API
   "${FAULT_MANAGER_ROOT}/src/rosbag_capture.cpp"                   # bag-recorder spawns its own node + executor, no shared rcl hash map
+  "${GATEWAY_ROOT}/src/ros2/transports/ros2_fault_service_transport.cpp"  # issue #399: single-threaded callback group at construction for synchronous service-client spin
 )
```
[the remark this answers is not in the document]
bburda: The allowlist entry was removed. The service clients are now hosted on a private rclcpp::Node instead of a dedicated callback group, so create_callback_group is no longer called and the regression gate passes with no allowlist entry needed.

src/ros2_medkit_gateway/src/ros2/transports/ros2_fault_service_transport.cpp
```diff
@@ -93,15 +102,23 @@ FaultResult Ros2FaultServiceTransport::report_fault(const std::string & fault_co
...
 
-  auto future = report_fault_client_->async_send_request(request);
-
-  if (future.wait_for(timeout) != std::future_status::ready) {
-    result.success = false;
-    result.error_message = "ReportFault service call timed out";
-    return result;
+  ros2_medkit_msgs::srv::ReportFault::Response::SharedPtr response;
+  {
+    std::lock_guard<std::mutex> lock(executor_mutex_);
```
mfaferek93: This lock spans the whole RPC (wait_for_service + spin) and all 7 methods share this one executor_mutex_. When fault_manager is down/hung, a caller holds it up to ~10s (5s wait_for_service + 5s spin at the default service_timeout_sec=5.0), serialising every other fault RPC - under the HTTP threadpool the whole fault API can stall. Normal-case is fine (report_fault is async server-side). Worth trimming wait_for_service to a small fixed value (~200ms), or splitting reads/writes onto a second executor so a hung backend can't block reads.

src/ros2_medkit_gateway/src/ros2/transports/ros2_fault_service_transport.cpp
```diff
@@ -93,15 +102,23 @@ FaultResult Ros2FaultServiceTransport::report_fault(const std::string & fault_co
...
+    if (!report_fault_client_->wait_for_service(timeout)) {
+      result.success = false;
+      result.error_message = "ReportFault service not available";
+      return result;
+    }
+    auto future = report_fault_client_->async_send_request(request);
+    if (executor_->spin_until_future_complete(future, timeout) != rclcpp::FutureReturnCode::SUCCESS) {
```
mfaferek93: On non-SUCCESS this returns without remove_pending_request, so abandoned requests accumulate in the client's pending map on repeated timeouts. The sibling transports clean up (ros2_service_transport.cpp:117, ros2_action_transport.cpp:206) - this one is the lone outlier. Pre-existing (old wait_for had it too), but worth fixing here: keep the FutureAndRequestId and call remove_pending_request on the timeout path. Same in all 7 methods.

src/ros2_medkit_gateway/src/ros2/transports/ros2_fault_service_transport.cpp
```diff
@@ -76,15 +92,8 @@ bool Ros2FaultServiceTransport::is_available() const {
...
-  std::lock_guard<std::mutex> lock(report_mutex_);
   FaultResult result;
-
-  auto timeout = std::chrono::duration<double>(service_timeout_sec_);
-  if (!report_fault_client_->wait_for_service(timeout)) {
-    result.success = false;
-    result.error_message = "ReportFault service not available";
-    return result;
-  }
+  const auto timeout = std::chrono::duration<double>(service_timeout_sec_);
```
mfaferek93: service_timeout_sec_ isn't clamped before it reaches spin_until_future_complete, which treats a negative timeout as "spin forever" - a misconfigured service_timeout_sec: -1 would hang here under executor_mutex_ and freeze all fault RPCs. The sibling transports guard this with std::max(timeout.count(), 0.0) (ros2_service_transport.cpp:113, ros2_topic_transport.cpp:125). Worth matching. Same in all 7 methods.

src/ros2_medkit_gateway/src/ros2/transports/ros2_fault_service_transport.cpp
bburda: Reworked the seven methods to share an `invoke_fault_service<Service>` helper. `wait_for_service` now runs outside `executor_mutex_` - the per-context graph listener that backs it is independent of the executor, so a hung backend can no longer hold the lock while another thread is just probing service discovery. The mutex now wraps only `async_send_request` + `spin_until_future_complete` + `remove_pending_request`, matching the "serialise the spin, nothing more" intent of the executor mutex.

src/ros2_medkit_gateway/src/ros2/transports/ros2_fault_service_transport.cpp
bburda: Fixed. The helper now calls `client->remove_pending_request(future.request_id)` on the non-SUCCESS path before returning the timeout error, so abandoned slots are dropped from the client's pending-request map - the pattern used by `ros2_service_transport.cpp` and `ros2_action_transport.cpp`. Applies to all seven methods via the helper.

src/ros2_medkit_gateway/src/ros2/transports/ros2_fault_service_transport.cpp
bburda: Fixed. The helper clamps via `std::max(timeout.count(), 0.0)` once, before either `wait_for_service` or `spin_until_future_complete` see the value - matches the guard in `ros2_service_transport.cpp:113` and `ros2_topic_transport.cpp:125`. A misconfigured `service_timeout_sec: -1` now degrades to "fail fast" instead of "spin forever under the mutex".

mfaferek93 (approved):
LGTM!

PR #403, opened 2026-05-18
feat(gateway): typed DTO contract for HTTP responses and OpenAPI schema

bburda:
# Pull Request

## Summary

The gateway's HTTP payloads were maintained as three independent, hand-written representations - the handler's JSON construction, the OpenAPI schema factories, and the `XMedkit` extension builder - with nothing keeping them in sync. They drifted: entity list responses advertised a minimal schema while the gateway actually emitted rich `x-medkit` metadata, so generated clients could not see those fields.

This introduces a typed DTO contract. Each payload is a plain C++17 struct described once by a `constexpr` field list. Three visitors fold over that single description:

- `JsonWriter` - struct to wire JSON
- `SchemaWriter` - type to OpenAPI schema
- `JsonReader` - request body to struct, with validation

Wire output and OpenAPI schema are therefore generated from one source and cannot drift. All HTTP handler domains (entities, faults, operations, configurations, data, locks, triggers, cyclic-subscriptions, bulk-data, logs, scripts, updates, auth, health) were migrated to the contract, and `components/schemas` is now generated from the DTO registry. The legacy `XMedkit` fluent builder and the ~50 hand-written schema factories were removed.

### Client-observable changes

- The OpenAPI `components/schemas` is regenerated from the DTOs; several schema names changed (for example the entity list/detail schemas). Clients generated from the previous spec need to be regenerated.
- Entity responses now always include the `type` discriminator (previously absent on list items).
- Optional fields previously emitted as empty strings or empty arrays are now omitted when empty. A few fields previously emitted as an explicit `null` when absent (the script execution `progress` / `started_at` / `completed_at` / `parameters` / `error`, and the script `parameters_schema`) are likewise now omitted; treat an absent key the same as `null`.
- Request-body validation failures on some PUT/POST endpoints now return the error code `invalid-request` (HTTP status 400 is unchanged).
- Synchronous operation-execution failures (when the underlying ROS 2 service call fails) now return the standard SOVD `GenericError` envelope (`error_code` / `vendor_code` / `parameters`) instead of the previous bespoke nested `{"error": {"code", "message", "details"}}` object (HTTP 500 unchanged).
- Aggregated list endpoints (data, operations, configurations, logs) now re-serialize peer items through the local DTO: peer items that fail validation are dropped and recorded in `x-medkit.peer_dropped_items`, and peer-supplied fields outside the local schema are no longer preserved verbatim.
- `GET /{entity}/data` now publishes the opaque `DataListResult` schema (like `GET /{entity}/faults`) instead of a typed `Collection<DataItem>`. The wire payload is unchanged for runtime (ROS 2) entities; for plugin-owned entities, vendor per-item fields (for example the OPC-UA plugin's `value` / `unit` / `data_type` / `writable`) are now returned verbatim, whereas the previous typed re-parse silently dropped any field outside `id` / `name` / `category`. Clients that generated a typed `DataItem` model for this route should treat the list response as a free-form object.

### Known limitation

The config/log list endpoints emit a richer collection-level `x-medkit` than the generic `Collection<T>` wrapper schema types it as. The wire payload stays a valid instance of the published schema; this is a schema-precision gap, documented in `design/dto_contract.rst`.

---

## Issue

Part of #338.

This PR makes the gateway spec accurate. #338 stays open until generated clients are regenerated against the corrected spec and downstream consumers updated, which is follow-up work in separate repositories.

---

## Type

- [ ] Bug fix
- [x] New feature or tests
- [x] Breaking change
- [ ] Documentation only

---

## Testing

- New `test_dto_contract` unit suite round-trips every registered DTO through all three visitors and validates each generated schema.
- Integration tests extended: live endpoint responses are validated against the served OpenAPI spec, and the typed `x-medkit` sub-schemas are asserted present in the spec.
- Full `ros2_medkit_gateway` and `ros2_medkit_integration_tests` suites pass (unit, integration, linters); clang-tidy is clean.

Reviewers can verify by building `ros2_medkit_gateway`, running `colcon test`, and hitting `GET /api/v1/docs` to confirm `components/schemas` contains the typed entity and `x-medkit` schemas.

---

## Checklist


mfaferek93 (commented):
(disregard - posted in error, no findings here)

mfaferek93 (commented):
(disregard - posted in error, no findings here)

src/ros2_medkit_gateway/include/ros2_medkit_gateway/dto/json_reader.hpp
```diff
@@ -0,0 +1,149 @@
...
+    if (!j.is_object()) {
+      errs.push_back({"", "expected a JSON object"});
+      return tl::make_unexpected(errs);
+    }
+    for_each_field<T>([&](const auto & f) {
+      const std::string key(f.key);
+      auto & member = obj.*(f.ptr);
+      using MemberT = std::decay_t<decltype(member)>;
+      const auto it = j.find(key);
+      if (it == j.end() || it->is_null()) {
```
mfaferek93: Treating null as absent here means {"data": null} on PUT configurations/{name} and PUT data/{name} now 400s. The old handlers accepted data:null as a value (body.contains("data") is true for null). Setting a parameter/value to null is legitimate and now breaks, and it isn't in the documented change list. Worth distinguishing present-but-null from absent for the json / optional<json> value members.

src/ros2_medkit_gateway/include/ros2_medkit_gateway/dto/scripts.hpp
```diff
@@ -0,0 +1,158 @@
...
+// value-range validation with ERR_INVALID_PARAMETER; parse_body provides the
+// enum check instead.
+// =============================================================================
+struct ScriptControlRequest {
+  std::string action;  // enum: stop | forced_termination
+};
+
+template <>
+inline constexpr auto dto_fields<ScriptControlRequest> =
+    std::make_tuple(field_enum("action", &ScriptControlRequest::action, kScriptControlActionValues));
```
mfaferek93: field_enum on action rejects anything outside {stop, forced_termination} at parse time, but control_execution is a virtual ScriptProvider - a plugin backend with custom actions (pause/resume/...) is now blocked at the gateway. Same pattern you reverted for ExecutionUpdateRequest.capability (71b51ae8); worth a plain field() here too and letting the provider validate.

src/ros2_medkit_gateway/include/ros2_medkit_gateway/dto/updates.hpp
```diff
@@ -0,0 +1,123 @@
...
+// Wire keys (from update_status_to_json in update_types.hpp):
+//
+//   phase - internal lifecycle phase, distinguishes prepare-completed from
+//           execute-completed (required; always emitted by update_status_to_json)
+//           enum: none | preparing | prepared | executing | executed | failed | deleting
+//
+// Uses field_enum: phase is a RESPONSE-side field; the handler does NOT perform
+// bespoke validation of the phase value in any request.
+// =============================================================================
+struct XMedkitUpdate {
```
mfaferek93, line 166: This nests phase under x-medkit (x-medkit.phase), but the old update_status_to_json emitted a flat top-level x-medkit-phase, so clients reading x-medkit-phase break silently. Consistent with the nested-x-medkit convention, but it's a breaking wire change not in the client-observable list - same as config list items moving x-medkit-source/x-medkit-node into a nested x-medkit object. Worth adding both to the change list.

src/ros2_medkit_gateway/include/ros2_medkit_gateway/dto/faults.hpp
```diff
@@ -0,0 +1,264 @@
...
+
+// =============================================================================
+// FaultListItem - flat fault list item emitted by fault_msg_conversions.cpp
+// (fault_to_json shape), wrapped in Collection<FaultListItem> for list endpoints.
+//
+// Wire keys (exact, from fault_msg_conversions.cpp):
+//   fault_code, severity, description, first_occurred, last_occurred,
+//   occurrence_count, status, reporting_sources, severity_label
+// =============================================================================
+struct FaultListItem {
```
mfaferek93, line 47: FaultListItem is schema-only: the fault list handlers emit items verbatim from the transport (fault_handlers.cpp:339), never through JsonWriter, so this schema is an unverified claim about that wire - the drift the contract was meant to remove. It matches fault_msg_conversions today but nothing enforces it. Also several always-present fields are marked optional here, so the schema is looser than the wire. Route items through the DTO, or add a test pinning the verbatim items to this schema.

src/ros2_medkit_gateway/include/ros2_medkit_gateway/dto/json_writer.hpp
```diff
@@ -0,0 +1,74 @@
...
+    return arr;
+  } else if constexpr (is_variant_v<U>) {
+    return std::visit(
+        [](const auto & alt) {
+          return encode_value(alt);
+        },
+        v);
+  } else {
+    // string / bool / integral / floating / nlohmann::json passthrough
+    return nlohmann::json(v);
```
mfaferek93, line 50: The terminal else has no static_assert, unlike decode_value (json_reader.hpp:88) and schema_of (schema_writer.hpp:61) which both end in static_assert(sizeof(U)==0). If a DTO sentinel ever leaks (a nested DTO used in a TU that didn't see its dto_fields specialization), is_dto_v is false and this silently takes the scalar branch instead of failing to compile. Defused today by include discipline + the registry-wide instantiation, but a matching else static_assert here would make the writer fail loud like the other two visitors.

src/ros2_medkit_gateway/include/ros2_medkit_gateway/dto/faults.hpp
```diff
@@ -0,0 +1,264 @@
...
+  std::optional<std::string> severity_label;  // enum: INFO|WARN|ERROR|CRITICAL|UNKNOWN
+};
+
+template <>
+inline constexpr auto dto_fields<FaultListItem> = std::make_tuple(
+    field("fault_code", &FaultListItem::fault_code), field("severity", &FaultListItem::severity),
+    field("description", &FaultListItem::description), field("first_occurred", &FaultListItem::first_occurred),
+    field("last_occurred", &FaultListItem::last_occurred), field("occurrence_count", &FaultListItem::occurrence_count),
+    field("status", &FaultListItem::status), field("reporting_sources", &FaultListItem::reporting_sources),
+    field("severity_label", &FaultListItem::severity_label));
```
mfaferek93: severity_label is a plain field() so no enum reaches the schema, but the comment on line 49 and the old OpenAPI spec constrained it to INFO|WARN|ERROR|CRITICAL|UNKNOWN (same for severity_filter in logs). On a PR about schema accuracy this drops a constraint the previous spec advertised. kFaultSeverityLabelValues exists but is unused - field_enum here would restore it (response-side enum is schema-only, no request-rejection risk).

src/ros2_medkit_gateway/include/ros2_medkit_gateway/dto/json_reader.hpp
bburda: Fixed. `JsonReader` now distinguishes a present-but-null value from an absent field: for free-form `json` / `optional<json>` members an explicit `null` is decoded as a value, so `PUT .../configurations/{id}` and `PUT .../data/{id}` with `{"data": null}` are accepted again (setting a parameter/value to null). Null on non-json members is still treated as absent. Regression tests added in `test_dto_contract`.

src/ros2_medkit_gateway/include/ros2_medkit_gateway/dto/scripts.hpp
bburda: Fixed. `ScriptControlRequest.action` now uses plain `field()`; the value is validated by the provider, so plugin script backends can accept actions beyond `stop`/`forced_termination`. Removed the now-unused vocabulary constant. Tests added (custom action parses; presence is still required).

src/ros2_medkit_gateway/include/ros2_medkit_gateway/dto/updates.hpp
bburda: Checked this against the base commit and both sub-claims look like false positives. The old code already nested these under `x-medkit`: `update_status` emitted `{"x-medkit": {"phase": ...}}`, and the config list items emitted `{"x-medkit": {"source": ..., "node": ...}}`. A grep of the base tree finds no flat `x-medkit-phase` / `x-medkit-source` / `x-medkit-node` keys anywhere, so this is not a new wire change and there is nothing to add to the change list. Happy to be corrected if I missed a path.

src/ros2_medkit_gateway/include/ros2_medkit_gateway/dto/faults.hpp
bburda: Addressed. `FaultListResult` stays opaque by design (plugin fault-list shapes vary), but I added a round-trip test pinning `fault_to_json` output to the `FaultListItem` schema (`JsonReader` then `JsonWriter` equality), so the in-tree fault-list wire can no longer drift from the published schema unnoticed.

src/ros2_medkit_gateway/include/ros2_medkit_gateway/dto/json_writer.hpp
bburda: Fixed. `encode_value`'s terminal branch now ends in `static_assert(sizeof(U) == 0)` like `decode_value` and `schema_of`, so an unsupported field type fails to compile instead of silently taking the scalar branch.

src/ros2_medkit_gateway/include/ros2_medkit_gateway/dto/faults.hpp
bburda: Fixed. Restored the `severity_label` response enum via `field_enum`, and added `UNKNOWN` to the vocabulary since `fault_to_json` emits it for out-of-range severities. The new `fault_to_json` round-trip test covers that case.

mfaferek93 (approved):

Issue #406, opened 2026-06-01
Release v0.5.0


bburda:
## Summary

Cut release **v0.5.0**. The last release (v0.4.0) shipped 2026-03-22; `main` has accumulated ~48 merged PRs since - multi-instance aggregation, per-entity provider plugins, a typed DTO router, pool-backed topic data, OPC-UA alarms, fault-route scoping, and sanitizer CI. This issue tracks the changelog, version bump, and pre-tag checklist.

Target date: **TBD**.

Part of a coordinated 0.5.0 bump across the open-core repos (web UI, MCP server, generated clients, Foxglove extension, demos).

## Highlights since 0.4.0

**New packages**
- `ros2_medkit_opcua` - OPC-UA gateway plugin bridging PLCs into the SOVD entity tree, with native OPC-UA Part 9 AlarmConditionType event subscriptions and SOVD ack/confirm operations (#365, #387).
- `ros2_medkit_sovd_service_interface` - new SOVD service-interface plugin (#331).

**Features**
- Multi-instance / cross-gateway aggregation: `AggregationManager`, `PeerClient`, `EntityMerger`, mDNS discovery, SSE proxy; peer fan-out across all per-entity resource handlers plus `/health` and `/version-info`; daisy-chain hierarchical aggregation with provenance via `x-medkit.contributors` (#327, #372, #374).
- Per-entity provider plugin interfaces (`DataProvider`, `OperationProvider`, `FaultProvider`) with dlsym discovery and auto-deduced capabilities (#355).
- Typed DTO contract + typed router: response/request bodies and OpenAPI schema derived from one descriptor (#338, #403).
- Plugin API v7: `notify_entities_changed` lifecycle hook + manifest fragment merging (#377); declarative `get_routes()` plugin API replacing `register_routes` (#347).
- Pool-backed `TopicDataProvider` with LRU/idle eviction and QoS matching; new `GET /health/subscription-pool` endpoint (#384).
- New endpoints: `GET /apps/{id}/belongs-to` (#398); entity context on `/faults/stream` SSE events (#400).
- SOVD entity-model alignment: namespaces map to Functions; removal of synthetic Area/Component heuristics (#327, #325).

**Build / CI**
- New `quality.yml` workflow: clang-tidy decoupled + ASan/UBSan and TSan sanitizer jobs running unit + integration tests (#348).
- Root `Dockerfile` + GHCR publish workflow (#343); centralized/hardened warning flags with selective `-Werror` (#357).

[...]

PR #407, opened 2026-06-01
Release v0.5.0: version bump and changelogs

bburda:
## Summary

Release preparation for **v0.5.0**:

- **Version bump** to `0.5.0` across all 15 packages, the `version.hpp` fallback constant, `docs/conf.py`, `docs/pyproject.toml`, and the version literals in the root / version-info API example responses (via `scripts/release.sh bump 0.5.0`).
- **Changelogs finalized**: the gateway `Unreleased` and OPC-UA `Forthcoming` sections are promoted to `0.5.0 (2026-06-01)`, `0.5.0` entries are added to the remaining packages, and an initial `CHANGELOG.rst` is added for the new `ros2_medkit_sovd_service_interface` package (wired into `docs/changelog.rst`).
- **Bug fix** in `scripts/release.sh`: it pointed at `include/ros2_medkit_gateway/version.hpp`, but the version constant lives at `.../core/version.hpp`, so `bump` silently skipped the `kGatewayVersion` fallback and `verify` never validated it. Path corrected; the fallback is now bumped and verified.

The 0.5.0 changelog calls out the breaking changes that already merged this cycle (typed router / `HandlerContext` API, typed provider ABI, fault-route scoping, plugin API v7 / `get_routes()`, entity-model shift to Functions, nested `x-medkit` payload fields).

---

## Issue

Part of #406

---

## Type

- [x] Bug fix (`release.sh` version.hpp path)
- [ ] New feature or tests
- [ ] Breaking change
- [x] Documentation only (changelogs + version)

---

## Testing

- `./scripts/release.sh verify 0.5.0` → all 15 packages + `version.hpp` + `docs/conf.py` + `docs/pyproject.toml` consistent at `0.5.0`.
- `colcon build --packages-up-to ros2_medkit_gateway` builds cleanly; `info.version` reported by `/api/v1/docs` is `0.5.0`.
- Sphinx docs build of the changelog page is clean (no changelog-related warnings; the only warnings are the pre-existing doxygen-XML/`requirements/verification` ones that need the doxygen + generation steps CI runs first).

---

## Checklist

- [x] Breaking changes are clearly described (announced in the gateway/per-package changelogs)
- [x] Tests were added or updated if needed (no code-logic change; version/changelog only)
- [x] Docs were updated if behavior or public API changed (changelogs + API example version literals)

### DIFF
--- a/docs/api/rest.rst
+++ b/docs/api/rest.rst
@@ -25,7 +25,7 @@ Server Capabilities
 
       {
         "name": "ROS 2 Medkit Gateway",
-        "version": "0.4.0",
+        "version": "0.5.0",
         "api_base": "/api/v1",
         "endpoints": [
           "GET /api/v1/health",
@@ -78,7 +78,7 @@ Server Capabilities
             "version": "1.0.0",
             "base_uri": "/api/v1",
             "vendor_info": {
-              "version": "0.4.0",
+              "version": "0.5.0",
               "name": "ros2_medkit"
             }
           }
--- a/scripts/release.sh
+++ b/scripts/release.sh
@@ -30,7 +30,7 @@ set -euo pipefail
 SCRIPT_DIR="$(cd "$(dirname "${BASH_SOURCE[0]}")" && pwd)"
 REPO_ROOT="$(cd "${SCRIPT_DIR}/.." && pwd)"
 SRC_DIR="${REPO_ROOT}/src"
-VERSION_HPP="${SRC_DIR}/ros2_medkit_gateway/include/ros2_medkit_gateway/version.hpp"
+VERSION_HPP="${SRC_DIR}/ros2_medkit_gateway/include/ros2_medkit_gateway/core/version.hpp"
 CONF_PY="${REPO_ROOT}/docs/conf.py"
 DOCS_PYPROJECT="${REPO_ROOT}/docs/pyproject.toml"
 
--- a/scripts/release.sh
+++ b/scripts/release.sh
@@ -35,6 +35,7 @@ CONF_PY="${REPO_ROOT}/docs/conf.py"
 DOCS_PYPROJECT="${REPO_ROOT}/docs/pyproject.toml"
 DOXYFILE="${REPO_ROOT}/docs/Doxyfile"
 QUALITY_DECL="${REPO_ROOT}/QUALITY_DECLARATION.md"
+REST_RST="${REPO_ROOT}/docs/api/rest.rst"
 
 usage() {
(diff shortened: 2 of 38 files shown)


docs/api/rest.rst
```diff
@@ -25,7 +25,7 @@ Server Capabilities
 
       {
         "name": "ROS 2 Medkit Gateway",
-        "version": "0.4.0",
+        "version": "0.5.0",
```
mfaferek93, line 28: These two version examples (here and :81) were bumped by hand and aren't covered by release.sh bump/verify, so the next release will leave them stale and verify won't catch it. Worth wiring into the script - but anchor on the vendor_info `version` key, since :78 has the SOVD `"version": "1.0.0"` you don't want rewritten.

scripts/release.sh
```diff
@@ -177,6 +196,32 @@ cmd_verify() {
...
+        else
+            echo "  OK: docs/Doxyfile PROJECT_NUMBER = ${doxy_version}"
+        fi
+        versions_seen+=("$doxy_version")
+    fi
+
+    # Check QUALITY_DECLARATION.md current version
+    if [ -f "$QUALITY_DECL" ]; then
+        local qd_version
+        qd_version=$(grep -oP 'current version is \*\*\K[0-9]+\.[0-9]+\.[0-9]+' "$QUALITY_DECL" || echo "unknown")
```
mfaferek93, line 230: Minor: on the no-arg verify path a not-found pattern falls through as "unknown" and only fails as a side effect of being a distinct string - an all-"unknown" file would false-pass. Cheap to treat "unknown" as a hard fail.

docs/api/rest.rst
bburda: Wired both example-response version literals into `release.sh bump`/`verify`. Each is anchored on its adjacent `name` key (`ROS 2 Medkit Gateway` for the root response, `ros2_medkit` for the version-info `vendor_info`), so the SOVD API `version` `1.0.0` has neither anchor and is left alone. Verified on a trial bump: both gateway literals are rewritten, the SOVD `1.0.0` stays untouched, and `verify <version>` now flags them if stale.

scripts/release.sh
bburda: Done. The no-arg `verify` path now treats any `unknown` (a version pattern that failed to match) as a hard failure, evaluated before the unique-count check, so an all-`unknown` run fails outright instead of collapsing to a single unique value and false-passing.

mfaferek93 (approved):
LGTM!

PR #408, opened 2026-06-03
Add a config_file argument to the gateway launchfile to allow using an external config file.

evTessellate:
…

# Pull Request

## Summary

Added an optional config_file argument to the ros2_mekdit_gateway to be able to set the gateway parameters from outside the ros2_mekdit_gateway. This feature is proposed for large projects where the gateway config lives in another bringup package and needs to be feed to the gateway. 

---

## Issue

No issues, its new feature.

---

## Type

- [ ] Bug fix
- [X ] New feature or tests
- [ ] Breaking change
- [ ] Documentation only

---

## Testing

How was this tested / how should reviewers verify it?:

I launched the gateway with and without the new config_file orverride:
```
launch:
  - include:
      file: $(find-pkg-share ros2_medkit_gateway)/launch/gateway.launch.py
```
and 
```
launch:
  - include:
      file: $(find-pkg-share ros2_medkit_gateway)/launch/gateway.launch.py
      arg:
        - name: config_file
          value: "$(find-pkg-share bringup_package)/config/medkit_config/gateway_params.yaml"    
```
In my `gateway_params.yaml` from my `bringup_package`, I enable the CORS parameters. So when I launched the gateway without the config file, the `rosout` log showed that CORS was disabled, and when I launched with the config_file, the log said CORS was enabled.

Also, I did all the test in https://github.com/selfpatch/ros2_medkit/blob/main/CONTRIBUTING.md with success.

Since no code had to be build, there wasn't that much build test to do.

---

## Checklist

### DIFF
--- a/src/ros2_medkit_gateway/launch/gateway.launch.py
+++ b/src/ros2_medkit_gateway/launch/gateway.launch.py
@@ -43,6 +43,10 @@ def generate_launch_description():
               f'{graph_provider_path} - plugin will not load')
         graph_provider_path = ''
 
+    declare_overide_config_arg = DeclareLaunchArgument(
+        'config_file', default_value=default_config,
+        description='Path to YAML config file for gateway parameters')
+
     declare_host_arg = DeclareLaunchArgument(
         'server_host', default_value='127.0.0.1',
         description='Host to bind REST server (127.0.0.1 or 0.0.0.0)')
@@ -74,10 +78,11 @@ def generate_launch_description():
         executable='gateway_node',
         name='ros2_medkit_gateway',
         output='screen',
-        parameters=[default_config, param_overrides],
+        parameters=[LaunchConfiguration('config_file'), param_overrides],
         arguments=['--ros-args', '--log-level', 'info'])
 
     return LaunchDescription([
+        declare_overide_config_arg,
         declare_host_arg,
         declare_port_arg,
         declare_refresh_arg,
--- a/src/ros2_medkit_gateway/launch/gateway.launch.py
+++ b/src/ros2_medkit_gateway/launch/gateway.launch.py
@@ -45,7 +45,8 @@ def generate_launch_description():
 
     declare_override_config_arg = DeclareLaunchArgument(
         'config_file', default_value=default_config,
-        description='Path to YAML config file for gateway parameters')
+        description='Path to YAML config file to override gateway parameters. Default config '
+                    'is the ros2_medkit_gateway/config/gateway_params.yaml.')
 
     declare_host_arg = DeclareLaunchArgument(
         'server_host', default_value='127.0.0.1',
@@ -78,7 +79,7 @@ def generate_launch_description():
         executable='gateway_node',
         name='ros2_medkit_gateway',
         output='screen',
-        parameters=[LaunchConfiguration('config_file'), param_overrides],
+        parameters=[default_config, LaunchConfiguration('config_file'), param_overrides],
         arguments=['--ros-args', '--log-level', 'info'])
 
     return LaunchDescription([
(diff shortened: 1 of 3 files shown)


src/ros2_medkit_gateway/launch/gateway.launch.py
```diff
@@ -74,10 +78,11 @@ def generate_launch_description():
         executable='gateway_node',
         name='ros2_medkit_gateway',
         output='screen',
-        parameters=[default_config, param_overrides],
+        parameters=[LaunchConfiguration('config_file'), param_overrides],
```
bburda: Worth documenting: when `config_file` is overridden the packaged `gateway_params.yaml` is no longer loaded at all - only the supplied file is. Any parameter absent from it falls back to the node's compiled-in `declare_parameter` defaults, not to the packaged YAML values. So the external file has to be a **complete** gateway config, not a fragment - a partial file would silently run with default discovery / auth / fault-manager / plugin settings.

`gateway.launch.py` doesn't have a launch-args table in the README yet (only `gateway_https.launch.py` does). Could you add one near the **Configuration Examples** section, e.g.:

````markdown
### Launch Arguments (`gateway.launch.py`)

| Launch Argument       | Default                               | Description                                                                                                  |
| --------------------- | ------------------------------------- | ------------------------------------------------------------------------------------------------------------ |
| `config_file`         | packaged `config/gateway_params.yaml` | Path to the YAML parameter file. **Replaces** the packaged config entirely - must be a complete gateway config, not a fragment. |
| `server_host`         | `127.0.0.1`                           | Host to bind the REST server (`127.0.0.1` or `0.0.0.0`).                                                      |
| `server_port`         | `8080`                                | Port for the REST API.                                                                                       |
| `refresh_interval_ms` | `30000`                               | Safety-backstop refresh interval in ms (graph events drive the primary refresh).                             |

Use a config file from another package:

```bash
ros2 launch ros2_medkit_gateway gateway.launch.py \
  config_file:=$(ros2 pkg prefix my_bringup)/share/my_bringup/config/gateway_params.yaml
```
````

File: src/ros2_medkit_gateway/launch/gateway.launch.py [hunk 2]
```diff
@@ -43,6 +43,10 @@ def generate_launch_description():
               f'{graph_provider_path} - plugin will not load')
         graph_provider_path = ''
 
+    declare_overide_config_arg = DeclareLaunchArgument(
```
bburda: nit: typo `overide` -> `override` in the variable name.

bburda (commented):
Nice, backward-compatible addition - defaulting `config_file` to the packaged config keeps existing launches unchanged.

One thing before merge: docs. `config_file` is a new public launch argument and isn't documented - the README has a launch-args table for `gateway_https.launch.py`, and `gateway.launch.py` should get an equivalent entry. Please also call out the replace semantics there (see inline).

Out of scope, your call: `gateway_https.launch.py` uses the same `parameters=[default_config, {...}]` pattern and could take the same `config_file` arg for parity.

src/ros2_medkit_gateway/launch/gateway.launch.py
evTessellate: You brought me good insights. My goal was to allow partial overrides with my new file. Therefore, I thus changed the way the override config is provided to do so. You can review the new change if you want. I tested the partial override by changing the default value in the default config file, and commenting the override config to see if the partial setting worked. It did, for both the http and the https gateway launchfiles. I also updated the documentation with a mention on the parameter override priorities.

bburda: Thanks, the partial-override approach is cleaner than the full-replace I originally assumed - `[default_config, config_file, param_overrides]` merges per-key, so the external file only needs the keys it changes. Defaulting it to the packaged YAML keeps existing launches unchanged, and the parity on `gateway_https.launch.py` is a nice touch. Docs read correctly now too.

bburda (approved):

Issue #409, opened 2026-06-06
Migrate multipart/form-data handling to the cpp-httplib >= 0.20 req.form API


bburda:
## Summary

The gateway's `multipart/form-data` handling (bulk-data and script uploads) is written against the cpp-httplib multipart API that existed up to 0.15: the `httplib::Request::files` map and the `httplib::MultipartFormData` / `httplib::MultipartFormDataItems` types. cpp-httplib 0.26 (shipped by Ubuntu 26.04 / ROS 2 Lyrical) removed that surface in favour of a new `req.form`-based API.

As a result, `medkit_find_cpp_httplib` (in `ros2_medkit_cmake`) caps the system cpp-httplib at `< 0.20` and falls back to the vendored 0.14.3 header on distros that ship 0.26+. This keeps Lyrical building, but pins us to the vendored copy, which then misses upstream fixes. The pin is meant to be temporary.

---

## Proposed solution

Migrate the multipart code paths to the cpp-httplib >= 0.20 `req.form` API, decoupling the gateway's typed `MultipartBody` from the cpp-httplib struct names so handlers no longer depend on the removed types. Affected sites:

- `src/openapi/route_registry.hpp` - `multipart_upload<T>()` reads `req.files` to populate `MultipartBody.parts`
- `include/ros2_medkit_gateway/http/response_types.hpp` - `MultipartBody::parts` is typed as `httplib::MultipartFormDataItems`
- `src/http/handlers/bulkdata_handlers.cpp` - consumes parts as `httplib::MultipartFormData`
- `src/http/handlers/script_handlers.cpp` - same

Once migrated, drop the `< 0.20` cap in `medkit_find_cpp_httplib` (both the pkg-config and `find_package` tiers) so the system package is used on Lyrical/Resolute.

---

## Additional context

Blocks using the system cpp-httplib on Ubuntu 26.04 / ROS 2 Lyrical (and any future distro shipping 0.20+). Until then the vendored 0.14.3 header is used on those distros.

PR #411, opened 2026-06-08
fix(gateway): keep httplib out of plugin-facing headers + add build-farm purity gate

### LINKED ISSUE #410: [BUG] Plugin libraries fail to build against the installed gateway: httplib.h not found
Reported via the bug template.

### Steps to reproduce
1. Build the gateway and install it (or use the multi-distro `Dockerfile`, which resolves dependencies via rosdep and skips `libcpp-httplib-dev`).
2. Build a gateway plugin (e.g. `ros2_medkit_opcua`) against the **installed** gateway, without the system `cpp-httplib` headers on the include path.
3. Compilation fails.

### Expected behavior
Plugins build against the installed gateway without needing cpp-httplib on their own include path. Plugins implement provider interfaces and exchange JSON / typed DTOs / `tl::expected` / the opaque `PluginRequest`/`PluginResponse` shim - they never use `httplib` types across the `.so` boundary, so cpp-httplib should be a gateway-internal implementation detail.

### Actual behavior
```
.../ros2_medkit_gateway/http/typed_router.hpp:17:10: fatal error: httplib.h: No such file or directory
   17 | #include <httplib.h>
```
A public, provider-facing header chain leaks `<httplib.h>`:

`core/providers/operation_provider.hpp` -> `dto/operations.hpp` -> `http/alternate_status.hpp` -> `http/typed_router.hpp` -> `<httplib.h>`

Normal CI does not catch this: the system `libcpp-httplib-dev` is installed (so the leaked include resolves) and the workspace is built as a single overlay. The failure surfaces only when a plugin is built against the **installed** gateway with cpp-httplib withheld - the gateway installs its vendored `httplib.h` under `share/` (not the include path), so a downstream consumer cannot resolve the leaked include. This is the build-farm / Docker topology, and the Docker Publish build has been red on all three distros since the leak landed.

### Environment
- ros2_medkit version: current `main`
- ROS 2 distro: Jazzy, Humble, Rolling (all affected; the leak is distro-independent)
- OS: Ubuntu 22.04 / 24.04

### Additional information
Root cause: `http/typed_router.hpp` mixes httplib-free handler-result markers (`Result`, `NoContent`, `Forwarded`, `ValidatorResult`, `ResponseAttachments`) with the only httplib-dependent type (`TypedRequest`). The DTO/provider chain only needs the markers but transitively pulls in `<httplib.h>`.

Fix: extract the httplib-free markers into a leaf header (`http/handler_result.hpp`) so the provider/DTO chain is httplib-free; keep `TypedRequest` in `typed_router.hpp`. Add a `g++ -M -MG` header-purity check (linter ctest + pre-push hook) plus a local build-farm-isolation script that builds each plugin against the installed gateway with cpp-httplib withheld, so this class of regression is caught before merge.

Related follow-up: #409 (the gateway's own multipart handling still pins cpp-httplib < 0.20; orthogonal to this header leak).

bburda:
# Pull Request

## Summary

Plugin libraries failed to build against the **installed** gateway with `fatal error: httplib.h: No such file or directory`. A provider-facing public header chain leaked `<httplib.h>`:

`core/providers/operation_provider.hpp` -> `dto/operations.hpp` -> `http/alternate_status.hpp` -> `http/typed_router.hpp` -> `<httplib.h>`

Plugins implement provider interfaces and exchange JSON / typed DTOs / `tl::expected` / the opaque `PluginRequest`/`PluginResponse` shim - they never use `httplib` types across the `.so` boundary, so cpp-httplib should stay a gateway-internal detail. Normal CI masked this (system `libcpp-httplib-dev` installed + single-overlay build); it only surfaced when a plugin was built against the installed gateway with cpp-httplib withheld (build-farm / Docker topology).

`http/typed_router.hpp` mixed httplib-free handler-result markers (`Result`, `NoContent`, `Forwarded`, `ValidatorResult`, `ResponseAttachments`) with the only httplib-dependent type (`TypedRequest`). This PR:

- extracts the httplib-free markers into a new leaf header `http/handler_result.hpp`; `typed_router.hpp` re-exports them and keeps only `TypedRequest`; `alternate_status.hpp` includes the leaf. The provider/DTO chain is now httplib-free. No ABI, wire, or behaviour change (markers moved verbatim).
- adds a `g++ -M -MG` header-purity scan (`gateway_plugin_header_purity` linter ctest + pre-push hook) that fails if any plugin-facing public header reaches `<httplib.h>`.
- adds `scripts/check_isolated_build.sh`, which reproduces the build-farm topology locally (each plugin built against the installed gateway with cpp-httplib withheld via a poisoned `-isystem` shadow header).
- documents the plugin-header purity contract.

---

## Issue

- closes #410

---

## Type

- [x] Bug fix
- [x] New feature or tests
- [ ] Breaking change
- [ ] Documentation only

---

## Testing

- Gateway builds and the existing unit suite passes (96 ctests).
- Header-purity scan: red on the leaking tree, green after the fix; fails loud on preprocessor errors.
- `scripts/check_isolated_build.sh` passes end-to-end on Jazzy: gateway installed, then `ros2_medkit_graph_provider`, `ros2_medkit_sovd_service_interface`, and `ros2_medkit_opcua` all build against the installed gateway with cpp-httplib withheld.
- The change is a pure header relocation (standard C++17), so it is distro-independent across Jazzy / Humble / Rolling.

Reviewers can verify the contract with `colcon test --packages-select ros2_medkit_gateway --ctest-args -L linter -R gateway_plugin_header_purity`, or run `scripts/check_isolated_build.sh`.

---

## Checklist

- [x] Breaking changes are clearly described (none - no ABI/wire/behaviour change)

---

Out of scope: the gateway still pins cpp-httplib `< 0.20` for its own multipart handling (tracked in #409); that is orthogonal to this header leak and not required to make plugins build.

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -71,6 +71,13 @@ repos:
         language: script
         pass_filenames: false
         always_run: true
+      - id: gateway-plugin-header-purity
+        name: plugin-facing headers must not depend on httplib
+        entry: ./src/ros2_medkit_gateway/scripts/check_headers_httplib_free.sh
+        language: script
+        stages: [pre-push]
+        pass_filenames: false
+        always_run: true
 
   # ── Incremental clang-tidy (pre-push only) ────────────────────────
   # Requires: pre-commit install --hook-type pre-push
--- a/scripts/check_isolated_build.sh
+++ b/scripts/check_isolated_build.sh
@@ -1,0 +1,76 @@
+#!/usr/bin/env bash
+# Copyright 2026 bburda
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+# Reproduce the ROS build-farm topology locally on Jazzy: build + install the
+# gateway into a throwaway prefix, then build each plugin LIBRARY against ONLY
+# the installed gateway with httplib genuinely WITHHELD (a poisoned shadow
+# <httplib.h> injected via -isystem ahead of /usr/include). A leaked <httplib.h>
+# in a plugin-facing public header (masked in normal CI by the colcon overlay +
+# system libcpp-httplib-dev) trips the poison's #error here, exactly like the
+# build farm / Docker Publish. The leak is distro-independent, so Jazzy alone
+# catches this class. Run before pushing changes to gateway public headers or
+# plugins.
+set -euo pipefail
+
+DISTRO="${ROS_DISTRO:-jazzy}"
+# The ament setup scripts reference optional env vars, so relax nounset while
+# sourcing (otherwise `set -u` aborts on AMENT_TRACE_SETUP_FILES et al.).
+set +u
+# shellcheck disable=SC1090
+source "/opt/ros/${DISTRO}/setup.bash"
+set -u
+
+# Cap build parallelism. The gateway has many heavy TUs (PCH + httplib +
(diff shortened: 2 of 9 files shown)


mfaferek93 (approved):
LGTM!

PR #417, opened 2026-06-11
feat(openapi): derive query parameters from a typed query contract

### LINKED ISSUE #416: [BUG] Generated OpenAPI spec omits all query parameters
# Bug report

The OpenAPI spec generated by the gateway contains **zero query parameters**, even though several handlers read query parameters at runtime via `req.query_param(...)`. Clients generated from the spec therefore cannot pass these filters type-safely.

### Steps to reproduce
1. Run the gateway and fetch the spec: `curl http://localhost:8080/api/v1/docs`
2. Count query parameters across all paths: `... | jq "[.paths[][]?.parameters[]? | select(.in==\"query\")] | length"`
3. Result is `0`.

### Expected behavior

Endpoints whose handlers read query parameters should declare them in the spec (`in: query`), so the generated TypeScript/Python clients expose them. Affected endpoints and the parameters their handlers read:
- `GET /faults`, `GET /{entity}/faults`: `status`, `include_muted`, `include_clusters`
- `DELETE /faults`: `status`
- `GET /{entity}/logs`: `severity`, `context`
- `GET /updates`: `origin`, `target-version`

### Actual behavior

The spec declares only path parameters; every operation gets `query?: never` in the generated TypeScript types. The gateway runtime still reads the parameters (e.g. `log_handlers.cpp` reads `severity`/`context`; `fault_handlers.cpp` reads `status`/`include_muted`/`include_clusters`), so filtering works over raw HTTP but is invisible to any generated client.

Root cause: query parameters were read ad-hoc in handler bodies with no link to the route OpenAPI metadata, so the spec generator never saw them. Request and response bodies are derived from typed DTOs, but query parameters had no equivalent typed contract.

This was masked for a long time: an earlier published TypeScript client shipped without its generated types, so consumer types silently degraded to `any` and query usage was never type-checked. Once the client shipped real types, the gap surfaced.

### Environment
- ros2_medkit version: 0.5.0
- ROS 2 distro: Jazzy / Humble / Rolling (spec generation is distro-independent)
- OS: Ubuntu 24.04

### Additional information

Fix approach: make query parameters a typed, descriptor-driven contract like request/response bodies. A query DTO whose `dto_fields` descriptor drives both the OpenAPI `parameters` and the handler typed query read, so the two cannot drift. A pre-commit guard bans raw query reads in handlers to keep the typed path the only way in.

bburda:
## Summary

Make query parameters a typed, descriptor-driven contract like request and response bodies, fixing the generated spec which carried zero query parameters.

Handlers read query parameters ad-hoc via `req.query_param()`, so the OpenAPI generator never saw them and the exported spec - and every client generated from it - had no query parameters even though the gateway reads them at runtime.

- `dto::QueryParamWriter<T>` emits the OpenAPI `parameters` (all `in: query`) for a query DTO from its `dto_fields`, mirroring `SchemaWriter`.
- `TypedRequest::query<T>()` parses the request query string into a typed query DTO via the same descriptor.
- `RouteEntry::query<T>()` declares those parameters on the route.

A handler can only read fields that exist on its query DTO, and those fields are exactly what the route advertises, so the parsed object and the spec cannot drift. Adding a parameter means adding a member - which appears in the spec automatically. A pre-commit guard bans raw query reads in handlers so the typed path is the only way in.

Applied to the endpoints whose handlers read query parameters:
- `GET /faults`, `GET /{entity}/faults`: `status`, `include_muted`, `include_clusters`
- `DELETE /faults`: `status`
- `GET /{entity}/logs`: `severity`, `context`
- `GET /updates`: `origin`, `target-version`

The exported spec now declares 26 query parameters (was 0).

---

## Issue

- closes #416

---

## Type

- [x] Bug fix
- [x] New feature or tests
- [ ] Breaking change
- [ ] Documentation only

---

## Testing

- `colcon build` clean; gateway unit suite green, including new `RouteRegistry` test (typed `.query<T>()` emits the parameters) and `TypedRequest` tests (typed `query<T>()` parsing of present/absent params).
- End-to-end: ran the gateway and fetched `/api/v1/docs`; the spec now declares 26 `in: query` parameters across the faults/logs/updates endpoints with the correct schema types (string/boolean) and `required: false`.
- clang-format / clang-tidy clean on changed files; the new pre-commit guard passes.

---

## Checklist

- [x] Breaking changes are clearly described (none - additive query parameters)
- [x] Docs were updated if behavior or public API changed (the generated OpenAPI spec is the API surface; it now includes the query parameters)

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -78,6 +78,12 @@ repos:
         stages: [pre-push]
         pass_filenames: false
         always_run: true
+      - id: gateway-handlers-typed-query
+        name: handlers must read query params via typed query<T>()
+        entry: ./src/ros2_medkit_gateway/scripts/check_handlers_typed_query.sh
+        language: script
+        pass_filenames: false
+        always_run: true
 
   # ── Incremental clang-tidy (pre-push only) ────────────────────────
   # Requires: pre-commit install --hook-type pre-push
--- a/src/ros2_medkit_gateway/include/ros2_medkit_gateway/dto/query.hpp
+++ b/src/ros2_medkit_gateway/include/ros2_medkit_gateway/dto/query.hpp
@@ -1,0 +1,108 @@
+// Copyright 2026 bburda
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+#pragma once
+
+// Query-parameter contract: the fourth descriptor-driven visitor.
+//
+// A "query DTO" is a plain struct with a `dto_fields<T>` specialization whose
+// members are query-parameter scalars (string / bool, optionally wrapped in
+// std::optional). The same descriptor drives two sides so they cannot drift:
+//
+//   * QueryParamWriter<T> - emits the OpenAPI `parameters` array (all
+//     `in: query`) for the route, mirroring SchemaWriter.
+//   * TypedRequest::query<T>() (see http/typed_router.hpp) - parses the request
+//     query string into a typed T, using `assign_query_field` below.
+//
+// Because a handler can only read fields that exist on T, and those same fields
+// are what the spec advertises, a handler cannot read an undeclared query
+// parameter. Adding a parameter means adding a member - which appears in the
+// spec automatically.
+
+#include <nlohmann/json.hpp>
+#include <optional>
+#include <string>
+#include <type_traits>
+#include <utility>
+
+#include "ros2_medkit_gateway/dto/contract.hpp"
+#include "ros2_medkit_gateway/dto/schema_writer.hpp"
+
+namespace ros2_medkit_gateway {
+namespace dto {
+
+/// Unwrap std::optional<U> -> U, identity otherwise. Query-parameter optionality
+/// is expressed via `required:false`, not a nullable schema, so the emitted
+/// `schema` uses the unwrapped scalar type.
+template <class M>
+struct query_value {
+  using type = M;
+};
+template <class U>
+struct query_value<std::optional<U>> {
+  using type = U;
+};
+template <class M>
+using query_value_t = typename query_value<M>::type;
+
+/// Emits the OpenAPI `parameters` array (all `in: query`) for a query DTO `T`,
+/// derived from the same `dto_fields<T>` descriptor the typed reader consumes.
+template <class T>
+struct QueryParamWriter {
+  static nlohmann::json parameters() {
+    nlohmann::json params = nlohmann::json::array();
+    for_each_field<T>([&](const auto & f) {
+      using FieldT = std::decay_t<decltype(f)>;
+      static_assert(!is_opaque_object_field_v<FieldT>, "query DTOs must not contain opaque-object fields");
+      using MemberT = std::decay_t<decltype(std::declval<T>().*(f.ptr))>;
+
+      nlohmann::json param;
+      param["name"] = std::string(f.key);
+      param["in"] = "query";
+      param["required"] = (f.presence == Presence::kRequired);
+      if (!f.description.empty()) {
+        param["description"] = std::string(f.description);
+      }
+      nlohmann::json schema = schema_of<query_value_t<MemberT>>();
+      if (f.enum_count > 0) {
+        nlohmann::json values = nlohmann::json::array();
+        for (std::size_t i = 0; i < f.enum_count; ++i) {
+          values.push_back(std::string(f.enum_values[i]));
+        }
+        schema["enum"] = std::move(values);
+      }
+      param["schema"] = std::move(schema);
+      params.push_back(std::move(param));
+    });
+    return params;
+  }
+};
+
+/// Assigns a raw query-string value to a typed query-DTO member. Only the scalar
+/// shapes a query parameter can take are supported; any other field type is a
+/// compile error so unsupported types fail loudly at the route, not silently at
+/// runtime.
+template <class M>
+void assign_query_field(M & member, const std::string & raw) {
+  if constexpr (std::is_same_v<M, std::string> || std::is_same_v<M, std::optional<std::string>>) {
+    member = raw;
+  } else if constexpr (std::is_same_v<M, bool> || std::is_same_v<M, std::optional<bool>>) {
+    member = (raw == "true");
+  } else {
+    static_assert(sizeof(M) == 0, "assign_query_field: unsupported query field type");
+  }
+}
+
+}  // namespace dto
+}  // namespace ros2_medkit_gateway
--- a/src/ros2_medkit_gateway/include/ros2_medkit_gateway/dto/query.hpp
+++ b/src/ros2_medkit_gateway/include/ros2_medkit_gateway/dto/query.hpp
@@ -29,10 +29,22 @@
 // are what the spec advertises, a handler cannot read an undeclared query
 // parameter. Adding a parameter means adding a member - which appears in the
 // spec automatically.
+//
+// Scope of the guarantee: the descriptor binds parameter NAMES and PRESENCE -
+// those cannot drift between the spec and the parser. It does NOT police a
+// parameter's VALUE. The emitted `schema` (enum membership, type/format) is
+// advisory metadata; enforcing it (e.g. rejecting `?status=bogus`) is the
+// handler's job, where the richer 400 payload lives. To keep the two type
+// universes aligned, query fields are constrained at compile time to the scalar
+// shapes the parser supports (string / bool, optionally std::optional), and a
+// query parameter can never be `required` (the parser always tolerates absence,
+// leaving the member at its default).
 
+#include <cctype>
 #include <nlohmann/json.hpp>
 #include <optional>
 #include <string>
+#include <tuple>
 #include <type_traits>
 #include <utility>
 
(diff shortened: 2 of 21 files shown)


src/ros2_medkit_gateway/src/http/rest_server.cpp
```diff
@@ -655,7 +655,8 @@ void RESTServer::setup_routes() {
         .tag("Faults")
         .summary(std::string("List faults for ") + et.singular)
         .description(std::string("Returns all active faults reported by this ") + et.singular + ".")
-        .operation_id(std::string("list") + capitalize(et.singular) + "Faults");
+        .operation_id(std::string("list") + capitalize(et.singular) + "Faults")
+        .query<dto::FaultListQuery>();
```
mfaferek93: This per-entity `/faults` route declares `.query<dto::FaultListQuery>()`, so the spec advertises `include_muted`/`include_clusters` - but the per-entity handler intentionally ignores them (only global `GET /faults` honors them). So the spec promises params the runtime silently drops, which is the same spec-vs-runtime drift this PR removes, just on the declared-but-unused axis. Use a narrower query DTO here (status only) so the route advertises exactly what the handler reads.

.pre-commit-config.yaml
```diff
@@ -78,6 +78,12 @@ repos:
         stages: [pre-push]
         pass_filenames: false
         always_run: true
+      - id: gateway-handlers-typed-query
+        name: handlers must read query params via typed query<T>()
+        entry: ./src/ros2_medkit_gateway/scripts/check_handlers_typed_query.sh
```
mfaferek93, line 83: `check_handlers_typed_query.sh` is wired only into pre-commit, and no CI workflow runs it (or runs `pre-commit`). So the gate that forces handlers onto the typed path - the thing preventing another zero-query-params regression - is local-and-optional; a raw `query_param()` read would still pass CI. Please run it as a CI step in `quality.yml` (or `pre-commit run --all-files`).

src/ros2_medkit_gateway/include/ros2_medkit_gateway/dto/query.hpp
```diff
@@ -0,0 +1,108 @@
...
+};
+
+/// Assigns a raw query-string value to a typed query-DTO member. Only the scalar
+/// shapes a query parameter can take are supported; any other field type is a
+/// compile error so unsupported types fail loudly at the route, not silently at
+/// runtime.
+template <class M>
+void assign_query_field(M & member, const std::string & raw) {
+  if constexpr (std::is_same_v<M, std::string> || std::is_same_v<M, std::optional<std::string>>) {
+    member = raw;
```
mfaferek93, line 166: The writer emits `schema.enum` (line 83), but this just assigns the raw string and never checks it against `f.enum_values`. So `?status=bogus` parses cleanly and rejection depends on each handler validating again - the drift-proof guarantee covers parameter names, not enum values. Either validate membership here (400 on a non-member) or scope the "cannot drift" comment to names.

src/ros2_medkit_gateway/include/ros2_medkit_gateway/dto/query.hpp
```diff
@@ -0,0 +1,108 @@
...
+  } else if constexpr (std::is_same_v<M, bool> || std::is_same_v<M, std::optional<bool>>) {
+    member = (raw == "true");
```
mfaferek93: `member = (raw == "true")` folds `True`, `1`, `yes`, and an empty flag-style `?x` all to false with no way to flag a malformed value, but the emitted schema says `type: boolean` (under which `1`/`0` are valid). A client following the spec can silently get the wrong filter. Suggest accepting `true`/`1` case-insensitively as truthy, `false`/`0`/empty as false, optionally 400 otherwise.

src/ros2_medkit_gateway/include/ros2_medkit_gateway/http/typed_router.hpp
```diff
@@ -94,6 +95,23 @@ class TypedRequest {
...
+  /// same `dto_fields<T>` descriptor that declares the parameters in the OpenAPI
+  /// spec (via `dto::QueryParamWriter<T>`). Absent parameters leave their member
+  /// at its default. This is the sanctioned way for handlers to read query
+  /// parameters: a handler cannot read a parameter the route did not declare,
+  /// because it can only access members of `T`.
+  template <class T>
+  T query() const {
+    T out{};
+    dto::for_each_field<T>([&](const auto & f) {
+      if (auto raw = query_param(f.key)) {
```
mfaferek93, line 108: `QueryParamWriter` emits `required:true` for any non-optional member, but this loop only assigns when the param is present and otherwise leaves the default - there's no missing-required check. So a route could advertise a mandatory query param the runtime silently treats as empty. No current DTO hits this, but either enforce required-presence here or constrain query DTOs so `required:true` is unrepresentable.

src/ros2_medkit_gateway/include/ros2_medkit_gateway/dto/query.hpp
```diff
@@ -0,0 +1,108 @@
...
+  } else {
+    static_assert(sizeof(M) == 0, "assign_query_field: unsupported query field type");
```
mfaferek93: The writer side (`schema_of`) can emit `type: integer`/`number`, but this reader static_asserts on anything that isn't string/bool. So adding an `int limit` member compiles on the writer side with a correct-looking schema, then fails here with an opaque `sizeof(M)==0`. Either support integers in the reader or static_assert in the writer so both sides advertise the same type set.

src/ros2_medkit_gateway/scripts/check_handlers_typed_query.sh
```diff
@@ -0,0 +1,51 @@
...
+# can only read fields that exist on its query DTO, and those fields are what the
+# route declares.
+#
+# To add a query parameter: define a query DTO (struct + dto_fields), declare it
+# on the route with .query<T>(), and read it with req.query<T>(). See
+# include/ros2_medkit_gateway/dto/query.hpp.
+set -euo pipefail
+
+SCRIPT_DIR="$(cd "$(dirname "${BASH_SOURCE[0]}")" && pwd)"
+HANDLERS_DIR="$(cd "${SCRIPT_DIR}/../src/http/handlers" && pwd)"
```
mfaferek93: The lint scans only `src/http/handlers/*.cpp`, so a query read in a `.hpp` handler, a helper outside this dir, or a plugin slips through, and `get_param_value(` in a comment/string is a false positive. Worth widening the scan (or anchoring it at the route layer) so the gate can't be bypassed.

src/ros2_medkit_gateway/scripts/check_handlers_typed_query.sh
```diff
@@ -0,0 +1,51 @@
...
+
+# Path reads (req.path), header reads (req.header), and fan-out (raw_for_framework
+# for path/Authorization) are legitimate; only query-string reads are banned.
+pattern='\.query_param\(|->query_param\(|get_param_value\('
+
+hits="$(grep -rnE "${pattern}" "${HANDLERS_DIR}" --include='*.cpp' || true)"
```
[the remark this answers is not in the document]
bburda: The guard now matches the accessor name directly (`(query_param|get_param_value|has_param)[[:space:]]*\(`), so `has_param(` and whitespace-evaded reads like `req . query_param (` are both caught. It scans the source handler layers (`src/http/handlers`, `src/core/http/handlers`) and the include handler layers, `.hpp` as well as `.cpp`. Whole-line `//` comments are filtered out, and a grep I/O error (exit >= 2) now fails the gate instead of passing silently.

src/ros2_medkit_gateway/src/http/rest_server.cpp
bburda: Added a narrower `FaultEntityListQuery` (status only) and pointed the per-entity `GET /{entity}/faults` route and `list_faults` at it, so the route advertises exactly what the handler reads. `include_muted`/`include_clusters` stay on the global `GET /faults` (`FaultListQuery`) only.

.pre-commit-config.yaml
bburda: The gate now runs in CI as a step in the `format-lint` job (`quality.yml`), next to the other regression gates, so a raw `query_param()` read fails CI and not just the local pre-commit.

src/ros2_medkit_gateway/include/ros2_medkit_gateway/dto/query.hpp
bburda: Scoped the guarantee to parameter names and presence: the file-header comment now states that value constraints (enum membership, format) are advisory schema metadata which the handler validates and 400s on (e.g. `read_fault_status_filter`). The enum emission in `QueryParamWriter` is marked advisory as well.

src/ros2_medkit_gateway/include/ros2_medkit_gateway/dto/query.hpp
bburda: Added `parse_query_bool`: `true`/`1` (case-insensitive) are truthy; everything else, including `false`, `0`, and an empty flag-style value, is false, matching the advertised `type: boolean`. Covered by a new `TypedQueryParsesBooleanSpellings` test.

src/ros2_medkit_gateway/include/ros2_medkit_gateway/http/typed_router.hpp
bburda: Constrained query DTOs so `required:true` is unrepresentable: `QueryParamWriter` now `static_assert`s `query_dto_all_optional<T>()`, which fails the build if any query field is `Presence::kRequired`. Since the parser never enforces presence, a required query parameter can no longer be declared.

src/ros2_medkit_gateway/include/ros2_medkit_gateway/dto/query.hpp
bburda: Made the type sets symmetric: `is_query_scalar_v` (string/bool) is now the single source of truth, `static_assert`ed by both the writer (`QueryParamWriter`) and the reader (`assign_query_field`). An `int` member now fails at the route with a clear message on both sides instead of the opaque `sizeof(M)==0`.

src/ros2_medkit_gateway/scripts/check_handlers_typed_query.sh
bburda: Widened the scan to `.hpp` as well as `.cpp` across both handler source dirs and both include handler dirs, so a query read in a handler header is caught (verified with a probe). Whole-line `//` comments are filtered, which drops the `get_param_value(`-in-a-comment false positive.

mfaferek93 (approved):

PR #419, opened 2026-06-15
SOVD service interface: aggregate peer faults

### LINKED ISSUE #418: SOVD service interface returns only local faults, not aggregated peer faults
The HTTP REST API fans out fault queries to aggregation peers (`AggregationManager::fan_out_get`), but the ROS 2 SOVD service interface returns only the local fault manager's faults.

In a federated setup (a gateway with `aggregation.enabled` and peer gateways), `/<prefix>/list_entity_faults` and the `list_all_faults` plugin path return only locally-managed faults. A ROS 2 client therefore cannot see faults that originate on peers, even though the HTTP `/api/v1/faults` and `/api/v1/<entity>/faults` endpoints already expose them via fan-out.

The ROS 2 service interface and the HTTP API should be consistent: both should surface aggregated peer faults.

bburda:
## Summary

The ROS 2 SOVD service interface returned only local fault-manager faults, while the HTTP API already fans out to aggregation peers. This wires `GatewayPluginContext::list_entity_faults` and `list_all_faults` to the aggregation manager (`fan_out_get`), mirroring the HTTP fault handler, so a gateway with `aggregation.enabled` exposes peer faults over the ROS 2 service path too - consistent with the HTTP `/api/v1/faults` and `/api/v1/<entity>/faults` endpoints.

- per-entity: `list_entity_faults` fans out `/api/v1/<type>/<id>/faults` to peers and merges.
- all-faults: `list_all_faults` fans out `/api/v1/faults` and merges.
- Aggregation is gated on `get_aggregation_manager() != nullptr` (no behavior change when aggregation is disabled).

Aggregated faults can carry `status` as a SOVD DTC object (e.g. `{"aggregatedStatus": "active", ...}`) rather than a plain string. `handle_list_entity_faults` previously called `value("status", string{})`, which throws `json::type_error.302` on an object. It now checks the JSON type first: strings pass through unchanged, objects use the `aggregatedStatus` field when present. Without this, a fanned-out service call over a peer with object-typed status would fail.

---

## Issue

- closes #418

---

## Type

- [x] Bug fix

---

## Testing

- New `test_plugin_context_aggregation` (mock peer server): per-entity and all-faults paths surface a peer fault when aggregation is enabled; local-only behavior unchanged when disabled.
- Gateway aggregation suite: `test_plugin_context_aggregation`, `test_aggregation_manager`, `test_aggregation_classification` - all pass.
- SOVD service interface package suite: 17 tests, 0 failures.
- Full gateway unit suite: `colcon test --packages-select ros2_medkit_gateway` - 0 failures.
- Clean build (`-Wall -Wextra -Wpedantic -Wshadow -Wconversion`), zero warnings; clang-tidy clean on the changed lines.

### DIFF
--- a/src/ros2_medkit_gateway/src/plugins/plugin_context.cpp
+++ b/src/ros2_medkit_gateway/src/plugins/plugin_context.cpp
@@ -16,9 +16,12 @@
 
 #include <mutex>
 #include <rclcpp/rclcpp.hpp>
+#include <string>
 #include <vector>
 
+#include "ros2_medkit_gateway/aggregation/aggregation_manager.hpp"
 #include "ros2_medkit_gateway/core/entity_validation.hpp"
+#include "ros2_medkit_gateway/core/faults/fault_scope.hpp"
 
 #include "ros2_medkit_gateway/core/condition_evaluator.hpp"
 #include "ros2_medkit_gateway/core/http/error_codes.hpp"
@@ -30,6 +33,29 @@
 #include "ros2_medkit_gateway/gateway_node.hpp"
 
 namespace ros2_medkit_gateway {
+
+namespace {
+
+/// Map a SovdEntityType to the plural URL path segment used in the REST API.
+/// Returns empty string for SERVER/UNKNOWN (no faults collection on those).
+std::string entity_type_to_url_segment(SovdEntityType type) {
+  switch (type) {
+    case SovdEntityType::APP:
+      return "apps";
+    case SovdEntityType::COMPONENT:
+      return "components";
+    case SovdEntityType::AREA:
+      return "areas";
+    case SovdEntityType::FUNCTION:
+      return "functions";
+    case SovdEntityType::SERVER:
+    case SovdEntityType::UNKNOWN:
+      return "";
+  }
+  return "";
+}
+
+}  // namespace
 
 // ---- Concrete implementation ----
 
@@ -76,28 +102,53 @@ class GatewayPluginContext : public RosPluginContext {
   }
 
   nlohmann::json list_entity_faults(const std::string & entity_id) const override {
-    if (!fault_manager_ || !fault_manager_->is_available()) {
-      return nlohmann::json::array();
-    }
-
-    // Determine source_id for fault filtering based on entity type
+    // Determine entity type (needed for both local query and peer path construction).
     auto entity = get_entity(entity_id);
     if (!entity) {
       return nlohmann::json::array();
     }
 
-    std::string source_id;
-    if (entity->type == SovdEntityType::COMPONENT) {
-      source_id = entity->namespace_path;
-    } else if (entity->type == SovdEntityType::APP) {
-      source_id = entity->fqn;
+    nlohmann::json local_faults = nlohmann::json::array();
+
+    if (fault_manager_ && fault_manager_->is_available()) {
+      // Mirror the HTTP per-entity fault handler: fetch all faults, then scope
+      // them to the entity's resolved App-FQN set (APP -> itself, COMPONENT ->
+      // hosted apps, AREA -> descendant apps, FUNCTION -> host apps) via the
+      // shared core helper. The fault manager returns a {"faults": [...]}
+      // object; this method's contract is a bare array, so extract and filter
+      // the inner array.
+      auto result = fault_manager_->list_faults("");
+      if (result.success && result.data.contains("faults") && result.data["faults"].is_array()) {
+        const auto & cache = node_->get_thread_safe_cache();
+        const auto source_fqns = faults::resolve_entity_source_fqns(cache, entity->type, entity_id);
+        local_faults = faults::filter_faults_by_sources(result.data["faults"], source_fqns);
+      }
     }
 
-    auto result = fault_manager_->list_faults(source_id);
-    if (result.success && result.data.is_array()) {
-      return result.data;
+    // Fan-out to peer gateways: fetch faults for this entity from each peer.
+    // The peer path mirrors the HTTP REST route: /api/v1/<type_plural>/<id>/faults.
+    // Auth: the ROS 2 service path has no inbound HTTP Authorization header to
+    // forward, so peers are queried with an empty credential - fan-out over this
+    // path only surfaces faults from peers that do not gate their /faults route.
+    auto * agg = node_->get_aggregation_manager();
+    if (agg) {
+      const std::string type_segment = entity_type_to_url_segment(entity->type);
+      if (!type_segment.empty()) {
+        const std::string peer_path = "/api/v1/" + type_segment + "/" + entity_id + "/faults";
+        auto fan_result = agg->fan_out_get(peer_path, "");
+        if (fan_result.merged_items.is_array()) {
+          for (const auto & item : fan_result.merged_items) {
+            local_faults.push_back(item);
+          }
+        }
+        if (fan_result.is_partial) {
+          RCLCPP_WARN(node_->get_logger(), "list_entity_faults('%s'): partial peer fan-out, %zu peer(s) failed",
+                      entity_id.c_str(), fan_result.failed_peers.size());
+        }
+      }
     }
-    return nlohmann::json::array();
+
+    return local_faults;
   }
 
   std::optional<PluginEntityInfo> validate_entity_for_route(const PluginRequest & req, PluginResponse & res,
@@ -207,14 +258,43 @@ class GatewayPluginContext : public RosPluginContext {
   }
 
   nlohmann::json list_all_faults() const override {
-    if (!fault_manager_ || !fault_manager_->is_available()) {
-      return nlohmann::json::object();
+    nlohmann::json response = nlohmann::json::object();
+
+    if (fault_manager_ && fault_manager_->is_available()) {
+      auto result = fault_manager_->list_faults("");
+      if (result.success) {
+        response = result.data;
+      }
     }
-    auto result = fault_manager_->list_faults("");
-    if (result.success) {
-      return result.data;
+
+    // Fan-out to peer gateways: fetch all faults from /api/v1/faults on each peer.
+    // Auth: see list_entity_faults - the ROS 2 service path forwards no inbound
+    // Authorization header, so peers are queried with an empty credential.
+    auto * agg = node_->get_aggregation_manager();
+    if (agg) {
+      auto fan_result = agg->fan_out_get("/api/v1/faults", "");
+      if (fan_result.merged_items.is_array()) {
+        // Ensure local response has a "faults" array to append to.
+        if (!response.contains("faults") || !response["faults"].is_array()) {
+          response["faults"] = nlohmann::json::array();
+        }
+        for (const auto & item : fan_result.merged_items) {
+          response["faults"].push_back(item);
+        }
+      }
+      if (fan_result.is_partial) {
+        RCLCPP_WARN(node_->get_logger(), "list_all_faults: partial peer fan-out, %zu peer(s) failed",
+                    fan_result.failed_peers.size());
+      }
     }
-    return nlohmann::json::object();
+
+    // "count" comes from the local-only query; after merging peer faults it
+    // would undercount, so keep it consistent with the merged array.
+    if (response.contains("count") && response.contains("faults") && response["faults"].is_array()) {
+      response["count"] = response["faults"].size();
+    }
+
+    return response;
   }
 
   void register_sampler(
--- a/src/ros2_medkit_gateway/test/test_plugin_context_aggregation.cpp
+++ b/src/ros2_medkit_gateway/test/test_plugin_context_aggregation.cpp
@@ -1,0 +1,742 @@
+// Copyright 2026 bburda
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+// Tests for plugin_context.cpp aggregation fan-out in:
+//   - list_entity_faults (peers queried via /api/v1/<type>/<id>/faults)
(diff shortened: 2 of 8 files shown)


src/ros2_medkit_gateway/src/plugins/plugin_context.cpp
```diff
@@ -76,28 +101,45 @@ class GatewayPluginContext : public RosPluginContext {
...
+    if (fault_manager_ && fault_manager_->is_available()) {
+      std::string source_id;
+      if (entity->type == SovdEntityType::COMPONENT) {
+        source_id = entity->namespace_path;
+      } else if (entity->type == SovdEntityType::APP) {
+        source_id = entity->fqn;
+      }
+
+      auto result = fault_manager_->list_faults(source_id);
+      if (result.success && result.data.is_array()) {
```
mfaferek93: `list_entity_faults` drops every local fault here: `result.data.is_array()` is always false because `FaultManager::list_faults` returns an object `{"faults":[...],"count":N}` (fault_manager.cpp:209), not an array. The old code had the same check and silently returned empty for COMPONENT/APP all along; now the method returns peer faults only, so the local+peer merge never happens. Read the array out: `if (result.success && result.data.contains("faults") && result.data["faults"].is_array()) local_faults = result.data["faults"];`

src/ros2_medkit_gateway/src/plugins/plugin_context.cpp
```diff
@@ -76,28 +101,45 @@ class GatewayPluginContext : public RosPluginContext {
...
-    if (result.success && result.data.is_array()) {
-      return result.data;
+    // Fan-out to peer gateways: fetch faults for this entity from each peer.
+    // The peer path mirrors the HTTP REST route: /api/v1/<type_plural>/<id>/faults.
+    auto * agg = node_->get_aggregation_manager();
+    if (agg) {
+      const std::string type_segment = entity_type_to_url_segment(entity->type);
+      if (!type_segment.empty()) {
+        const std::string peer_path = "/api/v1/" + type_segment + "/" + entity_id + "/faults";
+        auto fan_result = agg->fan_out_get(peer_path, "");
```
mfaferek93, line 138: Fan-out passes an empty `Authorization` header, but the HTTP mirror forwards the caller credential (`fault_handlers.cpp:446`). Against an authenticated peer every request 401s and the peer contributes nothing. Thread an auth/service token through the service request, or document this path as trusted/loopback-only. Reusing `fan_out_helpers.hpp`'s `merge_peer_items` (which already handles auth + `failed_peers`) would avoid re-deriving this inline.

src/ros2_medkit_gateway/src/plugins/plugin_context.cpp
```diff
@@ -76,28 +101,45 @@ class GatewayPluginContext : public RosPluginContext {
...
+        if (fan_result.merged_items.is_array()) {
```
mfaferek93, line 139: Both methods use only `merged_items` and drop `fan_result.is_partial` / `fan_result.failed_peers`, so a peer that is down, 401s, or times out silently truncates the fault list with no field and no log - whereas the HTTP path surfaces both into `x-medkit` (`fault_handlers.cpp:452`). For a diagnostics gateway that's a real hazard. At least `RCLCPP_WARN` on `is_partial` listing the failed peers; ideally carry `partial`/`failed_peers` in the service response.

src/ros2_medkit_gateway/src/plugins/plugin_context.cpp
```diff
@@ -76,28 +101,45 @@ class GatewayPluginContext : public RosPluginContext {
...
-      source_id = entity->namespace_path;
-    } else if (entity->type == SovdEntityType::APP) {
-      source_id = entity->fqn;
+    nlohmann::json local_faults = nlohmann::json::array();
+
+    if (fault_manager_ && fault_manager_->is_available()) {
+      std::string source_id;
+      if (entity->type == SovdEntityType::COMPONENT) {
+        source_id = entity->namespace_path;
+      } else if (entity->type == SovdEntityType::APP) {
```
mfaferek93: `source_id` is set only for COMPONENT and APP; FUNCTION and AREA fall through with `source_id == ""`, so `list_faults("")` returns ALL local faults as that entity's faults (once the extraction bug above is fixed). The HTTP path filters FUNCTION faults by host-app FQNs and rejects types without a faults collection (`validate_collection_access`, SOVD Table 8). Please mirror that scoping/validation, and skip the peer fan-out for types with no faults collection.

src/ros2_medkit_gateway/src/plugins/plugin_context.cpp
```diff
@@ -207,14 +249,31 @@ class GatewayPluginContext : public RosPluginContext {
...
+    auto * agg = node_->get_aggregation_manager();
+    if (agg) {
+      auto fan_result = agg->fan_out_get("/api/v1/faults", "");
+      if (fan_result.merged_items.is_array()) {
+        // Ensure local response has a "faults" array to append to.
+        if (!response.contains("faults") || !response["faults"].is_array()) {
+          response["faults"] = nlohmann::json::array();
+        }
+        for (const auto & item : fan_result.merged_items) {
+          response["faults"].push_back(item);
```
mfaferek93, line 282: Peer items are appended to `response["faults"]`, but `response` is `result.data` which also carries `"count"` from the local query - so `count` is now stale (local-only) while `faults` includes peers. Recompute `count = response["faults"].size()` after the merge, or drop it.

src/ros2_medkit_gateway/test/test_plugin_context_aggregation.cpp
```diff
@@ -0,0 +1,439 @@
...
+  EXPECT_EQ(faults.size(), 0u);
+}
+
+// =============================================================================
+// Tests: with aggregation enabled and a live mock peer
+// =============================================================================
+
+// @verifies REQ_INTEROP_013
+TEST_F(PluginContextWithPeerTest, ListEntityFaults_IncludesPeerFault) {
+  auto ctx = ros2_medkit_gateway::make_gateway_plugin_context(node_.get(), nullptr, nullptr);
```
mfaferek93, line 558: Every fixture builds the context with a null `fault_manager` (`make_gateway_plugin_context(node_.get(), nullptr, nullptr)`), so the local+peer MERGE is never exercised - only the peer-only path. That's exactly the gap that hides the `list_entity_faults` extraction bug. Please add a fixture with a real `fault_manager` returning local faults and assert both local and peer faults appear (plus FUNCTION/AREA, an authenticated peer, and a down peer).

src/ros2_medkit_gateway/test/test_plugin_context_aggregation.cpp
```diff
@@ -0,0 +1,439 @@
...
+
+  httplib::Server & server() {
+    if (!server_) {
+      server_ = std::make_unique<httplib::Server>();
+    }
+    return *server_;
+  }
+
+  int start() {
+    port_ = server_->bind_to_any_port("127.0.0.1");
```
mfaferek93, line 122: `start()` dereferences `server_` via `bind_to_any_port`, but `server_` is only created lazily in `server()`. It works because the fixtures call `server()` first, but a caller that does `start()` first segfaults. Create the server here too (or assert it is set).

src/ros2_medkit_gateway/src/plugins/plugin_context.cpp
```diff
@@ -76,28 +101,45 @@ class GatewayPluginContext : public RosPluginContext {
...
+        source_id = entity->namespace_path;
+      } else if (entity->type == SovdEntityType::APP) {
+        source_id = entity->fqn;
+      }
+
+      auto result = fault_manager_->list_faults(source_id);
+      if (result.success && result.data.is_array()) {
+        local_faults = result.data;
+      }
     }
```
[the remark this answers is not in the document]
bburda: Fixed in 86b69daa: the local branch now reads `result.data["faults"]` and scopes it to the entity's resolved App-FQN set through the shared `faults::resolve_entity_source_fqns` + `filter_faults_by_sources` helper (the same resolution the HTTP per-entity handler uses), so local faults are returned and merged with peer faults instead of being dropped. Covered by new tests that inject a real `FaultManager` via a fake `FaultServiceTransport`.

src/ros2_medkit_gateway/src/plugins/plugin_context.cpp
```diff
@@ -207,14 +249,31 @@ class GatewayPluginContext : public RosPluginContext {
...
+    if (agg) {
+      auto fan_result = agg->fan_out_get("/api/v1/faults", "");
+      if (fan_result.merged_items.is_array()) {
+        // Ensure local response has a "faults" array to append to.
+        if (!response.contains("faults") || !response["faults"].is_array()) {
+          response["faults"] = nlohmann::json::array();
+        }
+        for (const auto & item : fan_result.merged_items) {
+          response["faults"].push_back(item);
+        }
```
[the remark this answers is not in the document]
bburda: Fixed in 86b69daa: `list_all_faults` now recomputes `count = response["faults"].size()` after merging peer faults, so it no longer reflects only the local query. Asserted in `ListAllFaults_MergesLocalAndPeerFaults`.

src/ros2_medkit_gateway/test/test_plugin_context_aggregation.cpp
```diff
@@ -0,0 +1,439 @@
...
+    thread_ = std::thread([this]() {
+      server_->listen_after_bind();
+    });
+    return port_;
+  }
```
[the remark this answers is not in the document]
bburda: Fixed in 86b69daa: `start()` now calls `server()` to construct the server before `bind_to_any_port`, so calling `start()` before `server()` no longer dereferences a null pointer.

src/ros2_medkit_gateway/test/test_plugin_context_aggregation.cpp
```diff
@@ -0,0 +1,439 @@
...
+// =============================================================================
+
+// @verifies REQ_INTEROP_013
+TEST_F(PluginContextWithPeerTest, ListEntityFaults_IncludesPeerFault) {
+  auto ctx = ros2_medkit_gateway::make_gateway_plugin_context(node_.get(), nullptr, nullptr);
+  ASSERT_NE(ctx, nullptr);
+
+  // "brakes" is in the local manifest. No local fault manager (null), but the
+  // peer returns a PROTECTIVE_STOP fault via /api/v1/apps/brakes/faults.
+  auto faults = ctx->list_entity_faults("brakes");
```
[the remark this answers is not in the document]
bburda: Fixed in 86b69daa: added a `FakeFaultTransport` and tests that build the context with a real `FaultManager` - local scoped fault (out-of-scope fault excluded), component/area/function resolution, local+peer merge, and a failing peer where local faults still survive.

src/ros2_medkit_gateway/src/plugins/plugin_context.cpp
bburda: Fixed in 86b69daa: the local branch now reads `result.data["faults"]` and scopes it to the entity's resolved App-FQN set through the shared `faults::resolve_entity_source_fqns` + `filter_faults_by_sources` helper (the same resolution the HTTP per-entity handler uses), so local faults are returned and merged with peer faults instead of being dropped. Covered by new tests that inject a real `FaultManager` via a fake `FaultServiceTransport`.

src/ros2_medkit_gateway/src/plugins/plugin_context.cpp
bburda: Documented in 86b69daa: the ROS 2 service path has no inbound HTTP Authorization header to forward, so peers are queried with an empty credential - a comment at the call site notes that fan-out over this path only surfaces faults from peers that do not gate `/faults`. Reusing `merge_peer_items` is not a clean fit here because it derives the credential from the HTTP `raw_req`, which the service path does not have. Threading a configured service token is a reasonable follow-up if authenticated peers must be reachable over the service path.

src/ros2_medkit_gateway/src/plugins/plugin_context.cpp
bburda: Addressed in 86b69daa: both methods now `RCLCPP_WARN` on `fan_result.is_partial`, logging the failed-peer count, and a new failing-peer test asserts local faults survive a partial fan-out. Carrying `partial`/`failed_peers` in the service payload was left out for now to keep the bare array / `{"faults": [...]}` contract, but can be added if a consumer needs to distinguish a degraded view.

src/ros2_medkit_gateway/src/plugins/plugin_context.cpp
bburda: Fixed in 86b69daa: all four entity types now resolve their App-FQN set through the shared `faults::resolve_entity_source_fqns` (APP -> itself, COMPONENT -> hosted apps, AREA -> descendant apps, FUNCTION -> host apps) and filter by it - the same resolution the HTTP handler uses - so FUNCTION/AREA no longer return all faults. Types with no faults collection (SERVER/UNKNOWN) resolve to an empty set and the peer fan-out is skipped. New FUNCTION and AREA tests assert the scoping and that out-of-scope faults do not leak in.

src/ros2_medkit_gateway/src/plugins/plugin_context.cpp
bburda: Fixed in 86b69daa: `list_all_faults` now recomputes `count = response["faults"].size()` after merging peer faults, so it no longer reflects only the local query. Asserted in `ListAllFaults_MergesLocalAndPeerFaults`.

src/ros2_medkit_gateway/test/test_plugin_context_aggregation.cpp
bburda: Fixed in 86b69daa: added a `FakeFaultTransport` and tests that build the context with a real `FaultManager` - local scoped fault (out-of-scope fault excluded), component/area/function resolution, local+peer merge, and a failing peer where local faults still survive.

src/ros2_medkit_gateway/test/test_plugin_context_aggregation.cpp
bburda: Fixed in 86b69daa: `start()` now calls `server()` to construct the server before `bind_to_any_port`, so calling `start()` before `server()` no longer dereferences a null pointer.

mfaferek93 (approved):
LGTM!

PR #423, opened 2026-06-15
feat(action_status_bridge): add aborted-goal to faults bridge

### LINKED ISSUE #421: feat(fault_pack): promote aborted action goals to faults
For the core stacks (Nav2, MoveIt 2, ros2_control), the authoritative failure is the action result, not `/rosout` or `/diagnostics`. A `MoveGroup` / `NavigateToPose` abort emits no `/rosout` ERROR and there is no `/diagnostics`, so log- and diagnostic-based rules miss it entirely. Catch it generically via the goal-status channel.

## Scope
- Discover every action via its `*/_action/status` topic (`action_msgs/msg/GoalStatusArray`); rescan the graph for actions that appear later.
- `ABORTED (6)` -> fault (default SEVERITY_ERROR); `CANCELED (5)` -> fault only if `canceled_is_fault`; `SUCCEEDED (4)` -> PASSED to heal the action's ABORTED code.
- `fault_code` = `<PREFIX>_<ACTION>_ABORTED`; `source_id` = action name.
- Per-goal dedup keyed on `goal_id` so a latched terminal status reports once.
- Reporter local filtering disabled for these terminal events (an abort is discrete, not flapping) so a single recovery heals reliably.

## Acceptance
- An unreachable `MoveGroup` / `NavigateToPose` goal produces `ACTION_<name>_ABORTED` CONFIRMED in `list_faults`, `source` = action name.
- A subsequent successful goal heals it to HEALED (FaultManager `healing_enabled`).
- A MoveIt abort produces the action fault with zero `LOG_*` faults present (the action channel is the only signal).
- Unit tests: action-name extraction from the status topic, fault_code, uuid/dedup.

mfaferek93:
Adds `ros2_medkit_action_status_bridge`, a generic drop-in adapter that discovers every action via its `*/_action/status` topic and turns terminal `ABORTED` goal states into FaultManager faults, healing the code on `SUCCEEDED`.

This captures the authoritative failure channel for the core stacks (Nav2, MoveIt 2, ros2_control), where an aborted goal reaches neither `/rosout` nor `/diagnostics`.

- `source_id` is the action name; `fault_code` is `<PREFIX>_<ACTION>_ABORTED`.
- Per-goal dedup on `goal_id` so a latched terminal status reports once.
- New actions appearing later are picked up by a periodic rescan.

Unit tests cover action-name extraction, fault_code generation, and dedup.

Closes #421

### DIFF
--- a/src/ros2_medkit_action_status_bridge/CHANGELOG.rst
+++ b/src/ros2_medkit_action_status_bridge/CHANGELOG.rst
@@ -1,0 +1,11 @@
+^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
+Changelog for package ros2_medkit_action_status_bridge
+^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
+
+Forthcoming
+-----------
+* Initial release: generic action-status bridge. Watches every
+  ``/<action>/_action/status`` topic and turns ABORTED goals into FaultManager
+  faults (``<PREFIX>_<ACTION>_ABORTED``), heals on SUCCEEDED, with per-goal
+  dedup. Captures the terminal action verdict that the diagnostic and log
+  bridges cannot see.
--- a/src/ros2_medkit_action_status_bridge/CHANGELOG.rst
+++ b/src/ros2_medkit_action_status_bridge/CHANGELOG.rst
@@ -6,6 +6,17 @@ Forthcoming
 -----------
 * Initial release: generic action-status bridge. Watches every
   ``/<action>/_action/status`` topic and turns ABORTED goals into FaultManager
-  faults (``<PREFIX>_<ACTION>_ABORTED``), heals on SUCCEEDED, with per-goal
-  dedup. Captures the terminal action verdict that the diagnostic and log
-  bridges cannot see.
+  faults (``<PREFIX>_<ACTION>_ABORTED``), heals on a non-failing terminal state.
+  Captures the terminal action verdict that the diagnostic and log bridges
+  cannot see.
+* Fault state is per-ACTION, derived from the whole ``GoalStatusArray`` on each
+  message: a fault is raised only on the ``healthy -> failed`` transition and
+  healed only on ``failed -> healthy``, so it is order-independent and resilient
+  to dropped terminal messages. Per-goal dedup now only suppresses duplicate log
+  lines.
+* ``canceled_is_fault`` emits a status-aware ``<PREFIX>_<ACTION>_CANCELED`` code;
+  such a fault heals on the next non-failing terminal or when the canceled goal
+  ages out of the retained status array.
+* Vanished actions (lifecycle deactivate, one-shot nodes) are pruned on rescan.
+* Parameters are range-checked/normalized at load; an incompatible action status
+  QoS is now warned about instead of silently dropping faults.
--- a/src/ros2_medkit_action_status_bridge/CMakeLists.txt
+++ b/src/ros2_medkit_action_status_bridge/CMakeLists.txt
@@ -1,0 +1,112 @@
+# Copyright 2026 mfaferek93, bburda
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+cmake_minimum_required(VERSION 3.8)
+project(ros2_medkit_action_status_bridge)
+
+set(CMAKE_CXX_STANDARD 17)
+set(CMAKE_CXX_STANDARD_REQUIRED ON)
+set(CMAKE_EXPORT_COMPILE_COMMANDS ON)
+
+find_package(ros2_medkit_cmake REQUIRED)
+include(ROS2MedkitCcache)
+include(ROS2MedkitSanitizers)
+include(ROS2MedkitLinting)
+include(ROS2MedkitWarnings)
+
+option(ENABLE_COVERAGE "Enable code coverage reporting" OFF)
+if(ENABLE_COVERAGE)
+  message(STATUS "Code coverage enabled")
+  add_compile_options(--coverage -O0 -g)
+  add_link_options(--coverage)
+endif()
+
+find_package(ament_cmake REQUIRED)
+
+include(ROS2MedkitCompat)
+
+find_package(rclcpp REQUIRED)
+find_package(action_msgs REQUIRED)
+find_package(ros2_medkit_msgs REQUIRED)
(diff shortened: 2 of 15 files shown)


src/ros2_medkit_action_status_bridge/src/action_status_bridge_node.cpp
```diff
@@ -0,0 +1,437 @@
...
+
+ros2_medkit_fault_reporter::FaultReporter * ActionStatusBridgeNode::reporter_for(const std::string & action_name) {
+  std::lock_guard<std::mutex> lock(reporters_mutex_);
+  auto it = reporters_.find(action_name);
+  if (it != reporters_.end()) {
+    return it->second.get();
+  }
+  // Attribute the fault to the action (best stable identifier from the status
+  // topic alone). Multiple reporters on one node are safe (has_parameter guard).
+  auto reporter = std::make_unique<ros2_medkit_fault_reporter::FaultReporter>(this->shared_from_this(), action_name);
```
bburda: **[major] Faults attributed to the action name won't associate with any SOVD entity.**

`reporter_for()` constructs the `FaultReporter` with `source_id = action_name` (e.g. `/navigate_to_pose`). The gateway resolves a fault to an entity by node FQN: `source_matches_scope()` requires `source_id` to equal an entity's owned node FQN or be a strict path-child of it. An action name is the action-interface name, not the server node's FQN (e.g. `/bt_navigator`), so it matches no entity. These faults are then silently filtered out of `GET /apps/{id}/faults` (and the component/area variants), and per-entity triggers/snapshots never resolve.

`ros2_medkit_diagnostic_bridge` attributes to `get_fully_qualified_name()` for exactly this reason. Please resolve the action server's node FQN (e.g. from the status topic's publisher) and use that as `source_id`, or document this association gap explicitly in the README Limitations.

src/ros2_medkit_action_status_bridge/src/action_status_bridge_node.cpp
```diff
@@ -0,0 +1,437 @@
...
+      continue;
+    }
+    const std::string action_name = action_name_from_status_topic(topic);
+    {
+      std::lock_guard<std::mutex> lock(reporters_mutex_);
+      reporters_.erase(action_name);
+    }
+    {
+      std::lock_guard<std::mutex> lock(state_mutex_);
+      last_reported_state_.erase(action_name);
```
bburda, line 214: **[major] Pruning a still-failed action leaves a stuck, un-healable fault.**

`prune_vanished()` erases `last_reported_state_` and the reporter for a vanished action but never heals it. If the action was in `kFailed` (a fault is active in the FaultManager), the fault stays active and the bridge forgets it raised it. When the action reappears, `last_reported_state_` is absent and `apply_message()` treats absent state as healthy, so the next `SUCCEEDED` is a no-op and the `failed -> healthy` heal never fires — the fault is stuck until a fresh full abort->succeed cycle.

This is the Nav2 lifecycle-deactivate / one-shot-node case the README advertises, and it contradicts the README claim that a dropped terminal message cannot leave a fault stuck. Suggest sending PASSED for the raised code before erasing state when the pruned action is currently `kFailed`.

src/ros2_medkit_action_status_bridge/CMakeLists.txt
```diff
@@ -0,0 +1,112 @@
...
+  find_package(ament_cmake_gtest REQUIRED)
+
+  set(ament_cmake_clang_format_CONFIG_FILE "${CMAKE_CURRENT_SOURCE_DIR}/../../.clang-format")
+  list(APPEND AMENT_LINT_AUTO_EXCLUDE ament_cmake_uncrustify ament_cmake_cpplint ament_cmake_clang_tidy)
+  ament_lint_auto_find_test_dependencies()
+
+  ros2_medkit_clang_tidy()
+
+  include(ROS2MedkitTestDomain)
+  medkit_init_test_domains(START 135 END 139)
```
bburda: **[important] Test-domain range 135-139 reuses a band reserved for another package.**

The allocation table in `ROS2MedkitTestDomain.cmake` reserves `130-139` for `ros2_medkit_linux_introspection`. This package takes `135-139` of that band (the log-bridge change takes `130-134`). There is no DDS clash today because linux_introspection doesn't yet set a domain, but this violates the documented allocation and creates a latent cross-contamination once it does. Please allocate a new range and update the table comment in the same change (note DDS max domain is 232).

src/ros2_medkit_action_status_bridge/CHANGELOG.rst
```diff
@@ -0,0 +1,22 @@
+^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
+Changelog for package ros2_medkit_action_status_bridge
+^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
+
+Forthcoming
+-----------
+* Initial release: generic action-status bridge. Watches every
```
bburda, line 7: **[important] New package isn't registered in the docs build.**

Unlike `ros2_medkit_diagnostic_bridge`, this package isn't wired into the documentation: it is missing from `docs/changelog.rst` (an `.. include::` of this CHANGELOG), `docs/Doxyfile` INPUT, the `doxygenclass` block in `docs/api/cpp.rst`, and `docs/introduction.rst`. The class carries full Doxygen comments but won't appear in the published API docs. Please add the corresponding entries.

src/ros2_medkit_action_status_bridge/CMakeLists.txt
```diff
@@ -0,0 +1,112 @@
...
+
+  ament_add_gtest(test_action_status_bridge test/test_action_status_bridge.cpp)
```
bburda, line 100: **[important] No integration test for the live path.**

Only unit tests are registered here; `ros2_medkit_diagnostic_bridge` ships a launch_testing `test_integration.test.py` that brings up the fault manager + bridge and asserts a real fault lands. The bridge's core guarantee — that healing isn't throttled, which depends entirely on the `fault_reporter.local_filtering.enabled=false` pre-declaration — is never exercised end-to-end (every unit test calls `apply_message()` with a `nullptr` reporter). A regression in that wiring would raise-but-never-heal with a green suite. Please add an integration test covering discovery plus the raise/heal round-trip.

bburda (commented):
Review findings inline (severity tagged per comment).

src/ros2_medkit_action_status_bridge/src/action_status_bridge_node.cpp
mfaferek93: fixed

src/ros2_medkit_action_status_bridge/src/action_status_bridge_node.cpp
mfaferek93: fixed

src/ros2_medkit_action_status_bridge/CHANGELOG.rst
mfaferek93: fixed

src/ros2_medkit_action_status_bridge/CMakeLists.txt
mfaferek93: added

src/ros2_medkit_action_status_bridge/CMakeLists.txt
mfaferek93: fixed

bburda (dismissed):
LGTM!

bburda (approved):

PR #430, opened 2026-06-16
fix(fault_manager): make rosbag capture enablement crash-safe

### LINKED ISSUE #425: fix(fault_manager): make rosbag capture enablement crash-safe
Enabling black-box rosbag capture is the exact path that can crash the fault_manager on startup. `RosbagCapture`'s constructor validates the storage format by opening a throwaway bag, which throws `std::runtime_error` when the storage plugin (e.g. `rosbag2_storage_mcap`) is not available at runtime. That construction in `fault_manager_node.cpp` is not wrapped in try/catch and `main.cpp` has none either, so the exception propagates out of `main` and terminates the process. The published Docker image does not install `rosbag2_storage_mcap`, so a user who sets `snapshots.rosbag.enabled:=true` with the default format hits this.

Compounding: the default storage format is `mcap` in code but the shipped YAML config documents `sqlite3`, so code and docs disagree on what a user actually gets.

## Scope
- Wrap the `RosbagCapture` construction so a storage/format failure disables rosbag capture and logs a clear warning instead of terminating the node.
- Reconcile the default storage format so the code and the YAML config agree.
- Either ship the storage plugin in the binary/Docker distribution or default to a format that needs no extra plugin, so enabling capture works out of the box.

## Acceptance
- With `snapshots.rosbag.enabled:=true` and the storage plugin absent, the fault_manager stays up, logs a warning, and continues without rosbag capture.
- The default storage format is identical in code and in the shipped config.

mfaferek93:
Enabling black-box rosbag capture could crash the FaultManager on startup: `RosbagCapture`'s constructor validated the storage format by opening a throwaway bag, which threw `std::runtime_error` when the plugin (e.g. `rosbag2_storage_mcap`) was not installed, and the exception propagated out of `main` and terminated the process.

The constructor now resolves a usable backend without ever throwing on format:
- an unknown format, or a configured format whose plugin is missing, falls back to `sqlite3` (always shipped with rosbag2) for black-box capture;
- if no backend is usable, capture disables itself with a clear warning and the node keeps running (freeze-frame snapshots are independent of rosbag).

Also makes `sqlite3` the default storage format (zero extra setup, always available) and reconciles code and the shipped `snapshots.yaml`, which previously disagreed (code defaulted to `mcap`, the plugin that triggered the crash; config documented `sqlite3`). `mcap` stays an opt-in for Foxglove, protected by the fallback above.

Unit test updated: an unknown/unavailable format no longer throws; it degrades to `sqlite3` with capture still enabled.

Closes #425

### DIFF
--- a/src/ros2_medkit_fault_manager/src/rosbag_capture.cpp
+++ b/src/ros2_medkit_fault_manager/src/rosbag_capture.cpp
@@ -18,17 +18,38 @@
 #include <unistd.h>
 
 #include <algorithm>
+#include <cstdlib>
 #include <filesystem>
+#include <optional>
 #include <rosbag2_cpp/writer.hpp>
 #include <rosbag2_storage/storage_options.hpp>
 #include <set>
 #include <sstream>
+#include <utility>
 
 #include "ros2_medkit_fault_manager/time_utils.hpp"
 
 namespace ros2_medkit_fault_manager {
 
 namespace {
+
+/// Actionable install hint for an unavailable storage backend.
+std::string storage_plugin_hint(const std::string & format) {
+  if (format == "mcap") {
+    const char * distro = std::getenv("ROS_DISTRO");
+    const std::string d = (distro && *distro) ? distro : "$ROS_DISTRO";
+    return "install ros-" + d + "-rosbag2-storage-mcap";
+  }
+  return "check the rosbag2 storage plugin installation";
+}
+
+/// Bound the probe reason so a verbose pluginlib error does not flood the log.
+std::string truncate_reason(const std::string & reason, size_t max_len = 200) {
+  if (reason.size() <= max_len) {
+    return reason;
+  }
+  return reason.substr(0, max_len) + "...";
+}
 
 /// Custom deleter for rcutils_uint8_array_t that calls rcutils_uint8_array_fini
 struct Uint8ArrayDeleter {
@@ -85,7 +106,7 @@ std::shared_ptr<rosbag2_storage::SerializedBagMessage> create_bag_message(const 
 }  // namespace
 
 RosbagCapture::RosbagCapture(rclcpp::Node * node, FaultStorage * storage, const RosbagConfig & config,
-                             const SnapshotConfig & snapshot_config)
+                             const SnapshotConfig & snapshot_config, StorageProbeFn storage_probe)
   : node_(node), storage_(storage), config_(config), snapshot_config_(snapshot_config) {
   if (!node_) {
     throw std::invalid_argument("RosbagCapture requires a valid node pointer");
@@ -93,6 +114,10 @@ RosbagCapture::RosbagCapture(rclcpp::Node * node, FaultStorage * storage, const 
   if (!storage_) {
     throw std::invalid_argument("RosbagCapture requires a valid storage pointer");
   }
+
+  storage_probe_ = storage_probe ? std::move(storage_probe) : [this](const std::string & f) {
+    return default_storage_probe(f);
+  };
 
   if (!config_.enabled) {
     RCLCPP_INFO(node_->get_logger(), "RosbagCapture disabled");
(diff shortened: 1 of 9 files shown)


src/ros2_medkit_fault_manager/test/test_rosbag_capture.cpp
```diff
@@ -109,11 +109,17 @@ TEST_F(RosbagCaptureTest, ConstructorWithDisabledRosbag) {
 }
 
 // @verifies REQ_INTEROP_088
-TEST_F(RosbagCaptureTest, ConstructorThrowsOnInvalidFormat) {
+TEST_F(RosbagCaptureTest, ConstructorFallsBackOnUnknownFormat) {
+  // An unknown/unavailable format must NOT terminate the node; it degrades to
+  // sqlite3 (always shipped with rosbag2) and capture stays enabled.
   auto rosbag_config = create_rosbag_config();
   rosbag_config.format = "invalid_format";
```
bburda, line 117: This renamed test only exercises the *unknown-format-string* branch: `"invalid_format"` is rewritten to `sqlite3` by the first guard before the availability probe runs, so it never reaches the probe-fails-then-fallback path. The two branches that actually implement crash-safety are uncovered:

1. a *known* format (`mcap`) whose plugin is unavailable falling back to `sqlite3` (the exact scenario from the linked issue), and
2. no usable backend -> capture self-disables (`is_enabled() == false`) while the node keeps running.

Both are currently hard to test because the probe calls into rosbag2 and mcap is installed in CI. Suggest extracting `storage_format_available()` into an injectable seam (`protected virtual`, or a probe functor) so a test double can force a known format to report unavailable, then add one test per branch. As written, the primary scenario the change exists to fix has no direct coverage.

src/ros2_medkit_fault_manager/src/rosbag_capture.cpp
```diff
@@ -99,8 +99,31 @@ RosbagCapture::RosbagCapture(rclcpp::Node * node, FaultStorage * storage, const
...
+  // shipped with rosbag2); if no backend works, disable capture and keep running
+  // (freeze-frame snapshots are independent of rosbag).
+  if (config_.format != "sqlite3" && config_.format != "mcap") {
+    RCLCPP_WARN(node_->get_logger(), "Unknown rosbag storage format '%s'; using 'sqlite3'", config_.format.c_str());
+    config_.format = "sqlite3";
+  }
+  if (!storage_format_available(config_.format)) {
+    if (config_.format != "sqlite3" && storage_format_available("sqlite3")) {
+      RCLCPP_WARN(node_->get_logger(),
+                  "Rosbag storage format '%s' is unavailable (see debug log for the probe error); falling back "
```
bburda: When an explicitly-configured format (e.g. `mcap` for Foxglove) is unavailable, the operator is silently downgraded to `sqlite3` (whose bags Foxglove cannot open directly) and pointed at the debug log - which is off by default and, once enabled, shows a pluginlib error rather than the remedy. Consider surfacing the actionable fix at WARN for an explicitly-configured (non-default) format: name the package to install (`ros-$ROS_DISTRO-rosbag2-storage-mcap`) and include a truncated probe reason (have `storage_format_available()` return the captured `e.what()`, e.g. via `std::optional<std::string>`). Keep the DEBUG line for the default `sqlite3` probe noise.

src/ros2_medkit_fault_manager/src/rosbag_capture.cpp
```diff
@@ -99,8 +99,31 @@ RosbagCapture::RosbagCapture(rclcpp::Node * node, FaultStorage * storage, const
...
+                  "to 'sqlite3' for black-box capture",
+                  config_.format.c_str());
+      config_.format = "sqlite3";
+    } else {
+      RCLCPP_WARN(node_->get_logger(),
+                  "No usable rosbag storage backend ('%s' and 'sqlite3' both unavailable); black-box rosbag "
```
bburda: When the configured format is already `sqlite3` and its probe fails, `config_.format` is `"sqlite3"`, so this logs `('sqlite3' and 'sqlite3' both unavailable)`. It reads like a bug and obscures that the always-shipped baseline failed to load, which is a distinct, environment-level condition worth calling out on its own. Suggest special-casing `config_.format == "sqlite3"` with a dedicated message (e.g. "sqlite3 rosbag storage is unavailable; rosbag2 base install may be broken").

bburda (commented):
A few related items outside this diff worth adding:

- `docs/tutorials/snapshots.rst` (the rosbag format note) still tells users to manually switch to `sqlite3` when the mcap plugin is missing, and does not mention the new automatic fallback / self-disable behavior. Since that behavior is the main user-visible change here, the tutorial should be updated alongside it.
- `ros2_medkit_fault_manager/package.xml` declares a hard `<depend>rosbag2_storage_mcap</depend>`, which is inconsistent with treating mcap as opt-in with a fallback. For rosdep installs this keeps the plugin always present, so the new fallback path is only reachable in distributions that omit it. Reconciling it (drop, or make it optional/exec-only and document the install) would align packaging with the new design.
- The gateway bulk-data handler defaults the bag `format` to `"mcap"` when the metadata field is absent (`bulkdata_handlers.cpp`, mimetype resolution). That default is now inconsistent with `sqlite3`. Low impact because the format is normally persisted per bag, but worth aligning (or deriving the mimetype from the on-disk `.db3`/`.mcap` extension the handler already inspects).

src/ros2_medkit_fault_manager/src/rosbag_capture.cpp
```diff
@@ -99,8 +99,31 @@ RosbagCapture::RosbagCapture(rclcpp::Node * node, FaultStorage * storage, const
...
+  // shipped with rosbag2); if no backend works, disable capture and keep running
+  // (freeze-frame snapshots are independent of rosbag).
+  if (config_.format != "sqlite3" && config_.format != "mcap") {
+    RCLCPP_WARN(node_->get_logger(), "Unknown rosbag storage format '%s'; using 'sqlite3'", config_.format.c_str());
+    config_.format = "sqlite3";
+  }
+  if (!storage_format_available(config_.format)) {
+    if (config_.format != "sqlite3" && storage_format_available("sqlite3")) {
+      RCLCPP_WARN(node_->get_logger(),
+                  "Rosbag storage format '%s' is unavailable (plugin not installed); falling back to "
```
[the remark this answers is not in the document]
mfaferek93: The warning now includes the actual probe reason, so it no longer assumes the plugin is simply missing.

src/ros2_medkit_fault_manager/src/rosbag_capture.cpp
```diff
@@ -698,42 +721,27 @@ void RosbagCapture::post_fault_timer_callback() {
...
-
-    throw std::runtime_error("Rosbag storage format '" + config_.format +
-                             "' is not available. "
-                             "Install the plugin or use a different format. "
-                             "Error: " +
-                             std::string(e.what()));
+    ok = true;
+  } catch (const std::exception &) {
+    ok = false;
   }
```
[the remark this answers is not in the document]
mfaferek93: The probe now returns the failure reason instead of swallowing it, and it is surfaced in the fallback/disable warning.

src/ros2_medkit_fault_manager/test/test_rosbag_capture.cpp
mfaferek93: Extracted the probe into an injectable seam and added direct tests for the known-format fallback and the no-usable-backend self-disable.

src/ros2_medkit_fault_manager/src/rosbag_capture.cpp
mfaferek93: An explicitly-configured non-default format now warns with the install command and the truncated probe reason before falling back.

src/ros2_medkit_fault_manager/src/rosbag_capture.cpp
mfaferek93: Special-cased: a failed sqlite3 baseline now gets its own 'base install may be broken' message instead of 'sqlite3 and sqlite3 both unavailable'.

bburda (approved):

PR #431, opened 2026-06-17
feat(fault_manager): entity-scoped rosbag capture by default

### LINKED ISSUE #426: feat(fault_manager): dynamic fault-relevant rosbag capture with QoS fidelity
Today the rosbag black-box default (`snapshots.rosbag.topics: "config"`) resolves topics from `snapshots.yaml`; with no config file it subscribes to nothing, so a confirmed fault produces an empty bag. The existing `"all"` mode records the whole graph but is blunt (not fault-relevant), subscribes with a hardcoded `SensorDataQoS` (BEST_EFFORT / volatile, `rosbag_capture.cpp:297`) so RELIABLE / transient_local topics are captured unfaithfully, and the ring buffer is time-bounded only (`prune_buffer`, `rosbag_capture.cpp:403`) with no memory cap, so recording high-rate topics grows RAM without bound.

Make the black-box give fault-relevant value out of the box with zero per-stack config, while keeping full manual control.

## Scope
- New default topic-selection mode (entity-scoped): buffer continuously, and on fault confirmation write only the topics published/subscribed by the faulting entity (resolved from the fault `source_id` node FQN via `get_publisher_names_and_types_by_node` / `get_subscription_names_and_types_by_node`) plus always-on context (`/tf`, `/tf_static`). Keeps the pre-fault ring while keeping the on-disk bag small and relevant.
- QoS fidelity: subscribe with each topic's offered QoS instead of a fixed `SensorDataQoS`, matching reliability/durability/depth per topic (the pattern already exists in `snapshot_capture.cpp:199` via `get_publishers_info_by_topic`). Define a policy for topics whose publishers offer differing QoS.
- Bound the ring buffer by memory: add a byte/size cap on top of the time window, and ship sensible default `exclude_topics` patterns for high-bandwidth topics (image / compressed / points / depth).
- Preserve all existing modes (`config`, `explicit`, `all`, comma-separated list) and the `include_topics` / `exclude_topics` overrides, which keep applying on top of any mode. Manual modes stay literal (no entity flush-filter). Existing `topics: config` + `snapshots.yaml` behaviour is unchanged.

## Acceptance
- On a stack with no `snapshots.yaml`, a confirmed fault produces a non-empty bag containing the faulting entity's topics plus `/tf`, with no manual configuration.
- RELIABLE and transient_local topics are captured with matching QoS (verified on a stack that uses them, e.g. Nav2 `/tf`, action status).
- Ring buffer memory stays bounded while recording a high-rate topic.
- `topics: config` with an existing `snapshots.yaml` behaves exactly as before; `include_topics` / `exclude_topics` still add/remove on top of the default mode.

Relates to #424.

mfaferek93:
Stacked on #430 (#425); the diff against `feat/425-rosbag-crash-safe` is only the #426 changes. Rebase to `main` once #430 merges.

Today the black-box rosbag's default `topics: "config"` mode captures nothing unless a YAML maps the fault code to topics, so an enabled black box often produces an empty bag. This makes it useful out of the box.

- **Entity-scoped capture, now the default (`topics: "entity"`)**: the ring buffer subscribes broadly for pre-roll, but on fault confirmation only the faulting source node's topics (resolved from the fault's reporting source via the ROS graph) plus `/tf` and `/tf_static` are written to the bag. If the source cannot be resolved to a live node, it falls back to writing the full buffer (a broad bag beats an empty one). Manual modes (`config`/`all`/`explicit`/comma-list) are unchanged and write exactly what they subscribe to.
- **Faithful QoS (`qos_match`, default on)**: each subscription matches the publisher-offered QoS (reliable / transient-local where all publishers offer it) instead of forcing best-effort `SensorDataQoS`, which silently dropped messages on reliable topics. Falls back to `SensorDataQoS` when no publisher is known yet.
- **Bounded memory**: a ring-buffer RAM cap (`max_buffer_mb`, default 256) drops the oldest buffered messages past the limit, and broad modes auto-exclude high-bandwidth sensor topics (image/points/depth/compressed); `include_topics` re-adds any you need.

New integration test `test_rosbag_entity_scope` runs two publisher nodes and asserts a fault sourced at one yields a bag with that node's topics and not the other's. Unit tests cover the new modes and flags.

Closes #426

### DIFF
--- a/src/ros2_medkit_fault_manager/config/snapshots.yaml
+++ b/src/ros2_medkit_fault_manager/config/snapshots.yaml
@@ -86,13 +86,18 @@ rosbag:
   # Continue recording after fault is confirmed to capture aftermath
   duration_after_sec: 1.0
 
-  # Topic selection mode (default: "config")
+  # Topic selection mode (default: "entity")
   # Options:
+  #   "entity"   - DEFAULT. Subscribe broadly for pre-roll, but on fault confirmation
+  #                write only the faulting source node's topics (+ /tf, /tf_static),
+  #                resolved from the fault's reporting source. Useful black box with
+  #                zero per-topic config. Falls back to the full buffer if the source
+  #                cannot be resolved to a live node.
   #   "config"   - Use same topics as JSON snapshots (from this config file)
-  #   "all"      - Record all available topics (WARNING: high memory usage!)
+  #   "all"      - Record all available topics
   #   "auto"     - Alias for "all" (auto-discover all topics)
   #   "explicit" - Use only topics from include_topics list below
-  topics: "config"
+  topics: "entity"
 
   # Explicit topic list (only used when topics: "explicit")
   # include_topics:
@@ -106,6 +111,11 @@ rosbag:
   #   - /camera/image_raw
   #   - /camera/depth/image_raw
   #   - /pointcloud
+
+  # Auto-exclude high-bandwidth sensor topics (image/points/depth/compressed) in
+  # broad modes ("all"/"auto"/"entity") to bound memory (default: true).
+  # Excluded topics are dropped silently; list them in include_topics to re-add.
+  exclude_sensor_topics: true
 
   # Lazy start mode (default: false)
   # When true, ring buffer only starts when fault enters PREFAILED state
@@ -133,6 +143,16 @@ rosbag:
   # Maximum total storage for all bag files in MB (default: 500)
   # Oldest bags are deleted when this limit is exceeded
   max_total_storage_mb: 500
+
+  # Maximum in-memory ring buffer size in MB (default: 256)
+  # Oldest buffered messages are dropped once the buffer exceeds this, so a broad
+  # subscribe set on a busy robot cannot grow memory without bound.
+  max_buffer_mb: 256
+
+  # Match each topic's publisher-offered QoS for faithful capture (default: true)
+  # Captures reliable / transient-local topics without dropping; set false to force
+  # best-effort SensorDataQoS on every subscription.
+  qos_match: true
 
   # Auto-cleanup bag files when fault is cleared (default: true)
   # When true, bag files are deleted when ClearFault is called
--- a/src/ros2_medkit_fault_manager/src/rosbag_capture.cpp
+++ b/src/ros2_medkit_fault_manager/src/rosbag_capture.cpp
@@ -18,6 +18,7 @@
 #include <unistd.h>
 
 #include <algorithm>
+#include <array>
 #include <cstdlib>
 #include <filesystem>
 #include <optional>
@@ -211,11 +212,13 @@ void RosbagCapture::stop() {
 
   // Clear subscriptions
   subscriptions_.clear();
+  subscribed_topics_.clear();
 
   // Clear buffer
   {
     std::lock_guard<std::mutex> lock(buffer_mutex_);
     message_buffer_.clear();
+    buffer_bytes_ = 0;
   }
 
   RCLCPP_INFO(node_->get_logger(), "RosbagCapture stopped");
@@ -250,6 +253,10 @@ void RosbagCapture::on_fault_confirmed(const std::string & fault_code) {
   }
 
   RCLCPP_INFO(node_->get_logger(), "RosbagCapture: fault '%s' confirmed, flushing buffer to bag", fault_code.c_str());
+
+  // In "entity" mode, narrow the broadly-buffered messages to the faulting node's
+  // topics. Sets the filter used by flush_to_bag() and the post-fault write path.
+  resolve_entity_topics(fault_code);
 
   // Flush buffer to bag
   std::string bag_path = flush_to_bag(fault_code);
@@ -298,6 +305,9 @@ void RosbagCapture::on_fault_confirmed(const std::string & fault_code) {
 
     RCLCPP_INFO(node_->get_logger(), "Bag file created: %s (%.2f MB)", bag_path.c_str(),
                 static_cast<double>(bag_size) / (1024.0 * 1024.0));
+
+    std::lock_guard<std::mutex> lock(capture_topics_mutex_);
+    active_capture_topics_.clear();
   }
 }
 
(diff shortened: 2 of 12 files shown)


bburda (commented):

- package.xml does not declare rosbag2_py. This is the first test that imports it, so CI can fail with package not found. Please add it as a test_depend (launch_ros too, for clarity).
- The entity default and the new params (qos_match, max_buffer_mb, exclude_sensor_topics) are not in the docs: tutorials/snapshots.rst, config/fault-manager.rst (which also lists a "none" mode that does not exist), and the rosbag table in the gateway README. The default also changed from config to entity, worth a note for upgrades.
- diagnostic_bridge reports source_id as its own node FQN, so for faults coming from the bridge, entity mode scopes the bag to the bridge (/diagnostics) and not the faulting node. Worth documenting as a limitation.

src/ros2_medkit_fault_manager/src/rosbag_capture.cpp
```diff
@@ -509,6 +551,110 @@ std::vector<std::string> RosbagCapture::resolve_topics() const {
...
+  qos.reliability(all_reliable ? rclcpp::ReliabilityPolicy::Reliable : rclcpp::ReliabilityPolicy::BestEffort);
+  qos.durability(all_transient_local ? rclcpp::DurabilityPolicy::TransientLocal : rclcpp::DurabilityPolicy::Volatile);
+  return qos;
+}
+
+void RosbagCapture::resolve_entity_topics(const std::string & fault_code) {
+  std::set<std::string> topics;
+
+  if (config_.topics == "entity") {
+    auto fault = storage_->get_fault(fault_code);
```
bburda: get_fault() is outside the try/catch below (which only covers the graph loop). With the default sqlite backend it can throw (prepare/bind/step, e.g. on SQLITE_BUSY), and on_fault_confirmed runs on a bare std::thread with no catch, so a throw here ends the process. That is the same crash the parent crash-safety PR just fixed. Suggest wrapping this (or the whole function body) in try/catch and falling back to the empty filter, like the graph loop already does.

src/ros2_medkit_fault_manager/src/rosbag_capture.cpp
```diff
@@ -509,6 +551,110 @@ std::vector<std::string> RosbagCapture::resolve_topics() const {
...
+        // Graph query raced shutdown - keep whatever resolved so far.
+      }
+    }
+
+    if (!topics.empty()) {
+      // Always-on context that makes a scoped bag useful for replay.
+      topics.insert("/tf");
+      topics.insert("/tf_static");
+    } else {
+      RCLCPP_DEBUG(node_->get_logger(), "Entity mode: could not resolve topics for fault '%s'; writing full buffer",
```
bburda: Two things on the entity filter. It is built from the graph and not intersected with the subscribe set, so a faulting node's sensor or excluded topic can end up in the filter while it was never buffered, so it is silently missing from the bag. And this fallback to the full buffer is logged only at DEBUG, so an operator sees a broad bag with no reason. Suggest intersecting with the subscribed topics, and raising this to WARN plus an INFO that names the resolved scope.

src/ros2_medkit_fault_manager/src/rosbag_capture.cpp
```diff
@@ -18,18 +18,32 @@
...
 
 #include "ros2_medkit_fault_manager/time_utils.hpp"
 
 namespace ros2_medkit_fault_manager {
 
 namespace {
 
+/// High-bandwidth sensor topics auto-excluded from broad-mode capture to bound memory.
+/// Substring match; deliberately coarse - users re-add via include_topics if needed.
+bool is_high_bandwidth_topic(const std::string & topic) {
```
bburda: This substring match also catches /waypoints, /setpoints, /keypoints (they contain points) and similar low-bandwidth topics, which then get dropped silently in broad and entity modes. Could match whole path segments or suffixes instead (e.g. /image, /points, /compressed).

src/ros2_medkit_fault_manager/src/rosbag_capture.cpp
```diff
@@ -509,6 +551,110 @@ std::vector<std::string> RosbagCapture::resolve_topics() const {
...
 
+rclcpp::QoS RosbagCapture::resolve_topic_qos(const std::string & topic) const {
+  std::vector<rclcpp::TopicEndpointInfo> pubs;
+  try {
+    pubs = node_->get_publishers_info_by_topic(topic);
+  } catch (const std::exception &) {
+    return rclcpp::SensorDataQoS();
+  }
+  if (pubs.empty()) {
+    // No known publisher yet - best-effort default. Not upgraded once subscribed,
```
bburda, line 619: QoS is fixed at subscribe time. If a best-effort publisher appears later on a topic already subscribed as reliable, the subscription is incompatible and silently receives nothing from it. Probably fine, but worth a note in the comment, or re-resolving QoS on the discovery-retry path.

src/ros2_medkit_fault_manager/test/test_rosbag_capture.cpp
```diff
@@ -243,6 +243,29 @@ TEST_F(RosbagCaptureTest, ExcludeTopicsRespected) {
   EXPECT_NO_THROW(RosbagCapture(node_.get(), storage_.get(), rosbag_config, snapshot_config));
 }
 
+TEST_F(RosbagCaptureTest, EntityTopicsMode) {
```
bburda, line 305: These three new tests only check that the constructor does not throw, so they do not exercise entity narrowing, QoS matching or sensor exclusion (EntityTopicsMode behaves like AllTopicsMode here). The RAM cap has no test at all. Could expose should_capture_topic / resolve_topic_qos / is_high_bandwidth_topic to tests and assert the actual behavior.

src/ros2_medkit_fault_manager/config/snapshots.yaml
```diff
@@ -107,6 +112,11 @@ rosbag:
   #   - /camera/depth/image_raw
   #   - /pointcloud
 
+  # Auto-exclude high-bandwidth sensor topics (image/points/depth/compressed) in
+  # broad modes ("all"/"auto"/"entity") to bound memory (default: true).
+  # include_topics re-adds any you specifically need.
+  # exclude_sensor_topics: true
```
bburda: exclude_sensor_topics is commented out here while max_buffer_mb and qos_match below are not. Worth one style for the three new keys. Also worth saying that excluded topics are dropped silently and include_topics re-adds them.

src/ros2_medkit_fault_manager/src/rosbag_capture.cpp
```diff
@@ -565,6 +710,11 @@ std::string RosbagCapture::flush_to_bag(const std::string & fault_code) {
     // Write messages (no buffer_mutex_ held, writer_mutex_ only for brief access)
     size_t msg_count = 0;
     for (const auto & msg : messages_to_write) {
+      // Entity mode: write only the faulting node's topics (+ /tf). No-op otherwise.
+      if (!should_capture_topic(msg.topic)) {
+        continue;
+      }
+
```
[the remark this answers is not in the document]
mfaferek93: The active filter is now snapshotted once before the flush loop and checked against the local copy.

src/ros2_medkit_fault_manager/src/rosbag_capture.cpp
```diff
@@ -509,6 +551,108 @@ std::vector<std::string> RosbagCapture::resolve_topics() const {
...
+  std::vector<rclcpp::TopicEndpointInfo> pubs;
+  try {
+    pubs = node_->get_publishers_info_by_topic(topic);
+  } catch (const std::exception &) {
+    return rclcpp::SensorDataQoS();
+  }
+  if (pubs.empty()) {
+    // No known publisher yet (discovery race) - best-effort default, retried later.
+    return rclcpp::SensorDataQoS();
+  }
```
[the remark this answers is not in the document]
mfaferek93: Corrected the comment: there is no resubscribe path, and it now states the QoS is fixed at subscribe time.

src/ros2_medkit_fault_manager/src/rosbag_capture.cpp
mfaferek93: Wrapped the whole entity resolution (including get_fault) in try/catch, falling back to the full buffer on any throw.

src/ros2_medkit_fault_manager/src/rosbag_capture.cpp
mfaferek93: The entity filter is now intersected with the subscribed set, and the resolved scope and full-buffer fallback are logged at INFO/WARN.

src/ros2_medkit_fault_manager/src/rosbag_capture.cpp
mfaferek93: Now matches path segments (/image, /points, /depth, /compressed), so /waypoints, /setpoints and /keypoints are kept.

src/ros2_medkit_fault_manager/src/rosbag_capture.cpp
mfaferek93: Documented in the comment; QoS is intentionally not re-resolved on the discovery-retry path.

src/ros2_medkit_fault_manager/test/test_rosbag_capture.cpp
mfaferek93: Made is_high_bandwidth_topic a public helper with a direct unit test; entity narrowing and QoS stay covered by the launch test.

src/ros2_medkit_fault_manager/config/snapshots.yaml
mfaferek93: exclude_sensor_topics is now active like the other two keys, with a note that excluded topics are dropped silently.

bburda (approved):

Issue #436, opened 2026-06-18
Host reboot for the host Component


bburda:
## Summary

Implement LifecycleProvider restart/shutdown for the host Component (the machine the gateway runs
on), so a Component-level reset reboots or shuts down the host.

## Proposed solution

- Implement restart/force-restart/shutdown/force-shutdown for the host Component via a host control
  mechanism (for example systemd reboot/poweroff).
- Privileged, high-impact action: off by default, explicitly enabled in config, RBAC-gated, and
  advertised only when enabled.
- Remote host Components are handled by existing aggregation forwarding: the transition is
  forwarded to the peer gateway that owns the host.

Contributes coverage toward REQ_INTEROP_077-081 at the host/Component level.
Unit and integration tests, docs.

Issue #435, opened 2026-06-18
Process, container, and systemd restart via linux_introspection


bburda:
## Summary

Implement the LifecycleProvider for standard (non-lifecycle) nodes using the runtime substrate the
linux_introspection plugins already resolve (node -> PID, cgroup/container, systemd unit).

## Proposed solution

- signal path: graceful shutdown = SIGTERM to the node PID, force-shutdown = SIGKILL.
- systemd path: when a node maps to a systemd unit, stop/restart via the unit - this gives real
  start and respawn that raw signals cannot.
- container path: when a node runs in a container, restart via the container runtime.
- Otherwise restart is terminate plus respawn by an external supervisor where one is declared;
  if none, the start/restart URI is omitted (spec-compliant).
- Advertise only the transitions the resolved substrate can perform.

The plugins already map node -> PID / container / unit as read capabilities; this adds actuation.
Contributes coverage toward REQ_INTEROP_077-081 for process/container/systemd-backed nodes.
Unit and integration tests, docs.

Issue #434, opened 2026-06-18
ROS 2 lifecycle node actuation provider


bburda:
## Summary

Implement lifecycle control for rclcpp_lifecycle nodes through the LifecycleProvider contract.
Intra-process state transitions only (no process re-spawn).

## Proposed solution

- Detect lifecycle nodes at discovery by enumerating /{node}/change_state and /{node}/get_state
  services, record it on the App.
- Read state via GetState to enrich GET /status.
- Map transitions to ChangeState: shutdown via the lifecycle shutdown transition; restart as a
  deactivate -> cleanup -> configure -> activate cycle.
- Advertise only the transitions a given node can perform.

Contributes coverage toward REQ_INTEROP_077-081 for lifecycle-capable nodes.
Unit and integration tests, docs.

Issue #432, opened 2026-06-18
Configurable RBAC: define roles and permissions via config


bburda:
## Summary

The role-to-endpoint permission mapping is currently hardcoded. `AuthManager::check_authorization()`
only consults `AuthConfig::get_role_permissions()`, which is a compiled `method:path` map
(~600 lines in `auth_config.cpp`). Deployers cannot change which role an endpoint requires,
add permissions, or define their own roles without recompiling the gateway.

YAML config today only controls `require_auth_for` (none/write/all) and the client list
(`client:secret:role`). The four roles (viewer, operator, configurator, admin) and their
permissions are fixed.

## Proposed solution

The matching engine already exists, only the source of permissions is hardcoded:

- `check_authorization()` already does exact `method:path` match plus wildcard patterns
  via `matches_path()`. Feed it permissions loaded from config, layered over (or replacing)
  the built-in defaults.
- `JwtClaims.permissions` ("explicit permissions") is already issued into and decoded from
  the JWT, but `check_authorization()` never reads it. Wire it in so per-client explicit
  permissions are honored.
- Keep the current built-in mapping as the default when no config is provided, so existing
  deployments do not change behavior.

Open design point: `UserRole` is an enum (viewer/operator/configurator/admin). Custom
permissions for the existing roles is small. Defining new role names needs moving from the
enum to string-based roles / a role registry. We can ship custom-permissions first and decide
on custom-roles separately.

## Additional context

Relates to #259 (security hardening). New write endpoints currently need their role
assignment baked into the compiled defaults; this issue removes that limitation.

PR #437, opened 2026-06-18
feat(gateway): SOVD entity status and lifecycle control endpoints

### LINKED ISSUE #433: Entity status endpoints and lifecycle control contract (gateway core)
## Summary

Add the SOVD entity status and lifecycle endpoints to the gateway core. Status read works from
runtime state. The five control transitions are wired to a new provider interface with no
implementation in this issue, so they return 501 until a provider is registered.

Status/lifecycle applies to both Apps (a node, i.e. a service) and Components (the host/substrate).

## Proposed solution

Endpoints, dual-path /apps and /components:
- GET /{entity}/status - 200, status plus URIs only for the transitions the entity supports.
- PUT /{entity}/status/{start,restart,force-restart,shutdown,force-shutdown} - delegate to the
  provider; 501 when no provider owns the entity.

Read path (core, no plugin):
- App: ready when is_online, else notReady.
- Component: ready when the host is reachable (the runtime default Component is the host the
  gateway runs on; remote Components are reached through aggregation forwarding).
- Areas and Functions have no status endpoint.

Control path (interface only here):
- New per-entity LifecycleProvider interface, routed via the plugin manager like DataProvider/
  OperationProvider, plus a LifecycleManager that delegates to it. Substrate-owning plugins
  implement it, so the core stays mechanism-agnostic.
- Success -> 202 Accepted, Location points to the status resource, client polls GET /status.
  No separate execution resource.
- Errors: 403 insufficient-access-rights, 409 precondition-not-fulfilled, 501 when unsupported.

Contract:
- Typed DTO LifecycleStatusResponse (status plus optional transition URIs).
- Status is not a resource collection; advertise a status link in entity detail, update the root
  endpoint list and version-info.
- RBAC: GET = viewer, transitions = operator (current compiled defaults). Configurable roles in #432.

Verifies REQ_INTEROP_076. REQ_INTEROP_077-081 stay open until an actuation provider lands.
Unit and integration tests, docs, compliance matrix entry.

bburda:
## Summary

Adds the SOVD ISO 17978-3 entity status and lifecycle endpoints to the gateway core:

- `GET /{apps|components}/{id}/status` - returns `ready` / `notReady` plus the URIs of the
  transitions the entity supports.
- `PUT /{apps|components}/{id}/status/{start,restart,force-restart,shutdown,force-shutdown}`.

This is the core contract only. Status read works from runtime state: an App is `ready` when its
node is online, and the host Component is `ready` while it is reachable. The five control
transitions are routed to a new per-entity `LifecycleProvider` interface that has no implementation
in this PR, so control returns `501` until a substrate-owning plugin registers a provider. The
actuation work (ROS lifecycle nodes, process / container / systemd, host reboot) is out of scope
here and is tracked in follow-up issues #434, #435, #436.

Also in this PR:

- New `LifecycleProvider` interface plus a `LifecycleProviderErrorInfo` error struct, routed through
  `PluginManager` like the existing operation provider.
- `LifecycleStatusResponse` DTO, and a `status` link advertised on the app and component detail
  responses.
- RBAC: `GET` status is allowed for the viewer role and above; the PUT transitions for operator and
  above.
- Design doc, README, and `REQ_INTEROP_076` marked verified (077-081 stay open since control is a
  `501` stub).

Making the required role per endpoint configurable is tracked separately in #432. The version bump
is left to a maintainer at release time. No breaking changes; everything here is additive.

## Issue

- closes #433

## Type

- [x] New feature or tests

## Testing

Unit (GTest):

- DTO serialization, including the hyphenated wire keys (`force-restart`, `force-shutdown`).
- Provider routing in `PluginManager` (owned vs unowned entity).
- The status / lifecycle handlers: status read (`ready` / `notReady` for apps and the host
  component), 404 before 501 for an unknown entity, the five PUT transitions returning `501` with no
  provider, and the provider-present path via a mock provider (filled transition URIs, `202` +
  `Location`, and `precondition-not-fulfilled` -> 409).
- The new RBAC permissions, including a viewer-denied-PUT case.

Integration (launch_testing): `GET /apps/{id}/status` and `GET /components/{host}/status` return
`ready`, and `PUT /apps/{id}/status/restart` returns `501`. Tagged `@verifies REQ_INTEROP_076`.

clang-tidy is clean on the changed files.

## Checklist

- [x] Breaking changes are clearly described (none - additive)

### DIFF
--- a/src/ros2_medkit_gateway/design/lifecycle.rst
+++ b/src/ros2_medkit_gateway/design/lifecycle.rst
@@ -1,0 +1,295 @@
+Status and Lifecycle Endpoints
+==============================
+
+This document describes the design of the SOVD status/lifecycle endpoints
+in ros2_medkit_gateway. It covers the six REST routes, the status read
+semantics for each entity type, the ``LifecycleProvider`` plugin seam,
+error mapping, and SOVD requirement coverage.
+
+.. contents:: Table of Contents
+   :local:
+   :depth: 3
+
+Overview
+--------
+
+SOVD ISO 17978-3 defines a lifecycle model for software entities. A client
+reads the current lifecycle state via a GET and drives transitions via PUT
+sub-resources. The gateway implements these routes for **Apps** and
+**Components** only (Areas and Functions do not carry lifecycle state in the
+current SOVD profile).
+
+The six routes are:
+
++-------------------------------------------+--------+--------------------------------------+
+| Path                                      | Method | Description                          |
++===========================================+========+======================================+
+| ``/{entity}/{id}/status``                 | GET    | Read current lifecycle status        |
++-------------------------------------------+--------+--------------------------------------+
+| ``/{entity}/{id}/status/start``           | PUT    | Request transition to started state  |
++-------------------------------------------+--------+--------------------------------------+
+| ``/{entity}/{id}/status/restart``         | PUT    | Request controlled restart           |
++-------------------------------------------+--------+--------------------------------------+
+| ``/{entity}/{id}/status/force-restart``   | PUT    | Request forced restart               |
++-------------------------------------------+--------+--------------------------------------+
+| ``/{entity}/{id}/status/shutdown``        | PUT    | Request controlled shutdown          |
++-------------------------------------------+--------+--------------------------------------+
+| ``/{entity}/{id}/status/force-shutdown``  | PUT    | Request forced shutdown              |
++-------------------------------------------+--------+--------------------------------------+
+
+where ``{entity}`` is either ``apps`` or ``components``.
+
+Route Registration
+------------------
+
+The lifecycle routes are registered **outside** the four-entity-type loop in
+``rest_server.cpp::setup_routes()``, covering only ``apps`` and
+``components``:
+
+.. code-block:: cpp
+
+   for (const auto & et_lc :
+        std::vector<std::pair<const char *, const char *>>{
+            {"apps", "app"}, {"components", "component"}}) {
+     const std::string base_lc = "/" + et_lc.first + "/{" + et_lc.second + "_id}";
+
+     // PUT routes registered BEFORE GET /status to avoid the shorter path
+     // shadowing the more-specific fixed segments.
+     for (const auto & action :
+          {"start", "restart", "force-restart", "shutdown", "force-shutdown"}) {
+       reg.put<http::NoContent>(base_lc + "/status/" + action, ...);
+     }
+     reg.get<dto::LifecycleStatusResponse>(base_lc + "/status", ...);
+   }
+
+The PUT routes must be registered before the GET route. The typed
+``RouteRegistry`` matches fixed path segments (``/status/start``) before
+parameterized ones (``/status``), but registration order also affects the
+matching priority inside cpp-httplib's internal routing table.
+
+Status Read Semantics
+---------------------
+
+``GET /{entity}/{id}/status`` returns a ``LifecycleStatusResponse`` DTO:
+
+.. code-block:: text
+
+   {
+     "status": "ready" | "notReady",
+     "start":          "/api/v1/{entity}/{id}/status/start",    // optional
+     "restart":        "/api/v1/{entity}/{id}/status/restart",  // optional
+     "force-restart":  "/api/v1/{entity}/{id}/status/force-restart", // optional
+     "shutdown":       "/api/v1/{entity}/{id}/status/shutdown",       // optional
+     "force-shutdown": "/api/v1/{entity}/{id}/status/force-shutdown"  // optional
+   }
+
+The optional transition URI fields are only present when the registered
+``LifecycleProvider`` signals support for that transition. Each field value
+is the absolute path the client should call to trigger the transition.
+
+**App status (no provider registered):** derived from ``App::is_online`` in
+the entity cache. The entity cache marks an App online when it has at least
+one node in the ROS 2 graph. ``is_online = true`` maps to ``"ready"``;
+``false`` maps to ``"notReady"``.
+
+**Component status (no provider registered):** derived from
+``Component::host_metadata``. A Component populated by ``HostInfoProvider``
+carries OS, hostname, and architecture metadata. Presence of
+``host_metadata`` maps to ``"ready"``; absence maps to ``"notReady"``.
+Components created by runtime namespace grouping (``source: "synthetic"``)
+do not carry host metadata and therefore report ``"notReady"`` by default.
+
+When a ``LifecycleProvider`` is registered for the entity the provider's
+response takes precedence over the cache-derived default. The handler also
+fills absolute transition URIs from the provider-supplied optional fields.
+
+LifecycleProvider Interface
+---------------------------
+
+``LifecycleProvider`` is a per-entity plugin interface defined in
+``include/ros2_medkit_gateway/core/providers/lifecycle_provider.hpp``.
+It follows the same typed-DTO style as ``OperationProvider``.
+
+.. plantuml::
+   :caption: LifecycleProvider and handler relationships
+
+   @startuml lifecycle_provider
+
+   skinparam linetype ortho
+   skinparam classAttributeIconSize 0
+
+   package "core/providers/" {
+       class LifecycleProvider <<interface>> {
+           + get_status(entity_id): expected<LifecycleStatusResponse, LifecycleProviderErrorInfo>
+           + request_transition(entity_id, transition): expected<monostate, LifecycleProviderErrorInfo>
+       }
+   }
+
+   package "core/http/handlers/" {
+       class LifecycleHandlers {
+           + handle_get_status(req): Result<LifecycleStatusResponse>
+           + handle_transition(req, transition): Result<pair<NoContent, ResponseAttachments>>
+           - ctx_: HandlerContext
+           - plugin_mgr_: PluginManager*
+       }
+   }
+
+   package "core/plugins/" {
+       class PluginManager {
+           + get_lifecycle_provider_for_entity(entity_id): LifecycleProvider*
+       }
+   }
+
+   LifecycleHandlers --> PluginManager : queries for provider
+   LifecycleHandlers --> LifecycleProvider : calls get_status / request_transition
+   PluginManager ..> LifecycleProvider : returns registered instance
+
+   @enduml
+
+A substrate plugin (ROS 2 lifecycle node manager, process/container
+supervisor, host-level service manager) registers a ``LifecycleProvider``
+implementation via ``PluginManager``. The handler looks up the provider per
+entity at request time via ``plugin_mgr_->get_lifecycle_provider_for_entity(entity_id)``.
+
+**Fallback behavior when no provider is registered:**
+
+- ``GET /status`` returns a cache-derived ``LifecycleStatusResponse`` with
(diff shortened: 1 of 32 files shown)


src/ros2_medkit_gateway/src/plugins/plugin_manager.cpp
```diff
@@ -120,6 +121,7 @@ size_t PluginManager::load_plugins(const std::vector<PluginConfig> & configs) {
       lp.script_provider = result->script_provider;
       lp.data_provider = result->data_provider;
       lp.operation_provider = result->operation_provider;
+      lp.lifecycle_provider = dynamic_cast<LifecycleProvider *>(result->plugin.get());
```
mfaferek93: This resolves the lifecycle provider with `dynamic_cast` across the dlopen boundary, but every other provider (operation/fault/log/script/data/update) is taken from an `extern "C" get_X_provider` query in `GatewayPluginLoadResult` - precisely because RTTI/`dynamic_cast` is unreliable across a `.so` boundary (the typeinfo can differ between the plugin and the host). So a real lifecycle plugin can cast to null despite being a `LifecycleProvider`, silently never register, and lifecycle control 501s forever. Please add an `extern "C" get_lifecycle_provider` to the plugin ABI and resolve it like the others rather than via `dynamic_cast`.

src/ros2_medkit_gateway/src/http/handlers/lifecycle_handlers.cpp
```diff
@@ -0,0 +1,176 @@
...
+  const auto & cache = ctx_.node()->get_thread_safe_cache();
+  dto::LifecycleStatusResponse resp;
+
+  if (entity.type == EntityType::APP) {
+    auto app = cache.get_app(entity_id);
+    resp.status = (app && app->is_online) ? "ready" : "notReady";
+  } else {
+    // Component: ready only when host_metadata is present.
+    auto component = cache.get_component(entity_id);
+    resp.status = (component && component->host_metadata.has_value()) ? "ready" : "notReady";
```
mfaferek93: `host_metadata.has_value()` isn't a liveness signal - it's set once on the synthetic host component (`host_info_provider.cpp:101`) and never cleared, and Component has no online field. So every node-derived component reports `notReady` even when alive, and the host component can never go `notReady`, whereas the App path above correctly uses `is_online`. That contradicts the PR body ("ready while reachable"). Drive Component readiness from a real liveness signal (the bound app's `is_online` / node-graph reachability), or document this as a deliberate contract-only stub and fix the PR description and design doc.

src/ros2_medkit_gateway/src/http/handlers/lifecycle_handlers.cpp
```diff
@@ -0,0 +1,176 @@
...
+  std::string msg = e.message.size() > kMaxMsg ? e.message.substr(0, kMaxMsg) + "..." : e.message;
+
+  switch (e.code) {
+    case LifecycleProviderError::AccessDenied:
+      return make_error(403, ERR_INSUFFICIENT_ACCESS_RIGHTS, std::move(msg));
+    case LifecycleProviderError::PreconditionFailed:
+      return make_error(409, ERR_PRECONDITION_NOT_FULFILLED, std::move(msg));
+    case LifecycleProviderError::Unsupported:
+      return make_error(501, ERR_NOT_IMPLEMENTED, std::move(msg));
+    case LifecycleProviderError::EntityNotFound:
```
mfaferek93, line 49: `EntityNotFound` is bucketed here with `TransportError`/`Internal` and mapped to `ERR_PLUGIN_ERROR` with the provider's status (default 500), so a provider that legitimately reports a missing entity surfaces as a 500 plugin-error rather than a 404 - even though `ERR_ENTITY_NOT_FOUND` exists. Give it its own case: `case EntityNotFound: return make_error(404, ERR_ENTITY_NOT_FOUND, std::move(msg));`.

src/ros2_medkit_gateway/src/http/handlers/lifecycle_handlers.cpp
```diff
@@ -0,0 +1,176 @@
...
+  // Delegate to LifecycleProvider when one is registered for this entity.
+  if (plugin_mgr_) {
+    auto * provider = plugin_mgr_->get_lifecycle_provider_for_entity(entity_id);
+    if (provider) {
+      try {
+        auto result = provider->get_status(entity_id);
+        if (!result) {
+          return tl::make_unexpected(to_error_info(result.error()));
+        }
+        auto resp = *result;
```
mfaferek93, line 89: The provider's `status` is returned verbatim here, but the JSON writer only checks the enum on parse, not on write - so a provider returning `running` or `""` goes straight onto the wire and breaks the SOVD `ready`/`notReady` contract. Validate the provider-supplied status against `{ready, notReady}` (map or reject) before emitting.

src/ros2_medkit_gateway/src/core/auth/auth_config.cpp
```diff
@@ -365,6 +371,9 @@ const std::unordered_map<UserRole, std::unordered_set<std::string>> & AuthConfig
            "PUT:/api/v1/apps/*/scripts/*/executions/*",
            "DELETE:/api/v1/components/*/scripts/*/executions/*",
            "DELETE:/api/v1/apps/*/scripts/*/executions/*",
+           // Lifecycle control: apps and components (PUT)
+           "PUT:/api/v1/apps/*/status/*",
```
mfaferek93: This single `PUT:/.../status/*` rule lets the lowest write role (OPERATOR) invoke `shutdown` and `force-shutdown`, which tear an entity down, with no separation from the benign `start`/`restart` and no extra floor for the more aggressive `force-` variants. Worth a deliberate call: gate the destructive transitions behind CONFIGURATOR, or at least note in the rule comment that OPERATOR-level teardown is intended - the wildcard hides a fairly broad grant.

src/ros2_medkit_integration_tests/test/features/test_lifecycle.test.py
```diff
@@ -0,0 +1,136 @@
...
+        if TestLifecycleStatus._host_component_id is not None:
+            return TestLifecycleStatus._host_component_id
+
+        data = self.get_json('/components')
+        components = data.get('items', [])
+        self.assertGreater(
+            len(components), 0,
+            'Expected at least 1 host component, got none'
+        )
+        TestLifecycleStatus._host_component_id = components[0]['id']
```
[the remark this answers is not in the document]
bburda: Fixed. The test now asserts exactly one host Component (`assertEqual(len, 1)`), matching the runtime model and the other feature tests.

src/ros2_medkit_gateway/design/lifecycle.rst
```diff
@@ -0,0 +1,295 @@
...
+
+.. code-block:: cpp
+
+   for (const auto & et_lc :
+        std::vector<std::pair<const char *, const char *>>{
+            {"apps", "app"}, {"components", "component"}}) {
+     const std::string base_lc = "/" + et_lc.first + "/{" + et_lc.second + "_id}";
+
+     // PUT routes registered BEFORE GET /status to avoid the shorter path
+     // shadowing the more-specific fixed segments.
```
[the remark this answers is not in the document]
bburda: Fixed. The inline comment now states that GET and PUT are different HTTP methods, so neither can shadow the other and registration order is arbitrary.

src/ros2_medkit_gateway/design/lifecycle.rst
```diff
@@ -0,0 +1,295 @@
...
+          {"start", "restart", "force-restart", "shutdown", "force-shutdown"}) {
+       reg.put<http::NoContent>(base_lc + "/status/" + action, ...);
+     }
+     reg.get<dto::LifecycleStatusResponse>(base_lc + "/status", ...);
+   }
+
+The PUT routes must be registered before the GET route. The typed
+``RouteRegistry`` matches fixed path segments (``/status/start``) before
+parameterized ones (``/status``), but registration order also affects the
+matching priority inside cpp-httplib's internal routing table.
```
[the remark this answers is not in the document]
bburda: Fixed. Rewrote the paragraph: GET and PUT are different methods and cpp-httplib keeps a separate handler list per method, so registration order does not matter.

src/ros2_medkit_gateway/src/http/rest_server.cpp
```diff
@@ -1592,6 +1593,48 @@ void RESTServer::setup_routes() {
       .operation_id("revokeToken")
       .error_renderer(openapi::ErrorRenderer::kOAuth2Error);
 
+  // === Lifecycle (status) - apps and components only, outside the 4-type loop ===
+  // GET and PUT use different HTTP methods so neither can shadow the other.
+  // Registration order within the loop is arbitrary from the router's perspective.
+  for (const auto & et_lc :
+       std::vector<std::pair<const char *, const char *>>{{"apps", "app"}, {"components", "component"}}) {
+    const std::string base_lc = std::string("/") + et_lc.first + "/{" + et_lc.second + "_id}";
```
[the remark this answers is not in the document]
bburda: Fixed. Added a Lifecycle Endpoints section to `docs/api/rest.rst` covering the status read and the five PUT transition routes with their response and error shapes.

src/ros2_medkit_gateway/src/plugins/plugin_manager.cpp
bburda: Fixed. Added `extern "C" get_lifecycle_provider` to the plugin ABI; the loader resolves it via `dlsym` into `GatewayPluginLoadResult` and the manager now takes the provider from the load result, like the other providers, instead of `dynamic_cast` across the dlopen boundary. The in-process `add_plugin` path keeps `dynamic_cast` since RTTI is reliable within the same binary.

src/ros2_medkit_gateway/src/http/handlers/lifecycle_handlers.cpp
bburda: Fixed. Component readiness now uses a real liveness signal: the host Component is ready while reachable, and any other Component is ready when at least one of its hosted Apps is online. host_metadata is only used to identify the host Component. The design doc was updated to match.

src/ros2_medkit_gateway/src/http/handlers/lifecycle_handlers.cpp
bburda: Fixed. EntityNotFound now has its own case mapping to 404 entity-not-found instead of being bucketed with TransportError and Internal.

src/ros2_medkit_gateway/src/http/handlers/lifecycle_handlers.cpp
bburda: Fixed. The handler now validates the provider-supplied status against {ready, notReady} before writing it and returns 500 if it is neither.

src/ros2_medkit_gateway/src/core/auth/auth_config.cpp
bburda: Fixed. Replaced the single `status/*` wildcard with explicit per-transition rules. Operator gets start, restart and force-restart; the destructive shutdown and force-shutdown now require the configurator role.

mfaferek93 (approved):

Issue #427, opened 2026-06-15
feat(gateway): startup discovery summary + empty-graph diagnostics + self-entity


mfaferek93:
The gateway boots cleanly and serves an empty entity tree when it cannot see the target ROS graph (wrong `ROS_DOMAIN_ID`, mismatched `RMW_IMPLEMENTATION`, container network isolation, `ROS_LOCALHOST_ONLY`, blocked multicast). Today there is no startup summary, no empty-graph warning, no environment echo, and no gateway self-entity (`gateway_node.cpp` logs its config but never the discovered entity count). The result is a silent dead end: the user sees a blank tree with no hint where to look.

## Scope
- Startup discovery summary: after the first discovery cycle, log the number of discovered nodes / topics / medkit entities, the gateway URL, and a sample curl.
- Empty-graph diagnostics, surfaced to the USER (not console-only): when zero application entities are discovered, report the likely causes and echo the active environment (`ROS_DOMAIN_ID`, `RMW_IMPLEMENTATION`, `ROS_LOCALHOST_ONLY`) somewhere the UI/API can show it (for example on `/health` or as a gateway self-entity attribute), so a user staring at an empty tree sees "0 application nodes visible - domain X, rmw Y, localhost-only Z - check that the gateway shares the robot's domain / RMW / network" and knows where to look. Also log it loudly on the console.
- Gateway self-entity: the gateway always registers itself as an entity, so the tree is never literally blank and the "0 application entities" state is explicit rather than empty.
- Document the co-location requirement and multi-domain topology in the bringup quickstart (#424): the gateway must share `ROS_DOMAIN_ID` + `RMW_IMPLEMENTATION` + network with the target stack; multiple domains run one gateway per domain federated by peer aggregation.

## Acceptance
- On startup the gateway logs discovered node / topic / entity counts plus the URL and a sample curl.
- With no reachable application nodes, the gateway both logs a warning and exposes the same diagnostic via the UI/API (not console-only), including the active `ROS_DOMAIN_ID` / `RMW_IMPLEMENTATION` / `ROS_LOCALHOST_ONLY` and the likely causes.
- The gateway self-entity is always present; the entity tree is never empty even before any application node is discovered.

PR #438, opened 2026-06-18
feat(gateway): startup discovery summary + empty-graph warning

mfaferek93:
The gateway boots into a silent empty entity tree when it cannot see the target ROS graph (wrong `ROS_DOMAIN_ID`, mismatched `RMW_IMPLEMENTATION`, `ROS_LOCALHOST_ONLY`, container network isolation): the user gets a blank tree with no hint where to look.

This logs a one-time discovery summary a few seconds after startup (the delay lets DDS discovery settle): peer node / topic / medkit entity counts, the REST URL and a sample curl. When no application nodes are visible it additionally warns loudly with the active `ROS_DOMAIN_ID`, `RMW_IMPLEMENTATION` and `ROS_LOCALHOST_ONLY` and the likely causes, so a misconfigured bringup is diagnosable from the gateway's own console/logs.

The gateway's own nodes (the main node plus its internal helpers, which share the main node's FQN as a prefix, e.g. `<fqn>_sub`) and hidden nodes are excluded from the peer count, so "0 peer nodes" reliably means the gateway is isolated.

Verified at runtime: an empty domain logs the summary with 0 peers and the env-echo warning; a domain with one peer node logs 1 peer and no warning.

Scope: this is the startup-log diagnostic from #427. The `/health` diagnostic surface and the synthetic gateway self-entity are intentionally left out - the current web UI renders neither entity attributes nor `/health`, so they would not reach the user without separate UI work; a self-fault was considered but deferred to avoid empty-graph false positives at cold start.

Addresses #427.

### DIFF
--- a/src/ros2_medkit_gateway/src/gateway_node.cpp
+++ b/src/ros2_medkit_gateway/src/gateway_node.cpp
@@ -18,8 +18,10 @@
 #include <cctype>
 #include <chrono>
 #include <cinttypes>
+#include <cstdlib>
 #include <mutex>
 #include <set>
+#include <string>
 #include <string_view>
 #include <unordered_map>
 #include <unordered_set>
@@ -1484,6 +1486,63 @@ GatewayNode::GatewayNode(const rclcpp::NodeOptions & options) : Node("ros2_medki
   std::string protocol = tls_config_.enabled ? "HTTPS" : "HTTP";
   RCLCPP_INFO(get_logger(), "ROS 2 Medkit Gateway ready on %s://%s:%d", protocol.c_str(), server_host_.c_str(),
               server_port_);
+
+  // One-shot discovery summary a few seconds after startup. The delay lets DDS
+  // discovery settle so the counts (and any empty-graph warning) reflect a
+  // converged view rather than the cold-start moment.
+  startup_summary_timer_ = create_wall_timer(std::chrono::seconds(5), [this]() {
+    startup_summary_timer_->cancel();
+    startup_summary_timer_.reset();
+    log_startup_summary();
+  });
+}
+
+void GatewayNode::log_startup_summary() {
+  size_t topic_count = 0;
+  size_t peer_node_count = 0;
+  try {
+    const std::string self_fqn = get_fully_qualified_name();
+    for (const auto & entry : get_topic_names_and_types()) {
+      const std::string & topic = entry.first;
+      if (topic.find("/parameter_events") == std::string::npos && topic.find("/rosout") == std::string::npos) {
+        ++topic_count;
+      }
+    }
+    for (const auto & name : get_node_names()) {
+      // Count peers only: exclude the gateway's own nodes (the main node plus its
+      // internal helpers, which share the main node's FQN as a prefix, e.g.
+      // "<fqn>_sub") and hidden (_-prefixed) nodes.
+      if (name.rfind(self_fqn, 0) != 0 && name.rfind("/_", 0) != 0) {
+        ++peer_node_count;
+      }
+    }
+  } catch (const std::exception & e) {
+    RCLCPP_DEBUG(get_logger(), "Startup summary: graph query failed: %s", e.what());
+  }
+
+  const size_t entity_count = thread_safe_cache_.get_areas().size() + thread_safe_cache_.get_components().size() +
+                              thread_safe_cache_.get_apps().size() + thread_safe_cache_.get_functions().size();
+
+  const std::string protocol = tls_config_.enabled ? "https" : "http";
+  const std::string url = protocol + "://" + server_host_ + ":" + std::to_string(server_port_);
+
+  RCLCPP_INFO(get_logger(),
+              "Discovery summary: %zu peer node(s), %zu topic(s), %zu medkit entit%s. REST API: %s  Try: curl "
+              "%s/api/v1/areas",
+              peer_node_count, topic_count, entity_count, entity_count == 1 ? "y" : "ies", url.c_str(), url.c_str());
+
+  if (peer_node_count == 0) {
+    const char * domain = std::getenv("ROS_DOMAIN_ID");
+    const char * rmw = std::getenv("RMW_IMPLEMENTATION");
+    const char * localhost_only = std::getenv("ROS_LOCALHOST_ONLY");
+    RCLCPP_WARN(get_logger(),
+                "No application nodes are visible - the gateway sees only itself, so the entity tree is empty. "
+                "Active ROS environment: ROS_DOMAIN_ID=%s, RMW_IMPLEMENTATION=%s, ROS_LOCALHOST_ONLY=%s. Check that "
+                "the gateway shares the robot stack's ROS_DOMAIN_ID and RMW_IMPLEMENTATION and can reach it on the "
+                "network (in containers: matching domain/RMW and DDS discovery not blocked).",
+                domain ? domain : "0 (default)", rmw ? rmw : "(default)",
+                localhost_only ? localhost_only : "0 (default)");
+  }
 }
 
 GatewayNode::~GatewayNode() {
@@ -1554,6 +1613,10 @@ GatewayNode::~GatewayNode() {
   if (backstop_timer_) {
     backstop_timer_->cancel();
     backstop_timer_.reset();
+  }
+  if (startup_summary_timer_) {
+    startup_summary_timer_->cancel();
+    startup_summary_timer_.reset();
   }
   graph_event_.reset();
   // 10. Normal member destruction (managers safe - all transports stopped)
--- a/src/ros2_medkit_gateway/src/gateway_node.cpp
+++ b/src/ros2_medkit_gateway/src/gateway_node.cpp
@@ -1497,25 +1497,48 @@ GatewayNode::GatewayNode(const rclcpp::NodeOptions & options) : Node("ros2_medki
   });
 }
 
+size_t GatewayNode::count_peer_nodes(const std::vector<std::pair<std::string, std::string>> & nodes_and_namespaces,
+                                     const std::string & self_fqn) {
+  size_t count = 0;
+  for (const auto & [name, ns] : nodes_and_namespaces) {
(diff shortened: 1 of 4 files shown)


src/ros2_medkit_gateway/src/gateway_node.cpp
```diff
@@ -1484,6 +1486,63 @@ GatewayNode::GatewayNode(const rclcpp::NodeOptions & options) : Node("ros2_medki
...
+      }
+    }
+    for (const auto & name : get_node_names()) {
+      // Count peers only: exclude the gateway's own nodes (the main node plus its
+      // internal helpers, which share the main node's FQN as a prefix, e.g.
+      // "<fqn>_sub") and hidden (_-prefixed) nodes.
+      if (name.rfind(self_fqn, 0) != 0 && name.rfind("/_", 0) != 0) {
+        ++peer_node_count;
+      }
+    }
```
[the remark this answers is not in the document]
mfaferek93: Now resolves FQNs from get_node_names_and_namespaces() and excludes hidden nodes by their name part (the gateway's own nodes share its FQN as a prefix).

src/ros2_medkit_gateway/src/gateway_node.cpp
```diff
@@ -1484,6 +1486,63 @@ GatewayNode::GatewayNode(const rclcpp::NodeOptions & options) : Node("ros2_medki
...
+  } catch (const std::exception & e) {
+    RCLCPP_DEBUG(get_logger(), "Startup summary: graph query failed: %s", e.what());
+  }
+
+  const size_t entity_count = thread_safe_cache_.get_areas().size() + thread_safe_cache_.get_components().size() +
+                              thread_safe_cache_.get_apps().size() + thread_safe_cache_.get_functions().size();
+
+  const std::string protocol = tls_config_.enabled ? "https" : "http";
+  const std::string url = protocol + "://" + server_host_ + ":" + std::to_string(server_port_);
+
```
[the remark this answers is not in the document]
mfaferek93: Bind-all hosts (0.0.0.0 / :: / empty) are now translated to loopback for the sample curl via connectable_host().

src/ros2_medkit_gateway/src/gateway_node.cpp
```diff
@@ -1484,6 +1486,63 @@ GatewayNode::GatewayNode(const rclcpp::NodeOptions & options) : Node("ros2_medki
...
+    const char * rmw = std::getenv("RMW_IMPLEMENTATION");
+    const char * localhost_only = std::getenv("ROS_LOCALHOST_ONLY");
+    RCLCPP_WARN(get_logger(),
+                "No application nodes are visible - the gateway sees only itself, so the entity tree is empty. "
+                "Active ROS environment: ROS_DOMAIN_ID=%s, RMW_IMPLEMENTATION=%s, ROS_LOCALHOST_ONLY=%s. Check that "
+                "the gateway shares the robot stack's ROS_DOMAIN_ID and RMW_IMPLEMENTATION and can reach it on the "
+                "network (in containers: matching domain/RMW and DDS discovery not blocked).",
+                domain ? domain : "0 (default)", rmw ? rmw : "(default)",
+                localhost_only ? localhost_only : "0 (default)");
+  }
```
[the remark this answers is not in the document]
mfaferek93: Reworded to 'no application nodes visible on the ROS graph - the gateway sees only its own nodes', dropping the entity-tree-empty claim; the gate stays on the runtime peer count.

src/ros2_medkit_gateway/src/gateway_node.cpp
```diff
@@ -1484,6 +1486,63 @@ GatewayNode::GatewayNode(const rclcpp::NodeOptions & options) : Node("ros2_medki
...
+  // converged view rather than the cold-start moment.
+  startup_summary_timer_ = create_wall_timer(std::chrono::seconds(5), [this]() {
+    startup_summary_timer_->cancel();
+    startup_summary_timer_.reset();
+    log_startup_summary();
+  });
+}
+
+void GatewayNode::log_startup_summary() {
+  size_t topic_count = 0;
```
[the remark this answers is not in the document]
mfaferek93: Bind-all hosts (0.0.0.0 / :: / empty) are now translated to loopback for the sample curl via connectable_host().

src/ros2_medkit_gateway/src/gateway_node.cpp
```diff
@@ -1484,6 +1486,86 @@ GatewayNode::GatewayNode(const rclcpp::NodeOptions & options) : Node("ros2_medki
...
+        count_peer_nodes(get_node_graph_interface()->get_node_names_and_namespaces(), get_fully_qualified_name());
+  } catch (const std::exception & e) {
+    RCLCPP_DEBUG(get_logger(), "Startup summary: graph query failed: %s", e.what());
+  }
+
+  const size_t entity_count = thread_safe_cache_.get_areas().size() + thread_safe_cache_.get_components().size() +
+                              thread_safe_cache_.get_apps().size() + thread_safe_cache_.get_functions().size();
+
+  const std::string protocol = tls_config_.enabled ? "https" : "http";
+  const std::string url = protocol + "://" + connectable_host(server_host_) + ":" + std::to_string(server_port_);
```
bburda, line 1555: `connectable_host` returns IPv6 literals without brackets, so for an IPv6 bind this builds `http://::1:8080/api/v1/areas` (or `http://2001:db8::1:8080/...` for an explicit address). curl does not accept an unbracketed IPv6 host with a port; it needs `http://[::1]:8080/...`. Bracket the host when it contains `:`. The `[::]` input is also not covered by the test.

src/ros2_medkit_gateway/src/gateway_node.cpp
```diff
@@ -1484,6 +1486,86 @@ GatewayNode::GatewayNode(const rclcpp::NodeOptions & options) : Node("ros2_medki
...
+                                     const std::string & self_fqn) {
+  size_t count = 0;
+  for (const auto & [name, ns] : nodes_and_namespaces) {
+    if (!name.empty() && name.front() == '_') {
+      continue;  // hidden node (name part starts with '_')
+    }
+    const std::string fqn = (ns == "/") ? ("/" + name) : (ns + "/" + name);
+    // Exclude the gateway's own nodes: the main node plus its internal helpers,
+    // which share the main node's FQN as a prefix (e.g. "<fqn>_sub").
+    if (fqn.rfind(self_fqn, 0) == 0) {
```
bburda: `rfind(self_fqn, 0) == 0` is a plain prefix match, so a separate peer node whose name starts with the gateway name (for example `ros2_medkit_gateway_monitor`, or a second `ros2_medkit_gateway2`) is also dropped from the count, and an otherwise non-empty graph can end up on the 0-peer warning path. The internal helpers are `<fqn>_sub` / `<fqn>_fault_clients` (suffix after `_`, not a `/` sub-namespace), so a `self_fqn + "/"` boundary check would stop excluding them. Matching the exact FQN plus the known helper names is more reliable than a prefix match. The unit test only feeds own helpers; a case with a prefix-sharing peer that should still count would cover this.

src/ros2_medkit_gateway/src/gateway_node.cpp
```diff
@@ -1484,6 +1486,86 @@ GatewayNode::GatewayNode(const rclcpp::NodeOptions & options) : Node("ros2_medki
...
+    const char * domain = std::getenv("ROS_DOMAIN_ID");
+    const char * rmw = std::getenv("RMW_IMPLEMENTATION");
+    const char * localhost_only = std::getenv("ROS_LOCALHOST_ONLY");
+    RCLCPP_WARN(get_logger(),
+                "No application nodes are visible on the ROS graph - the gateway sees only its own nodes. "
+                "Active ROS environment: ROS_DOMAIN_ID=%s, RMW_IMPLEMENTATION=%s, ROS_LOCALHOST_ONLY=%s. Check that "
+                "the gateway shares the robot stack's ROS_DOMAIN_ID and RMW_IMPLEMENTATION and can reach it on the "
+                "network (in containers: matching domain/RMW and DDS discovery not blocked).",
+                domain ? domain : "0 (default)", rmw ? rmw : "(default)",
+                localhost_only ? localhost_only : "0 (default)");
```
bburda: When `ROS_LOCALHOST_ONLY` is unset this prints `0 (default)`, which reads as "localhost-only is off" and points the reader away from the network as the cause. `(unset)` is more accurate. The `0 (default)` for `ROS_DOMAIN_ID` is fine, since unset really means domain 0.

bburda (commented):
Outside the diff: `docs/troubleshooting.rst` has a "Gateway shows no areas or components" entry that lists the same causes this summary now logs at startup. One line there pointing to the new "Discovery summary" line and the empty-graph warning would help operators find it.

src/ros2_medkit_gateway/src/gateway_node.cpp
mfaferek93: connectable_host() now brackets IPv6 literals ([::1], [2001:db8::1]) and the test covers '::' / '[::]' / a bare IPv6 address.

src/ros2_medkit_gateway/src/gateway_node.cpp
mfaferek93: Now excludes by exact FQN plus the known helper names (_sub / _fault_clients), so a prefix-sharing peer like <fqn>_monitor still counts; added a test for it.

src/ros2_medkit_gateway/src/gateway_node.cpp
mfaferek93: Unset ROS_LOCALHOST_ONLY now prints (unset) instead of '0 (default)'; ROS_DOMAIN_ID keeps '0 (default)' since unset really is domain 0.

mfaferek93:
Good call - added a pointer in docs/troubleshooting.rst ('Gateway shows no areas or components') to the new Discovery summary line and the empty-graph warning.

bburda (approved):

PR #451, opened 2026-06-20
build(image): bundle CycloneDDS RMW as opt-in

### LINKED ISSUE #447: Prebuilt Docker image ships only FastDDS; bundle CycloneDDS RMW as opt-in
**Scope: prebuilt Docker image only.** Installing medkit from rosdistro (apt) onto an existing ROS 2 install is unaffected - no medkit package pins an RMW (verified: no `rmw_*` dep in any `package.xml`), so it uses whatever RMW the host already runs (a robot on CycloneDDS already has `rmw_cyclonedds_cpp`).

The prebuilt image is built on `ros-base`, which ships FastDDS only, so a containerized gateway attached to an external stack running CycloneDDS (Autoware and others) does not load at all.

**Fix**
Bundle `rmw_cyclonedds_cpp` in the image as an opt-in. FastDDS stays the default; switch with `RMW_IMPLEMENTATION=rmw_cyclonedds_cpp`. Image delta ~8 MB; Cyclone verified to load on a stock host (`net.core.rmem_max` 208 KB), so no tuned config is required for basic operation.

mfaferek93:
The prebuilt Docker image is built on `ros-base`, which ships FastDDS only. A containerized gateway attached to an external stack running CycloneDDS (Autoware and others) does not load at all. Note this is image-only: apt installs from rosdistro use the host's RMW and are unaffected (no medkit `package.xml` pins an RMW).

This bundles `rmw_cyclonedds_cpp` in the runtime stage. FastDDS stays the default; switch with `RMW_IMPLEMENTATION=rmw_cyclonedds_cpp`.

Verified: the package installs on `ros:jazzy-ros-base` (~8 MB delta incl. iceoryx) and Cyclone loads + publishes on a stock host (`net.core.rmem_max` 208 KB) with no tuned config. Full official image not rebuilt (single additive runtime apt line).

Closes #447

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -97,10 +97,14 @@ ENV ROS_DISTRO=${ROS_DISTRO}
 ENV COLCON_WS=/home/medkit/ws
 
 # Runtime dependencies only (nlohmann-json and cpp-httplib are compiled into
-# the binaries at build time, no need to install here)
+# the binaries at build time, no need to install here).
+# CycloneDDS RMW is bundled so the gateway can attach to a stack running
+# CycloneDDS (e.g. Autoware) without rebuilding. FastDDS stays the default;
+# switch with RMW_IMPLEMENTATION=rmw_cyclonedds_cpp at runtime.
 RUN apt-get update && apt-get install -y --no-install-recommends \
     ros-${ROS_DISTRO}-yaml-cpp-vendor \
     ros-${ROS_DISTRO}-example-interfaces \
+    ros-${ROS_DISTRO}-rmw-cyclonedds-cpp \
     libsqlite3-0 \
     libsystemd0 \
     libssl3 \


bburda (dismissed):
Verified. The new package is in the same RUN block that cleans apt lists, ros-<distro>-rmw-cyclonedds-cpp exists on humble/jazzy/lyrical, and the entrypoint keeps FastDDS as the default RMW (overridable), so this is additive and opt-in as described.

bburda (approved):
Verified the change across apt/layer mechanics, package availability, RMW default behavior, image size/security and docs. The new package sits in the same RUN that cleans apt lists, ROS_DISTRO is re-declared after FROM so it resolves in the runtime stage, ros-<distro>-rmw-cyclonedds-cpp exists on humble/jazzy/lyrical, the entrypoint keeps FastDDS as the default and honors an external RMW_IMPLEMENTATION override, and no doc claims the image is FastDDS-only. Additive, opt-in, correct.

PR #455, opened 2026-06-21
feat(gateway): accurate status read for apps and components

### LINKED ISSUE #454: Accurate status read for apps and components
## Summary

`GET /{entity}/status` reports `ready` or `notReady` from signals that are too coarse, so the
status does not reflect reality.

- App: `ready` comes from `App::is_online`, which is set to true whenever the node name appears in
  the ROS 2 graph (`get_node_names_and_namespaces`). It is unconditional on graph presence. A
  managed lifecycle node that is in the graph but `inactive` or `unconfigured` still reports
  `ready`, even though it does no work.
- Component: a non-host component is `ready` only when at least one hosted app is online, so a
  component with zero apps, or with all apps temporarily down, reports `notReady` even though the
  substrate is reachable. The SOVD spec defines `notReady` as "stopped, restarting, or
  unreachable", which a reachable substrate is not.

This issue makes the read accurate. It adds no control: the PUT transitions stay `501` and
actuation is tracked separately.

## Proposed solution

App status:

- For a managed lifecycle node (one that exposes `get_state` and `change_state`), derive the
  status from the lifecycle state read with `GetState`: `active` is `ready`, any other state or an
  unreachable service is `notReady`. This also covers liveness, because a dead node's service is
  gone and reads `notReady`.
- For a plain node with no lifecycle services, keep `is_online` (graph presence). That is the best
  signal at the ROS level for a plain node; application-level health is out of scope here.

Component status:

- A local component is `ready` while the gateway is serving the request, independent of how many
  hosted apps are online, including zero. Remove the "at least one hosted app online" rule. A down
  app is that app's own `notReady`, not the component's.
- Remote components keep current behavior: the status request is forwarded to the peer gateway.

Improves the fidelity of `GET /{entity}/status` (REQ_INTEROP_076). Read-only. Actuation (lifecycle
transitions, process, container, and host restart) is tracked separately.

Unit and integration tests, docs.

bburda:
## Summary

Make `GET /{apps|components}/{id}/status` report `ready` / `notReady` from real signals instead of
coarse ones. Read-only; the `PUT` lifecycle transitions stay `501` (actuation is tracked separately).

- App status: for a managed lifecycle node (one exposing `lifecycle_msgs` `get_state` and
  `change_state`), the status is read from the actual lifecycle state via a `GetState` service call
  (`active` is `ready`; any other state, or an unreachable/timed-out read, is `notReady`). This
  subsumes liveness, since a dead node's service is gone. Plain (unmanaged) nodes keep using
  `App::is_online` (presence in the ROS 2 graph), which is the best signal available for an
  unmanaged node.
- Component status: the synthetic host component (carrying `host_metadata`) is `ready` while the
  gateway is serving the request. Any other local component derives readiness from its hosted apps:
  it is `notReady` when it hosts apps and every one of them is offline (the subsystem is down), and
  `ready` otherwise, including when it hosts no apps. A single down hosted app is that app's own
  `notReady`, not enough to mark the component down. Remote components are unchanged (reached
  through aggregation forwarding).

The lifecycle read goes through a small injected `LifecycleStateReader` seam: a pure interface in
`gateway_core` (keeps the handler unit-testable with a stub and preserves core purity) plus a
`gateway_ros2` implementation that calls `GetState` on a private node and private single-threaded
executor spun inline (mirrors the existing `ros2_fault_service_transport` pattern, so it never
blocks or races the gateway executor). A `managed_lifecycle` `rclcpp_lifecycle` demo node and an
integration test prove the path end to end.

The third `LifecycleHandlers` constructor parameter is defaulted, so existing call sites and
non-lifecycle apps are unchanged. No API break.

## Issue

- closes #454

## Type

- [x] New feature or tests

## Testing

Unit (GTest):

- Pure helpers: `active` to `ready`, other state or `nullopt` to `notReady`; the lifecycle-service
  detector requires both `get_state` and `change_state`.
- `Ros2LifecycleStateReader` against an in-process `GetState` service: reads `active` / `inactive`,
  returns `nullopt` for an unreachable service.
- Handler: a managed node whose state is `inactive` reports `notReady` even though it is online
  (the discriminator); an active managed node reports `ready`; a plain node still uses `is_online`.
  Component reads return `ready` for the host component and a zero-app component, and `notReady` for
  a component that hosts apps which are all offline.

Integration (launch_testing):

- A `managed_lifecycle` node is launched and left unconfigured. It is present in the ROS 2 graph
  (so `is_online` alone would say `ready`), but `GET /apps/managed_lifecycle/status` returns
  `notReady`, proving the status is read from the lifecycle state via `GetState`. Tagged
  `@verifies REQ_INTEROP_076`.

clang-tidy is clean on the changed files.

## Checklist

- [x] Breaking changes are clearly described (none - additive; the new constructor parameter is defaulted)
- [x] Docs were updated if behavior or public API changed (design doc + README)

### DIFF
--- a/src/ros2_medkit_gateway/CMakeLists.txt
+++ b/src/ros2_medkit_gateway/CMakeLists.txt
@@ -41,6 +41,7 @@ find_package(yaml_cpp_vendor REQUIRED)
 medkit_find_yaml_cpp()
 find_package(ament_index_cpp REQUIRED)
 find_package(action_msgs REQUIRED)
+find_package(lifecycle_msgs REQUIRED)
 find_package(ros2_medkit_msgs REQUIRED)
 find_package(ros2_medkit_serialization REQUIRED)
 find_package(SQLite3 REQUIRED)
@@ -159,6 +160,7 @@ add_library(gateway_ros2 STATIC
   src/ros2/transports/ros2_service_transport.cpp
   src/ros2/transports/ros2_topic_subscription_transport.cpp
   src/ros2/transports/ros2_topic_transport.cpp
+  src/ros2/status/ros2_lifecycle_state_reader.cpp
   src/ros2/trigger_topic_subscriber.cpp
   src/default_script_provider.cpp
   src/discovery/discovery_manager.cpp
@@ -224,6 +226,7 @@ medkit_target_dependencies(gateway_ros2
   rcl_interfaces
   ament_index_cpp
   action_msgs
+  lifecycle_msgs
   ros2_medkit_msgs
   ros2_medkit_serialization
   rosidl_typesupport_cpp
@@ -773,6 +776,12 @@ if(BUILD_TESTING)
   target_link_libraries(test_lifecycle_status_helpers gateway_core)
   medkit_set_test_domain(test_lifecycle_status_helpers)
 
+  # GetState-backed lifecycle state reader (ROS 2 in-process service)
+  ament_add_gtest(test_ros2_lifecycle_state_reader test/test_ros2_lifecycle_state_reader.cpp)
+  target_link_libraries(test_ros2_lifecycle_state_reader gateway_ros2)
+  medkit_target_dependencies(test_ros2_lifecycle_state_reader rclcpp lifecycle_msgs)
+  medkit_set_test_domain(test_ros2_lifecycle_state_reader)
+
   # Add operation handler tests
   ament_add_gtest(test_operation_handlers test/test_operation_handlers.cpp)
   target_link_libraries(test_operation_handlers gateway_ros2)
@@ -978,6 +987,7 @@ if(BUILD_TESTING)
       test_tls_config
       test_fault_manager
       test_error_info
+      test_ros2_lifecycle_state_reader
       test_ros2_subscription_executor
       test_ros2_subscription_slot
       test_topic_data_provider_interface
--- a/src/ros2_medkit_gateway/CMakeLists.txt
+++ b/src/ros2_medkit_gateway/CMakeLists.txt
@@ -987,6 +987,7 @@ if(BUILD_TESTING)
       test_tls_config
       test_fault_manager
       test_error_info
+      test_lifecycle_status_helpers
       test_ros2_lifecycle_state_reader
       test_ros2_subscription_executor
       test_ros2_subscription_slot
--- a/src/ros2_medkit_gateway/CMakeLists.txt
+++ b/src/ros2_medkit_gateway/CMakeLists.txt
@@ -768,6 +768,11 @@ if(BUILD_TESTING)
   medkit_target_dependencies(test_lifecycle_handlers rclcpp)
   medkit_set_test_domain(test_lifecycle_handlers)
 
+  # Lifecycle status helpers (pure C++17, no ROS node)
+  ament_add_gtest(test_lifecycle_status_helpers test/test_lifecycle_status_helpers.cpp)
+  target_link_libraries(test_lifecycle_status_helpers gateway_core)
+  medkit_set_test_domain(test_lifecycle_status_helpers)
+
   # Add operation handler tests
   ament_add_gtest(test_operation_handlers test/test_operation_handlers.cpp)
   target_link_libraries(test_operation_handlers gateway_ros2)
--- a/src/ros2_medkit_gateway/design/lifecycle.rst
+++ b/src/ros2_medkit_gateway/design/lifecycle.rst
@@ -99,11 +99,15 @@ never blocks or races the gateway executor; it is, however, serialized by an int
 reachable-but-slow managed node holds that mutex across its spin and delays other concurrent
 ``/status`` reads for up to the (short) read timeout.
 
-**Component status:** a local component is ``"ready"`` while the gateway is serving the
-request (the substrate is reachable), independent of how many hosted apps are online,
-including zero. SOVD ``notReady`` means stopped, restarting, or unreachable, which a
-reachable local substrate is not; a down hosted app is that app's own ``notReady``. Remote
-components are reached through aggregation forwarding, so peer reachability drives their status.
+**Component status:** the synthetic host component (the one carrying ``host_metadata``,
+populated by ``HostInfoProvider``) is ``"ready"`` while the gateway is serving the request -
+its substrate is the gateway host, which is reachable by definition. Any other local
+component derives readiness from its hosted Apps: a component that hosts Apps is
+``"notReady"`` when every one of them is offline (its subsystem is down), and ``"ready"`` when
+at least one is online. A component with no hosted Apps is a pure grouping and stays
+``"ready"`` (the local substrate is reachable; a single down App is that App's own
+``"notReady"``, not enough to mark the component down). Remote components are reached through
+aggregation forwarding, so peer reachability drives their status.
 
 When a ``LifecycleProvider`` is registered for the entity the provider's
 response takes precedence over the cache-derived default. The handler also
--- a/src/ros2_medkit_gateway/design/lifecycle.rst
+++ b/src/ros2_medkit_gateway/design/lifecycle.rst
@@ -91,8 +91,13 @@ the status is read from the node's lifecycle state via ``lifecycle_msgs/srv/GetS
 ``active`` state is ``"ready"``; any other state, or an unreachable/timed-out read, is
 ``"notReady"``. This subsumes liveness, since a dead node's ``get_state`` service is gone. For a
 plain node with no lifecycle services, the status falls back to ``App::is_online`` (presence in
-the ROS 2 graph), which is the best signal available for an unmanaged node. The GetState read runs
-on a private node and executor (spun inline), so it never blocks or races the gateway executor.
+the ROS 2 graph), which is the best signal available for an unmanaged node. A node that is not
+online short-circuits to ``"notReady"`` without a GetState read (an offline node cannot be
+``active``), which also avoids a blocking read against a crashed managed node whose services still
+linger in the cache. The GetState read runs on a private node and executor (spun inline), so it
+never blocks or races the gateway executor; it is, however, serialized by an internal mutex, so a
+reachable-but-slow managed node holds that mutex across its spin and delays other concurrent
+``/status`` reads for up to the (short) read timeout.
 
 **Component status:** a local component is ``"ready"`` while the gateway is serving the
 request (the substrate is reachable), independent of how many hosted apps are online,
--- a/src/ros2_medkit_gateway/design/lifecycle.rst
+++ b/src/ros2_medkit_gateway/design/lifecycle.rst
@@ -86,10 +86,13 @@ The optional transition URI fields are only present when the registered
 ``LifecycleProvider`` signals support for that transition. Each field value
 is the absolute path the client should call to trigger the transition.
 
-**App status (no provider registered):** derived from ``App::is_online`` in
-the entity cache. The entity cache marks an App online when it has at least
-one node in the ROS 2 graph. ``is_online = true`` maps to ``"ready"``;
-``false`` maps to ``"notReady"``.
+**App status:** for a managed lifecycle node (one exposing ``get_state`` and ``change_state``),
+the status is read from the node's lifecycle state via ``lifecycle_msgs/srv/GetState``: the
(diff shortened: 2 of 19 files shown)


src/ros2_medkit_gateway/src/http/handlers/lifecycle_handlers.cpp
```diff
@@ -129,28 +136,28 @@ http::Result<dto::LifecycleStatusResponse> LifecycleHandlers::handle_get_status(
...
-          }
-        }
-      }
+    std::optional<std::string> get_state_path;
+    if (app && lifecycle_reader_) {
+      get_state_path = find_lifecycle_get_state_path(app->services);
+    }
+    if (app && lifecycle_reader_ && get_state_path.has_value()) {
+      // Managed lifecycle node: read the real state. "active" is the only ready state.
+      auto state = lifecycle_reader_->get_state(*get_state_path);
```
mfaferek93, line 153: This calls `get_state()` whenever the cached services still advertise GetState+ChangeState, without first checking `app->is_online`. For a managed node that has crashed (services linger in the last-cached `app->services`, but the node is gone from the graph), every poll runs `wait_for_service` to exhaustion then `spin_until_future_complete` - a guaranteed ~timeout blocking call - and it's taken under the reader's single global mutex, so one crashed managed node stalls every other concurrent `/status` read. A plain offline node short-circuits instantly via `is_online`; the managed one doesn't, which punishes exactly the failure case operators poll hardest. Suggest `if (app && !app->is_online) { resp.status = "notReady"; }` before the reader call - an offline node can't be active.

src/ros2_medkit_gateway/src/ros2/status/ros2_lifecycle_state_reader.cpp
```diff
@@ -0,0 +1,53 @@
...
+    executor_->remove_node(client_node_);
+  }
+}
+
+std::optional<std::string> Ros2LifecycleStateReader::get_state(const std::string & get_state_service_path) {
+  std::lock_guard<std::mutex> lock(mutex_);
+  const auto clamped = std::chrono::duration<double>(std::max(timeout_.count(), 0.0));
+
+  auto client = client_node_->create_client<lifecycle_msgs::srv::GetState>(get_state_service_path);
+  if (!client->wait_for_service(clamped)) {
```
mfaferek93: `wait_for_service` runs under `mutex_` (locked at line 37), so with cpp-httplib's thread pool a slow or unreachable node head-of-line-blocks every concurrent `/status` poll for up to the timeout. The transport this file's header cites - `invoke_fault_service` in ros2_fault_service_transport.cpp - deliberately runs `wait_for_service` *outside* the executor mutex (the graph listener backing it is independent); only `async_send_request` + the spin need serializing. Suggest moving the `lock_guard` down to just before `async_send_request`. Keep `create_client` under the lock though - creating a client mutates `client_node_`'s registry and would race concurrent callers otherwise.

src/ros2_medkit_gateway/src/ros2/status/ros2_lifecycle_state_reader.cpp
```diff
@@ -0,0 +1,53 @@
...
+  if (executor_ && client_node_) {
+    executor_->remove_node(client_node_);
+  }
+}
+
+std::optional<std::string> Ros2LifecycleStateReader::get_state(const std::string & get_state_service_path) {
+  std::lock_guard<std::mutex> lock(mutex_);
+  const auto clamped = std::chrono::duration<double>(std::max(timeout_.count(), 0.0));
+
+  auto client = client_node_->create_client<lifecycle_msgs::srv::GetState>(get_state_service_path);
```
mfaferek93: `create_client(get_state_service_path)` is called with no validation. An empty or malformed cached path throws `rclcpp::exceptions::InvalidServiceNameError`, and the default (no-provider) status branch in handle_get_status has no try/catch around the reader call (only the plugin-provider branch does), so it escapes onto the HTTP handler thread instead of degrading to notReady. A quick empty-path guard (or wrapping create_client in try/catch returning `std::nullopt`) keeps a bad name safe. Separately: creating+destroying a client per call churns DDS entities on every poll - caching clients keyed by service path would amortize it, but that's a nice-to-have.

src/ros2_medkit_gateway/src/ros2/status/ros2_lifecycle_state_reader.cpp
```diff
@@ -0,0 +1,53 @@
...
+namespace ros2_medkit_gateway {
+
+Ros2LifecycleStateReader::Ros2LifecycleStateReader(rclcpp::Node * host, std::chrono::duration<double> timeout)
+  : timeout_(timeout) {
+  client_node_ = std::make_shared<rclcpp::Node>(std::string(host->get_name()) + "_lifecycle_state_reader");
+  executor_ = std::make_shared<rclcpp::executors::SingleThreadedExecutor>();
+  executor_->add_node(client_node_);
+}
+
+Ros2LifecycleStateReader::~Ros2LifecycleStateReader() {
```
mfaferek93, line 31: The destructor doesn't lock `mutex_`, so it relies entirely on callers having drained before destruction. That holds today because `stop_rest_server()` joins the HTTP thread before the handler is destroyed, but the transport you mirror locks its executor mutex in its destructor precisely so an in-flight RPC can't race teardown (remove_node during an active spin is UB). Suggest locking `mutex_` here for parity and to make the class self-contained.

src/ros2_medkit_gateway/src/http/handlers/lifecycle_handlers.cpp
```diff
@@ -129,28 +136,28 @@ http::Result<dto::LifecycleStatusResponse> LifecycleHandlers::handle_get_status(
...
     }
-    resp.status = ready ? "ready" : "notReady";
+  } else {
+    // A local component is operational while the gateway is serving this
+    // request (the substrate is reachable). SOVD notReady means stopped,
+    // restarting, or unreachable; a reachable local substrate is none of
+    // these, regardless of how many hosted apps are online. A down hosted
+    // app is that app's own notReady, not the component's. Remote components
+    // are handled earlier by aggregation forwarding.
+    resp.status = "ready";
```
mfaferek93: Local components now report `ready` unconditionally, dropping #437's "ready iff at least one hosted app is online". The remote-forwarding guard in validate_entity_for_route makes this safe for reachability (remote components are forwarded to the owning peer before this runs), and the rationale is reasonable - just flagging the silent semantic shift: a local component hosting zero apps, or only down apps, now reads `ready` instead of `notReady`. Fine if that's the intended SOVD semantics (a down app is that app's own notReady); confirming it's a conscious decision.

src/ros2_medkit_integration_tests/test/features/test_lifecycle.test.py
```diff
@@ -116,6 +116,19 @@ def test_app_status_structure(self):
         self.assertIn('status', data)
         self.assertIn(data['status'], ['ready', 'notReady'])
 
+    def test_managed_lifecycle_unconfigured_is_not_ready(self):
```
mfaferek93, line 119: Nice - this proves the lifecycle reader is actually consulted (an unconfigured node is in the graph, so bare `is_online` would have said ready). But it only covers the notReady side. The demo node already supports `auto_activate:=true` (it self-configures + activates -> ready) and launch_helpers wires the param, yet nothing launches it activated, so the active=>ready happy path - the actual point of this PR - is proven only with the in-process stub, never through a real over-the-wire GetState round-trip. Could you add a case that activates the node and polls `/status` until `ready`?

src/ros2_medkit_gateway/design/lifecycle.rst
```diff
@@ -86,19 +86,19 @@ The optional transition URI fields are only present when the registered
...
-its hosted Apps is online (``App::is_online`` over the Component's hosted apps),
-and ``"notReady"`` otherwise. ``host_metadata`` is only used to identify the host
-Component; it is not itself a liveness signal.
+**App status:** for a managed lifecycle node (one exposing ``get_state`` and ``change_state``),
+the status is read from the node's lifecycle state via ``lifecycle_msgs/srv/GetState``: the
+``active`` state is ``"ready"``; any other state, or an unreachable/timed-out read, is
+``"notReady"``. This subsumes liveness, since a dead node's ``get_state`` service is gone. For a
+plain node with no lifecycle services, the status falls back to ``App::is_online`` (presence in
+the ROS 2 graph), which is the best signal available for an unmanaged node. The GetState read runs
+on a private node and executor (spun inline), so it never blocks or races the gateway executor.
```
mfaferek93: "spun inline, so it never blocks or races the gateway executor" is true for the host executor, but it omits that the read serializes *other* status reads: a single global mutex is held across wait_for_service + spin, so a slow or unreachable managed node stalls concurrent `/status` polls for up to the timeout. Worth one caveat line here so the cost isn't a surprise.

src/ros2_medkit_gateway/src/http/handlers/lifecycle_handlers.cpp
bburda: Added the short-circuit: an offline app returns `notReady` before the reader call. This fires for manifest/hybrid apps; in `runtime_only` an app that leaves the graph is dropped from the cache rather than kept as offline. The transient case (a just-crashed managed node still cached as online with a dead service) is covered by moving `wait_for_service` outside the mutex, so a stalled read no longer blocks other `/status` polls.

src/ros2_medkit_gateway/src/ros2/status/ros2_lifecycle_state_reader.cpp
bburda: Moved `wait_for_service` outside the mutex. `create_client`, `async_send_request`, the spin, and the client teardown stay serialized under it; only the unbounded service wait is unlocked now.

src/ros2_medkit_gateway/src/ros2/status/ros2_lifecycle_state_reader.cpp
bburda: Added an empty-path guard and wrapped `create_client` in try/catch returning `nullopt`, so a malformed cached path degrades to `notReady` instead of escaping onto the handler thread. Client caching is tracked in a separate follow-up issue.

src/ros2_medkit_gateway/src/ros2/status/ros2_lifecycle_state_reader.cpp
bburda: Locked the mutex in the destructor before `remove_node`, so an in-flight read cannot race teardown.

src/ros2_medkit_gateway/src/http/handlers/lifecycle_handlers.cpp
bburda: Reconsidered this. Local components no longer report `ready` unconditionally: the host component stays `ready`, but any other local component is `notReady` when it hosts apps and all of them are offline, and `ready` otherwise (including when it hosts no apps). This keeps a manifest subsystem in the entity tree and lets its status show it is down.

src/ros2_medkit_integration_tests/test/features/test_lifecycle.test.py
bburda: Added `test_managed_lifecycle_activated_is_ready`: it launches the lifecycle node with `auto_activate:=true` and polls `/status` until `ready`, proving the active->ready path over a real `GetState` round-trip.

src/ros2_medkit_gateway/design/lifecycle.rst
bburda: Added the caveat to the design doc: the read is serialized by an internal mutex, so a reachable-but-slow managed node delays other concurrent `/status` reads up to the read timeout. Also documented the offline short-circuit.

src/ros2_medkit_gateway/CMakeLists.txt
```diff
@@ -973,6 +987,7 @@ if(BUILD_TESTING)
       test_tls_config
       test_fault_manager
       test_error_info
+      test_ros2_lifecycle_state_reader
       test_ros2_subscription_executor
```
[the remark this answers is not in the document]
bburda: Added `test_lifecycle_status_helpers` to the coverage target list.

mfaferek93 (approved):

PR #456, opened 2026-06-21
Bound concurrent snapshot capture under fault storms

### LINKED ISSUE #441: Bound concurrent snapshot capture under fault storms
## Summary

When several different faults confirm at about the same time, the fault_manager starts one capture thread per fault with no upper bound.

`src/ros2_medkit_fault_manager/src/fault_manager_node.cpp` (around line 426) starts a `std::thread` per confirmation that runs snapshot capture and rosbag capture, tracked in `capture_threads_`. There is a recapture cooldown for the same fault code, but nothing bounds the number of concurrent capture threads across different faults. The rosbag side has a single writer (`active_writer_` under `writer_mutex_`, `recording_post_fault_`), so concurrent captures also contend there.

Measured with a fresh gateway per fault count: peak memory grows with the number of concurrent faults (about +0.5 MiB at N=1 up to +5.8 MiB at N=16), and above about N=4 the process does not return to its pre-fault memory within a 20 s window. Reproduced with a benchmark harness in selfpatch_demos (fault lane, `--faults 1,2,4,8,16`).

## Proposed solution

- Replace the unbounded thread-per-fault model with a bounded capture thread pool or a semaphore.
- Use a bounded queue with a clear policy when it is full (drop oldest or reject), so memory has an upper bound under a fault storm.

## Additional context

Capture is already off the service callback thread, so the fault path is not blocked. The problem is unbounded concurrency: N simultaneous faults cost N threads and N buffers. The single rosbag writer already serializes its part; the snapshot path does not.

bburda:
## Summary

Replace the unbounded thread-per-fault snapshot/rosbag capture in the fault manager with a bounded worker pool draining a bounded queue, so peak memory and thread count have a hard, configurable upper bound under a fault storm.

Previously, each fault confirmation spawned one `std::thread` to run snapshot + rosbag capture, with no ceiling on concurrent captures across different faults: N simultaneous faults cost N threads and N capture buffers (measured peak memory grew from about +0.5 MiB at N=1 to about +5.8 MiB at N=16, and above roughly N=4 the process did not return to its pre-fault footprint within 20 s). The `capture_threads_` vector also accumulated finished thread handles for the node's entire lifetime.

### What changed

- **`CaptureThreadPool`** - a fixed pool of N workers draining a bounded FIFO queue of fault codes. Hard concurrency bound = pool size; queued items are cheap strings. Full-queue policy is configurable: `reject_newest` (drop the incoming capture) or `drop_oldest` (evict the oldest pending capture). Drops are logged (throttled) and counted.
- **Three new parameters** under `snapshots.`: `capture_pool_size` (default `2`), `capture_queue_depth` (default `16`), `capture_queue_full_policy` (default `reject_newest`).
- **Confirm path** enqueues into the pool instead of spawning a thread; the cooldown check, enqueue, and cooldown stamp/erase are atomic under one lock, and a dropped capture is not silently put on cooldown.
- **Teardown fix**: the destructor now joins the pool before `RosbagCapture::stop()`. `stop()` is not a barrier against an in-flight `on_fault_confirmed()` (it clears subscriptions/buffers), so joining workers first removes a latent use-after-free window.

### Behavior note

Under a sustained storm that exceeds `pool_size + queue_depth`, some captures are now dropped per the configured policy (the intended bound). Defaults are sized to absorb the documented N<=16 storm without dropping. This is not an API change; existing snapshot/rosbag behavior is unchanged below the bound.

---

## Issue

- closes #441

---

## Type

- [ ] Bug fix
- [x] New feature or tests
- [ ] Breaking change
- [ ] Documentation only

---

## Testing

- Unit tests for `CaptureThreadPool`: bounded-and-reaches concurrency, both full-queue policies with exact drop counts and eviction order, FIFO ordering, `pool_size=1` serialization, callback-exception isolation, enqueue-after-shutdown, shutdown discards-pending while completing in-flight, idempotent shutdown, and deterministic `capture_fn` release after join. All gates are mutex/condition-variable based (no sleeps), C++17.
- Parameter-validation tests: clamping (`< 1` -> 1) and policy parsing (including the `drop_oldest` path).
- Integration test: confirms a burst of 8 distinct faults and verifies the node stays responsive (no crash/UAF) and snapshots are still captured through the bounded pool.
- Full `ros2_medkit_fault_manager` unit + integration + linter suites pass; clang-tidy is clean on the changed files.

Reviewers can verify with `colcon test --packages-select ros2_medkit_fault_manager` (unit + linters) and inspect the bound via the `CaptureThreadPool` unit tests.

---

## Checklist

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/docs/tutorials/snapshots.rst
+++ b/docs/tutorials/snapshots.rst
@@ -90,6 +90,15 @@ Configure snapshot capture via fault manager parameters:
      - ``10``
      - Maximum number of snapshots stored per fault code. When the limit is reached,
        new snapshots for that fault are rejected. Set to 0 for unlimited.
+   * - ``snapshots.capture_pool_size``
+     - ``2``
+     - Max concurrent capture threads under a fault storm (>= 1).
+   * - ``snapshots.capture_queue_depth``
+     - ``16``
+     - Max pending captures before the full-queue policy applies (>= 1).
+   * - ``snapshots.capture_queue_full_policy``
+     - ``reject_newest``
+     - Policy when the queue is full: ``reject_newest`` or ``drop_oldest``.
 
 Advanced Configuration
 ----------------------
--- a/docs/tutorials/snapshots.rst
+++ b/docs/tutorials/snapshots.rst
@@ -92,7 +92,9 @@ Configure snapshot capture via fault manager parameters:
        new snapshots for that fault are rejected. Set to 0 for unlimited.
    * - ``snapshots.capture_pool_size``
      - ``2``
-     - Max concurrent capture threads under a fault storm (>= 1).
+     - Max concurrent capture threads under a fault storm (>= 1). This parallelizes
+       freeze-frame snapshot capture only; rosbag capture is single-writer and
+       records one fault at a time regardless of this value.
    * - ``snapshots.capture_queue_depth``
      - ``16``
      - Max pending captures before the full-queue policy applies (>= 1).
--- a/src/ros2_medkit_fault_manager/config/fault_manager.yaml
+++ b/src/ros2_medkit_fault_manager/config/fault_manager.yaml
@@ -20,7 +20,10 @@ fault_manager:
     # storage backend is usable).
     snapshots.rosbag.enabled: false
 
-    # Capture concurrency bound under fault storms (issue #441):
+    # Capture concurrency bound under fault storms (issue #441). These govern the
+    # shared capture pool, which is created when snapshots OR rosbag is enabled, so
+    # they also bound rosbag capture. Note: pool_size parallelizes freeze-frame
+    # snapshots only - rosbag is single-writer and records one fault at a time.
     # snapshots.capture_pool_size: 2                  # max concurrent capture threads (>= 1)
     # snapshots.capture_queue_depth: 16               # max pending captures (>= 1)
     # snapshots.capture_queue_full_policy: reject_newest  # reject_newest | drop_oldest
--- a/src/ros2_medkit_fault_manager/config/fault_manager.yaml
+++ b/src/ros2_medkit_fault_manager/config/fault_manager.yaml
@@ -19,3 +19,8 @@ fault_manager:
     # to entity-scoped capture and is crash-safe (falls back / self-disables if no
     # storage backend is usable).
     snapshots.rosbag.enabled: false
+
+    # Capture concurrency bound under fault storms (issue #441):
+    # snapshots.capture_pool_size: 2                  # max concurrent capture threads (>= 1)
+    # snapshots.capture_queue_depth: 16               # max pending captures (>= 1)
+    # snapshots.capture_queue_full_policy: reject_newest  # reject_newest | drop_oldest
--- a/src/ros2_medkit_fault_manager/src/capture_thread_pool.cpp
+++ b/src/ros2_medkit_fault_manager/src/capture_thread_pool.cpp
@@ -23,12 +23,17 @@ namespace ros2_medkit_fault_manager {
 
 CaptureThreadPool::CaptureThreadPool(std::size_t pool_size, std::size_t queue_depth, QueueFullPolicy full_policy,
                                      rclcpp::Logger logger, std::function<void(const std::string &)> capture_fn)
-  : queue_depth_(queue_depth)
+  : queue_depth_(queue_depth == 0 ? 1 : queue_depth)
   , full_policy_(full_policy)
   , logger_(std::move(logger))
   , capture_fn_(std::move(capture_fn)) {
-  workers_.reserve(pool_size);
-  for (std::size_t i = 0; i < pool_size; ++i) {
+  // Defensive clamp: this is a public class. Callers (the node) validate to >= 1,
+  // but a stray 0 must not UB. With 0 workers nothing drains the queue; with
+  // queue_depth 0 + kDropOldest, enqueue() would call front() on an empty deque.
+  // queue_depth_ is clamped in the initializer above.
+  const std::size_t worker_count = pool_size == 0 ? 1 : pool_size;
+  workers_.reserve(worker_count);
+  for (std::size_t i = 0; i < worker_count; ++i) {
     workers_.emplace_back([this] {
       worker_loop();
     });
@@ -75,6 +80,11 @@ void CaptureThreadPool::shutdown() noexcept {
   cv_.notify_all();
   for (auto & worker : workers_) {
     if (worker.joinable()) {
+      // join() only throws std::system_error, and only for a self-join
+      // (resource_deadlock_would_occur). shutdown() runs on the owner/destructor
+      // thread, never on a worker, so that cannot happen here. The catch keeps
+      // shutdown() noexcept and, in the impossible-in-practice case that one join
+      // fails, still lets us join the remaining workers instead of terminating.
       try {
         worker.join();
       } catch (const std::system_error & e) {
--- a/src/ros2_medkit_fault_manager/src/capture_thread_pool.cpp
+++ b/src/ros2_medkit_fault_manager/src/capture_thread_pool.cpp
@@ -1,0 +1,120 @@
+// Copyright 2026 bburda
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+#include "ros2_medkit_fault_manager/capture_thread_pool.hpp"
+
+#include <system_error>
+#include <utility>
+
+#include "rclcpp/logging.hpp"
+
+namespace ros2_medkit_fault_manager {
+
+CaptureThreadPool::CaptureThreadPool(std::size_t pool_size, std::size_t queue_depth, QueueFullPolicy full_policy,
+                                     rclcpp::Logger logger, std::function<void(const std::string &)> capture_fn)
+  : queue_depth_(queue_depth)
+  , full_policy_(full_policy)
+  , logger_(std::move(logger))
+  , capture_fn_(std::move(capture_fn)) {
+  workers_.reserve(pool_size);
+  for (std::size_t i = 0; i < pool_size; ++i) {
+    workers_.emplace_back([this] {
+      worker_loop();
+    });
+  }
+}
+
+CaptureThreadPool::~CaptureThreadPool() {
+  shutdown();
+}
+
+EnqueueOutcome CaptureThreadPool::enqueue(const std::string & fault_code) {
+  std::lock_guard<std::mutex> lock(queue_mutex_);
+  if (stop_) {
+    return {EnqueueResult::kRejectedShuttingDown, std::nullopt};
+  }
+  if (queue_.size() < queue_depth_) {
+    queue_.push_back(fault_code);
+    cv_.notify_one();
+    return {EnqueueResult::kAccepted, std::nullopt};
+  }
+  // Queue full.
+  if (full_policy_ == QueueFullPolicy::kRejectNewest) {
+    dropped_captures_.fetch_add(1, std::memory_order_relaxed);
+    return {EnqueueResult::kDroppedNewest, std::nullopt};
+  }
+  // kDropOldest: evict the oldest pending job.
+  std::string evicted = std::move(queue_.front());
+  queue_.pop_front();
+  queue_.push_back(fault_code);
+  dropped_captures_.fetch_add(1, std::memory_order_relaxed);
+  cv_.notify_one();
+  return {EnqueueResult::kEvictedOldest, std::move(evicted)};
+}
+
+void CaptureThreadPool::shutdown() noexcept {
+  {
+    std::lock_guard<std::mutex> lock(queue_mutex_);
+    if (stop_) {
+      return;
+    }
+    stop_ = true;
+    queue_.clear();  // discard pending jobs atomically with setting stop_
+  }
+  cv_.notify_all();
+  for (auto & worker : workers_) {
+    if (worker.joinable()) {
+      try {
+        worker.join();
+      } catch (const std::system_error & e) {
+        RCLCPP_ERROR(logger_, "CaptureThreadPool worker join failed: %s", e.what());
+      }
+    }
+  }
+  // Release captured shared_ptrs deterministically, before owning objects vanish.
+  capture_fn_ = nullptr;
+}
+
+std::size_t CaptureThreadPool::pending_size() const {
+  std::lock_guard<std::mutex> lock(queue_mutex_);
+  return queue_.size();
+}
+
+void CaptureThreadPool::worker_loop() {
+  for (;;) {
+    std::string job;
+    {
+      std::unique_lock<std::mutex> lock(queue_mutex_);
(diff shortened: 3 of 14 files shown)


src/ros2_medkit_fault_manager/src/fault_manager_node.cpp
```diff
@@ -210,6 +244,22 @@ FaultManagerNode::FaultManagerNode(const rclcpp::NodeOptions & options) : Node("
...
+    auto snap = snapshot_capture_;
+    auto bag = rosbag_capture_;
+    capture_pool_ = std::make_unique<CaptureThreadPool>(
+        static_cast<std::size_t>(capture_pool_size_), static_cast<std::size_t>(capture_queue_depth_),
+        capture_queue_full_policy_, get_logger(), [snap, bag](const std::string & fault_code) {
+          if (snap) {
+            snap->capture(fault_code);
+          }
+          if (bag) {
+            bag->on_fault_confirmed(fault_code);
```
mfaferek93, line 264: The pool documents `capture_fn` must be thread-safe for pool_size concurrent calls (capture_thread_pool.hpp:63) and pool_size defaults to 2, but `RosbagCapture::on_fault_confirmed` isn't safe for that. The cooldown only dedupes the *same* fault code, so two different codes confirming in a storm (the #441 scenario) run this concurrently, and then race: `recording_post_fault_` is a check-then-act (rosbag_capture.cpp:249 load vs :272 store), both reset/reopen the single `active_writer_`, and `current_fault_code_`/`current_bag_path_` (plain strings, no mutex) are written here while the post-fault timer callback reads them - a data race. The `if(!active_writer_) break` guard avoids a crash, so the symptom is bag corruption/truncation, not a SIGSEGV. Snapshot capture is fine; this is rosbag-only and rosbag is off by default. Not a regression (the old thread-per-fault model raced harder), but this PR now blesses it via the documented contract + a >1 default. Suggest serializing the rosbag leg under one mutex spanning on_fault_confirmed, or defaulting pool_size to 1 when rosbag is enabled, and aligning the header contract.

src/ros2_medkit_fault_manager/src/fault_manager_node.cpp
```diff
@@ -400,46 +443,70 @@ void FaultManagerNode::handle_report_fault(
...
+        const auto now = std::chrono::steady_clock::now();
+        // RCLCPP_WARN_THROTTLE needs a non-const Clock lvalue (Humble/Lyrical
+        // compat); mirror rosbag_capture.cpp's local-copy pattern. Cast the
+        // uint64_t counter to unsigned long long + %llu to avoid -Wuseless-cast
+        // (uint64_t == unsigned long on LP64).
+        rclcpp::Clock throttle_clock(*get_clock());
+        switch (outcome.result) {
+          case EnqueueResult::kAccepted:
+            if (cooldown_enabled) {
+              last_capture_times_[fault_code] = now;
```
mfaferek93, line 498: `last_capture_times_` is inserted per accepted/evicted capture but pruned only on clear/auto-clear or a drop_oldest eviction - never by time. A storm of *distinct* fault codes (the #441 case, and what test_27 sends) leaves one permanent entry per confirmed code with the default 60s cooldown on, so this map grows unbounded on the exact path the pool is meant to bound - the "hard upper bound on peak memory" doesn't fully hold. Cheap fix: while you hold `last_capture_mutex_` here, sweep entries whose elapsed time already exceeds the cooldown (they can never gate a capture again).

src/ros2_medkit_fault_manager/src/fault_manager_node.cpp
```diff
@@ -400,46 +443,70 @@ void FaultManagerNode::handle_report_fault(
...
-      if (snapshot_recapture_cooldown_sec_ > 0.0) {
-        std::lock_guard<std::mutex> lock(last_capture_mutex_);
-        auto it = last_capture_times_.find(request->fault_code);
+    // Capture snapshots/rosbag when a fault confirms. Bounded pool (issue #441):
+    // the cooldown check, enqueue, and cooldown update are atomic under
+    // last_capture_mutex_ to prevent a concurrent same-fault double-enqueue.
+    if (just_confirmed && capture_pool_) {
+      const std::string fault_code = request->fault_code;
+      const bool cooldown_enabled = snapshot_recapture_cooldown_sec_ > 0.0;
+      std::unique_lock<std::mutex> cd_lock(last_capture_mutex_, std::defer_lock);
```
mfaferek93, line 460: The comment above says the cooldown check + enqueue + update are atomic under `last_capture_mutex_` to prevent a same-fault double-enqueue, but the lock is only taken when cooldown is enabled (defer_lock, locked conditionally). With `recapture_cooldown_sec=0` the enqueue runs lock-free. In practice it's harmless today because the node is spun single-threaded, which is what actually serializes handle_report_fault - so please either always take the lock around enqueue, or scope the comment to that (single-threaded executor / cooldown-enabled), since the stated invariant isn't what holds it.

src/ros2_medkit_fault_manager/test/test_integration.test.py
```diff
@@ -956,6 +956,78 @@ def test_26_cleared_fault_can_be_reactivated(self):
...
+
+        def keep_publishing():
+            while not stop_publishing.is_set():
+                default_pub.publish(temp_msg)
+                time.sleep(0.05)
+
+        pub_thread = threading.Thread(target=keep_publishing)
+        pub_thread.start()
+        try:
+            for code in fault_codes:
```
mfaferek93: This sends 8 distinct faults against the default pool (pool_size=2, queue_depth=16, cap 18), so the bound is never exceeded - the full-queue policy, eviction, and `dropped_captures` paths that #441 exists to enforce never fire here, and the test only proves liveness. Suggest overriding `snapshots.capture_pool_size`/`capture_queue_depth` to small values (e.g. 1/2) in this launch so the burst genuinely overflows, otherwise `survives_fault_storm` overpromises (the bound is covered only in the C++ unit tests).

src/ros2_medkit_fault_manager/test/test_integration.test.py
```diff
@@ -956,6 +956,78 @@ def test_26_cleared_fault_can_be_reactivated(self):
...
+            snap_request = GetSnapshots.Request()
+            snap_request.fault_code = code
+            snap_request.topic = ''
+            snap_response = self._call_service(self.get_snapshots_client, snap_request)
+            if snap_response.success and len(snap_response.data) > 0:
+                data = json.loads(snap_response.data)
+                if data.get('topics'):
+                    captured += 1
+        print(f'Storm capture: {captured}/{len(fault_codes)} faults captured topic data')
+        self.assertGreater(captured, 0, 'No storm fault captured topic data on /test/default_data')
```
mfaferek93: `assertGreater(captured, 0)` is nearly a tautology here: a steady publisher + 5s drain makes at least one of 8 captures a near-certainty, and the same assertion passes against the old unbounded thread-per-fault model - so it adds little regression signal for the new pool. Consider an invariant tied to the bound instead, e.g. captured+dropped accounts for all faults, or at least pool_size captures land.

src/ros2_medkit_fault_manager/test/test_capture_thread_pool.cpp
```diff
@@ -0,0 +1,311 @@
...
+    std::unique_lock<std::mutex> lock(m);
+    cv.wait(lock, [&] {
+      return ran >= 2;
+    });
+  }
+  pool.shutdown();
+  EXPECT_EQ(ran, 2);
+}
+
+TEST_F(CaptureThreadPoolTest, EnqueueAfterShutdownRejected) {
```
mfaferek93, line 261: This checks enqueue() only *after* shutdown() has fully returned. The interleaving that the kRejectedShuttingDown guard actually exists for - a fault arriving via enqueue() concurrently with the destructor setting stop_ - isn't exercised anywhere (ShutdownDiscardsPendingCompletesInFlight runs shutdown on a side thread but doesn't concurrently enqueue). The guard looks correct; this is just a coverage gap worth closing given the concurrency bar here. Suggest a producer thread hammering enqueue() during shutdown(), asserting every job is either run or cleanly rejected (clean under TSan).

docs/tutorials/snapshots.rst
```diff
@@ -90,6 +90,15 @@ Configure snapshot capture via fault manager parameters:
      - ``10``
      - Maximum number of snapshots stored per fault code. When the limit is reached,
        new snapshots for that fault are rejected. Set to 0 for unlimited.
+   * - ``snapshots.capture_pool_size``
+     - ``2``
+     - Max concurrent capture threads under a fault storm (>= 1).
```
mfaferek93: `capture_pool_size` is documented as "max concurrent capture threads" that bounds memory, which is accurate for freeze-frame snapshots, but the rosbag path is single-buffer: `RosbagCapture::on_fault_confirmed` gates on one `recording_post_fault_` flag over a shared writer, so only one rosbag records at a time and other concurrent confirmations log "Already recording... skipping" (not via `capture_queue_full_policy`). Worth one line that `capture_pool_size` parallelizes snapshots only and doesn't add rosbag concurrency, so raising it for black-box capture doesn't surprise with unexplained skips.

src/ros2_medkit_fault_manager/config/fault_manager.yaml
```diff
@@ -19,3 +19,8 @@ fault_manager:
     # to entity-scoped capture and is crash-safe (falls back / self-disables if no
     # storage backend is usable).
     snapshots.rosbag.enabled: false
+
+    # Capture concurrency bound under fault storms (issue #441):
+    # snapshots.capture_pool_size: 2                  # max concurrent capture threads (>= 1)
```
mfaferek93, line 27: These `snapshots.capture_*` params also bound rosbag capture - the pool is created when snapshots OR rosbag is enabled, and its job calls both `snapshot.capture()` and `rosbag.on_fault_confirmed()`. So with `snapshots.enabled: false` but `rosbag.enabled: true` they still govern rosbag concurrency/queueing, but they're documented only under Snapshot config. One clarifying line that `capture_*` governs the shared snapshot+rosbag pool would save a rosbag-only operator the confusion.

src/ros2_medkit_fault_manager/src/fault_manager_node.cpp
bburda: Serialized the rosbag leg: the pool job now takes a mutex around `on_fault_confirmed()`, so two distinct fault codes confirming concurrently can no longer race on the single writer / recording state. Snapshot capture is independent and stays parallel, so I kept `pool_size > 1` and the pool's thread-safety contract now holds (the job serializes its own non-thread-safe resource). Fixed in 529ec768.

src/ros2_medkit_fault_manager/src/fault_manager_node.cpp
bburda: Bounded the map: while holding `last_capture_mutex_` on each confirm, entries older than the cooldown are now swept (they can never gate a capture again), so it's bounded by the distinct faults seen within the cooldown window instead of growing one entry per code. Fixed in 529ec768.

src/ros2_medkit_fault_manager/src/fault_manager_node.cpp
bburda: Rewrote the comment to state the real invariant: `handle_report_fault` runs on the single-threaded executor, so confirmations are serialized; `last_capture_mutex_` only guards `last_capture_times_` and is taken when the cooldown is enabled. Dropped the inaccurate same-fault-double-enqueue claim. 529ec768.

src/ros2_medkit_fault_manager/test/test_integration.test.py
bburda: The launch now caps the pool small (pool_size=1, queue_depth=2, capacity 3), and the test keeps the publisher idle during the burst so each on-demand capture blocks the full timeout; a 12-fault burst then overflows capacity deterministically and the reject_newest drop path fires. 39d95aea.

src/ros2_medkit_fault_manager/test/test_integration.test.py
bburda: Replaced the near-tautological captured>0 with a bound-tied assertion: the test now asserts the pool logged a `Capture queue full` drop (via `proc_output.assertWaitFor`) plus node liveness. That drop signal is absent on the old unbounded model, so it's the real regression guard. 39d95aea.

src/ros2_medkit_fault_manager/test/test_capture_thread_pool.cpp
bburda: Added `ConcurrentEnqueueDuringShutdownIsClean`: a producer thread hammers `enqueue()` while `shutdown()` runs concurrently, asserting the shutdown transition is observed, post-shutdown enqueues reject cleanly, and no job runs that was never accepted. Race-clean under TSan. b7153a81.

docs/tutorials/snapshots.rst
bburda: Documented that `capture_pool_size` parallelizes freeze-frame snapshot capture only; rosbag is single-writer and records one fault at a time regardless of pool size. f373a7c1.

src/ros2_medkit_fault_manager/config/fault_manager.yaml
bburda: Documented that the `capture_*` parameters govern the shared capture pool, which is created when snapshots or rosbag is enabled, so they bound rosbag too (README, config reference, and the commented config knobs). f373a7c1.

src/ros2_medkit_fault_manager/test/test_capture_thread_pool.cpp
```diff
@@ -0,0 +1,292 @@
...
+
+#include <algorithm>
+#include <condition_variable>
+#include <cstddef>
+#include <functional>
+#include <mutex>
+#include <set>
+#include <string>
+#include <thread>
+#include <vector>
```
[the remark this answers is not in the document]
bburda: Already addressed in 40b3b43f - `<memory>` and `<stdexcept>` are now included explicitly.

src/ros2_medkit_fault_manager/src/capture_thread_pool.cpp
```diff
@@ -0,0 +1,120 @@
...
+  cv_.notify_all();
+  for (auto & worker : workers_) {
+    if (worker.joinable()) {
+      try {
+        worker.join();
+      } catch (const std::system_error & e) {
+        RCLCPP_ERROR(logger_, "CaptureThreadPool worker join failed: %s", e.what());
+      }
+    }
+  }
```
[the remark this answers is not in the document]
bburda: `shutdown()` runs only on the owner/destructor thread, never on a worker, so `join()` can only throw `std::system_error(resource_deadlock_would_occur)` on a self-join, which cannot happen here; the `joinable()` check covers the not-joinable case. No worker is ever left joinable. The catch only keeps `shutdown()` noexcept and, in the impossible-in-practice case, still lets the remaining workers be joined. Documented inline at capture_thread_pool.cpp:83-92.

src/ros2_medkit_fault_manager/src/capture_thread_pool.cpp
```diff
@@ -0,0 +1,120 @@
...
+  , full_policy_(full_policy)
+  , logger_(std::move(logger))
+  , capture_fn_(std::move(capture_fn)) {
+  workers_.reserve(pool_size);
+  for (std::size_t i = 0; i < pool_size; ++i) {
+    workers_.emplace_back([this] {
+      worker_loop();
+    });
+  }
+}
```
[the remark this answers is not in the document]
bburda: Already addressed in 40b3b43f - the constructor clamps both pool_size and queue_depth to >= 1, covered by the `ZeroSizesClampToOne` test.

mfaferek93 (approved):

PR #457, opened 2026-06-21
feat(gateway): bound the executor and HTTP server thread pools

### LINKED ISSUE #440: Bound the rclcpp executor and HTTP server thread pools
## Summary

The gateway runs about 50 threads at idle (about 53 on a Nav2 stack). Most of them are two pools that size themselves to the host CPU count, not to the work.

- The rclcpp executor is created as `MultiThreadedExecutor executor;` in `src/ros2_medkit_gateway/src/main.cpp` (around line 77), with no thread count. The default uses `std::thread::hardware_concurrency()`.
- The vendored cpp-httplib server uses `CPPHTTPLIB_THREAD_POOL_COUNT = max(8, hardware_concurrency - 1)`. `HttpServerManager` does not override `new_task_queue` (`src/ros2_medkit_gateway/src/http/http_server.cpp`), so the pool stays at that default.

On a 16-core host this is about 39 threads across the two pools. On a 64-core host it would be much larger for the same workload.

Under HTTP load the request p95 latency stays low (about 2.3 ms at 32 concurrent clients). So this is thread and scheduling overhead, not a latency problem. Measured with a benchmark harness in selfpatch_demos (load lane, thread census).

## Proposed solution

- Add a parameter for the executor thread count and pass it through `rclcpp::ExecutorOptions`. Default to a small bounded value instead of all cores.
- Bound the HTTP pool by setting `server->new_task_queue` to a fixed-size `ThreadPool`, also from a parameter.

## Additional context

Neither pool has a parameter today. The thread count grows with the host core count, so the footprint is bigger than needed on many-core machines.

bburda:
## Summary

The rclcpp `MultiThreadedExecutor` and the vendored cpp-httplib request pool both sized themselves to the host CPU count, so the gateway ran ~50 threads at idle (about 53 on a Nav2 stack) and far more on many-core hosts for the same workload. Under load, request p95 latency stayed low (~2.3 ms at 32 concurrent clients), so this was thread/scheduling overhead, not a latency problem.

This bounds both pools and makes them configurable:

- **`server.executor_threads`** (default `2`): thread count for the main `MultiThreadedExecutor`, replacing rclcpp's default (host cores, min 2). The executor only delivers the node's own callbacks plus the fast service-response callbacks; the blocking wait for an RPC runs on the cpp-httplib pool thread, not an executor thread, so a small executor cannot starve or deadlock RPC handlers.
- **`server.http_thread_pool_size`** (default `3`): a fixed-size cpp-httplib `ThreadPool` installed via `new_task_queue`, replacing `max(8, cores - 1)`.

Both are resolved through a shared `clamp_thread_count()` helper into a bounded range (`[1, 256]` for the executor, `[1, 1024]` for the HTTP pool) so a mis-set value can neither drop the pool to zero nor spawn a pathological thread count.

Each active SSE stream pins one HTTP worker for its lifetime, so **`sse.max_clients` default is lowered from 10 to 2** (kept at or below the pool) to keep SSE from starving ordinary requests. Raise `http_thread_pool_size` and `sse.max_clients` together for more concurrent streams.

---

## Issue

- closes #440

---

## Type

- [ ] Bug fix
- [x] New feature or tests
- [x] Breaking change
- [ ] Documentation only

Behavior change to surface: `sse.max_clients` default drops 10 -> 2, and both thread pools no longer scale with the host core count. Deployments relying on the old defaults should set the parameters explicitly.

---

## Testing

- **Unit (GTest):** new `test_http_server_thread_pool` starts a real loopback server and proves the pool caps handler concurrency at its size (pool=1 serializes, pool=2 caps at 2), and that `clamp_thread_count` floors/caps out-of-range values. Full gateway unit suite green (2306 tests, 0 failures).
- **Integration (launch_testing):** new `test_thread_pool_bounds` launches the gateway with bounded pools and verifies discovery + data sampling work and that ordinary requests are still served while an SSE stream holds a pool worker. Existing SSE / auth / updates integration tests pass under the new defaults.
- clang-format, cpplint, ament-copyright, flake8, and clang-tidy all clean.

---

## Checklist

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/src/ros2_medkit_gateway/src/gateway_node.cpp
+++ b/src/ros2_medkit_gateway/src/gateway_node.cpp
@@ -157,7 +157,16 @@ GatewayNode::GatewayNode(const rclcpp::NodeOptions & options) : Node("ros2_medki
   // together for more concurrent SSE. Both are clamped to a bounded range on
   // read (see clamp_thread_count).
   declare_parameter("server.executor_threads", 2);
-  declare_parameter("server.http_thread_pool_size", 3);
+  declare_parameter("server.http_thread_pool_size", 4);
+
+  // cpp-httplib keep-alive idle timeout (seconds). The library default (5s)
+  // parks a request-pool worker on an idle keep-alive connection for that long;
+  // with the small bounded pool above, a burst of short-lived client connections
+  // (e.g. a client polling several endpoints per cycle) can pin every worker and
+  // stall ordinary requests for up to one timeout per cycle. Defaulting to 2s
+  // recovers workers quickly while retaining connection reuse for legitimate
+  // clients. Clamped to [1, 3600] on read (see clamp_keep_alive_timeout).
+  declare_parameter("server.keep_alive_timeout_sec", 2);
   // Safety-backstop refresh interval. Primary discovery refresh is driven
   // by rclcpp graph events; this only controls the periodic forced
   // refresh used when a graph event would otherwise be missed.
--- a/src/ros2_medkit_gateway/src/gateway_node.cpp
+++ b/src/ros2_medkit_gateway/src/gateway_node.cpp
@@ -145,6 +145,19 @@ GatewayNode::GatewayNode(const rclcpp::NodeOptions & options) : Node("ros2_medki
   // Declare parameters with defaults
   declare_parameter("server.host", "127.0.0.1");
   declare_parameter("server.port", 8080);
+
+  // Thread-pool bounds (issue #440). Both pools default to a small fixed size
+  // instead of scaling with the host core count, so the footprint stays the
+  // same on a 4-core SBC and a 64-core server. The main executor only delivers
+  // the node's own callbacks (timers, graph events, subscriptions, and fast RPC
+  // response callbacks); the blocking wait for an RPC happens on the cpp-httplib
+  // pool thread, not an executor thread, so a small executor cannot starve it.
+  // Each active SSE stream holds one HTTP pool thread for its lifetime, so
+  // sse.max_clients is kept at or below http_thread_pool_size - raise both
+  // together for more concurrent SSE. Both are clamped to a bounded range on
+  // read (see clamp_thread_count).
+  declare_parameter("server.executor_threads", 2);
+  declare_parameter("server.http_thread_pool_size", 3);
   // Safety-backstop refresh interval. Primary discovery refresh is driven
   // by rclcpp graph events; this only controls the periodic forced
   // refresh used when a graph event would otherwise be missed.
@@ -158,7 +171,10 @@ GatewayNode::GatewayNode(const rclcpp::NodeOptions & options) : Node("ros2_medki
   declare_parameter("cors.max_age_seconds", 86400);
 
   // SSE (Server-Sent Events) parameters
-  declare_parameter("sse.max_clients", 10);         // Limit concurrent SSE connections to prevent resource exhaustion
+  // Concurrent SSE connections. Each one pins an HTTP pool worker for its
+  // lifetime, so this defaults at or below server.http_thread_pool_size (3) to
+  // leave a worker for ordinary requests; raise both together for more SSE.
+  declare_parameter("sse.max_clients", 2);
   declare_parameter("sse.max_subscriptions", 100);  // Maximum active cyclic subscriptions across all entities
   declare_parameter("sse.max_duration_sec", 3600);  // Maximum subscription duration in seconds (1 hour default)
 
--- a/src/ros2_medkit_gateway/src/gateway_node.cpp
+++ b/src/ros2_medkit_gateway/src/gateway_node.cpp
@@ -30,6 +30,7 @@
 #include "ros2_medkit_gateway/core/aggregation/network_utils.hpp"
 #include "ros2_medkit_gateway/core/data/topic_data_provider.hpp"
 #include "ros2_medkit_gateway/core/entity_validation.hpp"
+#include "ros2_medkit_gateway/core/thread_pool_config.hpp"
 #include "ros2_medkit_gateway/param_utils.hpp"
 #include "ros2_medkit_gateway/plugins/ros_plugin_context.hpp"
 
@@ -148,16 +149,25 @@ GatewayNode::GatewayNode(const rclcpp::NodeOptions & options) : Node("ros2_medki
 
   // Thread-pool bounds (issue #440). Both pools default to a small fixed size
   // instead of scaling with the host core count, so the footprint stays the
-  // same on a 4-core SBC and a 64-core server. The main executor only delivers
-  // the node's own callbacks (timers, graph events, subscriptions, and fast RPC
-  // response callbacks); the blocking wait for an RPC happens on the cpp-httplib
-  // pool thread, not an executor thread, so a small executor cannot starve it.
-  // Each active SSE stream holds one HTTP pool thread for its lifetime, so
-  // sse.max_clients is kept at or below http_thread_pool_size - raise both
-  // together for more concurrent SSE. Both are clamped to a bounded range on
-  // read (see clamp_thread_count).
+  // same on a 4-core SBC and a 64-core server.
+  //
+  // executor_threads: the main executor only delivers the node's own callbacks
+  // (timers, graph events, subscriptions) and the service-response callbacks
+  // that complete operation/action RPC futures. All of these run on the node's
+  // default callback group, which is mutually-exclusive, so they serialize
+  // through a single thread regardless of this count - raising it buys no
+  // RPC-response parallelism. The reason a small executor is safe is solely that
+  // the blocking wait for an RPC runs on the cpp-httplib pool thread (off the
+  // executor), never on an executor thread, so it cannot deadlock the executor.
+  //
+  // http_thread_pool_size: each active SSE stream pins one worker for its
+  // lifetime and each cold-/data wait parks one for up to
+  // topic_sample_timeout_sec, so the pool defaults at or above
+  // sse.max_clients + data_provider.cold_wait_cap (2 + 4 = 6) to avoid
+  // starvation; main.cpp warns at startup if it is set below that sum. Both
+  // values are clamped to a bounded range on read (see clamp_thread_count).
   declare_parameter("server.executor_threads", 2);
-  declare_parameter("server.http_thread_pool_size", 4);
+  declare_parameter("server.http_thread_pool_size", 6);
 
   // cpp-httplib keep-alive idle timeout (seconds). The library default (5s)
   // parks a request-pool worker on an idle keep-alive connection for that long;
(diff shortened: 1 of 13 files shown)


src/ros2_medkit_gateway/src/gateway_node.cpp
```diff
@@ -145,6 +145,28 @@ GatewayNode::GatewayNode(const rclcpp::NodeOptions & options) : Node("ros2_medki
...
+  // same on a 4-core SBC and a 64-core server. The main executor only delivers
+  // the node's own callbacks (timers, graph events, subscriptions, and fast RPC
+  // response callbacks); the blocking wait for an RPC happens on the cpp-httplib
+  // pool thread, not an executor thread, so a small executor cannot starve it.
+  // Each active SSE stream holds one HTTP pool thread for its lifetime, so
+  // sse.max_clients is kept at or below http_thread_pool_size - raise both
+  // together for more concurrent SSE. Both are clamped to a bounded range on
+  // read (see clamp_thread_count).
+  declare_parameter("server.executor_threads", 2);
+  declare_parameter("server.http_thread_pool_size", 4);
```
mfaferek93: The default pool (4) is below the sizing rule this PR documents in server.rst: `sse.max_clients(2) + cold_wait_cap(4) + headroom` >= 6 (the rst then claims "the defaults leave headroom", which they don't). The cold-topic `/data` path parks the httplib worker in `buf_cv.wait_for()` for up to `topic_sample_timeout_sec` (~2s), capped at `data_provider.cold_wait_cap` - which also defaults to 4. So a single burst of cold-topic data requests can park all 4 workers at once (the guard 503s only the 5th), leaving zero for SSE / `/health` / discovery until the sample timeout - the guard meant to protect the pool can consume it. Suggest raising this default to satisfy the formula against the shipped `cold_wait_cap`/`max_clients`, or lowering `cold_wait_cap` in lockstep.

src/ros2_medkit_gateway/src/http/rest_server.cpp
```diff
@@ -81,8 +82,25 @@ RESTServer::RESTServer(GatewayNode * node, const std::string & host, int port, c
...
+  // pool in [1, 1024]: a pool of 0 would queue every request forever, and a
+  // typo'd huge value would spawn that many OS threads. Each active SSE stream
+  // holds one worker for its lifetime, so the default pool is kept at or above
+  // sse.max_clients; if you raise sse.max_clients, raise
+  // server.http_thread_pool_size to match. clamp_keep_alive_timeout keeps the
+  // keep-alive timeout in [1, 3600]s: with a small pool, the cpp-httplib default
+  // (5s) lets a burst of short-lived client connections pin every worker, so we
+  // shorten it (default 2s) to recover workers quickly while retaining reuse.
+  const auto http_thread_pool_size =
+      clamp_thread_count(node_->get_parameter("server.http_thread_pool_size").as_int(), 1, 1024);
```
mfaferek93, line 97: `clamp_thread_count` bounds the pool in isolation [1,1024] but nothing relates it to `sse.max_clients` or `data_provider.cold_wait_cap`. The invariant `pool >= max_clients + cold_wait_cap` is documented in several comments + the rst, but a config that violates it produces silent total starvation rather than a warning: e.g. `http_thread_pool_size=1` with the default `sse.max_clients=2` pins the one worker on the first SSE stream forever, and the server can serve nothing else (not even the second client's rejection). Worth a startup warn/coerce when `pool < sse.max_clients + cold_wait_cap` (and clamp `sse.max_clients` too - it's read raw, unlike the pool knobs). Separately: bulk-data `binary_download` streams via set_content_provider and holds a worker for the whole transfer, uncounted against any cap - with the new pool of 4, two slow downloads + two SSE streams pin every worker. That endpoint should be in the budget too (harmless only because the pool used to be host-core-sized).

src/ros2_medkit_gateway/src/main.cpp
```diff
@@ -69,13 +71,23 @@ int main(int argc, char ** argv) {
...
-    rclcpp::executors::MultiThreadedExecutor executor;
+    // MultiThreadedExecutor for the gateway node. Issue #440: the thread count
+    // is bounded by the server.executor_threads parameter (default 2) instead of
+    // rclcpp's default (host cores, minimum 2), so the footprint does not grow
+    // with the host core count. The executor only delivers the gateway node's
+    // own callbacks - timers, graph events, log/fault subscriptions, and the
+    // (fast) service-response callbacks for operation/action RPCs. The blocking
+    // wait for those RPCs happens on the cpp-httplib pool thread (a separate
+    // server_thread_), not on an executor thread, and the fault transport runs
+    // on its own private executor - so a small executor here cannot starve or
```
mfaferek93: The no-deadlock reasoning is correct (the wait runs on the httplib worker, never an executor thread), but the rest is imprecise: the future is completed by a service-response callback delivered by the main executor, and every client registers on the node's default callback group (no custom groups), which is mutually-exclusive. So all RPC responses + timers + graph events serialize through a single executor thread regardless of `executor_threads`, and raising it buys no RPC-response parallelism. Worth rewording so a future maintainer doesn't add executor threads expecting response concurrency - the safety here comes purely from the wait being off-executor.

src/ros2_medkit_gateway/src/gateway_node.cpp
```diff
@@ -158,7 +180,10 @@ GatewayNode::GatewayNode(const rclcpp::NodeOptions & options) : Node("ros2_medki
   declare_parameter("cors.max_age_seconds", 86400);
 
   // SSE (Server-Sent Events) parameters
-  declare_parameter("sse.max_clients", 10);         // Limit concurrent SSE connections to prevent resource exhaustion
+  // Concurrent SSE connections. Each one pins an HTTP pool worker for its
+  // lifetime, so this defaults at or below server.http_thread_pool_size (3) to
```
mfaferek93: Stale number: this says SSE "defaults at or below server.http_thread_pool_size (3)" but the shipped default is 4 (same `(3)` in gateway_params.yaml:178 and the PR body). Looks like a leftover from before the widen-pool commit. Please update both to 4 - and note the real headroom, since with cold_wait_cap=4 the effective headroom is smaller than `pool - max_clients` implies.

src/ros2_medkit_gateway/src/gateway_node.cpp
```diff
@@ -158,7 +180,10 @@ GatewayNode::GatewayNode(const rclcpp::NodeOptions & options) : Node("ros2_medki
...
+  // leave a worker for ordinary requests; raise both together for more SSE.
+  declare_parameter("sse.max_clients", 2);
```
mfaferek93, line 199: `sse.max_clients` drops 10 -> 2 here. The rationale (each stream pins a worker, small pool) is sound, but it's a meaningful default reduction - a multi-panel dashboard or multiple clients now get only 2 concurrent SSE streams out of the box and the 3rd is rejected. Worth calling out as a behavior change in the PR/changelog, and maybe 3-4 if you bump the pool to match the sizing rule.

src/ros2_medkit_integration_tests/test/features/test_thread_pool_bounds.test.py
```diff
@@ -0,0 +1,205 @@
...
+from ros2_medkit_test_utils.gateway_test_case import GatewayTestCase
+from ros2_medkit_test_utils.launch_helpers import create_test_launch
+
+# Bounded pool sizes under test. Both are set to values DISTINCT from the
+# shipped defaults (http=4, executor=2, keep_alive=2) so that a regression which
+# ignored the parameters would change observable behaviour. The HTTP pool is
+# larger than the single SSE stream opened below so there is headroom for regular
+# requests, and the keep-alive timeout is set short so a recently-used connection
+# cannot pin a worker for long.
+HTTP_THREAD_POOL_SIZE = 5
```
mfaferek93, line 53: This launches with `pool=5`, `executor_threads=3`, one SSE stream and small bursts - a config that satisfies every invariant, so it can only pass. The scenarios the PR's safety actually rests on aren't exercised: `pool <= sse.max_clients` (the documented starvation case), `cold_wait_cap` concurrent cold-`/data` requests draining the pool, a bulk-data download held during a burst, and the `executor_threads` bound itself (set but never observed/asserted). A test that opens `max_clients` streams (or saturates cold-waits) at a pool that can't cover them - and asserts the server still serves `/health` or correctly 503s - would actually guard the invariant.

src/ros2_medkit_gateway/src/gateway_node.cpp
bburda: Agreed. The default `http_thread_pool_size` is now 6, which covers `sse.max_clients (2) + data_provider.cold_wait_cap (4)`. The docs state that as the sizing rule, and the gateway logs a startup warning if the pool is set below `sse.max_clients + cold_wait_cap`.

src/ros2_medkit_gateway/src/http/rest_server.cpp
bburda: All three addressed:
- `sse.max_clients` is now clamped on read (it was taken raw, so a negative value wrapped to a huge `size_t`).
- The gateway logs a startup warning when `http_thread_pool_size < sse.max_clients + data_provider.cold_wait_cap`.
- The thread-pool docs now call out that a bulk-data download also holds a worker for the whole transfer, uncounted by any cap, so it must be budgeted for.

src/ros2_medkit_gateway/src/main.cpp
bburda: Reworded both the comment and the docs: the node's default callback group is mutually-exclusive, so RPC-response callbacks, timers and graph events serialize through a single thread regardless of `executor_threads` - raising it buys no response parallelism. The note now states the safety comes solely from the blocking RPC wait running off the executor.

src/ros2_medkit_gateway/src/gateway_node.cpp
bburda: Fixed. The stale `(3)` in `gateway_node.cpp` and `gateway_params.yaml` now read `6`, and the headroom wording reflects the `sse.max_clients + cold_wait_cap` budget rather than `pool - max_clients`.

src/ros2_medkit_gateway/src/gateway_node.cpp
bburda: Kept at 2: with the pool at 6 the budget is tight (`max_clients (2) + cold_wait_cap (4) = 6`), so raising it would require a larger pool. The reduction from 10 is a deliberate behavior change and is now called out in the PR description.

src/ros2_medkit_integration_tests/test/features/test_thread_pool_bounds.test.py
bburda: Added `test_thread_pool_starvation`, which launches an under-budget pool (`http_thread_pool_size=2`, `sse.max_clients=2`) and asserts the guards the safety actually rests on rather than a config that can only pass: (a) the startup warning fires, and (b) the configured `executor_threads` is actually applied, read back from the gateway's startup log - so the executor bound is observed, not just set. The integration-test domain pool was widened to 229 to fit the extra test.

mfaferek93 (approved):

Issue #458, opened 2026-06-21
Fault bridges: default-config crash (action_status_bridge) + log_bridge self-noise


mfaferek93:
Two fault-bridge robustness issues found together.

## 1. action_status_bridge fails to start with its own default config

Same empty-untyped-list crash as #444 (log_bridge), but in `action_status_bridge` - the #444 fix did not cover this package.

`config/action_status_bridge.yaml` ships `exclude_actions: []` and `include_only_actions: []`. Empty YAML lists are untyped, so launching the node with this file aborts at startup:

```
terminate called after throwing an instance of 'rclcpp::exceptions::InvalidParameterValueException'
  what():  parameter_value_from failed for parameter 'exclude_actions': No parameter value set
[ros2run]: Aborted
```

The node code is correct (`declare_parameter<std::vector<std::string>>("exclude_actions", {})` / `include_only_actions` in `action_status_bridge_node.cpp`); only the shipped YAML breaks it. Reproduced live: with the shipped config the node aborts (exit 250); commenting the two lines out, it starts clean (`ActionStatusBridge started ... Watching action ...`).

**Fix:** comment out the two empty-list lines in `action_status_bridge.yaml` with non-empty examples - the same fix applied to log_bridge in #444 / PR #449.

## 2. log_bridge promotes the medkit stack's own logs (self-noise)

`log_bridge` only self-excludes its own node and `/log_bridge` (`log_bridge_node.cpp:98-99`: `source_id == own_node_name_ || source_id == "/log_bridge"`). It does **not** exclude the rest of the medkit stack, so when `fault_manager` (or the gateway) logs to `/rosout`, log_bridge promotes those lines into faults - medkit reporting on medkit, a noise/feedback loop.

Lower priority than the crash (robustness/noise, not a hard failure).

**Fix:** by default also skip the medkit stack's own nodes (e.g. `fault_manager`, the gateway) from promotion, or document a recommended `exclude_nodes` for them.

Issue #444, opened 2026-06-20
log_bridge fails to start with its own default config


mfaferek93:
`config/log_bridge.yaml` ships `exclude_nodes: []` and `include_only_nodes: []`. Empty YAML lists are untyped, so launching the node with this file fails (`parameter 'exclude_nodes' has no value` / type mismatch). The node code is correct (`declare_parameter<std::vector<std::string>>(..., {})` at `log_bridge_node.cpp:81-82`); the shipped YAML is what breaks it.

**Repro**
```
ros2 run ros2_medkit_log_bridge log_bridge_node --ros-args --params-file config/log_bridge.yaml
```

**Fix**
Remove the two empty-list lines from `log_bridge.yaml` (the code default already applies), or ship a non-empty example value.

PR #459, opened 2026-06-21
fix(action_status_bridge): start with the default config

mfaferek93:
Part 1 of #458. Same empty-untyped-list crash as #444 (log_bridge), in `action_status_bridge` - the #444 fix did not cover this package.

`config/action_status_bridge.yaml` ships `exclude_actions: []` and `include_only_actions: []`. Empty YAML lists are untyped, so launching the node with this file aborts at startup:

```
terminate called after throwing an instance of 'rclcpp::exceptions::InvalidParameterValueException'
  what():  parameter_value_from failed for parameter 'exclude_actions': No parameter value set
[ros2run]: Aborted
```

The node code is correct (`declare_parameter<std::vector<std::string>>(...)`); only the shipped YAML breaks it. This comments out the two lines with non-empty examples so the typed declared defaults apply.

Verified live: with the shipped config the node aborts (exit 250); with this fix it starts clean (`ActionStatusBridge started ... Watching action ...`).

Addresses #458 (part 1). The log_bridge self-noise item in #458 is left as a follow-up: its fix touches `log_bridge.yaml`, which #449 already modifies, so it is better done after #449 lands to avoid a conflict.

### DIFF
--- a/src/ros2_medkit_action_status_bridge/config/action_status_bridge.yaml
+++ b/src/ros2_medkit_action_status_bridge/config/action_status_bridge.yaml
@@ -13,8 +13,10 @@ action_status_bridge:
     # Normalized to UPPER_SNAKE at load.
     code_prefix: "ACTION"
     # Action-name substrings to skip (unanchored substring match).
-    exclude_actions: []
+    # Defaults to empty; uncomment with a non-empty list to enable, e.g.:
+    # exclude_actions: ["/some_action"]
     # If non-empty, ONLY watch actions matching these substrings.
-    include_only_actions: []
+    # Defaults to empty; uncomment with a non-empty list to enable, e.g.:
+    # include_only_actions: ["/navigate_to_pose"]
     # Max remembered (goal_id:status) keys for log dedup.
     dedup_capacity: 4096
--- a/src/ros2_medkit_action_status_bridge/test/test_integration.test.py
+++ b/src/ros2_medkit_action_status_bridge/test/test_integration.test.py
@@ -15,10 +15,12 @@
 
 """End-to-end integration tests for ros2_medkit_action_status_bridge."""
 
+import os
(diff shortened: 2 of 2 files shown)


bburda (commented):
Out-of-diff note:
- The packaged config is not loaded by any test, so this startup crash has no regression guard. `test/test_integration.test.py` launches the bridge with inline params (`parameters=[{'rescan_period_sec': 1.0}]`), and the launch file that loads `config/action_status_bridge.yaml` is never run by the suite (the GTest can't load a params file either). Pointing the integration test's node at the packaged config (`parameters=[config_file, {'rescan_period_sec': 1.0}]`, config first) would turn the existing `assertExitCodes` into a guard so an empty-untyped-list value can't silently bring the crash back.

mfaferek93:
Added the regression guard you suggested: the integration test now launches the bridge with parameters=[packaged config_file, {rescan_period_sec: 1.0}] (config first), so assertExitCodes turns into a guard - if an empty-untyped-list value ever comes back in the shipped action_status_bridge.yaml, the node aborts and the test fails. Verified the node starts clean with the config + override.

bburda (approved):
LGTM. Matches the #444 fix pattern: commenting out the empty untyped lists lets the node's typed `declare_parameter` defaults apply, so the packaged config no longer aborts at startup.

The integration test now loads the packaged config first (`parameters=[config_file, {rescan_period_sec: 1.0}]`), turning the existing `assertExitCodes` into a real regression guard. Verified the config values mirror the node's declared defaults exactly (`code_prefix=ACTION`, `aborted_severity=ERROR`, `heal_on_succeeded=true`, etc.), so loading the file doesn't change the existing test assertions, and no other shipped `config/*.yaml` carries the same empty-untyped-list pattern.

PR #460, opened 2026-06-21
fix(log_bridge): skip the medkit stack's own nodes by default

### LINKED ISSUE #458: Fault bridges: default-config crash (action_status_bridge) + log_bridge self-noise

mfaferek93:
Part 2 of #458 (part 1 is #459).

`log_bridge` only self-excluded its own node and `/log_bridge` (`log_bridge_node.cpp`), so when the rest of the medkit stack logs to `/rosout` - notably `fault_manager` logging a confirmed fault - log_bridge promotes those lines into new faults about medkit itself, a self-noise feedback loop.

This skips the medkit stack's own infrastructure nodes (`fault_manager`, `ros2_medkit_gateway`, `diagnostic_bridge`, `action_status_bridge`) by default, gated by a new `exclude_medkit_stack` param (default `true`; set `false` only to debug medkit's own logs). Substring match, so namespaced nodes (`/robot1/fault_manager`) are covered. Ordinary application nodes are unaffected.

Verified: `colcon build` clean, clang-format-18 clean, and 2 new regression tests pass (`NodeEligibility_ExcludesMedkitStackByDefault`, `NodeEligibility_MedkitStackExclusionDisablable`) alongside the existing eligibility tests (5/5).

Addresses #458 (part 2). Together with #459 this resolves #458.

### DIFF
--- a/src/ros2_medkit_log_bridge/README.md
+++ b/src/ros2_medkit_log_bridge/README.md
@@ -100,3 +100,10 @@ ros2 launch ros2_medkit_log_bridge log_bridge.launch.py
 against the node FQN: `planner` matches `/planner_server` and
 `/robot1/planner_server`. Use a longer, more specific substring (e.g.
 `/planner_server`) to avoid accidental matches.
+
+`exclude_medkit_stack` likewise matches **unanchored substrings** (on the raw
+logger name, so a namespaced `robot1.fault_manager` is still caught). A user
+node whose name contains one of these tokens - e.g. `fault_manager` or
+`diagnostic_bridge` - is therefore also skipped; set `exclude_medkit_stack:
+false` (and use `exclude_nodes` for the real medkit nodes) if that collides
+with your naming.
--- a/src/ros2_medkit_log_bridge/README.md
+++ b/src/ros2_medkit_log_bridge/README.md
@@ -94,6 +94,7 @@ ros2 launch ros2_medkit_log_bridge log_bridge.launch.py
 | `include_only_nodes` | `[]` | if set, only promote nodes whose FQN matches |
 | `max_tracked_nodes` | `512` | cap on per-node reporters; least-recently-used nodes evicted past this |
 | `report_cooldown_sec` | `5.0` | per-fault_code forward debounce; `0.0` disables |
+| `exclude_medkit_stack` | `true` | skip medkit's own infrastructure nodes (`fault_manager`, gateway, the other bridges) so their logs do not feed back as faults; set `false` to debug medkit's own logs |
 
 `exclude_nodes` / `include_only_nodes` match as **unanchored substrings**
 against the node FQN: `planner` matches `/planner_server` and
--- a/src/ros2_medkit_log_bridge/config/log_bridge.yaml
+++ b/src/ros2_medkit_log_bridge/config/log_bridge.yaml
@@ -21,3 +21,7 @@ log_bridge:
     # forwarded immediately; the same code within the window is suppressed.
     # 0.0 disables. Tames ERROR/FATAL floods, which bypass the per-node filter.
     report_cooldown_sec: 5.0
+    # Skip the medkit stack's own infrastructure nodes (fault_manager, gateway,
+    # the other bridges). Their /rosout lines would otherwise feed back as
+    # faults about medkit itself. Set false only to debug medkit's own logs.
+    exclude_medkit_stack: true
--- a/src/ros2_medkit_log_bridge/src/log_bridge_node.cpp
+++ b/src/ros2_medkit_log_bridge/src/log_bridge_node.cpp
@@ -88,6 +88,7 @@ void LogBridgeNode::load_parameters() {
   if (report_cooldown_sec_ < 0.0) {
     report_cooldown_sec_ = 0.0;
   }
+  exclude_medkit_stack_ = declare_parameter<bool>("exclude_medkit_stack", true);
 }
 
 void LogBridgeNode::log_callback(const rcl_interfaces::msg::Log::ConstSharedPtr & msg) {
@@ -148,6 +149,16 @@ std::string LogBridgeNode::node_source_id(const std::string & log_name) {
 }
 
 bool LogBridgeNode::node_is_eligible(const std::string & source_id) const {
+  // Skip the medkit stack's own infrastructure nodes by default: promoting
+  // their /rosout lines would report faults about medkit itself (e.g. the
+  // fault_manager logging a confirmed fault feeds back as a new fault).
+  if (exclude_medkit_stack_) {
+    static const std::vector<std::string> kMedkitStack = {"fault_manager", "ros2_medkit_gateway", "diagnostic_bridge",
+                                                          "action_status_bridge"};
+    if (contains_substr(kMedkitStack, source_id)) {
+      return false;
+    }
+  }
   if (!include_only_nodes_.empty() && !contains_substr(include_only_nodes_, source_id)) {
     return false;
   }
--- a/src/ros2_medkit_log_bridge/src/log_bridge_node.cpp
+++ b/src/ros2_medkit_log_bridge/src/log_bridge_node.cpp
@@ -103,7 +103,7 @@ void LogBridgeNode::log_callback(const rcl_interfaces::msg::Log::ConstSharedPtr 
   // Skip the medkit stack's own infrastructure nodes (else fault_manager's own
   // /rosout lines feed back as faults about medkit). Matched on the raw logger
   // name, which keeps the namespace (node_source_id collapses it to the ns).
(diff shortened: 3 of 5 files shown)


src/ros2_medkit_log_bridge/src/log_bridge_node.cpp
```diff
@@ -148,6 +149,16 @@ std::string LogBridgeNode::node_source_id(const std::string & log_name) {
...
 bool LogBridgeNode::node_is_eligible(const std::string & source_id) const {
+  // Skip the medkit stack's own infrastructure nodes by default: promoting
+  // their /rosout lines would report faults about medkit itself (e.g. the
+  // fault_manager logging a confirmed fault feeds back as a new fault).
+  if (exclude_medkit_stack_) {
+    static const std::vector<std::string> kMedkitStack = {"fault_manager", "ros2_medkit_gateway", "diagnostic_bridge",
+                                                          "action_status_bridge"};
+    if (contains_substr(kMedkitStack, source_id)) {
+      return false;
+    }
```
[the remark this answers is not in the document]
mfaferek93: Good catch - fixed. The check now runs against the raw logger name (msg->name) via a static is_medkit_stack_logger, before node_source_id collapses a namespaced logger to its namespace. So "robot1.fault_manager" is now matched on the "fault_manager" substring.

src/ros2_medkit_log_bridge/test/test_log_bridge.cpp
```diff
@@ -219,6 +219,23 @@ TEST_F(LogBridgeTest, NodeEligibility_IncludeOnlySubstring) {
   EXPECT_FALSE(node->node_is_eligible("/amcl"));
 }
 
+TEST_F(LogBridgeTest, NodeEligibility_ExcludesMedkitStackByDefault) {
+  auto node = make_node_with({});
+  // medkit's own infrastructure must not feed its own logs back as faults
+  EXPECT_FALSE(node->node_is_eligible("/fault_manager"));
+  EXPECT_FALSE(node->node_is_eligible("/robot1/fault_manager"));
+  EXPECT_FALSE(node->node_is_eligible("/ros2_medkit_gateway"));
```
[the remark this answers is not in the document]
mfaferek93: Fixed - the tests now exercise the real path: they call is_medkit_stack_logger with the actual logger-name forms, including the namespaced "robot1.fault_manager", so the namespaced case is genuinely covered.

src/ros2_medkit_log_bridge/config/log_bridge.yaml
```diff
@@ -21,3 +21,7 @@ log_bridge:
     # forwarded immediately; the same code within the window is suppressed.
     # 0.0 disables. Tames ERROR/FATAL floods, which bypass the per-node filter.
     report_cooldown_sec: 5.0
+    # Skip the medkit stack's own infrastructure nodes (fault_manager, gateway,
+    # the other bridges). Their /rosout lines would otherwise feed back as
+    # faults about medkit itself. Set false only to debug medkit's own logs.
+    exclude_medkit_stack: true
```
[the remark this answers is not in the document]
mfaferek93: Added - the log_bridge README config table now lists exclude_medkit_stack (default true) with its meaning.

src/ros2_medkit_log_bridge/test/test_log_bridge.cpp
```diff
@@ -219,6 +219,23 @@ TEST_F(LogBridgeTest, NodeEligibility_IncludeOnlySubstring) {
   EXPECT_FALSE(node->node_is_eligible("/amcl"));
 }
 
+// Matched on the raw logger name (msg->name), which is what log_callback feeds
+// it - including the namespaced form "robot1.fault_manager" that node_source_id
+// would otherwise collapse to "/robot1".
+TEST_F(LogBridgeTest, MedkitStackLogger_MatchesOwnInfra) {
```
bburda, line 225: Both new tests call the static `is_medkit_stack_logger`, which never reads `exclude_medkit_stack`. The param toggle is untested: nothing exercises the default-true gate in `log_callback` or the disable path (`exclude_medkit_stack=false`). A small instance predicate (like `node_is_eligible`) would let a `make_node_with({rclcpp::Parameter("exclude_medkit_stack", false)})` test assert the toggle, mirroring `Cooldown_ZeroDisables`.

src/ros2_medkit_log_bridge/README.md
```diff
@@ -94,6 +94,7 @@ ros2 launch ros2_medkit_log_bridge log_bridge.launch.py
 | `include_only_nodes` | `[]` | if set, only promote nodes whose FQN matches |
 | `max_tracked_nodes` | `512` | cap on per-node reporters; least-recently-used nodes evicted past this |
 | `report_cooldown_sec` | `5.0` | per-fault_code forward debounce; `0.0` disables |
+| `exclude_medkit_stack` | `true` | skip medkit's own infrastructure nodes (`fault_manager`, gateway, the other bridges) so their logs do not feed back as faults; set `false` to debug medkit's own logs |
```
bburda, line 97: `is_medkit_stack_logger` matches an unanchored substring on the raw logger name against a hardcoded list, so a user node whose name contains `fault_manager` or `diagnostic_bridge` is also skipped and never becomes a fault. The README documents this substring caveat for `exclude_nodes`/`include_only_nodes` but not here. Either add the note, or anchor the match to the node segment (still catching `robot1.fault_manager`).

bburda (commented):
- The PR description lists the new tests as `NodeEligibility_ExcludesMedkitStackByDefault` and `NodeEligibility_MedkitStackExclusionDisablable`, but the diff adds `MedkitStackLogger_MatchesOwnInfra` and `MedkitStackLogger_AllowsApplicationNodes`. The default and disable tests named there aren't present.

src/ros2_medkit_log_bridge/test/test_log_bridge.cpp
mfaferek93: Added - a new instance predicate skips_medkit_stack() applies the exclude_medkit_stack gate to is_medkit_stack_logger and is used in log_callback; two tests now assert the default-on (MedkitStack_SkippedByDefault) and disabled (MedkitStack_ToggleDisables) paths, mirroring Cooldown_ZeroDisables.

src/ros2_medkit_log_bridge/README.md
mfaferek93: Documented - the README now notes exclude_medkit_stack matches unanchored substrings (on the raw logger name, so robot1.fault_manager is caught), and that a user node whose name contains one of these tokens is also skipped; set exclude_medkit_stack: false if it collides.

mfaferek93:
On the review note about test names: the PR description was stale - the actual tests are MedkitStackLogger_MatchesOwnInfra / _AllowsApplicationNodes (the matcher, incl. namespaced robot1.fault_manager) plus the new MedkitStack_SkippedByDefault / MedkitStack_ToggleDisables (the exclude_medkit_stack gate). Description updated to match.

bburda (approved):

PR #461, opened 2026-06-22
fix(launch): enable web UI CORS by default in the native gateway launch

mfaferek93:
The native `gateway.launch.py` / `bringup.launch.py` ran the gateway with CORS off (the gateway's `cors.allowed_origins` defaults to empty). So the documented path of running the web UI next to the gateway fails with "Failed to fetch" on a native bringup - the browser calls the gateway from a different origin and is blocked. The image CORS fix (#452) only covered `docker/gateway_docker_params.yaml`, not the native launch.

This adds a `cors_allowed_origins` launch arg (comma-separated) defaulting to the web UI origins (`http://localhost:3000,http://localhost:5173`), resolved into the gateway's `cors.allowed_origins` via an OpaqueFunction. Empty disables CORS. A wildcard is intentionally not the default (with auth off and write methods enabled it would let any site drive cross-origin writes), matching #452.

Verified live: launching `gateway.launch.py` logs `CORS enabled - origins: [http://localhost:3000, http://localhost:5173], credentials: false`; `curl -H "Origin: http://localhost:3000"` gets `Access-Control-Allow-Origin: http://localhost:3000`; `curl -H "Origin: http://evil.com"` gets no ACAO header (blocked). flake8 clean.

### DIFF
--- a/src/ros2_medkit_gateway/launch/gateway.launch.py
+++ b/src/ros2_medkit_gateway/launch/gateway.launch.py
@@ -18,7 +18,7 @@ from ament_index_python.packages import get_package_prefix
 from ament_index_python.packages import get_package_share_directory
 from ament_index_python.packages import PackageNotFoundError
 from launch import LaunchDescription
-from launch.actions import DeclareLaunchArgument
+from launch.actions import DeclareLaunchArgument, OpaqueFunction
 from launch.substitutions import LaunchConfiguration
 from launch_ros.actions import Node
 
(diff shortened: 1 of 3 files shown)


src/ros2_medkit_gateway/launch/gateway.launch.py
```diff
@@ -64,28 +64,44 @@ def generate_launch_description():
...
-        package='ros2_medkit_gateway',
-        executable='gateway_node',
-        name='ros2_medkit_gateway',
-        output='screen',
-        parameters=[default_config, LaunchConfiguration('config_file'), param_overrides],
-        arguments=['--ros-args', '--log-level', 'info'])
+    # Resolve the cors arg into a string list at launch time (a comma-separated
+    # LaunchConfiguration cannot be passed straight through as a string-array
+    # parameter). param_overrides is applied last, so cors_allowed_origins is the
+    # single override point for CORS; set it empty for the secure (off) default.
```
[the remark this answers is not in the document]
mfaferek93: Fixed the comment - it now says the default is non-empty (CORS on for the web UI) and that setting it empty disables CORS.

src/ros2_medkit_gateway/launch/gateway.launch.py
```diff
@@ -64,28 +64,50 @@ def generate_launch_description():
...
+        # gateway config uses; config.cpp filters the empty string, so CORS ends
+        # up off without the crash.
+        origins = [o.strip() for o in
+                   LaunchConfiguration('cors_allowed_origins').perform(context).split(',')
+                   if o.strip()] or ['']
+        param_overrides = {
+            'server.host': LaunchConfiguration('server_host'),
+            'server.port': LaunchConfiguration('server_port'),
+            'refresh_interval_ms': LaunchConfiguration('refresh_interval_ms'),
+            'cors.allowed_origins': origins,
```
bburda: `cors.allowed_origins` is added to `param_overrides`, the last entry in `parameters=[default_config, config_file, param_overrides]`, so it wins per key. A `config_file` that sets `cors.allowed_origins` is now silently overridden by the launch default (`localhost:3000,5173`); on current main `param_overrides` did not touch CORS, so `config_file` controlled it. Inject `cors.allowed_origins` only when `cors_allowed_origins` differs from its declared default, so an explicit launch arg still wins but a `config_file` value is respected when no arg is passed.

bburda (commented):
- `gateway_https.launch.py` also launches `gateway_node` but sets no `cors.allowed_origins`, so it keeps the `[""]` default from `gateway_params.yaml`. The HTTPS dev path still serves CORS off, so the web UI hits the same "Failed to fetch" this fixes for the HTTP path. Worth a follow-up to give it the same arg and default.

src/ros2_medkit_gateway/launch/gateway.launch.py
mfaferek93: Fixed. cors.allowed_origins is no longer unconditionally in param_overrides: a new cors_override() injects it only when the cors_allowed_origins arg is explicitly set (it then wins) or when the default config is in use (web UI default). With no arg and a custom config_file, nothing is injected, so the file's cors.allowed_origins is respected. Verified live: config_file with [http://custom.example] + no arg -> gateway reports [http://custom.example], localhost:3000 blocked (launch default no longer overrides); explicit arg still wins.

mfaferek93:
On the gateway_https.launch.py finding: applied the same fix there - it now declares cors_allowed_origins with the same default and folds it in via cors_override, so the HTTPS dev path serves the web UI out of the box instead of CORS-off. Verified: gateway_https.launch.py comes up on TLS (curl -k -> 200) with CORS enabled for localhost:3000/5173.

bburda (approved):

PR #462, opened 2026-06-22
Incremental embedded-hardened entity cache + bounded discovery allocation under churn (#442)

### LINKED ISSUE #442: Incremental EntityCache refresh instead of full rebuild each tick
## Summary

Each discovery refresh rebuilds the whole entity cache.

`GatewayNode::refresh_entity_cache()` (`src/ros2_medkit_gateway/src/gateway_node.cpp`, around lines 1885-2081) runs the discovery and merge pipeline to build fresh area, component, app and function vectors, then calls `ThreadSafeEntityCache::update_all(...)`. `update_all` replaces all primary vectors and calls `rebuild_all_indexes()`, which rebuilds all seven hash maps on every tick. There is no diff or change detection.

The refresh interval is the main CPU lever. A 200 ms refresh costs about 4x the CPU of the 1000 ms default, while memory differences across configs stay within a few MiB. Measured with a benchmark harness in selfpatch_demos (sweep lane).

## Proposed solution

- Compare the new entity set against the cached one and apply only the changes (add, remove, update), instead of rebuilding and swapping every refresh.
- Skip the work when nothing changed. On a stable graph most refresh ticks produce the same entities.

## Additional context

Full rebuild every tick is wasted work on a stable graph. The cost scales with graph size and refresh frequency, which is why a fast refresh is expensive.

bburda:
# Pull Request

## Summary

Rework the gateway entity cache for incremental, embedded-friendly refresh and
bound the discovery pipeline's per-graph-event allocation under graph churn.

**Entity cache (object pool + flat indexes):**
- `ThreadSafeEntityCache` now stores entities in a fixed-capacity `SlotStore`
  object pool with stable slots and a free-list, indexed by pre-reserved
  open-addressed `FlatHashMap`s instead of `std::vector` + `std::unordered_map`.
- `update_all` reconciles the incoming discovery output against the current
  state by id (add / remove / change only) and skips work entirely when nothing
  changed; `generation` advances only on a real change.
- Capacity is reserved at startup via `entity_cache.capacity` (default 256,
  clamped); steady-state refresh performs zero structural allocations in the
  cache layer. Cache stats are exposed on `/health` as `x-medkit-entity-cache`,
  with a one-shot WARN if capacity is exceeded.

**Discovery (bound per-graph-event allocation):**
- Operation `type_info` schemas are resolved lazily by the `/operations` handler
  and cached as shared, immutable per-type objects in `TypeIntrospection`, so
  they are no longer rebuilt and deep-copied on every discovery refresh.
- Graph-event refreshes are debounced via `discovery.refresh_debounce_ms`
  (default 1000 ms) so bursts of graph changes coalesce into one refresh.

Net effect under graph churn: gateway CPU drops ~4x (0.45 -> 0.11 cores) and the
gateway-side per-refresh allocation is largely eliminated.

## Issue

- Closes #442

(Not "closes" - see "Memory growth under churn: what this does and does not fix"
below. The gateway-side contribution is addressed here; the dominant residual is
a DDS-middleware property, outside this repo's scope.)

## Memory growth under churn: what this does and does not fix

The `benchmark churn` gate measures total-process USS slope under node churn.
Per-allocation profiling (operator-new counters per refresh phase) and isolation
experiments attribute the growth as follows:

- The entity-cache rebuild was **not** the leak. An in-repo allocation-counter
  test proves `update_all` does zero structural allocations under steady-state
  churn, yet removing all cache/refresh allocation left the USS slope unchanged.
- ~50% of the growth is **glibc multi-arena retention** (freed memory retained
  per-arena, and the arena count scales with thread count). This is addressed by
  bounding the executor / HTTP thread pools (separate work), which lowers the
  arena count; no allocator tuning is added here.
- Resource handles are **not** leaked: under churn the gateway's open file
  descriptors stay flat (18) and thread count stays flat (50), and the
  subscription count does not grow. The growth is allocation churn, not a
  pub/sub / handle leak.
- ~50% is **DDS participant-churn growth in the middleware**, not the gateway: a
  bare rclpy node (no gateway code) leaks ~24 KB/s under FastDDS and ~31 KB/s
  under CycloneDDS, and ~24 KB/s even when node names are reused. It is the
  participant create/destroy churn itself, independent of the gateway and of the
  RMW implementation.

So the gate's threshold is not reachable by any ROS 2 process under this churn
because the DDS layer alone exceeds it. The gateway-side improvements here (CPU,
cache structure, lazy schemas, debounce) are real and worthwhile; the residual
memory behaviour is a middleware concern (the RMW is swappable).

Long-run trajectory (bare node, FastDDS, ~40 min): net-growing but decelerating (first third ~64 KB/s, last third ~12 KB/s) with a periodic DDS lease-purge sawtooth (~8 MB reclaimed around the 26-minute mark); 0.4 MB -> 61.5 MB over 42 min, not fully plateaued. The gate's 10-minute window captures only the steep initial ramp, so it overstates the steady-state slope. This is a bare rclpy node (no gateway code), i.e. purely the DDS layer.

## Type

- [x] New feature or tests
- [x] Bug fix (performance / allocation under churn)
- [ ] Breaking change
- [ ] Documentation only

## Testing

- Full gateway unit suite green (2327 tests, 0 failures).
- New unit tests: `FlatHashMap`, `SlotStore`, entity `operator==`,
  `ThreadSafeEntityCache` incremental reconcile, an isolated allocation-counter
  gate (proves zero structural allocations under steady-state churn), and
  `TypeIntrospection` shared service/action `type_info` caching.
- Generation-counter tests updated to the change-gated contract.
- clang-tidy clean on changed files; gateway_core ROS-free purity preserved;
  Sphinx docs build clean.
- Docker `benchmark churn` gate exercised; findings summarised above.

## Checklist

- [x] Breaking changes are clearly described (no breaking API changes; entity
      detail `operations[]` no longer eagerly embeds schemas - schemas remain on
      the `/operations` resource)

### DIFF
--- a/docs/config/server.rst
+++ b/docs/config/server.rst
@@ -259,6 +259,15 @@ Performance Tuning
      - Safety-backstop refresh interval (ms). Primary discovery is graph-event
        driven (polled every 100 ms); this timer forces a full refresh only if
        a graph event is missed. Range: 100-60000 (0.1s-60s).
+   * - ``entity_cache.capacity``
+     - int
+     - ``256``
+     - Fixed capacity of the thread-safe entity cache object pool, reserved once
+       at startup. Steady-state refresh cycles perform zero structural allocations
+       in the cache layer as long as the live entity count stays within this
+       value. Valid range: 16-1000000 (values outside this range are clamped with
+       a warning). Raise it for graphs larger than the default; exceeding the
+       reserved capacity is harmless but triggers a one-shot WARN log.
 
 Lower values shorten the worst-case recovery window if a graph event is missed
 but increase idle CPU. The default rarely fires on a stable graph because the
--- a/docs/config/server.rst
+++ b/docs/config/server.rst
@@ -267,7 +267,8 @@ Performance Tuning
        full discovery pipeline. Coalesces graph events so a refresh runs at most
        once per this interval, bounding per-event allocation. A pending change is
        serviced within this interval plus one 100 ms poll. Range: 0-60000
-       (0 disables debouncing; clamped with a warning if out of range).
+       (0 disables debouncing). A value outside the range is rejected with a
+       warning and the default (1000) is used.
    * - ``entity_cache.capacity``
      - int
      - ``256``
@@ -276,7 +277,11 @@ Performance Tuning
        in the cache layer as long as the live entity count stays within this
        value. Valid range: 16-1000000 (values outside this range are clamped with
        a warning). Raise it for graphs larger than the default; exceeding the
-       reserved capacity is harmless but triggers a one-shot WARN log.
+       reserved capacity is harmless but triggers a one-shot WARN log. Note that a
+       refresh that fully turns the entity set over allocates the new slots before
+       freeing the absent old ones, so the pool transiently holds up to ~2x the
+       steady-state live count - size capacity to about twice the expected peak to
+       keep full-turnover ticks allocation-free.
 
 Lower values shorten the worst-case recovery window if a graph event is missed
 but increase idle CPU. The default rarely fires on a stable graph because the
--- a/docs/config/server.rst
+++ b/docs/config/server.rst
@@ -259,6 +259,15 @@ Performance Tuning
      - Safety-backstop refresh interval (ms). Primary discovery is graph-event
        driven (polled every 100 ms); this timer forces a full refresh only if
        a graph event is missed. Range: 100-60000 (0.1s-60s).
+   * - ``discovery.refresh_debounce_ms``
+     - int
+     - ``1000``
+     - Debounce for graph-event-driven refreshes (ms). Under graph churn the
+       rclcpp graph event fires many times per second; each refresh runs the
+       full discovery pipeline. Coalesces graph events so a refresh runs at most
+       once per this interval, bounding per-event allocation. A pending change is
+       serviced within this interval plus one 100 ms poll. Range: 0-60000
+       (0 disables debouncing; clamped with a warning if out of range).
    * - ``entity_cache.capacity``
      - int
      - ``256``
--- a/src/ros2_medkit_gateway/include/ros2_medkit_gateway/core/util/flat_hash_map.hpp
+++ b/src/ros2_medkit_gateway/include/ros2_medkit_gateway/core/util/flat_hash_map.hpp
@@ -78,10 +78,16 @@ class FlatHashMap {
   // -------------------------------------------------------------------------
 
   /// Ensure the table can hold at least `capacity` entries at max_load.
+  ///
+  /// reserve() establishes the baseline capacity and is NOT a structural growth:
+  /// the post-reserve grew_ flag is cleared so grew() reports only reallocations
+  /// that occur because the live set later outgrew the reserved capacity. This
+  /// matches SlotStore, whose reserve() likewise never sets grew().
   void reserve(size_t capacity) {
     size_t needed = next_pow2(static_cast<size_t>(static_cast<double>(capacity) / kMaxLoad + 1.0));
     if (needed > table_.size()) {
       rehash(needed);
+      grew_ = false;  // baseline allocation, not a growth
     }
   }
 
@@ -348,22 +354,37 @@ class FlatHashMap {
     return p;
   }
 
-  /// Grow to at least double current capacity if (occupied+deleted) load is too high.
-  /// O(1): uses the tracked `used_` counter rather than scanning the table.
+  /// Keep the table below max load. When (occupied+deleted) load is too high we
+  /// distinguish two causes, which matters for steady-state add/remove churn:
+  ///   - If the LIVE entries (size_) still fit at the current capacity, the load
+  ///     is driven by accumulated tombstones - rehash to the SAME capacity to
+  ///     vacuum them. This does not grow the table, so grew() stays false and
+  ///     churn over a bounded set never reallocates.
+  ///   - Otherwise the live set genuinely outgrew the table - double capacity.
+  /// O(1): uses the tracked size_/used_ counters rather than scanning the table.
   void ensure_capacity() {
     if (table_.empty()) {
       rehash(4);
       return;
     }
-    if (static_cast<double>(used_ + 1) > kMaxLoad * static_cast<double>(table_.size())) {
-      rehash(table_.size() * 2);
+    const double cap = static_cast<double>(table_.size());
+    if (static_cast<double>(used_ + 1) <= kMaxLoad * cap) {
+      return;  // within load - nothing to do
+    }
+    if (static_cast<double>(size_ + 1) <= kMaxLoad * cap) {
+      rehash(table_.size());  // tombstone-driven: vacuum in place, no growth
+    } else {
+      rehash(table_.size() * 2);  // live set outgrew capacity: grow
     }
   }
 
   /// Rehash into a new table of the given size (must be power-of-two).
-  /// Vacuums tombstones (used_ resets to size_ after rehash). Sets grew_ = true.
+  /// Vacuums tombstones (used_ resets to size_ after rehash). Sets grew_ only
+  /// when the new capacity exceeds the old one - a same-size vacuum preserves
+  /// capacity and must not be reported as a structural growth.
   void rehash(size_t new_cap) {
     new_cap = next_pow2(new_cap);
+    const bool grew_capacity = new_cap > table_.size();
     std::vector<Entry> new_table(new_cap);
     for (Entry & old_e : table_) {
       if (old_e.state != kOccupied) {
@@ -380,7 +401,9 @@ class FlatHashMap {
     }
     table_ = std::move(new_table);
     used_ = size_;  // tombstones vacuumed
-    grew_ = true;
+    if (grew_capacity) {
+      grew_ = true;
+    }
   }
 
   // -------------------------------------------------------------------------
--- a/src/ros2_medkit_gateway/include/ros2_medkit_gateway/core/util/flat_hash_map.hpp
+++ b/src/ros2_medkit_gateway/include/ros2_medkit_gateway/core/util/flat_hash_map.hpp
@@ -1,0 +1,395 @@
+// Copyright 2026 bburda
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+#pragma once
+
+#include <cstddef>
+#include <cstdint>
+#include <functional>
+#include <string>
+#include <vector>
+
+namespace ros2_medkit_gateway {
+
+// ---------------------------------------------------------------------------
+// flat_reset_value: clears container-valued entries in place (retains capacity)
+// Generic overload is a no-op (scalars, integral types, etc.)
+// ---------------------------------------------------------------------------
+
+template <typename V>
+inline void flat_reset_value(V & /*value*/) noexcept {
+}
+
+template <typename T, typename A>
+inline void flat_reset_value(std::vector<T, A> & value) noexcept {
+  value.clear();
+}
+
+inline void flat_reset_value(std::string & value) noexcept {
+  value.clear();
+}
+
+// ---------------------------------------------------------------------------
+// FlatHashMap<K,V,Hash>
+//
+// Open-addressed hash table with linear probing, tombstone deletion, and
+// ZERO structural allocation in steady state after reserve().
+//
+// Invariants:
+//   - Table size is always a power of two.
+//   - max_load = 0.70: (occupied + deleted) / capacity triggers a grow.
+//   - On grow, a brand-new table is allocated, all OCCUPIED entries rehashed,
+//     tombstones are vacuumed, and grew_ is set to true.
+//   - reset() retains the backing array (no deallocation) and calls
+//     flat_reset_value() on each entry's value so container-valued maps
+//     retain their heap buffers across refills.
+//   - used_ tracks (occupied + deleted) in O(1) so ensure_capacity() never
+//     scans the table.
+// ---------------------------------------------------------------------------
+
+template <typename K, typename V, typename Hash = std::hash<K>>
+class FlatHashMap {
+ public:
+  FlatHashMap() = default;
+
+  explicit FlatHashMap(size_t capacity) {
+    reserve(capacity);
+  }
+
+  ~FlatHashMap() = default;
+  FlatHashMap(const FlatHashMap &) = default;
+  FlatHashMap & operator=(const FlatHashMap &) = default;
+  FlatHashMap(FlatHashMap &&) noexcept = default;
+  FlatHashMap & operator=(FlatHashMap &&) noexcept = default;
+
+  // -------------------------------------------------------------------------
+  // Capacity management
+  // -------------------------------------------------------------------------
+
+  /// Ensure the table can hold at least `capacity` entries at max_load.
+  void reserve(size_t capacity) {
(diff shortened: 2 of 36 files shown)


src/ros2_medkit_gateway/include/ros2_medkit_gateway/core/util/slot_store.hpp
```diff
@@ -0,0 +1,161 @@
...
+  void assign(uint32_t slot, T && value) {
+    items_[static_cast<size_t>(slot)] = std::move(value);
+    if (live_[static_cast<size_t>(slot)] == 0u) {
+      live_[static_cast<size_t>(slot)] = 1u;
+      ++live_count_;
+    }
+  }
+
+  /// Release the slot: reset payload to T{}, mark dead, push to free-list.
+  void free(uint32_t slot) {
```
mfaferek93, line 94: free() has no liveness guard: a double-free underflows live_count_ (after which collect_live() reserve()s ~SIZE_MAX) and pushes the slot onto the free-list twice, so two later alloc_slot()s return the same slot. reconcile() frees each id once today so it's not hit, but a pool primitive whose double-free corrupts state is fragile. Suggest `if (live_[slot] == 0u) return;` up front to make it idempotent.

src/ros2_medkit_gateway/include/ros2_medkit_gateway/core/util/slot_store.hpp
```diff
@@ -0,0 +1,161 @@
...
+      grew_ = true;
+    }
+    return static_cast<uint32_t>(items_.size() - 1u);
+  }
+
+  /// Copy value into the slot and mark it live. Assigning over an already-live
+  /// slot updates the payload in place and does NOT change live_count_; only a
+  /// dead -> live transition bumps the count. This makes assign() safe for the
+  /// incremental-reconcile update path that overwrites existing live slots.
+  void assign(uint32_t slot, const T & value) {
```
mfaferek93, line 79: assign() on a dead (free-listed) slot flips it live but never removes it from the free-list, so the next alloc_slot() hands the same slot to a second entity. Current callers only assign to slots from alloc_slot(), but the dead->live path the doc sanctions can alias a slot. Suggest assert(live_[slot] != 0u) here (use alloc_slot() for new slots), or swap-remove from the free-list when reviving.

src/ros2_medkit_gateway/src/gateway_node.cpp
```diff
@@ -354,6 +361,7 @@ GatewayNode::GatewayNode(const rclcpp::NodeOptions & options) : Node("ros2_medki
   server_host_ = get_parameter("server.host").as_string();
   server_port_ = static_cast<int>(get_parameter("server.port").as_int());
   refresh_interval_ms_ = static_cast<int>(get_parameter("refresh_interval_ms").as_int());
+  refresh_debounce_ms_ = static_cast<int>(get_parameter("discovery.refresh_debounce_ms").as_int());
```
mfaferek93: This narrows the int64 param to int before the 0..60000 check at line 407, so a value above INT_MAX can wrap into range and bypass the clamp (e.g. 4294967296 -> 0, silently disabling debounce). entity_cache.capacity already does it right: clamp<int64_t> first, then cast.

docs/config/server.rst
```diff
@@ -259,6 +259,24 @@ Performance Tuning
...
        a graph event is missed. Range: 100-60000 (0.1s-60s).
+   * - ``discovery.refresh_debounce_ms``
+     - int
+     - ``1000``
+     - Debounce for graph-event-driven refreshes (ms). Under graph churn the
+       rclcpp graph event fires many times per second; each refresh runs the
+       full discovery pipeline. Coalesces graph events so a refresh runs at most
+       once per this interval, bounding per-event allocation. A pending change is
+       serviced within this interval plus one 100 ms poll. Range: 0-60000
+       (0 disables debouncing; clamped with a warning if out of range).
```
mfaferek93: "clamped" is inaccurate here: the implementation resets to the default 1000 on out-of-range (gateway_node.cpp:411), so 70000 becomes 1000, not 60000. Suggest "values outside the range are rejected with a warning and the default (1000) is used". Keep "clamped" for entity_cache.capacity, which genuinely uses std::clamp.

src/ros2_medkit_serialization/src/type_introspection.cpp
```diff
@@ -85,4 +85,44 @@ TopicTypeInfo TypeIntrospection::get_type_info(const std::string & type_name) {
...
 
+std::shared_ptr<const nlohmann::json> TypeIntrospection::get_service_type_info(const std::string & service_type) {
+  {
+    std::lock_guard<std::mutex> lock(cache_mutex_);
+    auto it = service_type_info_cache_.find(service_type);
+    if (it != service_type_info_cache_.end()) {
+      return it->second;
+    }
+  }
+  // get_type_info is itself cached + throws on failure; build the wrapper once.
```
mfaferek93: get_type_info() does not throw: on a failed/transient resolve it returns an empty schema. That empty wrapper is then try_emplace'd permanently below, so a type not yet resolvable on the first /operations request stays empty for the process lifetime. Suggest only memoizing a fully-resolved wrapper and letting empties fall through. (The comment is also backwards - it never throws, so the call-site try/catch is dead.)

src/ros2_medkit_gateway/include/ros2_medkit_gateway/core/util/flat_hash_map.hpp
```diff
@@ -0,0 +1,418 @@
...
+  size_t slot(uint64_t h) const {
+    return static_cast<size_t>(h) & (table_.size() - 1);
+  }
+
+  static size_t next_pow2(size_t n) {
+    if (n == 0) {
+      return 1;
+    }
+    size_t p = 1;
+    while (p < n) {
```
mfaferek93, line 357: next_pow2() has no overflow guard: for n > 2^63, p <<= 1 wraps to 0 and this loop spins forever. reserve() feeds it capacity/kMaxLoad. Capacity is config-clamped today, but this is a general primitive - guard with `if (n > (size_t(1) << 63)) return n;` or use std::bit_ceil.

src/ros2_medkit_gateway/src/gateway_node.cpp
```diff
@@ -1425,9 +1459,28 @@ GatewayNode::GatewayNode(const rclcpp::NodeOptions & options) : Node("ros2_medki
...
-    if (!graph_event_->check_and_clear()) {
+    // Coalesce graph events: under graph churn (nodes/topics created and
+    // destroyed repeatedly) the event fires many times per second, and each
+    // refresh_cache() runs the full discovery pipeline. Debounce so we run at
+    // most one refresh per `refresh_debounce_ms` (default 1000 ms) - the ROS
+    // graph rarely changes faster than that in practice, and coalescing
+    // bounds the per-event allocation churn (issue #442). A pending event is
+    // always serviced within refresh_debounce_ms + one 100 ms tick.
+    if (graph_event_->check_and_clear()) {
+      graph_dirty_.store(true, std::memory_order_relaxed);
```
mfaferek93: The debounce path (dirty latch, coalescing window, the 0ms disable, and cold-start fire-immediately) has no test coverage anywhere. A regression that drops the latch would be invisible. Suggest factoring the decision into a pure helper and unit-testing coalescing / 0ms / cold-start, or an integration test toggling the graph faster than the window and asserting at most one refresh per window.

src/ros2_medkit_gateway/test/test_entity_cache_alloc.cpp
```diff
@@ -0,0 +1,119 @@
...
+  c.update_all({}, host, ticks[1], {});
+  g_net.store(0);
+  g_armed.store(true);
+  for (std::size_t i = 2; i < kTicks; ++i) {
+    c.update_all({}, std::vector<Component>{}, std::move(ticks[i]), {});
+  }
+  g_armed.store(false);
+  // Only update_all ran while armed; no copies, no container mutation, no getter calls.
+  const long kSlack = 4;  // tiny; tolerates allocator bookkeeping noise only. Do NOT inflate.
+  EXPECT_LE(g_net.load(), kSlack) << "cache layer allocated under steady-state churn (regression to rebuild?)";
```
mfaferek93, line 150: This gate can pass vacuously: the armed churn uses all-SSO ids and empty vectors, so nothing on the reconcile path can allocate, and there is no positive control proving the new/delete interposer is live. Suggest one populated/non-SSO tick (long ids + a few service/action full_paths so rebuild_operation_index copies real keys) plus an arming check that EXPECT_GT(g_net, 0) on a deliberate allocation before this steady-state assert.

src/ros2_medkit_gateway/src/core/models/thread_safe_entity_cache.cpp
```diff
@@ -17,9 +17,169 @@
...
+    if (uint32_t * slot = index.find(e.id)) {
+      const uint32_t s = *slot;
+      seen_[s] = 1u;
+      // Compare against current payload; only assign (and flag) on a real diff.
+      if (!(store[s] == e)) {
+        store.assign(s, std::move(e));
+        changed = true;
+      }
+    } else {
+      const uint32_t s = store.alloc_slot();
```
mfaferek93, line 54: reconcile() allocates a slot for every new id here before the second loop frees absent old ids, and alloc_slot() can't reuse a slot freed later in the same pass. A full-turnover tick therefore transiently holds old+new slots (~2x peak live). Worth reflecting in the capacity-sizing / WARN guidance for the 256 default.

src/ros2_medkit_serialization/src/type_introspection.cpp
```diff
@@ -85,4 +85,44 @@ TopicTypeInfo TypeIntrospection::get_type_info(const std::string & type_name) {
...
+  auto assembled = std::make_shared<nlohmann::json>();
+  (*assembled)["request"] = get_type_info(service_type + "_Request").schema;
+  (*assembled)["response"] = get_type_info(service_type + "_Response").schema;
+  std::shared_ptr<const nlohmann::json> shared = std::move(assembled);
+  {
+    std::lock_guard<std::mutex> lock(cache_mutex_);
+    auto [it, inserted] = service_type_info_cache_.try_emplace(service_type, shared);
+    return it->second;
+  }
```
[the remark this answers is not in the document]
bburda: Fixed in 3e26d458: return `service_type_info_cache_.try_emplace(...).first->second` directly, dropping the unused `inserted` binding.

src/ros2_medkit_serialization/src/type_introspection.cpp
```diff
@@ -85,4 +85,44 @@ TopicTypeInfo TypeIntrospection::get_type_info(const std::string & type_name) {
...
+  auto assembled = std::make_shared<nlohmann::json>();
+  (*assembled)["goal"] = get_type_info(action_type + "_Goal").schema;
+  (*assembled)["result"] = get_type_info(action_type + "_Result").schema;
+  (*assembled)["feedback"] = get_type_info(action_type + "_Feedback").schema;
+  std::shared_ptr<const nlohmann::json> shared = std::move(assembled);
+  {
+    std::lock_guard<std::mutex> lock(cache_mutex_);
+    auto [it, inserted] = action_type_info_cache_.try_emplace(action_type, shared);
+    return it->second;
+  }
```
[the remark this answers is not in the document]
bburda: Fixed in 3e26d458: same change for `get_action_type_info` - return the try_emplace iterator value, no unused `inserted`.

src/ros2_medkit_serialization/src/type_introspection.cpp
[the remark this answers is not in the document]
bburda: Fixed in 3e26d458: corrected the comment - `get_type_info` returns empty schemas for unknown types and does not throw.

src/ros2_medkit_gateway/src/core/models/thread_safe_entity_cache.cpp
```diff
@@ -17,9 +17,169 @@
...
+// Construction / capacity
+// ============================================================================
+
+ThreadSafeEntityCache::ThreadSafeEntityCache(size_t capacity) {
+  reserve(capacity);
+}
+
+void ThreadSafeEntityCache::reserve(size_t capacity) {
+  std::unique_lock lock(mutex_);
+
```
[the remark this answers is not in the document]
bburda: Fixed in 0edefe1f: `reserve()` is now monotonic - a non-growing request is a no-op, so it never shrinks `seen_`/`capacity_` while the index containers stay grown.

src/ros2_medkit_gateway/test/test_entity_cache_alloc.cpp
```diff
@@ -0,0 +1,80 @@
...
```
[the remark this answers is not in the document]
bburda: Not adopted (validated by implementing it): the steady-state net is legitimately negative (~-1/tick) because payloads are pre-built before arming, so a symmetric bound false-fails. Kept the upper bound (the rebuild-regression guard) and documented why in 3e26d458; the vacuity concern is addressed separately by the positive control + non-SSO tick in 0edefe1f.

src/ros2_medkit_gateway/include/ros2_medkit_gateway/core/util/slot_store.hpp
bburda: Fixed in 0edefe1f: `free()` is now idempotent (`if (live_[slot]==0) return;`), so a double free cannot underflow `live_count_` or double-push the free-list.

src/ros2_medkit_gateway/include/ros2_medkit_gateway/core/util/slot_store.hpp
bburda: Fixed in 0edefe1f: `alloc_slot()` now marks the slot live and removes it from the free-list atomically, and `assign()` asserts liveness - so assign can no longer revive a free-listed slot and alias it.

src/ros2_medkit_gateway/src/gateway_node.cpp
bburda: Fixed in 0edefe1f: the parameter is read as int64 and range-checked before narrowing to int, so an out-of-range value can no longer wrap into range and bypass the check.

docs/config/server.rst
bburda: Fixed in 0edefe1f: the doc now says an out-of-range value is rejected with a warning and the default (1000) is used (not clamped).

src/ros2_medkit_serialization/src/type_introspection.cpp
bburda: Fixed in 0edefe1f: `get_type_info` no longer memoizes a failed resolve, and the service/action wrappers cache only a resolved result - so a type unresolvable on the first request can resolve on a later one instead of being pinned empty.

src/ros2_medkit_gateway/include/ros2_medkit_gateway/core/util/flat_hash_map.hpp
bburda: Fixed in 0edefe1f: `next_pow2()` saturates at the highest representable power of two instead of shifting past the top bit (no infinite loop for n > 2^63).

src/ros2_medkit_gateway/src/gateway_node.cpp
bburda: Fixed in 0edefe1f: the debounce decision is now a pure `decide_graph_refresh()` helper with unit tests covering coalescing, the 0 ms disable, and cold-start (test_refresh_debounce.cpp).

src/ros2_medkit_gateway/test/test_entity_cache_alloc.cpp
bburda: Fixed in 0edefe1f: added a positive control (asserts the new/delete interposer counts) plus non-SSO ids and a service per app so the reconcile/operation-index path copies real heap keys - the gate is no longer vacuous.

src/ros2_medkit_gateway/src/core/models/thread_safe_entity_cache.cpp
bburda: Fixed in 0edefe1f: documented the ~2x transient pool usage on full-turnover refreshes in the capacity guidance (server.rst and gateway_params.yaml) - size capacity to ~twice the expected peak.

mfaferek93 (approved):

Issue #395, opened 2026-04-28
[BUG] handle_get_fault returns cross-entity faults for synthetic components (scope leak)


bburda:
# Bug report

### Steps to reproduce

1. Launch a gateway with at least two synthetic / runtime-discovered components (any deployment without a manifest declaring the components is enough).
2. Trigger a fault on an app hosted by component A (for example by injecting a fault from a sensor node).
3. Query the same fault code on a *different* synthetic component B that does **not** host that app:
   ```
   curl -s "localhost:8080/api/v1/components/$COMP_B_ID/faults/$FAULT_CODE"
   ```

### Expected behavior

`404 Resource Not Found` (or similar) - the fault was not reported by any app hosted by component B, so the request should not return a body containing it.

### Actual behavior

`200 OK` with the fault body that was reported by component A. Cross-entity scope leak: any client with read access to component B can pull faults that were reported by entirely different components, as long as the fault_code matches.

### Root cause

`FaultHandlers::handle_get_fault` (`src/ros2_medkit_gateway/src/http/handlers/fault_handlers.cpp:631`) passes `entity_info.namespace_path` as `source_id` to `FaultManager::get_fault_with_env`. The match logic in `fault_manager.cpp:280` is:

```cpp
if (!source_id.empty()) {
  bool matches = false;
  for (const auto & src : result.fault.reporting_sources) {
    if (src.rfind(source_id, 0) == 0) { matches = true; break; }
  }
  if (!matches) {
    result.success = false;
    result.error_message = "Fault not found for source: " + source_id;
    ...
  }
}
```

Synthetic components have empty `fqn` *and* empty `namespace_path`. The guard `if (!source_id.empty())` short-circuits, the `reporting_sources` check is skipped entirely, and the manager returns the raw fault without any scope filter. The handler then forwards it to the client.

Manifest components with a non-empty `namespace_path` work correctly because the prefix match against `reporting_sources` still runs.

### Severity / class

Different class from #393 - that one is a *false negative* (synthetic components return empty results from `/logs` and `/bulk-data`). This one is a *false positive* / scope leak: a synthetic component returns faults that belong to other entities. Potentially a security-relevant boundary issue if any RBAC role allows `viewer` access to specific components but not others.

`handle_clear_fault` (DELETE per-entity) takes the same path and is presumably leaky in the same way - clearing a fault scoped to component B may delete a fault reported by component A. Worth verifying once the fix is in flight.

[...]

PR #466, opened 2026-06-22
Fix action_status_bridge fault source attribution + subscription destructor (#465)

### LINKED ISSUE #465: [BUG] action_status_bridge: faults attributed to discovery placeholder node name + missing subscription destructor
# Bug report

Two related defects in `ActionStatusBridgeNode` that make its integration test flaky and mis-attribute faults.

### 1. Faults attributed to the DDS discovery placeholder node name

`server_fqn_for_action()` resolves the action server's node FQN from `get_publishers_info_by_topic(<action>/_action/status)` and only skips publishers whose `node_name()` is empty. During DDS discovery the participant is known before its node name/namespace propagate, so rcl returns the placeholders `_NODE_NAME_UNKNOWN_` / `_NODE_NAMESPACE_UNKNOWN_` (non-empty), and the bridge builds `source_id = "_NODE_NAMESPACE_UNKNOWN_/_NODE_NAME_UNKNOWN_"`.

`reporter_for()` then caches the `FaultReporter` built from that unresolved `source_id` for the node's lifetime, so faults stay permanently attributed to the placeholder even after discovery completes.

### 2. Missing subscription destructor

`ActionStatusBridgeNode` holds status subscriptions and a rescan timer whose callbacks capture `this`, but declares no destructor. On teardown those callbacks can fire on a partially destroyed node and crash (SIGABRT).

### Steps to reproduce
1. Start an action server and the `action_status_bridge`.
2. Trigger an action result the bridge maps to a fault (e.g. ABORTED) shortly after startup, while DDS discovery is still settling.
3. Observe the fault's `reporting_sources` (placeholder instead of the server FQN), and intermittent crashes on shutdown.

### Expected behavior
- `reporting_sources` contains the action server node FQN (e.g. `/test_action_status_client`).
- Clean shutdown.

### Actual behavior
- `reporting_sources` contains `_NODE_NAMESPACE_UNKNOWN_/_NODE_NAME_UNKNOWN_` and never corrects.
- Intermittent SIGABRT on teardown.

Both surface as flaky failures of `test_integration`.

### Environment
- ros2_medkit version: current main
- ROS 2 distro: Jazzy / Humble / Lyrical
- OS: Ubuntu 24.04

### Additional information
Fix: treat the placeholder as unresolved in `server_fqn_from_endpoint()` and do not permanently cache a reporter built from an unresolved FQN (re-resolve until the real node name appears); add a destructor that resets the rescan timer and subscriptions before the node is destroyed.

bburda:
# Pull Request

## Summary

Fix two defects in `ActionStatusBridgeNode` that mis-attribute faults and make
its integration test flaky.

**1. Faults attributed to the DDS discovery placeholder node name.**
`server_fqn_for_action()` resolved the action server FQN from the status-topic
publisher and only skipped publishers with an empty `node_name()`. During DDS
discovery rcl returns the placeholders `_NODE_NAME_UNKNOWN_` /
`_NODE_NAMESPACE_UNKNOWN_` (non-empty), so the bridge built
`source_id = "_NODE_NAMESPACE_UNKNOWN_/_NODE_NAME_UNKNOWN_"`, and `reporter_for()`
cached that reporter for the node's lifetime, permanently mis-attributing the
fault even after discovery completed.

- Extract `server_fqn_from_endpoint(name, ns)` which treats the placeholders (and
  an empty name) as unresolved and returns `""`. `server_fqn_for_action()`
  returns `""` until a publisher resolves.
- `reporter_for()` uses a fallback reporter (keyed by action name) while
  unresolved so a fault still fires, but does not mark it resolved; it re-resolves
  on each status message and upgrades to the real node FQN once discovery
  completes (`resolved_actions_` tracks which reporters are final).

**2. Missing subscription destructor.**
The node holds status subscriptions and a rescan timer whose callbacks capture
`this` but declared no destructor, risking SIGABRT on teardown when a callback
fires on a partially destroyed node. Add a destructor that resets the rescan
timer and clears the subscriptions before the rest of the node is destroyed, and
delete the copy/move special members (subscription destructor pattern).

## Issue

- closes #465

## Type

- [x] Bug fix
- [ ] New feature or tests
- [ ] Breaking change
- [ ] Documentation only

## Testing

- New unit coverage for `server_fqn_from_endpoint` (resolved + placeholder cases).
- `test_action_status_bridge` (unit) and `test_integration` (launch_testing) pass
  locally.
- clang-tidy clean on the changed files (incl. special-member-functions after
  adding the destructor).

## Checklist

- [x] Breaking changes are clearly described (none)
- [x] Docs were updated if behavior or public API changed (no public API change)

### DIFF
--- a/src/ros2_medkit_action_status_bridge/src/action_status_bridge_node.cpp
+++ b/src/ros2_medkit_action_status_bridge/src/action_status_bridge_node.cpp
@@ -20,6 +20,9 @@
 #include <chrono>
 #include <cstdio>
 #include <functional>
+#include <optional>
+#include <utility>
+#include <vector>
 
 #include "action_msgs/msg/goal_status.hpp"
 #include "ros2_medkit_msgs/msg/fault.hpp"
@@ -95,7 +98,8 @@ ActionStatusBridgeNode::~ActionStatusBridgeNode() {
 }
 
 void ActionStatusBridgeNode::load_parameters() {
-  const int aborted_severity = declare_parameter<int>("aborted_severity", ros2_medkit_msgs::msg::Fault::SEVERITY_ERROR);
+  const int aborted_severity =
+      static_cast<int>(declare_parameter<int>("aborted_severity", ros2_medkit_msgs::msg::Fault::SEVERITY_ERROR));
   if (aborted_severity < ros2_medkit_msgs::msg::Fault::SEVERITY_INFO ||
       aborted_severity > ros2_medkit_msgs::msg::Fault::SEVERITY_CRITICAL) {
     RCLCPP_WARN(get_logger(), "aborted_severity %d out of range [0,3]; clamping to ERROR", aborted_severity);
@@ -122,7 +126,7 @@ void ActionStatusBridgeNode::load_parameters() {
   include_only_actions_ =
       declare_parameter<std::vector<std::string>>("include_only_actions", std::vector<std::string>{});
 
-  const int dedup_capacity = declare_parameter<int>("dedup_capacity", 4096);
+  const int dedup_capacity = static_cast<int>(declare_parameter<int>("dedup_capacity", 4096));
   if (dedup_capacity <= 0) {
     RCLCPP_WARN(get_logger(), "dedup_capacity %d not positive; using 4096", dedup_capacity);
     logged_capacity_ = 4096;
@@ -187,6 +191,12 @@ void ActionStatusBridgeNode::rescan_actions() {
   }
 
   prune_vanished(present);
+
+  // Now that discovery may have resolved more server FQNs, register them on any
+  // fault still attributed only to the action-name fallback. This is the reliable
+  // upgrade path: the status message that raised such a fault can be the last one
+  // the server sends, so the per-message path alone may never run again.
+  upgrade_unresolved_attribution();
 }
 
 void ActionStatusBridgeNode::prune_vanished(const std::map<std::string, std::string> & present_topics) {
@@ -219,6 +229,7 @@ void ActionStatusBridgeNode::prune_vanished(const std::map<std::string, std::str
       }
       reporters_.erase(action_name);
       resolved_actions_.erase(action_name);
+      pending_attribution_.erase(action_name);
     }
     {
       std::lock_guard<std::mutex> lock(state_mutex_);
@@ -279,6 +290,20 @@ ActionStatusBridgeNode::apply_message(const std::string & action_name, const act
       std::lock_guard<std::mutex> lock(state_mutex_);
       last_reported_state_[action_name] = ActionState::kFailed;
     }
+    // Decision state (no IPC): if this fault is not yet attributed to the
+    // resolved server FQN, flag it so the FQN is registered once discovery
+    // settles. The healthy->failed transition that would otherwise carry the FQN
+    // may never recur (the latched terminal status is often the server's last
+    // message), so the rescan timer upgrades it instead. Updated regardless of
+    // `reporter` (the null reporter is a test-only seam).
+    {
+      std::lock_guard<std::mutex> lock(reporters_mutex_);
+      if (resolved_actions_.count(action_name) == 0) {
+        pending_attribution_[action_name] = canceled;
+      } else {
+        pending_attribution_.erase(action_name);
+      }
+    }
     return ActionState::kFailed;
   }
 
@@ -294,6 +319,10 @@ ActionStatusBridgeNode::apply_message(const std::string & action_name, const act
     {
       std::lock_guard<std::mutex> lock(state_mutex_);
       last_reported_state_[action_name] = ActionState::kHealthy;
+    }
+    {
+      std::lock_guard<std::mutex> lock(reporters_mutex_);
+      pending_attribution_.erase(action_name);  // healed: no FQN upgrade needed
     }
     return ActionState::kHealthy;
   }
@@ -321,15 +350,27 @@ void ActionStatusBridgeNode::status_callback(const std::string & action_name,
     }
   }
 
-  apply_message(action_name, *msg, reporter_for(action_name));
+  const auto lookup = reporter_for(action_name);
+  if (lookup.reaffirm_canceled.has_value()) {
+    // The server FQN just resolved while this action's fault is already active.
+    // Re-report the active code on the resolved reporter so the FQN is added to
+    // the fault's reporting_sources (EVENT_UPDATED on the already-confirmed
+    // fault: no new snapshot, no transition) - this is what lets the gateway
+    // associate the fault with the server's SOVD entity.
+    const bool canceled = *lookup.reaffirm_canceled;
+    lookup.reporter->report(fault_code_for(action_name, canceled), aborted_severity_,
+                            std::string("Action ") + action_name + (canceled ? " canceled" : " aborted"));
+    RCLCPP_INFO(get_logger(), "Action %s fault attribution resolved to server FQN", action_name.c_str());
+  }
+  apply_message(action_name, *msg, lookup.reporter);
 }
 
-ros2_medkit_fault_reporter::FaultReporter * ActionStatusBridgeNode::reporter_for(const std::string & action_name) {
+ActionStatusBridgeNode::ReporterLookup ActionStatusBridgeNode::reporter_for(const std::string & action_name) {
   std::lock_guard<std::mutex> lock(reporters_mutex_);
   auto it = reporters_.find(action_name);
   const bool have = it != reporters_.end();
   if (have && resolved_actions_.count(action_name) != 0) {
-    return it->second.get();  // already attributed to the resolved server FQN
+    return {it->second.get(), std::nullopt};  // already attributed to the resolved server FQN
   }
   // Attribute the fault to the action SERVER's node FQN, not the action name:
   // the gateway resolves a fault to a SOVD entity by node FQN, and an action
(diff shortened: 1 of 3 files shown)


src/ros2_medkit_action_status_bridge/src/action_status_bridge_node.cpp
```diff
@@ -316,34 +327,61 @@ void ActionStatusBridgeNode::status_callback(const std::string & action_name,
...
+    reporters_[action_name] = std::move(reporter);
+    return raw;
+  }
+  // Resolved: (re)create the reporter attributed to the real server FQN and mark
+  // it resolved so it is never re-resolved again.
+  auto reporter = std::make_unique<ros2_medkit_fault_reporter::FaultReporter>(this->shared_from_this(), fqn);
   auto * raw = reporter.get();
-  reporters_.emplace(action_name, std::move(reporter));
+  reporters_[action_name] = std::move(reporter);
+  resolved_actions_.insert(action_name);
```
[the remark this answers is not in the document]
bburda: Confirmed. The fault stayed attributed to the action-name fallback because the server FQN was only attached on the healthy->failed transition, which does not recur for an already-failed action. A latched terminal status is also often the server's last message, so "re-resolve on the next status message" may never fire either.

Fixed by tracking the failure kind in `pending_attribution_` when a fault is raised on the fallback, then re-reporting the active code on the resolved reporter once discovery settles. That registers the server FQN on the still-active fault (EVENT_UPDATED: adds the FQN to `reporting_sources`, no new snapshot, no state transition). The upgrade also runs from the rescan timer, not only the per-message path, so it works when no further status message arrives.

src/ros2_medkit_action_status_bridge/src/action_status_bridge_node.cpp
```diff
@@ -310,40 +350,124 @@ void ActionStatusBridgeNode::status_callback(const std::string & action_name,
...
+  if (fqn.empty()) {
+    // DDS discovery has not resolved the server's node name yet. Use a fallback
+    // reporter (keyed by action name) so a fault still fires immediately (and its
+    // snapshot is captured at fault time); it is upgraded to the real FQN once
+    // discovery settles. Reuse the existing fallback reporter rather than
+    // recreating it.
+    if (have) {
+      return {it->second.get(), std::nullopt};
+    }
+    auto reporter = std::make_unique<ros2_medkit_fault_reporter::FaultReporter>(this->shared_from_this(), action_name);
```
mfaferek93: This fallback reporter is keyed by the action name, so source_id = /navigate_to_pose enters the fault's reporting_sources - and on the manager side that set is append-only. The later re-report under the FQN (lines 361 / 441) only ADDS the FQN; it never removes the action name. The per-entity /faults path filters via fault_in_source_scope (fault_scope.cpp:125), which requires ALL reporting_sources to match the node scope, so the lingering action-name source permanently excludes the fault from the server node entity (mixed-source faults are rejected, per the #395 behavior). Net: the whole upgrade path (pending_attribution_ / reaffirm / upgrade_unresolved_attribution) doesn't actually achieve attribution. Since the set is add-only and the scope check is strict-AND, the only in-bridge fix is to not emit the action-name source at all - defer the first report until the FQN resolves (keep pending_attribution_ to raise/snapshot, but hold reporter->report() until an FQN exists).

src/ros2_medkit_action_status_bridge/src/action_status_bridge_node.cpp
```diff
@@ -310,40 +350,124 @@ void ActionStatusBridgeNode::status_callback(const std::string & action_name,
...
   }
 
-  apply_message(action_name, *msg, reporter_for(action_name));
+  const auto lookup = reporter_for(action_name);
+  if (lookup.reaffirm_canceled.has_value()) {
+    // The server FQN just resolved while this action's fault is already active.
+    // Re-report the active code on the resolved reporter so the FQN is added to
+    // the fault's reporting_sources (EVENT_UPDATED on the already-confirmed
+    // fault: no new snapshot, no transition) - this is what lets the gateway
+    // associate the fault with the server's SOVD entity.
```
mfaferek93: This comment says the re-report is what associates the fault with the server's SOVD entity, but it isn't: reporting_sources is append-only and entity filtering is strict ALL-must-match, so the lingering action-name source keeps the fault out of the node entity. Worth updating/dropping once the source emission is fixed, so it doesn't read as if the gap is closed.

src/ros2_medkit_action_status_bridge/src/action_status_bridge_node.cpp
```diff
@@ -310,40 +350,124 @@ void ActionStatusBridgeNode::status_callback(const std::string & action_name,
...
+    const bool canceled = *lookup.reaffirm_canceled;
+    lookup.reporter->report(fault_code_for(action_name, canceled), aborted_severity_,
+                            std::string("Action ") + action_name + (canceled ? " canceled" : " aborted"));
+    RCLCPP_INFO(get_logger(), "Action %s fault attribution resolved to server FQN", action_name.c_str());
+  }
+  apply_message(action_name, *msg, lookup.reporter);
```
mfaferek93: lookup.reporter is a raw FaultReporter* dereferenced here (and at the re-report on line 361) outside reporters_mutex_, while prune_vanished() erases reporters_[action] under the lock on the rescan-timer thread -> use-after-free if this ever runs under a multi-threaded executor. Safe under the single-threaded rclcpp::spin in main.cpp today; worth pinning that invariant in a comment, or returning/holding a shared_ptr.

src/ros2_medkit_action_status_bridge/test/test_action_status_bridge.cpp
```diff
@@ -260,6 +276,33 @@ TEST_F(ActionStatusBridgeTest, Apply_PerActionIsolation) {
   EXPECT_EQ(node->apply_message("/nav", one_goal(3, GoalStatus::STATUS_SUCCEEDED), nullptr), State::kHealthy);
 }
 
+// --- attribution upgrade bookkeeping (unresolved raise -> pending -> cleared) ---
+
+TEST_F(ActionStatusBridgeTest, Apply_UnresolvedRaiseFlagsPendingAttribution) {
```
mfaferek93: Nice to see pending_attribution_ bookkeeping covered. The actual upgrade outcome still has no test, though: reporter_for returning reaffirm when an FQN resolves for an action with an active fallback fault, the re-report under the FQN, and - most importantly - whether the resulting fault actually resolves to the server node entity (it currently doesn't; see the bridge comment). Suggest a seam to inject a resolved FQN and assert the end-state attribution.

src/ros2_medkit_action_status_bridge/src/action_status_bridge_node.cpp
bburda: Confirmed - checked `fault_scope.cpp`: the per-entity `/faults` filter (`fault_in_source_scope`) is strict-AND over `reporting_sources`, and the manager's set is append-only, so a lingering action-name source keeps the fault out of the server node entity even after the FQN report.

Dropped the re-attribution entirely. The reporter source is now fixed at the first report (server FQN if discovery resolved, otherwise the action name so the fault still fires on time) and never swapped. Reporting the fault on time is the priority here. Doing both - reporting on time and correct entity attribution - needs a way to supersede a provisional source, which is tracked as a separate follow-up.

src/ros2_medkit_action_status_bridge/src/action_status_bridge_node.cpp
bburda: Removed along with the re-report. The comment now states that the source is fixed at the first report and not re-attributed, with the append-only + strict-AND reason.

src/ros2_medkit_action_status_bridge/src/action_status_bridge_node.cpp
bburda: The reaffirm re-report path is gone. The remaining raw `FaultReporter*` use in `apply_message` is now pinned with a comment: under the single-threaded `rclcpp::spin` in main.cpp, status callbacks and the rescan timer run on the same thread, so `prune_vanished()` never erases `reporters_` concurrently. The invariant is flagged for any future multi-threaded executor.

src/ros2_medkit_action_status_bridge/test/test_action_status_bridge.cpp
bburda: With the re-attribution removed there is no upgrade outcome left to test. Added a stickiness test asserting the reporter is created once per action and never swapped (the guarantee that replaces the upgrade). The server-FQN source path stays covered by the integration test (`reporting_sources` contains the server node FQN). Asserting the exact fallback source_id would need an accessor on `FaultReporter` in another package, left out to keep this change bridge-scoped.

src/ros2_medkit_action_status_bridge/src/action_status_bridge_node.cpp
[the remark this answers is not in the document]
bburda: Update: the re-report / upgrade approach described above was reverted. It could not achieve attribution because `reporting_sources` is append-only and the per-entity scope filter is strict-AND, so the provisional action-name source kept the fault out of the server node entity. The reporter source is now fixed at the first report (server FQN if discovery resolved, otherwise the action-name fallback) and not re-attributed; correct attribution for the slow-discovery case is tracked as a follow-up.

mfaferek93 (approved):

PR #472, opened 2026-06-23
fix(action_status_bridge): retry deferred fault reports instead of dropping them

### LINKED ISSUE #471: [BUG] action_status_bridge loses first fault when FaultManager service not yet discovered (startup race)
# Bug report

The action-status bridge silently loses the first fault of a failing action when the FaultManager's `report_fault` service has not been discovered yet at the instant the bridge observes the terminal status. This happens during the startup discovery window and is exposed by DDS discovery timing (seen failing on Humble, passing on Jazzy).

### Steps to reproduce
1. Launch `fault_manager_node` and `action_status_bridge_node` together (same startup).
2. Publish an `ABORTED` `GoalStatusArray` on a `*/_action/status` topic with `TRANSIENT_LOCAL` (latched) durability, so the bridge receives it the instant it subscribes.
3. Query `/fault_manager/list_faults` for the expected `ACTION_<NAME>_ABORTED` fault.

This is the `ros2_medkit_action_status_bridge` `test_integration` launch test (`test_01_aborted_goal_raises_fault`, `test_02_succeeded_goal_heals_fault`).

### Expected behavior

The bridge raises the `ACTION_<NAME>_ABORTED` fault in the FaultManager (and heals it on a later `SUCCEEDED`). A high-severity fault must not be lost just because the FaultManager service was discovered a few milliseconds late.

### Actual behavior

The fault never reaches the FaultManager; `list_faults` returns nothing and the test fails:

```
[action_status_bridge] WARN  FaultManager service not available, dropping high-severity fault 'ACTION_TEST_ACTION_ABORTED'
[action_status_bridge] INFO  Action /test_action -> fault ACTION_TEST_ACTION_ABORTED
...
AssertionError: unexpectedly None : expected a ACTION_TEST_ACTION_ABORTED fault
AssertionError: unexpectedly None : ACTION_TEST_ACTION_ABORTED fault not HEALED in time
```

**Root cause.** In `apply_message()`, on the `healthy -> failed` transition the bridge calls `reporter->report()` (fire-and-forget) and then sets `last_reported_state_[action] = kFailed` **unconditionally**. If the report service is not yet discovered, `FaultReporter::report()` drops the report - but the bridge has already recorded the action as reported. The latched status message is delivered exactly once (a rescan skips already-subscribed topics and does not re-deliver it), so the dropped report is never retried and the fault is lost permanently. The follow-up `SUCCEEDED` then has no raised fault to heal.

### Environment
- ros2_medkit version: main
- ROS 2 distro: Humble (timing-exposed; logic affects all distros)
- OS: Ubuntu 22.04

### Additional information

Fix direction: track the desired (observed) per-action state separately from the last reported state, and only commit `last_reported_state_` once the report is actually deliverable (service ready). Re-attempt undelivered transitions on the existing rescan tick so a report dropped during the discovery window is retried once the service appears. Scope: `ros2_medkit_action_status_bridge` only; the shared `FaultReporter` fire-and-forget contract is unchanged.

bburda:
# Pull Request

## Summary

Two fixes (see commits):

### 1. action_status_bridge: don't lose / promptly deliver faults when the FaultManager service isn't discovered yet (closes #471)

The bridge silently lost the first fault of a failing action when the FaultManager's `report_fault` service was not yet discovered at the instant the bridge observed the terminal status (startup discovery window; exposed by DDS timing - failing on Humble, passing on Jazzy).

Root cause: on the `healthy -> failed` transition `apply_message()` called the fire-and-forget `reporter->report()` and then committed `last_reported_state_ = kFailed` **unconditionally**. If the service was not ready, `FaultReporter` dropped the request - but the bridge had already recorded the action as reported. The latched (`transient_local`) status is delivered to a subscription exactly once and a rescan skips already-subscribed topics, so the dropped report was never retried and the high-severity fault was lost permanently.

Fix:
- Track the latest observed action state (`desired_state_`) separately from what the FaultManager was told (`last_reported_state_`); deliver via `reconcile()` only when they differ **and** the report is deliverable (service ready; null reporter = unit-test seam), committing only after delivery.
- **Deliver as fast as the discovery floor allows:** the FaultManager captures its freeze-frame snapshot when it receives the report, so delivery latency makes that snapshot stale. The happy path reports synchronously in the status callback; a deferred report is retried on a dedicated **fast retry timer** (`retry_period_sec`, default 50 ms, armed only while pending), decoupled from the slow discovery rescan, so it lands within one short tick of the service appearing. The executor is deliberately not blocked. The README documents this delivery-latency model and its level-triggered boundaries.
- `reconcile()` holds `state_mutex_` across gate -> report -> commit so the status callback and retry timer can't double-report under a multi-threaded executor.
- The prune-vanished heal is gated on service readiness and warns (instead of silently dropping) when a vanished action's heal can't be delivered.
- The shared `FaultReporter` fire-and-forget contract is unchanged.

### 2. gateway: fix flaky SSE handler test under ASan

`SSEFaultHandlerTest.StreamSnapshotsEntityContextAtEnqueue` flaked under AddressSanitizer. `enqueue_event()` spun for a fixed budget after publishing; under sanitizer slowdown the handler's `on_fault_event` (which snapshots the owning entity) could run after that budget, so the test cleared the cache before the snapshot was taken and the replayed event omitted `x-medkit`. Added `SSEFaultHandler::events_received()` (a monotonic counter, like `connected_clients()`) and made `enqueue_event()` wait until the handler has actually consumed the event. Test-only timing fix; no product behavior change.

---

## Issue

- closes #471

---

## Type

- [x] Bug fix
- [ ] New feature or tests
- [ ] Breaking change
- [ ] Documentation only

---

## Testing

- action_status_bridge: 31 unit tests pass (incl. deferred raise/heal delivered once the service is ready, and the retry pass not dropping while the service is down); the `test_integration` launch test passes (raise + heal); linters + `clang-tidy` clean. CI `Pixi (humble)` (originally failing) and Humble/Jazzy/Lyrical/TSan green on the prior commits.
- gateway SSE: full `test_sse_fault_handler` (19 tests) passes; the snapshot test passes 30/30 in a repeat loop; `clang-format` + `clang-tidy` clean on the changed files.

---

## Checklist

### DIFF
--- a/src/ros2_medkit_action_status_bridge/src/action_status_bridge_node.cpp
+++ b/src/ros2_medkit_action_status_bridge/src/action_status_bridge_node.cpp
@@ -76,13 +76,21 @@ ActionStatusBridgeNode::ActionStatusBridgeNode(const rclcpp::NodeOptions & optio
   : Node("action_status_bridge", options) {
   load_parameters();
 
+  // Fast retry timer, created up front but kept disarmed (cancelled) until a
+  // report is actually deferred, so an idle bridge does no periodic work.
+  retry_timer_ = create_wall_timer(std::chrono::duration<double>(retry_period_sec_), [this]() {
+    reconcile_pending();
+  });
+  retry_timer_->cancel();
+
   rescan_actions();
   rescan_timer_ = create_wall_timer(std::chrono::duration<double>(rescan_period_sec_), [this]() {
     rescan_actions();
   });
 
-  RCLCPP_INFO(get_logger(), "ActionStatusBridge started (prefix=%s, aborted_severity=%u, rescan=%.1fs)",
-              code_prefix_.c_str(), static_cast<unsigned>(aborted_severity_), rescan_period_sec_);
+  RCLCPP_INFO(get_logger(), "ActionStatusBridge started (prefix=%s, aborted_severity=%u, rescan=%.1fs, retry=%.0fms)",
+              code_prefix_.c_str(), static_cast<unsigned>(aborted_severity_), rescan_period_sec_,
+              retry_period_sec_ * 1000.0);
 }
 
 ActionStatusBridgeNode::~ActionStatusBridgeNode() {
@@ -92,6 +100,7 @@ ActionStatusBridgeNode::~ActionStatusBridgeNode() {
   // (subscription destructor pattern). subs_ is only mutated from the (now
   // stopped) rescan path, so no lock is needed here.
   rescan_timer_.reset();
+  retry_timer_.reset();
   subs_.clear();
 }
 
@@ -112,6 +121,16 @@ void ActionStatusBridgeNode::load_parameters() {
   if (rescan_period_sec_ <= 0.0) {
     RCLCPP_WARN(get_logger(), "rescan_period_sec %.3f not positive; using 2.0", rescan_period_sec_);
     rescan_period_sec_ = 2.0;
+  }
+
+  // Fast retry cadence for delivering reports deferred while the FaultManager
+  // service was undiscovered. Kept short (default 50 ms) so the FaultManager
+  // freeze-frame snapshot is taken as close to the event as the discovery floor
+  // allows; the timer is armed only while a delivery is pending.
+  retry_period_sec_ = declare_parameter<double>("retry_period_sec", 0.05);
+  if (retry_period_sec_ <= 0.0) {
+    RCLCPP_WARN(get_logger(), "retry_period_sec %.3f not positive; using 0.05", retry_period_sec_);
+    retry_period_sec_ = 0.05;
   }
 
   code_prefix_ = to_upper_snake(declare_parameter<std::string>("code_prefix", "ACTION"), kMaxActionPart);
@@ -190,10 +209,10 @@ void ActionStatusBridgeNode::rescan_actions() {
 
   prune_vanished(present);
 
-  // Re-attempt any fault delivery that was deferred because the FaultManager
-  // service was not discovered when the (latched) status arrived during startup.
-  // The latched message is delivered only once, so without this a dropped report
-  // would never be retried.
+  // Backstop the fast retry timer: re-attempt any deferred delivery and clean up
+  // healed-and-vanished actions on the (slow) rescan tick too, so a missed timer
+  // arming can never strand a pending report. The fast timer is what keeps the
+  // FaultManager freeze-frame contemporaneous; this is just belt-and-suspenders.
   reconcile_pending();
 }
 
@@ -216,14 +235,26 @@ void ActionStatusBridgeNode::prune_vanished(const std::map<std::string, std::str
       std::lock_guard<std::mutex> lock(reporters_mutex_);
       // If a still-failed action vanishes (e.g. Nav2 lifecycle deactivate), heal
       // its fault before forgetting it - otherwise the fault stays stuck active
-      // in the FaultManager and the bridge can no longer clear it.
+      // in the FaultManager and the bridge can no longer clear it. Gate the heal
+      // on service readiness: unlike a live heal, a vanished action cannot be
+      // retried (its state is dropped here), so if the FaultManager service is
+      // not discovered at this instant the heal cannot be delivered. Surface that
+      // narrow case with a warning instead of dropping it silently. See the
+      // delivery-latency note in the README.
       auto rit = reporters_.find(action_name);
       if (was_failed && heal_on_succeeded_ && rit != reporters_.end()) {
-        rit->second->report_passed(fault_code_for(action_name, false));
-        if (canceled_is_fault_) {
-          rit->second->report_passed(fault_code_for(action_name, true));
+        if (rit->second->is_service_ready()) {
+          rit->second->report_passed(fault_code_for(action_name, false));
+          if (canceled_is_fault_) {
+            rit->second->report_passed(fault_code_for(action_name, true));
+          }
+          RCLCPP_INFO(get_logger(), "Action '%s' vanished while failed; healed before dropping", action_name.c_str());
+        } else {
+          RCLCPP_WARN(get_logger(),
+                      "Action '%s' vanished while failed but FaultManager service is not ready; its heal could not be "
+                      "delivered and the fault may remain active",
+                      action_name.c_str());
         }
-        RCLCPP_INFO(get_logger(), "Action '%s' vanished while failed; healed before dropping", action_name.c_str());
       }
       reporters_.erase(action_name);
     }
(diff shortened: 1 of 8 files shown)


src/ros2_medkit_action_status_bridge/test/test_action_status_bridge.cpp
```diff
@@ -289,6 +289,27 @@ TEST_F(ActionStatusBridgeTest, ReporterFor_StickyCreatedOnceNeverSwapped) {
...
+  // No FaultManager runs in this unit test, so the report service is not ready.
+  ASSERT_FALSE(reporter->is_service_ready());
+
+  // An ABORTED whose report cannot be delivered must NOT commit the reported
+  // state (otherwise a later SUCCEEDED would "heal" a fault the manager never
+  // received) and must stay pending so a rescan can retry it once the service
+  // appears.
+  EXPECT_EQ(node->apply_message("/nav", one_goal(1, GoalStatus::STATUS_ABORTED), reporter), State::kUnknown);
+  EXPECT_FALSE(access.reported_failed("/nav"));  // not falsely recorded as delivered
+  EXPECT_TRUE(access.pending_failed("/nav"));    // remembered for retry, not dropped
```
mfaferek93, line 310: This proves the report is deferred and kept pending, but not the actual fix: that reconcile_pending() retries and delivers it once the service is ready. Worth adding the converse here, e.g. a second apply_message via the nullptr (deliverable) seam asserting reported_failed("/nav") == true and pending_failed("/nav") == false, plus the same for a deferred heal. reconcile_pending() has no test seam today, so the retry loop is entirely unexercised.

src/ros2_medkit_action_status_bridge/src/action_status_bridge_node.cpp
```diff
@@ -256,49 +264,106 @@ ActionStatusBridgeNode::apply_message(const std::string & action_name, const act
...
+  if (desired.net == reported) {
+    return ActionState::kUnknown;  // already in sync, nothing to report
+  }
+
+  // A real reporter can only deliver once the FaultManager service is
+  // discovered; until then keep the transition pending (retried by
+  // reconcile_pending() on the next rescan) instead of dropping it. A null
+  // reporter is the unit-test decision-only seam: delivery is a no-op that
+  // still commits the transition.
+  const bool deliverable = (reporter == nullptr) || reporter->is_service_ready();
```
mfaferek93, line 346: reconcile() reads desired/reported under the lock, releases it for this gate and the report() call, then re-locks to commit. With reconcile_pending() now a second caller next to the status callback, a multi-threaded executor could let both pass this gate and double-report. report() is fire-and-forget (async_send_request), so holding state_mutex_ across gate+report+commit would close the window cleanly. If you'd rather keep report() out of the lock, worth documenting the non-atomic invariant here; the comment at status_callback only covers reporter-pointer lifetime.

src/ros2_medkit_action_status_bridge/src/action_status_bridge_node.cpp
```diff
@@ -256,49 +264,106 @@ ActionStatusBridgeNode::apply_message(const std::string & action_name, const act
...
+      return ActionState::kUnknown;  // nothing desired yet
+    }
+    desired = dit->second;
+    auto rit = last_reported_state_.find(action_name);
+    // Absent reported state is treated as healthy: a first SUCCEEDED must not
+    // heal a fault that was never raised.
+    reported = (rit == last_reported_state_.end()) ? ActionState::kHealthy : rit->second;
+  }
+
+  if (desired.net == reported) {
```
mfaferek93, line 338: Level-triggered edge case: if an action ABORTS then SUCCEEDS while the service is still undiscovered, the ABORTED defers without committing, the SUCCEEDED overwrites desired to healthy, and this short-circuit then drops the failure edge with no raise ever sent. It doesn't hit the latched single-goal case (only the last sample arrives), but it does contradict the 'never silently dropped' framing. Either document the tradeoff, or latch an observed-while-undeliverable flag so the retry still emits a raise+heal pair once the service appears.

src/ros2_medkit_action_status_bridge/src/action_status_bridge_node.cpp
```diff
@@ -223,6 +230,7 @@ void ActionStatusBridgeNode::prune_vanished(const std::map<std::string, std::str
     {
       std::lock_guard<std::mutex> lock(state_mutex_);
       last_reported_state_.erase(action_name);
+      desired_state_.erase(action_name);
```
mfaferek93, line 264: The heal a few lines up (report_passed on vanish-while-failed) isn't gated on is_service_ready() the way reconcile() now is. If the service isn't ready at prune time, that heal is dropped and never retried, leaving an orphan raised fault in the FaultManager. Same deferred-drop class this PR fixes on the reconcile path; worth gating it the same way, or at least noting the gap.

src/ros2_medkit_action_status_bridge/test/test_action_status_bridge.cpp
bburda: Good catch. Added the converse coverage:
- `Reconcile_DeferredRaiseDeliveredOnceServiceReady` - a deferred raise is delivered (`reported_failed` true, `pending_failed` false) once the report becomes deliverable.
- `Reconcile_DeferredHealDeliveredOnceServiceReady` - same for a deferred heal.
- `ReconcilePending_RetriesWithoutDroppingWhileServiceDown` - drives the retry pass and asserts the pending fault is retried, not dropped, while the service stays down.

Added `run_reconcile_pending()` and `reconcile_deliverable()` test seams so the retry loop is exercised directly; end-to-end delivery+heal is also covered by `test_integration`.

src/ros2_medkit_action_status_bridge/src/action_status_bridge_node.cpp
bburda: Closed rather than documented: `reconcile()` now holds `state_mutex_` across the gate, the `report()`/`report_passed()` call and the commit, so the status callback and the new retry timer cannot both pass the gate and double-report under a multi-threaded executor. `report()` is async (fire-and-forget), so the critical section stays short, and `reporters_mutex_` is never held across this section so there is no lock-order inversion.

src/ros2_medkit_action_status_bridge/src/action_status_bridge_node.cpp
bburda: Agreed this is the level-triggered model boundary. Rather than change the net-state model, documented it explicitly in the README ("Delivery latency and freeze-frame"): a flap that occurs entirely while the report channel is undiscovered collapses to the net healthy state (there is nothing to freeze-frame after the fact). The new fast retry timer (default 50 ms) shrinks the window to roughly one tick, so the action would have to fail and recover within that window while the service is undiscovered. A flap while the channel is up still emits a raise+heal per cycle. Reworded the "never silently dropped" framing to match.

src/ros2_medkit_action_status_bridge/src/action_status_bridge_node.cpp
bburda: Fixed: the prune-vanished heal is now gated on `is_service_ready()`. Because a vanished action state is dropped here and cannot be retried, the narrow case where the service is undiscovered at that exact instant is surfaced with a warning instead of being dropped silently, and the residual-orphan possibility is documented in the README delivery-latency note. Live (non-vanish) heals go through the retryable reconcile path.

mfaferek93 (approved):

PR #484, opened 2026-06-30
fix(fault_manager): clamp debounce counter and latch confirmed/healed status

### LINKED ISSUE #428: Unbounded debounce counter of faults
## Summary

When using the ros2_medkit reporter, manager and gateway, I realized that some of my faults seem to be "stuck" in a PREPASSED state even when a fault was occurring. After some time, they appeared as CONFIRMED faults. I thus investigated the code and found something unexpected: the de-bounce counter is unbounded (except by the maximum integer value). Therefore, my periodic heal() calls increase the de-bounce counter to a very large value, so it then takes many fault reports to bring it back down. As a result, the system builds up inertia which, from my experience with integral controllers in control theory, could lead to undesired behavior.

Here are the debounce_counter increment lines that are only limited by the int maximum value:

- https://github.com/selfpatch/ros2_medkit/blob/be1b03a6219c7f076383972b9dbc6ac4bb1541c1/src/ros2_medkit_fault_manager/src/sqlite_fault_storage.cpp#L400
- https://github.com/selfpatch/ros2_medkit/blob/be1b03a6219c7f076383972b9dbc6ac4bb1541c1/src/ros2_medkit_fault_manager/src/sqlite_fault_storage.cpp#L453`

I then looked into the DEM AUTOSAR counter logic (https://www.autosar.org/fileadmin/standards/R20-11/CP/AUTOSAR_SWS_DiagnosticEventManager.pdf) and it appears that there should be thresholds on the PASSED and FAILED statuses that serve as their confirmation criteria, set at -128 and 128 in that standard. From my understanding, an ECU that reports a fault applies a multiplicative value to heal or fail faster. In summary, it is a system where a gain determines the number of occurrences before passing or failing, but with bounded thresholds.

I understand that confirmation_threshold and healing_threshold simulate that gain, but since the debounce counter is unbounded, the logic differs from the DEM AUTOSAR specification.

So I was wondering if there is a reason why you don't put thresholds on the de-bounce counter? If so, I might stop trying to heal constantly when the system is healthy. Though, I still feel there could be some race conditions on initialization that would prevent the system from healing normally. If not, I propose adding bounds equal to confirmation_threshold and healing_threshold.

---

Add bounds on the debounce counter equal to the confirmation and healing thresholds. A new hysteresis system should be designed to ensure that healing once from a fault does not put the system into an automatic PREPASSED state. I believe the DEM AUTOSAR standard has a solution for this.

---

Any extra details, links, screenshots, etc.

mfaferek93:
## Summary

Fixes the unbounded debounce counter from #428. A periodic heal heartbeat on a healthy system drove the counter toward `INT32_MAX` (it only saturated at the integer limits), so when a real fault arrived it had to count all the way back before confirming - arbitrary confirmation lag.

## Fix

- Clamp the counter to `[confirmation_threshold, healing_threshold]` (an AUTOSAR-DEM-style bounded FDC). Worst-case lag is now `healing_threshold - confirmation_threshold` reports instead of unbounded.
- Add hysteresis (Schmitt trigger): once `CONFIRMED` or `HEALED`, the status latches until the counter walks all the way to the opposite threshold, so a single report can't flip a confirmed fault back to a pending state.
- Applied to both storage backends (`SqliteFaultStorage` and `InMemoryFaultStorage`); the manager uses both.
- No DB schema change.

## Tests

- Updated two tests that encoded the old behaviour (a confirmed fault silently downgrading on a heal).
- Added regression tests for the runaway-clamp and the latch on both backends.
- `colcon test` for `ros2_medkit_fault_manager`: 0 failures. 

Thanks to @evTessellate for the detailed report and the AUTOSAR DEM analysis in #428.

Closes #428

### DIFF
--- a/src/ros2_medkit_fault_manager/src/fault_storage.cpp
+++ b/src/ros2_medkit_fault_manager/src/fault_storage.cpp
@@ -148,8 +148,9 @@ bool InMemoryFaultStorage::report_fault_event(const std::string & fault_code, ui
       ++state.occurrence_count;
     }
 
-    // Decrement debounce counter (towards confirmation), clamped at the
-    // confirmation threshold so a heal heartbeat can't drive it off to INT32_MIN.
+    // Decrement debounce counter (towards confirmation), clamped at the confirmation
+    // threshold so a long burst of FAILED events can't drive it to INT32_MIN (which
+    // would then delay later healing).
     if (state.debounce_counter > config.confirmation_threshold) {
       --state.debounce_counter;
     }
--- a/src/ros2_medkit_fault_manager/src/fault_storage.cpp
+++ b/src/ros2_medkit_fault_manager/src/fault_storage.cpp
@@ -60,7 +60,9 @@ bool sanitize_debounce_config(DebounceConfig & config) {
     config.confirmation_threshold = -1;
     valid = false;
   }
-  if (config.healing_threshold <= 0) {
+  // healing_threshold == 0 is valid: it means "heal on a single PASSED event" (the counter reaches
+  // the threshold at 0). Only a negative threshold is rejected.
+  if (config.healing_threshold < 0) {
     config.healing_threshold = 3;
     valid = false;
   }
--- a/src/ros2_medkit_fault_manager/src/fault_storage.cpp
+++ b/src/ros2_medkit_fault_manager/src/fault_storage.cpp
@@ -19,6 +19,54 @@
 
 namespace ros2_medkit_fault_manager {
 
+int32_t clamp_debounce_counter(int32_t counter, const DebounceConfig & config) {
+  // Manual min/max rather than std::clamp: well-defined even if a bad config has lo > hi
+  // (sanitize_debounce_config normally prevents that, but the storage layer must never UB).
+  if (counter < config.confirmation_threshold) {
+    return config.confirmation_threshold;
+  }
+  if (counter > config.healing_threshold) {
+    return config.healing_threshold;
+  }
+  return counter;
+}
+
+std::string compute_debounce_status(int32_t counter, const std::string & current_status,
+                                    const DebounceConfig & config) {
+  namespace msg = ros2_medkit_msgs::msg;
+  if (counter <= config.confirmation_threshold) {
+    return msg::Fault::STATUS_CONFIRMED;
+  }
+  if (config.healing_enabled && counter >= config.healing_threshold) {
+    return msg::Fault::STATUS_HEALED;
+  }
+  // Hysteresis latch: a confirmed or healed fault stays put until the counter reaches the opposite
+  // threshold (handled above). A single opposite-direction report must not flip it to a pending state.
+  if (current_status == msg::Fault::STATUS_CONFIRMED || current_status == msg::Fault::STATUS_HEALED) {
+    return current_status;
+  }
+  if (counter < 0) {
+    return msg::Fault::STATUS_PREFAILED;
+  }
+  if (counter > 0) {
+    return msg::Fault::STATUS_PREPASSED;
+  }
+  return current_status;  // counter == 0 keeps the current status (avoids flapping at the boundary)
+}
+
+bool sanitize_debounce_config(DebounceConfig & config) {
+  bool valid = true;
+  if (config.confirmation_threshold >= 0) {
+    config.confirmation_threshold = -1;
+    valid = false;
+  }
+  if (config.healing_threshold <= 0) {
+    config.healing_threshold = 3;
+    valid = false;
+  }
+  return valid;
+}
+
 ros2_medkit_msgs::msg::Fault FaultState::to_msg() const {
   ros2_medkit_msgs::msg::Fault msg;
   msg.fault_code = fault_code;
@@ -49,24 +97,9 @@ DebounceConfig InMemoryFaultStorage::get_debounce_config() const {
 }
 
 void InMemoryFaultStorage::update_status(FaultState & state, const DebounceConfig & config) {
-  // Note: CLEARED faults are handled in report_fault_event() before this is called
-
-  if (state.debounce_counter <= config.confirmation_threshold) {
-    state.status = ros2_medkit_msgs::msg::Fault::STATUS_CONFIRMED;
-  } else if (config.healing_enabled && state.debounce_counter >= config.healing_threshold) {
-    state.status = ros2_medkit_msgs::msg::Fault::STATUS_HEALED;
-  } else if (state.status == ros2_medkit_msgs::msg::Fault::STATUS_CONFIRMED ||
-             state.status == ros2_medkit_msgs::msg::Fault::STATUS_HEALED) {
-    // Hysteresis latch: a confirmed or healed fault stays put until the counter
-    // walks all the way to the opposite threshold (handled above). A single report
-    // must not flip it back to a pending state.
-  } else if (state.debounce_counter < 0) {
-    state.status = ros2_medkit_msgs::msg::Fault::STATUS_PREFAILED;
-  } else if (state.debounce_counter > 0) {
-    state.status = ros2_medkit_msgs::msg::Fault::STATUS_PREPASSED;
-  }
-  // Note: debounce_counter == 0 keeps current status (hysteresis behavior).
-  // This avoids rapid status flapping at the zero crossing boundary.
+  // CLEARED faults are handled in report_fault_event() before this is called.
+  // Delegate to the shared state machine so both backends behave identically.
+  state.status = compute_debounce_status(state.debounce_counter, state.status, config);
 }
 
 bool InMemoryFaultStorage::report_fault_event(const std::string & fault_code, uint8_t event_type, uint8_t severity,
@@ -148,12 +181,9 @@ bool InMemoryFaultStorage::report_fault_event(const std::string & fault_code, ui
       ++state.occurrence_count;
     }
 
-    // Decrement debounce counter (towards confirmation), clamped at the confirmation
-    // threshold so a long burst of FAILED events can't drive it to INT32_MIN (which
-    // would then delay later healing).
-    if (state.debounce_counter > config.confirmation_threshold) {
-      --state.debounce_counter;
-    }
+    // Decrement towards confirmation, clamped to the thresholds (a long FAILED burst can't
+    // run the counter off to INT32_MIN and delay later healing).
+    state.debounce_counter = clamp_debounce_counter(state.debounce_counter - 1, config);
 
     // Add source if not already present
     state.reporting_sources.insert(source_id);
@@ -177,11 +207,9 @@ bool InMemoryFaultStorage::report_fault_event(const std::string & fault_code, ui
     // PASSED event
     state.last_passed_time = timestamp;
 
-    // Increment debounce counter (towards healing), clamped at the healing
-    // threshold so a heal heartbeat can't drive it off to INT32_MAX.
-    if (state.debounce_counter < config.healing_threshold) {
-      ++state.debounce_counter;
-    }
+    // Increment towards healing, clamped to the thresholds (a heal heartbeat can't run the
+    // counter off to INT32_MAX and then delay a real fault from confirming).
+    state.debounce_counter = clamp_debounce_counter(state.debounce_counter + 1, config);
   }
 
   // Update status based on debounce counter
@@ -435,6 +463,18 @@ std::vector<ros2_medkit_msgs::msg::Fault> InMemoryFaultStorage::get_all_faults()
   }
 
   return result;
+}
+
+size_t InMemoryFaultStorage::reclassify_healed_as_cleared() {
+  std::lock_guard<std::mutex> lock(mutex_);
+  size_t changed = 0;
+  for (auto & [code, state] : faults_) {
+    if (state.status == ros2_medkit_msgs::msg::Fault::STATUS_HEALED) {
+      state.status = ros2_medkit_msgs::msg::Fault::STATUS_CLEARED;
+      ++changed;
+    }
+  }
+  return changed;
 }
 
 }  // namespace ros2_medkit_fault_manager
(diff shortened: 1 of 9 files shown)


src/ros2_medkit_fault_manager/src/sqlite_fault_storage.cpp
```diff
@@ -395,8 +395,10 @@ bool SqliteFaultStorage::report_fault_event(const std::string & fault_code, uint
         ++new_count;
       }
 
-      // Decrement debounce counter with saturation
-      if (debounce_counter > std::numeric_limits<int32_t>::min()) {
+      // Decrement debounce counter, clamped at the confirmation threshold (lower
+      // bound). Without this floor it only stops at INT32_MIN, so after a long heal
+      // heartbeat the counter has to climb all the way back before a fault confirms.
```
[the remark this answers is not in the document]
mfaferek93: Good catch, reworded. This floor guards against a long burst of FAILED events driving the counter to INT32_MIN (which would then delay later healing), not a heal heartbeat.

src/ros2_medkit_fault_manager/src/fault_storage.cpp
```diff
@@ -143,8 +148,9 @@ bool InMemoryFaultStorage::report_fault_event(const std::string & fault_code, ui
       ++state.occurrence_count;
     }
 
-    // Decrement debounce counter (towards confirmation) with saturation
-    if (state.debounce_counter > std::numeric_limits<int32_t>::min()) {
+    // Decrement debounce counter (towards confirmation), clamped at the
+    // confirmation threshold so a heal heartbeat can't drive it off to INT32_MIN.
```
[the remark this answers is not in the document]
mfaferek93: Reworded - same as the SQLite side: the floor protects against a long FAILED burst hitting INT32_MIN and delaying later healing.

src/ros2_medkit_fault_manager/src/sqlite_fault_storage.cpp
```diff
@@ -448,14 +455,22 @@ bool SqliteFaultStorage::report_fault_event(const std::string & fault_code, uint
...
       }
     } else {
-      // PASSED event - increment debounce counter with saturation
-      if (debounce_counter < std::numeric_limits<int32_t>::max()) {
+      // PASSED event - increment debounce counter, clamped at the healing threshold
+      // (upper bound). Without this ceiling a periodic heal heartbeat on a healthy
+      // system drives the counter to INT32_MAX, so a real fault then takes a huge
+      // number of reports to confirm.
+      if (debounce_counter < config.healing_threshold) {
         ++debounce_counter;
```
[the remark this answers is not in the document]
mfaferek93: Intentional, and the docstring now says so. healing_threshold is the counter's upper bound and has to apply even when healing is off, otherwise PASSED events still run the counter away. healing_enabled only gates whether reaching the bound produces a HEALED status.

src/ros2_medkit_fault_manager/src/fault_storage.cpp
```diff
@@ -170,8 +176,9 @@ bool InMemoryFaultStorage::report_fault_event(const std::string & fault_code, ui
     // PASSED event
     state.last_passed_time = timestamp;
 
-    // Increment debounce counter (towards healing) with saturation
-    if (state.debounce_counter < std::numeric_limits<int32_t>::max()) {
+    // Increment debounce counter (towards healing), clamped at the healing
+    // threshold so a heal heartbeat can't drive it off to INT32_MAX.
+    if (state.debounce_counter < config.healing_threshold) {
       ++state.debounce_counter;
     }
```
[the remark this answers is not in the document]
mfaferek93: Same as the SQLite side - the upper clamp must apply regardless of healing_enabled to stop the runaway; documented in DebounceConfig.

src/ros2_medkit_fault_manager/test/test_sqlite_storage.cpp
```diff
@@ -514,24 +514,80 @@ TEST_F(SqliteFaultStorageTest, ConfirmationPersistsAfterReopen) {
...
+}
+
+// A confirmed fault must not be un-confirmed by a heal heartbeat when auto-healing
+// is off, and the counter must stay bounded.
+TEST_F(SqliteFaultStorageTest, ConfirmedFaultSurvivesHealHeartbeat) {
+  rclcpp::Clock clock;
+  DebounceConfig config;
+  config.confirmation_threshold = -1;
+  config.healing_enabled = false;  // healing_threshold (default 3) still bounds the counter
+
```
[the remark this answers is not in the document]
mfaferek93: Done - set healing_threshold explicitly so the test doesn't rely on the default.

src/ros2_medkit_fault_manager/test/test_fault_manager.cpp
```diff
@@ -451,20 +451,21 @@ TEST_F(FaultStorageTest, ClearedFaultReactivationRestartsDebounce) {
 
 TEST_F(FaultStorageTest, PassedEventIncrementsCounter) {
   rclcpp::Clock clock;
+  // confirmation_threshold = -3 so 2 FAILED stays PREFAILED (a confirmed fault is
+  // latched and would not move to PREPASSED on a heal).
+  DebounceConfig config;
+  config.confirmation_threshold = -3;
```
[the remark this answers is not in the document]
mfaferek93: Done - healing_threshold set explicitly here too.

src/ros2_medkit_fault_manager/src/sqlite_fault_storage.cpp
```diff
@@ -406,6 +409,11 @@ bool SqliteFaultStorage::report_fault_event(const std::string & fault_code, uint
         new_status = ros2_medkit_msgs::msg::Fault::STATUS_CONFIRMED;
       } else if (debounce_counter <= config.confirmation_threshold) {
         new_status = ros2_medkit_msgs::msg::Fault::STATUS_CONFIRMED;
+      } else if (current_status == ros2_medkit_msgs::msg::Fault::STATUS_HEALED) {
```
bburda: This branch latches only HEALED. But the in-memory update_status (fault_storage.cpp:58) latches both CONFIRMED and HEALED. The latch can leave a CONFIRMED fault with a positive counter. Example: with critical_immediate_confirm=true and confirmation_threshold=-3, a CRITICAL FAILED confirms the fault at counter -1. Then a few PASSED events raise the counter to +2, and the status stays CONFIRMED because of the latch. The next normal FAILED lowers the counter, but it is still positive. The code then reaches `counter > 0 -> PREPASSED` (line 420). So one FAILED changes a confirmed fault to PREPASSED, and it disappears from the default fault list (this list returns only CONFIRMED). Now the sqlite backend gives a different result than the in-memory backend for the same events. Please do not just add one more latch in this branch. The same state machine is written three times, and that is why it became different. Move the logic from the in-memory version (it is the correct one) into one function compute_debounce_status(counter, current_status, config), and call it from both sqlite branches and from update_status. Then both backends always behave the same.

src/ros2_medkit_fault_manager/src/sqlite_fault_storage.cpp
```diff
@@ -448,14 +456,22 @@ bool SqliteFaultStorage::report_fault_event(const std::string & fault_code, uint
...
+      // system drives the counter to INT32_MAX, so a real fault then takes a huge
+      // number of reports to confirm.
+      if (debounce_counter < config.healing_threshold) {
         ++debounce_counter;
       }
 
       std::string new_status = current_status;
       if (config.healing_enabled && debounce_counter >= config.healing_threshold) {
         new_status = ros2_medkit_msgs::msg::Fault::STATUS_HEALED;
+      } else if (current_status == ros2_medkit_msgs::msg::Fault::STATUS_CONFIRMED) {
```
bburda: Same problem as the FAILED branch, but the other way around. This branch latches only CONFIRMED. So a HEALED fault with a negative counter (FAILED events lowered the counter while HEALED was latched) takes one PASSED, the counter goes up but is still negative, and the code reaches `counter < 0 -> PREFAILED` (line 476). The in-memory backend keeps it HEALED. The shared compute_debounce_status from the comment above also removes this branch and fixes it.

src/ros2_medkit_fault_manager/src/sqlite_fault_storage.cpp
```diff
@@ -395,8 +395,11 @@ bool SqliteFaultStorage::report_fault_event(const std::string & fault_code, uint
         ++new_count;
       }
 
-      // Decrement debounce counter with saturation
-      if (debounce_counter > std::numeric_limits<int32_t>::min()) {
+      // Decrement debounce counter, clamped at the confirmation threshold (lower
+      // bound). Without this floor a long burst of FAILED events drives it to
+      // INT32_MIN, and a later heal then has to climb all the way back before the
+      // fault can clear.
+      if (debounce_counter > config.confirmation_threshold) {
```
bburda: This clamp works in one direction only. It stops the counter from going further out of range, but it never brings a value that is already out of range back. A database from an older build has a runaway counter (the same problem this PR fixes). After an upgrade the counter needs very many events (up to about INT32) to come back into range. The counter is read at line 366 and used only inside this function. So please clamp it right after the read: `debounce_counter = clamp_debounce_counter(debounce_counter, config);`. This uses the same helper, and it also makes the next +1/-1 safe from overflow. The in-memory backend does not save state, so it is not affected.

src/ros2_medkit_fault_manager/test/test_sqlite_storage.cpp
```diff
@@ -514,24 +514,81 @@ TEST_F(SqliteFaultStorageTest, ConfirmationPersistsAfterReopen) {
...
+  DebounceConfig config;
+  config.confirmation_threshold = -1;
+  config.healing_enabled = false;
+  config.healing_threshold = 3;  // upper clamp on the counter, even with healing disabled
+
+  storage_->report_fault_event("FAULT_1", ReportFault::Request::EVENT_FAILED, Fault::SEVERITY_ERROR, "Test", "/node1",
+                               clock.now(), config);
+  ASSERT_EQ(storage_->get_fault("FAULT_1")->status, Fault::STATUS_CONFIRMED);
+
+  for (int i = 0; i < 20; ++i) {
```
bburda, line 588: This loop brings the fault to CONFIRMED with counter +3 and then stops. So it never sends the FAILED event that would show the missing CONFIRMED latch above. Please add one more FAILED here and check that the status is still CONFIRMED (now it becomes PREPASSED). Please also add the opposite case: a HEALED fault with a negative counter, then one PASSED, and check that it stays HEALED. Add both cases for the in-memory backend too, so both backends are tested the same way.

bburda (commented):
Notes that do not fit on a diff line:
- design/index.rst (Status Lifecycle) and README.md (How It Works) still say the status only depends on the counter value. The new latch breaks that rule. Now a CONFIRMED fault keeps its status while the counter is above the confirmation threshold. Please update both files to explain the counter clamp and the latch.
- Threshold checks only look at the sign of each value (fault_manager_node.cpp, entity_threshold_resolver.cpp). So confirmation_threshold == 0 or healing_threshold == 0 are accepted. With these values the counter gets stuck, and the two backends end with different status for the same config (the in-memory backend checks confirm first, the sqlite PASSED branch checks heal first). Please require confirmation_threshold < 0 < healing_threshold. Check the final merged config, not only the per-entity override, because the override is merged field by field. If the config is wrong, log a warning and use a safe default. Do not reject it.
- A HEALED row saved with healing_enabled=true and later read with healing_enabled=false changes in two opposite ways: a FAILED keeps HEALED (latch), but a PASSED changes it to PREPASSED. Please change a stale HEALED to CLEARED one time at startup when healing is disabled. This is a small UPDATE, no schema change.
- Please document the latch delay in the DebounceConfig comment and the design doc. A fault that becomes active again stays out of the default list (CONFIRMED only) for up to healing_threshold - confirmation_threshold reports. During this time occurrence_count and last_occurred still show the activity.

src/ros2_medkit_fault_manager/src/sqlite_fault_storage.cpp
mfaferek93: Done. Extracted compute_debounce_status(counter, current_status, config) and call it from both SQLite branches and the in-memory update_status, so CONFIRMED/HEALED latch identically on both backends. Your CRITICAL + positive-counter example is now a regression test (ConfirmedLatchSurvivesFailedAtPositiveCounter) on both.

src/ros2_medkit_fault_manager/src/sqlite_fault_storage.cpp
mfaferek93: Fixed by the same shared function - the PASSED branch no longer carries its own latch, so a HEALED fault with a negative counter stays HEALED. Covered by HealedLatchSurvivesPassedAtNegativeCounter on both backends.

src/ros2_medkit_fault_manager/src/sqlite_fault_storage.cpp
mfaferek93: Done. clamp_debounce_counter() now runs right after reading the counter, so a runaway value from an older DB is pulled back into range on first touch and re-confirms in a bounded number of events. Regression: RunawayCounterFromOldDbRecovers (pokes 100000 into the DB, confirms in 4).

src/ros2_medkit_fault_manager/test/test_sqlite_storage.cpp
mfaferek93: Added. The test now sends the extra FAILED and asserts still-CONFIRMED, plus the opposite HEALED + PASSED-at-negative case, on both backends.

mfaferek93:
Thanks, all four notes addressed in the latest commit:
- Docs: design/index.rst (Status Lifecycle) and README (How It Works) now describe the counter clamp, the latch, and the re-confirmation delay.
- Threshold validation: the node validates the merged config (confirmation_threshold < 0 < healing_threshold), logs a warning and falls back to safe defaults instead of rejecting; per-entity overrides are checked against the global config at load, and sanitize_debounce_config also runs defensively per resolve.
- Stale HEALED: reclassify_healed_as_cleared() runs once at startup when healing is disabled (UPDATE only, no schema change).
- Latch delay: documented in the DebounceConfig comment and the design doc.

bburda (dismissed):

mfaferek93:
Heads up on the threshold validation: I relaxed it to `confirmation_threshold < 0 <= healing_threshold` rather than strictly `> 0`. `healing_threshold == 0` is a valid, intentional config (action_status_bridge uses it to heal on a single PASSED event); since both backends now share `compute_debounce_status`, 0 no longer causes the backend divergence - the counter reaches the threshold at 0 and heals. Requiring `> 0` regressed the action_status_bridge integration test. Added a regression test for the single-event-heal case.

bburda (approved):

Issue #480, opened 2026-06-30
Gateway secure-config option + hardening profile


mfaferek93:
Provide a documented secure config preset wiring the existing JWT/RBAC/TLS: `auth.enabled`, `require_auth_for=all`, TLS, CORS restricted to explicit origins, rate limiting; plus a hardening checklist. The gateway default stays dev-friendly (open, no auth); the secure preset is opt-in and is what packaged deployments apply.

- Acceptance: enabling the preset refuses unauthenticated writes and serves over TLS; the default (no preset) stays dev-friendly.

mfaferek93:
Landed on `main` via #485. Opt-in secure field profile added: `config/gateway_params.secure.yaml` (JWT auth, TLS, restricted CORS, token-bucket rate limiting; disabled by default, fails closed when enabled with invalid config) plus `design/hardening.rst` checklist. Closing as done.

Issue #479, opened 2026-06-30
Multi-alarm mapping + alarm text


mfaferek93:
Add `event_alarms[*].mappings` routing one source's events to distinct fault codes by condition identity (ConditionName/SourceNode/EventType, first-match, source-level fallback). Populate fault description from the event Message + associated values. Add ConditionName to the EventFilter.

- Acceptance: two distinct alarm conditions on one source surface as two distinct faults with their configured texts + associated values.

mfaferek93:
Landed on `main` via #485, with follow-ups #506 (message-substring routing) and #501 (severity alias). `AlarmMapping` maps multiple alarms per source by condition identity (`condition_name`/`source_node`/`event_type`) plus a `match_message` case-sensitive substring matcher and `associated_values` for extra event fields; see `node_map.cpp::resolve_alarm`. Closing as done.

Issue #478, opened 2026-06-30
Active-condition replay on reconnect


mfaferek93:
Add `condition_replay_strategy` (method/read/auto/off). On `ConditionRefresh` rejection, browse the alarm sources, read ActiveState/Retain, drive the condition state machine, and reconcile the fault set on (re)subscribe.

- Acceptance: trigger an alarm, restart the gateway, the active alarm is reported again without a new transition.

mfaferek93:
Landed on `main` via #485 (merged 2026-07-02, commit 18e0935). The OPC UA client now performs active-condition replay on reconnect via the `condition_replay_strategy` parameter (`method`/`read`/`auto`/`off`): `ConditionRefresh` (Server method i=3875) is invoked on subscribe and every reconnect, with a read-based fallback that does not false-clear on a failed scan. See src/ros2_medkit_plugins/ros2_medkit_opcua README section "Active-condition replay on reconnect (issue #389/#478)". Closing as done (#485 wrote a multi-issue Closes line so GitHub only auto-closed #477).

PR #485, opened 2026-06-30
OPC UA + gateway hardening (security, reconnect replay, multi-alarm, secure profile)

### LINKED ISSUE #477: OPC UA client security (SecurityPolicy, certificates, user auth)
Add to the OPC UA client config + connect path:

- SecurityPolicy (None/Basic256Sha256/Aes128Sha256RsaOaep/Aes256Sha256RsaPss), MessageSecurityMode (None/Sign/SignAndEncrypt), client application-instance certificate (cert+key paths, application URI), server certificate trust store + reject-untrusted, user identity (Anonymous/Username-Password/X.509). Anonymous/None stays opt-in.
- Acceptance: connect to a server with Basic256Sha256 + SignAndEncrypt + username/password and a trusted client cert; read tags + receive events; Anonymous/None still works.

mfaferek93:
Closes #477 #478 #479 #480

- OPC UA client security: SecurityPolicy / MessageSecurityMode / client application-instance cert / server trust store / user identity (anonymous, username-password, X.509). Anonymous + None stay the default.
- Active-condition replay on reconnect with a configurable strategy; read-based fallback for servers that reject ConditionRefresh (does not false-clear on a failed scan).
- Multi-alarm mapping by condition identity (ConditionName / SourceNode / EventType) + associated values.
- Opt-in secure config profile (auth/TLS/CORS/rate-limit) for field deployments; fails closed when enabled with invalid config.

Tested: 196 unit tests pass. Real-PLC acceptance for the security/replay/alarm paths is pending hardware validation.

### DIFF
--- a/src/ros2_medkit_gateway/design/hardening.rst
+++ b/src/ros2_medkit_gateway/design/hardening.rst
@@ -1,0 +1,104 @@
+Gateway hardening (secure field profile)
+========================================
+
+The gateway ships every transport and access control needed for a hardened
+deployment - JWT authentication with RBAC, TLS/HTTPS, restricted CORS, and
+token-bucket rate limiting - but they are **disabled by default** so local
+development works out of the box. A gateway exposed on a plant network with the
+defaults is wide open: unauthenticated reads and writes over cleartext HTTP.
+
+For any deployment reachable from an untrusted network, start from the secure
+field profile preset ``config/gateway_params.secure.yaml`` instead of
+``config/gateway_params.yaml``:
+
+.. code-block:: bash
+
+   ros2 run ros2_medkit_gateway gateway_node \
+     --ros-args --params-file gateway_params.secure.yaml \
+     -p auth.jwt_secret:="$MEDKIT_JWT_SECRET" \
+     -p 'auth.clients:=["operator:'"$OP_SECRET"':operator"]'
+
+What the secure profile turns on
+--------------------------------
+
+============================ ============== ===========================================
+Control                      Default        Secure profile
+============================ ============== ===========================================
+``auth.enabled``             false          true
+``auth.require_auth_for``    write          all (auth on reads + writes)
+``server.tls.enabled``       false          true (HTTPS, min TLS 1.3)
+``cors.allowed_origins``     ``[""]``       explicit origin list (no wildcard)
+``rate_limiting.enabled``    false          true (global + per-client + per-endpoint)
+``scripts.allow_uploads``   true           false (manifest-defined scripts only)
+``docs.enabled``             true           false (reduced surface)
+``bulk_data.max_upload``     100 MiB        25 MiB
+``locking`` on operations    none           lock required before mutation
+============================ ============== ===========================================
+
+Credential and certificate provisioning
+----------------------------------------
+
+1. **TLS certificate.** Provision a real server certificate + private key and
+   point ``server.tls.cert_file`` / ``server.tls.key_file`` at them. The key
+   file must be ``chmod 600`` and owned by the gateway service user. For a
+   dev/test box only, ``scripts/generate_dev_certs.sh`` emits a self-signed
+   ``cert.pem`` / ``key.pem`` / ``ca.pem`` (never use these in production).
+
+2. **JWT secret.** Generate a high-entropy secret of at least 32 characters
+   (HS256) or provision an RS256 key pair. Inject it at deploy time from a
+   secret store or environment variable - do not commit it to source control.
+
+3. **Role-scoped clients.** Create the minimum set of clients in
+   ``auth.clients`` (``client_id:client_secret:role``). Roles, least to most
+   privileged: ``viewer`` (read), ``operator`` (+ trigger ops / ack faults /
+   publish), ``configurator`` (+ modify configs), ``admin`` (+ auth
+   management). Rotate secrets periodically.
+
+4. **Obtain a token** and call the API over HTTPS:
+
+   .. code-block:: bash
+
+      curl -sk -X POST https://gateway:8443/api/v1/auth/authorize \
+        -H 'Content-Type: application/json' \
+        -d '{"client_id":"operator","client_secret":"...","grant_type":"client_credentials"}'
+      # use the returned access_token:
+      curl -sk https://gateway:8443/api/v1/faults -H "Authorization: Bearer $TOKEN"
+
+Hardening checklist
+-------------------
+
+Before exposing a gateway on a shared / plant network, confirm:
+
+- [ ] ``auth.enabled: true`` and ``auth.require_auth_for`` is ``all`` (or
+  ``write`` only if unauthenticated reads are explicitly acceptable).
+- [ ] ``auth.jwt_secret`` is set to a >= 32-char secret injected from a secret
+  store (not the placeholder, not in version control).
+- [ ] ``auth.clients`` lists only the role-scoped clients you need; default /
+  example credentials removed; secrets rotated.
+- [ ] ``server.tls.enabled: true`` with a real certificate; private key is
+  ``chmod 600``; ``min_version`` is ``1.3`` (or ``1.2`` only for legacy
+  clients).
+- [ ] ``cors.allowed_origins`` is an explicit list (no ``*``); ``*`` is never
+  combined with ``allow_credentials: true``.
+- [ ] ``rate_limiting.enabled: true`` with per-client and mutating-endpoint
+  limits tuned to the deployment.
+- [ ] ``scripts.allow_uploads: false`` unless remote script upload is a
+  required, reviewed capability.
+- [ ] ``bulk_data.max_upload_size`` bounded to what the deployment needs.
+- [ ] If peer aggregation is used: ``aggregation.require_tls: true`` and
+  ``forward_auth`` only enabled when every peer is trusted.
+- [ ] Bind ``server.host`` to a management interface where the network layout
+  allows, and place the gateway behind the plant firewall / segmentation.
+- [ ] Back the gateway with persistent storage on a volume with restricted
+  permissions (faults DB, triggers DB, rosbag snapshots).
+
+OPC-UA plugin (southbound) hardening
+------------------------------------
+
+The gateway controls the northbound REST surface; the OPC-UA plugin controls
+the southbound connection to the PLC. Harden both. The plugin supports
+SecurityPolicy (Basic256Sha256 / Aes128 / Aes256), MessageSecurityMode
+(Sign / SignAndEncrypt), a client application-instance certificate, a server
+trust store with reject-untrusted, and user identity (anonymous /
+username-password / X.509). See ``ros2_medkit_opcua`` README, section
+"OPC-UA client security".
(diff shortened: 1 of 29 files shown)


bburda (commented):

- `ros2_medkit_opcua/package.xml` has no `test_depend`/`exec_depend` on `ros2_medkit_fault_manager`, but the new `test_opcua_secured.test.py` launches `fault_manager_node`. The secured CI job passes only because it explicitly names `--packages-up-to ... ros2_medkit_fault_manager`; per the project's integration-test dependency rule, an isolated per-package build would not guarantee fault_manager is built first and the required secured gate would fail. Add `<test_depend>ros2_medkit_fault_manager</test_depend>`.

- The C event trampoline `on_event_trampoline_c` (opcua_client.cpp) calls the user callback with no try/catch, and `run_iterate` catches only `opcua::BadStatus`. This PR adds new throwing work reachable from that callback (`resolve_alarm`, associated-value string concatenation in `on_event`, `conditions_.emplace`), so a `std::bad_alloc` or any non-BadStatus exception unwinds through open62541's C frames - undefined behavior. Wrap the callback dispatch in `try { ... } catch (...) { /* log, swallow */ }`.

src/ros2_medkit_plugins/ros2_medkit_opcua/src/opcua_poller.cpp
```diff
@@ -266,61 +324,284 @@ void OpcuaPoller::setup_event_subscriptions() {
...
+        condition_refresh_warned_ = true;
+      } else if (condition_refresh_warned_) {
+        condition_refresh_warned_ = false;
+      }
+      return;
+    case ConditionReplayStrategy::Read:
+      read_fallback_replay();
+      return;
+    case ConditionReplayStrategy::Auto:
+      if (try_condition_refresh()) {
```
bburda, line 379: On the Method and Auto-success paths a successful `ConditionRefresh` returns here with no reconcile step; only the read-fallback path runs `reconcile_after_read`. ConditionRefresh re-reports only currently-active (Retain=true) conditions, and `conditions_` is not cleared on reconnect, so a condition that cleared entirely during the outage is never replayed and never reconciled - its SOVD fault stays CONFIRMED until a manual clear. Auto is the default and any ConditionRefresh-capable server (the open62541 reference server, S7-1500) takes this success path, so the default config stuck-actives alarms that clear while offline. Reconcile the tracked set against the refresh result on the success paths too.

src/ros2_medkit_plugins/ros2_medkit_opcua/src/node_map.cpp
```diff
@@ -364,7 +415,12 @@ bool NodeMap::load(const std::string & yaml_path) {
     {
       std::set<std::pair<std::string, std::string>> event_keys;
       for (const auto & cfg : event_alarms_) {
-        event_keys.emplace(cfg.entity_id, cfg.fault_code);
+        if (!cfg.fault_code.empty()) {
+          event_keys.emplace(cfg.entity_id, cfg.fault_code);
+        }
+        for (const auto & m : cfg.mappings) {
+          event_keys.emplace(cfg.entity_id, m.fault_code);
```
bburda: This guard checks detection-vs-event collisions but not event-vs-event: two conditions that resolve to the same `(entity_id, fault_code)` - via a catch-all source-level `fault_code`, or two mappings sharing a code - are allowed. Clears go by `fault_code` only (`send_clear_fault(fault_code)`), so when one such condition clears it clears the shared SOVD fault while the sibling is still active; the sibling's runtime is already CONFIRMED, so no re-raise fires and the active alarm silently disappears from `/faults`. Reject a `fault_code` shared by more than one condition identity under one source, or make the clear path condition/entity-scoped.

src/ros2_medkit_plugins/ros2_medkit_opcua/src/opcua_client.cpp
```diff
@@ -199,6 +226,216 @@ OpcuaClient::~OpcuaClient() {
...
+      trust.push_back(std::move(b));
+    }
+
+    opcua::ClientConfig cc(cert, key, trust, {});
+
+    // Force the requested SecurityPolicy. An empty URI lets open62541 pick the
+    // highest the server offers; we want explicit selection so a downgraded
+    // server endpoint cannot silently weaken the channel.
+    const std::string policy_uri = OpcuaClient::security_policy_uri(cfg.security_policy);
+    if (cfg.security_policy != SecurityPolicy::None && !policy_uri.empty()) {
```
bburda, line 413: The policy URI is pinned only when `security_policy != None`, but `security_mode` is applied unconditionally and `requires_secure_channel` is true if either is non-None. So `security_mode: Sign`/`SignAndEncrypt` with `security_policy` left at the default `None` enters the secure branch yet leaves `securityPolicyUri` unset, and open62541 negotiates a server-chosen policy - the exact silent downgrade the comment above says this pin prevents. Reject the mixed case: a non-None mode requires a non-None policy and vice versa.

src/ros2_medkit_plugins/ros2_medkit_opcua/src/opcua_client.cpp
```diff
@@ -199,6 +226,216 @@ OpcuaClient::~OpcuaClient() {
...
+    // The application URI must match the URI SAN in the client certificate or
+    // the server rejects the channel with BadCertificateUriInvalid.
+    if (!cfg.application_uri.empty()) {
+      UA_String_clear(&cc.handle()->clientDescription.applicationUri);
+      cc.handle()->clientDescription.applicationUri = UA_STRING_ALLOC(cfg.application_uri.c_str());
+    }
+
+    // Trust handling: by default validate the server certificate against the
+    // trust list. When reject_untrusted is false, accept any server
+    // certificate (lab / trust-on-first-use only).
```
bburda: `UA_CertificateVerification_AcceptAll` (line 403) accepts any server certificate on every connection and disables expiry, chain, hostname, and URI-SAN checks - that is accept-anything, not trust-on-first-use. TOFU pins the cert on first sight and rejects a changed cert afterwards; this accepts a different attacker cert on each reconnect, so `SignAndEncrypt` over it is a channel encrypted to a MITM. The same "TOFU / trust-on-first-use" wording is in opcua_client.hpp and the README. Relabel it "accepts any server certificate, no authentication (MITM-exposed), lab only".

src/ros2_medkit_gateway/design/hardening.rst
```diff
@@ -0,0 +1,104 @@
...
+Credential and certificate provisioning
+----------------------------------------
+
+1. **TLS certificate.** Provision a real server certificate + private key and
+   point ``server.tls.cert_file`` / ``server.tls.key_file`` at them. The key
+   file must be ``chmod 600`` and owned by the gateway service user. For a
+   dev/test box only, ``scripts/generate_dev_certs.sh`` emits a self-signed
+   ``cert.pem`` / ``key.pem`` / ``ca.pem`` (never use these in production).
+
+2. **JWT secret.** Generate a high-entropy secret of at least 32 characters
```
bburda, line 47: This covers not committing the JWT secret and injecting it from a secret store or env, but not that a secret set as a ROS 2 parameter is readable over the unauthenticated DDS control plane: any DDS peer on the domain can `ros2 param get` it, and it is published on `/parameter_events` at declare/set time. The secure profile hardens only the northbound HTTP surface, not DDS. This applies to `auth.jwt_secret` and to the OPC UA plugin's `plugins.opcua.password`. Recommend env-var / secret-store injection (the OPC UA plugin already supports `OPCUA_PASSWORD`) and a note on DDS control-plane hardening (SROS2).

src/ros2_medkit_gateway/design/hardening.rst
```diff
@@ -0,0 +1,104 @@
...
+What the secure profile turns on
+--------------------------------
+
+================================ ============== ===========================================
+Control                          Default        Secure profile
+================================ ============== ===========================================
+``auth.enabled``                 false          true
+``auth.require_auth_for``        write          all (auth on reads + writes)
+``server.tls.enabled``           false          true (HTTPS, min TLS 1.3)
+``cors.allowed_origins``         ``[""]``       explicit origin list (no wildcard)
```
bburda: The documented default here is `[""]`, but the code declares `cors.allowed_origins` as an empty list `[]` (`gateway_node.cpp`, `declare_parameter("cors.allowed_origins", std::vector<std::string>{})`). Align the doc with the actual default.

src/ros2_medkit_plugins/ros2_medkit_opcua/src/opcua_client.cpp
mfaferek93: Relabeled in 2ed1695 / 0dec989: the accept-any block and README no longer call it trust-on-first-use - it is now labeled accept-any (INSECURE, lab only), nothing is pinned. The secure default (reject_untrusted=true with a trust list) is unchanged; real TOFU pinning was not added since the code never functionally claimed it.

src/ros2_medkit_plugins/ros2_medkit_opcua/src/opcua_client.cpp
mfaferek93: Fixed in 2ed1695: apply_security_config now rejects a contradictory pairing (exactly one of security_policy / security_mode being None) with an error before connecting, so Sign/SignAndEncrypt with policy None no longer passes with the URI left unpinned.

src/ros2_medkit_gateway/design/hardening.rst
mfaferek93: Fixed in 2ed1695: added a warning that a secret set as a ROS 2 parameter is readable over the unauthenticated DDS control plane (and via ps/cmdline); it recommends env-var injection plus SROS2 access control or a firewalled ROS_DOMAIN_ID.

src/ros2_medkit_plugins/ros2_medkit_opcua/src/opcua_poller.cpp
mfaferek93: Fixed in 2ed1695: on_event now accumulates the replayed ConditionIds between RefreshStart and RefreshEnd and reconciles on RefreshEnd, clearing tracked-active conditions absent from the burst. Not a naive post-refresh reconcile, since the refresh events arrive asynchronously.

src/ros2_medkit_plugins/ros2_medkit_opcua/src/node_map.cpp
mfaferek93: Fixed in 2ed1695: the guard now also rejects a duplicate (entity_id, fault_code) among event_alarms (a counting check, not a dedup set), in addition to the detection-vs-event cross-check.

src/ros2_medkit_gateway/design/hardening.rst
mfaferek93: Fixed in 2ed1695: corrected the cors.allowed_origins default in the doc from [""] to [] to match the code.

bburda (dismissed):

bburda (approved):

PR #486, opened 2026-06-30
Shared fault-detection: threshold + status-bit + fault-code enum

### LINKED ISSUE #481: Shared fault-detection: threshold + status-bit + fault-code enum
A reusable fault-detection module so any value source maps raw values to faults the same way, matching how supervisory systems read device faults.

- Detection modes per mapped point, composable in the node-map: numeric threshold (above/below), status-word bit decode (named bit -> fault), fault-code value enum mapping (code -> fault code + text).
- One evaluator: given a value + the point's detection config, decide raise/clear + fault code/severity/text. Reused by the OPC UA plugin and other protocol plugins.
- Migrate the OPC UA plugin's existing threshold logic onto this evaluator so it gains bit-decode + enum for free.
- Acceptance: a status-word bit and a fault-code enum each raise a fault with the mapped code/text from the shared evaluator; numeric threshold still works; unit-tested.

mfaferek93:
Closes #481

New header-only `ros2_medkit_fault_detection` evaluator with three composable modes (numeric threshold, status-word bit decode, fault-code enum) and an edge-detecting transition tracker. The OPC UA plugin is migrated onto it and gains `status_bits` / `fault_enum` node-map blocks (no threshold regression). Load-time global fault-code uniqueness prevents cross-entry flapping and extends the event_alarms collision check.

Tested: 56 unit tests incl. poller-level integration.

### DIFF
--- a/src/ros2_medkit_plugins/ros2_medkit_opcua/README.md
+++ b/src/ros2_medkit_plugins/ros2_medkit_opcua/README.md
@@ -169,12 +169,47 @@ nodes:
     writable: true                # Allow writes via x-plc-operations
     min_value: 0.0                # Optional: range validation for writes
     max_value: 100.0
-    alarm:                        # Optional: map to SOVD fault
+    alarm:                        # Optional: numeric threshold -> SOVD fault
       fault_code: PLC_LOW_LEVEL
       severity: WARNING
       message: Tank level below minimum
       threshold: 100.0
       above_threshold: false      # Alarm when value < threshold
+
+  # Status-word bit decode: one integer register, one fault per named bit.
+  # Mutually exclusive per node with `alarm:` / `fault_enum:`.
+  - node_id: "ns=2;s=Pump.StatusWord"
+    entity_id: pump
+    data_name: status_word
+    data_type: int
+    status_bits:
+      - bit: 3                     # bit position, 0 = least significant
+        fault_code: PUMP_OVERLOAD
+        severity: ERROR
+        message: Pump motor overload
+      - bit: 7
+        fault_code: FILTER_DIRTY
+        severity: WARNING          # message defaults to the fault_code
+
+  # Fault-code enum: a register value selects at most one fault + text.
+  - node_id: "ns=2;s=Vfd.FaultCode"
+    entity_id: vfd
+    data_name: fault_code
+    data_type: int
+    fault_enum:
+      ok_value: 0                  # value meaning "no fault" (default 0)
+      codes:
+        - code: 10
+          fault_code: VFD_OVERVOLTAGE
+          severity: ERROR
+          message: DC bus overvoltage
+        - code: 11
+          fault_code: VFD_OVERCURRENT
+          severity: ERROR
+
+# The three modes (`alarm`, `status_bits`, `fault_enum`) are evaluated by the
+# shared `ros2_medkit_fault_detection` module so every protocol plugin maps raw
+# values to faults identically. They are mutually exclusive on a single node.
 
 # Native OPC-UA AlarmConditionType events (issue #386). Subscribes to alarms
 # defined inside the PLC (Siemens Program_Alarm / ProDiag, Beckhoff TF6100,
--- a/src/ros2_medkit_plugins/ros2_medkit_opcua/README.md
+++ b/src/ros2_medkit_plugins/ros2_medkit_opcua/README.md
@@ -210,6 +210,15 @@ nodes:
 # The three modes (`alarm`, `status_bits`, `fault_enum`) are evaluated by the
 # shared `ros2_medkit_fault_detection` module so every protocol plugin maps raw
 # values to faults identically. They are mutually exclusive on a single node.
+#
+# Status words: expose the register as an UNSIGNED OPC-UA type (Byte / UInt16 /
+# UInt32 / UInt64). A signed type (Int16/Int32) whose top bit is set is
+# sign-extended when widened to the 64-bit decode register, which sets every
+# bit above the register width and raises spurious bit faults. `bit:` positions
+# are 0-based and must be < 64; higher positions are dead config and rejected at
+# load. Fault codes must be globally unique across all `alarm` / `status_bits` /
+# `fault_enum` entries (the loader rejects duplicates); reusing a code on two
+# nodes makes its fault flap.
 
 # Native OPC-UA AlarmConditionType events (issue #386). Subscribes to alarms
 # defined inside the PLC (Siemens Program_Alarm / ProDiag, Beckhoff TF6100,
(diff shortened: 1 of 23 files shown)


src/ros2_medkit_plugins/ros2_medkit_opcua/src/node_map.cpp
```diff
@@ -353,25 +545,81 @@ bool NodeMap::load(const std::string & yaml_path) {
...
+      std::set<std::string> seen_codes;
+      for (const auto & fault : detection_faults) {
+        if (!seen_codes.insert(fault.second).second) {
+          RCLCPP_ERROR(rclcpp::get_logger("opcua.node_map"),
+                       "fault_code '%s' is emitted by more than one detection entry "
+                       "(threshold/status_bits/fault_enum); fault codes must be globally unique because "
+                       "the shared fault tracker is keyed by code alone (a collision flaps raise/clear "
+                       "every poll cycle) - rename one of them",
+                       fault.second.c_str());
+          return false;
```
bburda, line 677: This `return false` (and the detection-mode reject at ~370 and the collision reject below) is ignored by the caller: `OpcuaPlugin::configure()` only logs on a failed `load()` and starts the poller anyway (opcua_plugin.cpp:145-151, no early return). By the time this uniqueness check runs, `entries_` is already fully populated, so a duplicate-code map still runs and the shared tracker flaps raise/clear every poll cycle - exactly what this check is meant to prevent, and contrary to the README line "the loader rejects duplicates". A failed load also skips `build_entity_defs()` while `entity_defs_` is not cleared, so the poller raises faults for entities that SOVD discovery does not expose. Make a failed `load()` fail plugin bring-up (or refuse to start the poller).

src/ros2_medkit_plugins/ros2_medkit_opcua/src/node_map.cpp
```diff
@@ -290,14 +355,141 @@ bool NodeMap::load(const std::string & yaml_path) {
...
+              continue;
+            }
+            if (rule.width > 0 && br.bit >= rule.width) {
+              RCLCPP_WARN(logger,
+                          "status_bits entry on node_id=%s has bit=%u >= status_word_width=%u; "
+                          "this fault can never fire - skipping",
+                          entry.node_id_str.c_str(), br.bit, rule.width);
+              continue;
+            }
+            br.fault = {b["fault_code"].as<std::string>(),
```
bburda: This `.as<std::string>()` (and the `severity`/`message` beside it, the `fault_enum` string fields at ~470, and the existing `alarm.fault_code` at ~375) is unguarded, so a missing or wrong-typed string field throws `YAML::TypedBadConversion`, hits the outer catch, and aborts the whole file load - discarding every other valid point on the PLC. The numeric `bit`/`code` fields already skip-with-warn (~415, ~465); do the same for the string fields so one malformed entry does not take down the entire node map. Combined with the ignored `load()` failure above, one bad field leaves the plugin running a partial map with no discovery.

bburda (commented):

- The new package `ros2_medkit_fault_detection` has no `CHANGELOG.rst`, and the PR does not wire it into `docs/changelog.rst` (which aggregates every package changelog) or `docs/design/index.rst`. Every other package in the workspace ships a CHANGELOG; add one plus its include so the release/docs aggregation stays complete.

src/ros2_medkit_plugins/ros2_medkit_opcua/src/node_map.cpp
```diff
@@ -353,25 +545,81 @@ bool NodeMap::load(const std::string & yaml_path) {
...
+
+      // 2. Cross-pipeline collision with event_alarms by (entity_id, fault_code).
       std::set<std::pair<std::string, std::string>> event_keys;
       for (const auto & cfg : event_alarms_) {
         event_keys.emplace(cfg.entity_id, cfg.fault_code);
       }
-      for (const auto & entry : entries_) {
-        if (entry.alarm.has_value() && event_keys.count({entry.entity_id, entry.alarm->fault_code}) > 0) {
+      for (const auto & [entity_id, code] : detection_faults) {
+        if (event_keys.count({entity_id, code}) > 0) {
```
bburda: This cross-pipeline guard only rejects the same `(entity_id, fault_code)`. A polled detection code X on `entity_a` and an `event_alarms` code X on `entity_b` both pass it, but the fault manager keys and clears faults by `fault_code` alone (`clear_fault(fault_code)`, `get_fault(fault_code)`), so both pipelines drive the one global fault X and a detection clear wipes the event's fault - the exact flap this block says it prevents. Reject any code shared between the detection set and `event_alarms` regardless of entity, or make the clear path entity-scoped.

src/ros2_medkit_plugins/ros2_medkit_opcua/README.md
```diff
@@ -169,13 +169,58 @@ nodes:
...
+          message: DC bus overvoltage
+        - code: 11
+          fault_code: VFD_OVERCURRENT
+          severity: ERROR
+
+# The three modes (`alarm`, `status_bits`, `fault_enum`) are evaluated by the
+# shared `ros2_medkit_fault_detection` module so every protocol plugin maps raw
+# values to faults identically. They are mutually exclusive on a single node.
+#
+# Status words: expose the register as an UNSIGNED OPC-UA type (Byte / UInt16 /
```
bburda: This section tells operators to change the OPC-UA type to avoid sign-extension, but the loader also accepts a `status_word_width:` key (node_map.cpp:393) that masks the high bits in-config, and it is documented nowhere. The `fault_enum` catch-all override keys `unknown_fault_code` / `unknown_severity` / `unknown_message` (node_map.cpp:480-485) are also undocumented, so an operator sees an auto-derived `<ENTITY>_<DATA>_UNMAPPED` fault with no explanation. Document these keys.

src/ros2_medkit_plugins/ros2_medkit_opcua/src/node_map.cpp
mfaferek93: Fixed in c2798e4: configure() now throws on load failure; PluginManager disables the plugin, so the poller never starts on a rejected map.

src/ros2_medkit_plugins/ros2_medkit_opcua/src/node_map.cpp
mfaferek93: Fixed in c2798e4: a guarded parse_string() warns and skips that node/rule instead of aborting the whole file.

src/ros2_medkit_plugins/ros2_medkit_opcua/src/node_map.cpp
mfaferek93: Fixed in c2798e4: uniqueness is now global by fault_code across all detection codes and event_alarms, entity ignored (matching the code-only keying in the fault manager).

src/ros2_medkit_plugins/ros2_medkit_opcua/README.md
mfaferek93: Fixed in c2798e4: documented status_word_width and the unknown_fault_code / unknown_severity / unknown_message overrides.

bburda (approved):

PR #487, opened 2026-06-30
Tamper-evident append-only fault audit log

### LINKED ISSUE #483: Tamper-evident append-only fault audit log
Persist an append-only, hash-chained audit log of fault state transitions so the fault history is verifiable and any edit or deletion is detectable.

- Append-only table, one immutable row per fault state-transition (occurred/confirmed/cleared/ack); no UPDATE/DELETE.
- On insert compute `record_hash = sha256(prev_hash + canonical(event))`; store `record_hash` + `prev_hash` + monotonic `seq`; maintain a persisted chain head. Hash is set once at insert, never recomputed.
- Provide a verify routine over the persisted chain and a read API returning the ordered log + current head.
- Configurable: enable flag (default off, since it adds write + storage overhead); which transitions to log (all vs confirmed-only); retention/rotation that seals a segment (persisting its final head as an anchor) before pruning, so the remaining history stays verifiable.
- Acceptance: each fault transition appends a chained row when enabled; verify confirms an untampered chain; editing or deleting any past row makes verify fail; with the feature off there is no audit-log overhead.

mfaferek93:
Closes #483

Optional (off by default) append-only, hash-chained audit log of fault state transitions in the fault manager: one immutable row per transition, `sha256(prev_hash + canonical(event))` with a persisted chain head, a verify routine, retention that seals a segment anchor before pruning, and append-only triggers.

Threat model is documented honestly: the chain is unkeyed and lives in a single writable file, so it detects non-recomputing tampering and accidental loss; a key/signature or external anchoring is a follow-up.

Tested: 121 unit tests incl. truncation and tamper cases.


src/ros2_medkit_fault_manager/README.md
```diff
@@ -109,6 +110,28 @@ patterns:
...
+
+**Completeness is an integrity property.** `verify()` proves nothing was *deleted* from the chain, but it cannot prove a transition that was *never appended*. So a silently dropped append is a hole `verify()` can never see. Every transition on the write path is therefore audited (occurred, timer/threshold confirmations, auto-heal, and clears), and an append failure is never swallowed silently: it increments a dropped-writes health counter and clears an "audit healthy" flag. With `audit_log.fail_closed` set, an append failure is a **hard error** that aborts the operation, so a compliance-strict deployment learns the audit broke instead of losing records unnoticed. The default (`fail_closed=false`) keeps fault processing running but still surfaces the gap via the health counter.
+
+`verify()` walks the persisted chain oldest-first and recomputes every link: editing a row breaks its `record_hash`, deleting a row breaks the next row's `prev_hash` linkage, and deleting the newest row (the head row is read straight from the DB) is caught by the persisted-head check.
```
bburda: This is not fully true, and it costs no more than casual tampering, not the "recompute the entire chain" the threat model describes below. A write-capable actor can delete the newest row (drop the delete trigger, disclosed) and then repoint the head with `UPDATE audit_chain_head SET seq=2, record_hash=<the stored H2>`. The head table is intentionally not trigger-protected, and H2 is already stored in row 2, so no hash is recomputed. verify() then walks rows 1..2, the persisted-head check matches (2, H2), and returns ok. The newest event is gone and verify passes. Please correct this sentence and the "proves nothing was deleted" claim on line 119.

src/ros2_medkit_fault_manager/README.md
```diff
@@ -109,6 +110,28 @@ patterns:
...
+
+**Retention/rotation**: when more than `audit_log.retention_max_records` rows are retained, the oldest segment is *sealed* (its final `seq` + hash are persisted as an anchor) and then pruned. The surviving tail still verifies because the oldest retained row links back to the sealed anchor.
```
bburda: External prefix truncation is indistinguishable from this legitimate pruning, so "the surviving tail still verifies" also covers a forged truncation. audit_anchors has no trigger; a write-capable actor can `DELETE FROM audit_log WHERE seq<=7` (delete trigger dropped, disclosed) and `INSERT INTO audit_anchors` a row whose last_hash is row 8's prev_hash, which is already stored. verify()'s oldest-row branch accepts it with no recompute. So verify() cannot tell external prefix truncation from retention pruning, which again falsifies "verify() proves nothing was deleted" (line 119). Worth stating that truncation of either end is not detected.

src/ros2_medkit_fault_manager/src/fault_audit_log.cpp
```diff
@@ -0,0 +1,598 @@
...
+  out += ",\"transition\":";
+  out += nlohmann::json(event.transition).dump();
+  out += ",\"severity\":";
+  out += std::to_string(static_cast<int>(event.severity));
+  out += ",\"status\":";
+  out += nlohmann::json(event.status).dump();
+  out += ",\"source\":";
+  out += nlohmann::json(event.source_id).dump();
+  out += ",\"description\":";
+  out += nlohmann::json(event.description).dump();
```
bburda: `nlohmann::json(...).dump()` uses the strict UTF-8 handler and throws on any non-UTF-8 byte. `description` is unvalidated and `source_id` is only checked non-empty, both taken from the ReportFault request. A fault reported with a non-UTF-8 byte makes canonicalize throw, so append throws: the transition is dropped from the chain, audit_healthy latches false, and under fail_closed it aborts the operation. verify() also calls canonicalize on every row, so one planted bad-UTF-8 row makes verify() throw instead of returning ok=false. Make canonicalize total over arbitrary bytes, but keep it injective: do not use error_handler_t::replace, because lossy U+FFFD replacement maps distinct byte strings to the same canonical form and would let a byte-level edit pass verify. Length-prefix the raw bytes, or hash raw bytes.

src/ros2_medkit_fault_manager/src/fault_audit_log.cpp
```diff
@@ -0,0 +1,598 @@
...
+  if (sqlite3_open_v2(db_path.c_str(), &db_, flags, nullptr) != SQLITE_OK) {
+    std::string error = db_ ? sqlite3_errmsg(db_) : "unknown error";
+    if (db_) {
+      sqlite3_close(db_);
+      db_ = nullptr;
+    }
+    throw std::runtime_error("audit: failed to open '" + db_path + "': " + error);
+  }
+
+  exec_or_throw(db_, "PRAGMA journal_mode=WAL;", "enable WAL");
```
bburda: `db_` is a raw `sqlite3*` and the destructor does not run when the constructor throws, so the open handle leaks on the three throwing paths that do not close it first: this PRAGMA exec_or_throw (line 117), initialize_schema() (line 129), and load_head_locked() (line 130). Reachable on a locked, corrupt, or read-only audit DB. In production create_audit_log rethrows and the process dies, but test_fault_audit_log constructs FaultAuditLog directly, so a construction-failure test trips LSan in the ASan/UBSan CI job. Close db_ on every throwing path, or hold it in a unique_ptr with a closing deleter before the fallible init runs.

src/ros2_medkit_fault_manager/src/fault_audit_log.cpp
```diff
@@ -0,0 +1,598 @@
...
+    store_head_locked(ChainHead{new_seq, record_hash});
+    exec_or_throw(db_, "COMMIT", "commit append");
+  } catch (...) {
+    exec_or_throw(db_, "ROLLBACK", "rollback append");
+    throw;
+  }
+
+  head_ = ChainHead{new_seq, record_hash};
+
+  rotate_if_needed_locked();
```
bburda: rotate_if_needed_locked() runs after COMMIT and outside the try/catch, but the record and head are already durably written. If rotate throws (SQLITE_BUSY on a WAL checkpoint, disk-full on the prune DELETE), append() throws for a write that actually succeeded. The caller then counts it as a dropped write and clears audit_healthy, and under fail_closed aborts a real fault operation whose audit row was written. Only with retention>0, but then every at-capacity append is exposed. Move rotation into the append transaction, or make a post-commit rotate failure not look like a lost append.

src/ros2_medkit_fault_manager/src/fault_audit_log.cpp
```diff
@@ -0,0 +1,598 @@
...
+      // prune without a valid anchor.
+      return;
+    }
+    boundary_hash = stmt.column_text(0);
+    sealed_at_ns = stmt.column_int64(1);
+  }
+
+  exec_or_throw(db_, "BEGIN IMMEDIATE", "begin rotate");
+  try {
+    Stmt anchor(db_,
```
bburda: audit_anchors gets one row per pruned segment and is never deleted, but verify() only ever reads the single anchor at last_seq == oldest_retained_seq - 1. Past the retention limit every append prunes one row and seals one anchor, so audit_log stays at the cap while audit_anchors grows without bound. Retention then bounds the wrong table: on a long-running target with steady transitions this is unbounded disk growth and flash wear, and it eventually fills the disk and turns every append into a failure. Keep only the latest anchor (delete older anchors when sealing a new one).

src/ros2_medkit_fault_manager/src/fault_manager_node.cpp
```diff
@@ -347,6 +379,129 @@ std::unique_ptr<FaultStorage> FaultManagerNode::create_storage() {
...
+    const uint64_t dropped = audit_dropped_writes_.load(std::memory_order_relaxed);
+    RCLCPP_ERROR(get_logger(), "Failed to append audit record for '%s' (%s): %s [audit dropped_writes=%" PRIu64 "]",
+                 fault.fault_code.c_str(), transition, e.what(), dropped);
+    if (audit_fail_closed_) {
+      // Compliance-strict: refuse to proceed as if nothing happened. Abort the
+      // operation so the broken audit cannot be silently outlived.
+      RCLCPP_FATAL(get_logger(),
+                   "audit_log.fail_closed is set: aborting operation because the audit append failed for '%s' (%s)",
+                   fault.fault_code.c_str(), transition);
+      throw;
```
bburda, line 561: Two problems with this rethrow. First, it does not abort the operation "as if nothing happened": the fault-store transition was already committed before audit_transition ran (report_fault_event / clear_fault run first), so the state change persists while the audit row is missing, which is exactly the hole the feature targets. Second, the throw leaves the service callback and main.cpp runs rclcpp::spin with no try/catch, so it propagates to std::terminate (SIGABRT). response->accepted was already set to true and no response is sent, so the client hangs to timeout, and ~FaultManagerNode never runs so the logging_deactivated marker is skipped. The same throw from the auto_confirm timer crashes the node from a timer. If fail-stop is intended, reject the request with a defined status or move the node to an error lifecycle state instead of terminating.

src/ros2_medkit_fault_manager/include/ros2_medkit_fault_manager/fault_audit_log.hpp
```diff
@@ -0,0 +1,170 @@
...
+
+/// Canonical transition kinds. Stored verbatim, so they are part of the hash.
+constexpr const char * kTransitionOccurred = "occurred";
+constexpr const char * kTransitionConfirmed = "confirmed";
+constexpr const char * kTransitionCleared = "cleared";
+/// Auto-recovery: a fault reached the healing threshold via PASSED events. Kept
+/// distinct from kTransitionCleared so an automatic recovery is not mistaken for
+/// a manual clear in the timeline.
+constexpr const char * kTransitionHealed = "healed";
+/// Audit-log lifecycle markers (CIR (EU) 2024/2690 sec. 3.2: activation /
```
bburda: CIR (EU) 2024/2690 scopes DNS/TLD, cloud, data-centre, CDN, MSP/MSSP, marketplace, search, social, and trust-service providers, not a robotics or automotive fault manager. Its Annex 3.2 logging content is real, but the regulation does not govern this system class, so citing it here (and at fault_manager_node.cpp:331 and :444) can mislead an integrator into thinking the feature confers NIS2 standing. Keep the activation/deactivation markers as good practice, but drop or reframe the regulatory citation.

bburda (commented):

- The audit log has no runtime read/verify/health surface. `verify()`, `read()`, `head()`, `record_count()`, and the dropped-writes / audit-healthy counters are reachable only as C++ getters (the test-only accessor); no runtime path calls `verify()`, and nothing exposes the log or its health outside the process. The README (line 119, and the `fail_closed` row of the parameter table) presents the dropped-writes counter as the safety net for the completeness gap, but an operator has no way to observe audit health or run `verify()` without stopping the node and using an offline tool that reimplements `canonicalize()`. The history and its verify/health result should be reachable over the gateway API (a read plus a verify/health resource), with an optional fault-manager ROS service for verify/health, so the log is observable at runtime.

mfaferek93:
Agreed, and this is intentional for this PR. #487 lands only the write-time primitive: the append-only hash chain plus the in-process `verify()` / `read()` / head APIs. Surfacing the log and its verify/health at runtime (a read plus a verify/health resource over the gateway API, and optionally a fault-manager service) is deliberately out of scope here and tracked as follow-up work.

I'll also tighten the README so the dropped-writes / audit-healthy signals aren't presented as operator-observable at this stage (they are in-process only until that runtime surface lands), and I'll address the concrete correctness findings in the diff in a follow-up commit.

src/ros2_medkit_fault_manager/README.md
mfaferek93: Fixed in 6579aaf: README now states that dropping the newest row and repointing the head is casual tampering, undetectable by design (tamper-evident only).

src/ros2_medkit_fault_manager/README.md
mfaferek93: Fixed in 6579aaf: README now states a forged prefix-truncation is indistinguishable from legitimate pruning.

src/ros2_medkit_fault_manager/src/fault_audit_log.cpp
mfaferek93: Fixed in 6579aaf: canonicalize() dumps with the replace error handler, so non-UTF-8 bytes no longer throw; valid-UTF-8 output is byte-identical, so existing chains still verify.

src/ros2_medkit_fault_manager/src/fault_audit_log.cpp
mfaferek93: Fixed in 6579aaf: db_ is now a unique_ptr<sqlite3, ...>, closed even when the constructor throws.

src/ros2_medkit_fault_manager/src/fault_audit_log.cpp
mfaferek93: Fixed in 6579aaf: rotate_if_needed_locked() is in its own try/catch; a post-COMMIT rotation failure is tracked in rotation_failures_, not reported as a dropped append.

src/ros2_medkit_fault_manager/src/fault_audit_log.cpp
mfaferek93: Fixed in 6579aaf: rotation deletes anchors below the boundary, so audit_anchors stays at one row; added guard-gated triggers on it as defense-in-depth.

src/ros2_medkit_fault_manager/src/fault_manager_node.cpp
mfaferek93: Fixed in 6579aaf: reworded. The fault-store commit is in a separate DB, so fail_closed is a broken-audit alarm requiring operator action, not a rollback; no atomicity is claimed.

src/ros2_medkit_fault_manager/include/ros2_medkit_fault_manager/fault_audit_log.hpp
mfaferek93: Agreed, mis-fit reference. Removed the CIR (EU) 2024/2690 citation in 6579aaf; the logging description stands on its own.

bburda (approved):

PR #488, opened 2026-06-30
Asset-identity model + merge-by-identity

### LINKED ISSUE #482: Asset-identity model + merge-by-identity
Add identity fields to the Component entity (manufacturer, model, serial, hardware_rev, firmware/software version, network endpoint, role) plus an extensible map. Add a merge step combining identity from multiple sources by an identity key, with precedence and per-field provenance. Feeds the existing discovery.

- Align the field model to AAS concepts (asset<->shell, identity<->Nameplate, hierarchy<->BoM, endpoint<->AID) so the same model maps cleanly to/from AAS without rework.
- Decisions: typed fields vs generic map (likely both); precedence order; identity key (serial / order-code+slot / endpoint / configured id).
- Acceptance: an asset shows merged identity from more than one source, with provenance per field.

mfaferek93:
Closes #482

Adds an AAS-Nameplate-aligned `AssetIdentity` (manufacturer, model, serial, hardware/firmware/software version, network endpoint, role) plus an extensible map to the Component entity. Identity is merged across sources by a configurable precedence ranked on `Component.source`, with per-field provenance, wired through the discovery and peer-aggregation paths. JSON is backward compatible: identity is emitted only when set.

Tested: identity / merge-pipeline / peer-client / entity-merger suites pass.

### DIFF
--- a/src/ros2_medkit_gateway/README.md
+++ b/src/ros2_medkit_gateway/README.md
@@ -1584,20 +1584,22 @@ records **per-field provenance** (which source set each field) under `_provenanc
 - **Precedence** is config-driven (`IdentityMergeConfig::source_precedence`, highest first)
   and is deliberately **decoupled from the structural `MergePolicy`**: a manifest can be the
   authoritative *structure* source while a live protocol read is the authoritative *identity*
-  source. Default order: protocol device-info (`opcua`, `s7`, `ethernet_ip`, `modbus`, `ads`,
-  `profinet`) > `manifest` > `config` > runtime sources. A higher-authority source overrides a
+  source. Authority is ranked on each contributing entity's canonical `Component.source` tag
+  ("manifest", "plugin", "runtime", "node", "config", or a protocol-class tag a provider sets
+  such as "opcua"), **not** the free-form discovery-layer name. Default order: protocol
+  device-info (`opcua`, `s7`, `ethernet_ip`, `modbus`, `ads`, `profinet`, and the generic
+  `plugin`) > `manifest` > `config` > runtime sources. A higher-authority source overrides a
   field; lower-authority sources only fill gaps; unknown sources rank lowest. Empty values never
   overwrite.
-- **Identity key.** `compute_identity_key()` derives the key that decides whether two records
-  describe the same asset. Strategies: `serial`, `order_code_slot` (`model` + `extra["slot"]`),
-  `endpoint`, `configured_id`, and `auto` (serial -> order-code+slot -> endpoint -> configured id).
-  The discovery pipeline merges Components by `Component.id`, which is the `configured_id`
-  strategy; the other strategies let sources that assign ids (CSV/manifest import, protocol
-  probes) correlate records that arrive under different ids before they reach the pipeline.
+
+Components are correlated for merging by `Component.id`; identity is merged whenever two
+sources contribute the same Component id (in the discovery pipeline, and gap-filled across
+peer aggregation).
 
 The pipeline calls `merge_identity` inside the `IDENTITY` field group for Components, seeding
-provenance with the base (highest-priority) layer. Configure via
-`MergePipeline::set_identity_merge_config()`.
+provenance with the base (highest-priority) layer's source tag. Configure via
+`MergePipeline::set_identity_merge_config()`; precedence entries must equal the canonical
+`Component.source` values you expect at runtime.
 
 ## Demo Nodes
 
--- a/src/ros2_medkit_gateway/README.md
+++ b/src/ros2_medkit_gateway/README.md
@@ -1544,6 +1544,61 @@ Runtime mode mapping:
 /standalone_node                -> Component: <hostname>, App: standalone_node
 ```
 
+### Asset Identity (nameplate) and merge-by-identity
+
+Each Component can carry an **asset identity**: the stable nameplate of the physical
+or logical asset it represents (an ECU, PLC, drive, sensor), separate from its runtime
+ROS 2 footprint. The model lives in `core/discovery/models/asset_identity.hpp`.
+
+Typed fields (queryable, AAS-mappable) plus a generic `extra` key-value map for
+vendor-specific attributes we do not model up front:
+
+| Field | JSON key | AAS Nameplate (IDTA 02006) |
+|-------|----------|----------------------------|
+| `manufacturer` | `manufacturer` | ManufacturerName |
+| `model` (order code) | `model` | ManufacturerProductDesignation |
+| `serial_number` | `serialNumber` | SerialNumber |
+| `hardware_revision` | `hardwareRevision` | HardwareVersion |
+| `firmware_version` | `firmwareVersion` | FirmwareVersion |
+| `software_version` | `softwareVersion` | SoftwareVersion |
+| `network_endpoint` | `networkEndpoint` | Asset Interface Description (AID) endpoint |
+| `role` | `role` | functional role (BoM/usage context) |
+| `extra` | `extra` | additional Nameplate properties |
+
+The whole block maps to an Asset Administration Shell later without rework: the asset
+itself is the shell (`Component.id` as localId), this block is the **Nameplate** submodel,
+the Component hierarchy is the **BoM** submodel, and `network_endpoint` is an **AID** entry.
+
+Identity is emitted under `x-medkit.identity` only when populated, so identity-less
+components keep their existing JSON unchanged.
+
+**Why typed fields and a map?** Typed fields are the common, queryable, standard-mappable
+attributes; the `extra` map is the escape hatch for vendor keys (rack/slot, MAC, asset tag)
+that do not warrant a schema change. Both merge with the same rules.
+
+**Merge-by-identity with provenance.** A single asset is usually described by several
+sources (a hand-authored manifest, a live protocol device-info read, runtime discovery).
+`merge_identity()` (`core/discovery/identity_merge.hpp`) combines them field by field and
+records **per-field provenance** (which source set each field) under `_provenance`.
+
+- **Precedence** is config-driven (`IdentityMergeConfig::source_precedence`, highest first)
+  and is deliberately **decoupled from the structural `MergePolicy`**: a manifest can be the
+  authoritative *structure* source while a live protocol read is the authoritative *identity*
+  source. Default order: protocol device-info (`opcua`, `s7`, `ethernet_ip`, `modbus`, `ads`,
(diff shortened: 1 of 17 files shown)


src/ros2_medkit_gateway/README.md
```diff
@@ -1544,6 +1544,63 @@ Runtime mode mapping:
...
+  authoritative *structure* source while a live protocol read is the authoritative *identity*
+  source. Authority is ranked on each contributing entity's canonical `Component.source` tag
+  ("manifest", "plugin", "runtime", "node", "config", or a protocol-class tag a provider sets
+  such as "opcua"), **not** the free-form discovery-layer name. Default order: protocol
+  device-info (`opcua`, `s7`, `ethernet_ip`, `modbus`, `ads`, `profinet`, and the generic
+  `plugin`) > `manifest` > `config` > runtime sources. A higher-authority source overrides a
+  field; lower-authority sources only fill gaps; unknown sources rank lowest. Empty values never
+  overwrite.
+
+Components are correlated for merging by `Component.id`; identity is merged whenever two
```
bburda, line 1595: This correlates sources by `Component.id` string equality, not by any nameplate field. Two sources describing one physical asset under different ids (a manifest id vs a protocol-derived id) never merge, and peers that id the same ECU differently produce duplicate assets. The linked issue left the identity key open (serial / order-code+slot / endpoint / configured id) and this resolves it to `Component.id` without saying so. Please decide the correlation key explicitly and align the feature name and docs with it.

bburda (commented):

- No code populates `Component.identity` on a live gateway. The manifest model has no identity field and the manifest parser never sets one; `host_info_provider.cpp` sets only `host_metadata`; discovery plugins tag `source="plugin"` and never set identity. `set_identity_merge_config()` has no caller. Outside unit tests the only writers of `.identity` are the peer parse and the merge itself. So `x-medkit.identity` is never emitted at runtime, peers receive empty identities, and the whole merge/precedence/provenance path is unreachable in production. If model-only is the intended scope, please say so and add at least one producer (a manifest identity field, or the host provider) so the acceptance criterion (an asset shows merged identity from more than one source) can actually be met.

- This overlaps an existing open issue for host hardware identity in `x-medkit` (#358), which asked for manufacturer/serial/board read from `/sys/firmware/devicetree/base/model`, `/etc/machine-id`, and DMI. This PR supersedes the data-model half of it: the nested `x-medkit.identity` here is a superset of its flat `x-medkit.serial`/`hardware_id`/`board` proposal. It does not cover the producer half, since nothing reads those host files yet. Please decide the disposition of #358: close it as superseded by this model, or repurpose it to the host-producer follow-up. If repurposed, it should target `x-medkit.identity` and drop the flat keys, and note that `/etc/machine-id` is a software UUID that belongs in `identity.extra`, not `serialNumber`.

src/ros2_medkit_gateway/include/ros2_medkit_gateway/core/discovery/identity_merge.hpp
```diff
@@ -0,0 +1,172 @@
...
+ *
+ * Default precedence (highest first): a live protocol device-info read (a `plugin`
+ * source, or a protocol-specific source tag) beats the hand-authored `manifest`,
+ * which beats whatever runtime discovery guessed. The protocol-specific tags lead the
+ * list so that a provider which sets a concrete `Component.source` (e.g. "opcua") is
+ * honoured; the generic "plugin" tag covers the common case where the plugin layer
+ * stamps every plugin entity with source="plugin".
+ */
+struct IdentityMergeConfig {
+  std::vector<std::string> source_precedence{"opcua",    "s7",     "ethernet_ip", "modbus", "ads",
```
bburda, line 50: `PluginLayer::discover()` overwrites `comp.source = "plugin"` for every plugin component (src/ros2_medkit_gateway/src/discovery/layers/plugin_layer.cpp:77), so a discovery plugin can never surface `source="opcua"/"s7"/"modbus"/etc`. These six leading tags are unreachable at runtime; two protocol plugins both get `plugin`, tie in `source_rank`, and fall back to layer order, not the documented per-protocol precedence. The README and the struct docstring describe a mechanism the runtime prevents. Either let plugins keep their own source tag (PluginLayer must stop overwriting it), or drop the protocol tags and document that only `plugin`/`manifest`/`runtime` are reachable.

src/ros2_medkit_gateway/src/core/aggregation/entity_merger.cpp
```diff
@@ -171,6 +173,15 @@ std::vector<Component> EntityMerger::merge_components(const std::vector<Componen
...
       }
 
+      // Merge the remote asset-identity nameplate into the local one. The peer is
+      // tagged as the lowest authority (peer:<name> is not in the default precedence),
+      // so this is a gap-fill: empty local fields are populated from the remote, but a
+      // value the local side already knows is never overridden. Stamp local provenance
+      // first so existing local fields are protected during the merge.
+      static const discovery::IdentityMergeConfig kIdentityConfig;
+      discovery::stamp_identity_provenance(merged.identity, merged.source);
+      discovery::merge_identity(merged.identity, remote_comp.identity, peer_source(), kIdentityConfig);
```
bburda, line 183: The peer identity is merged as `peer:<name>`, which is never in the precedence list, so it always ranks lowest and only gap-fills. But the same block (line 190) routes all runtime state of the colliding Component to the peer as the authoritative owner. The peer is usually closest to the hardware and holds the live nameplate read, yet a stale local value always wins and no config can invert it (the peer tag is synthesized, never listed). Identity authority points opposite to routing authority.

bburda, line 183: `merge_identity` re-attributes every field it writes to `peer:<name>` and ignores `remote_comp.identity.provenance`. A field the peer sourced from `opcua` shows provenance `peer:<name>` after this collision gap-fill, but keeps `opcua` after the no-collision wholesale copy on line 194. For an audit feature this attribution is inconsistent; carry the remote per-field provenance through the merge instead of overwriting it.

src/ros2_medkit_gateway/include/ros2_medkit_gateway/core/discovery/models/asset_identity.hpp
```diff
@@ -0,0 +1,178 @@
...
+      j["networkEndpoint"] = network_endpoint;
+    }
+    if (!role.empty()) {
+      j["role"] = role;
+    }
+    if (!extra.empty()) {
+      j["extra"] = extra;
+    }
+    if (!provenance.empty()) {
+      j["_provenance"] = provenance;
```
bburda, line 122: Field keys are camelCase (`serialNumber`, `hardwareRevision`) but `_provenance` keys are snake_case (`serial_number`, ...). A consumer correlating a field to its source has to translate case, and the mapping is untested at the wire level. Emit provenance under the same camelCase keys as the fields.

src/ros2_medkit_gateway/src/gateway_node.cpp
```diff
@@ -1963,6 +1963,19 @@ void GatewayNode::trigger_reentrant_notification_for_testing(const EntityChangeS
...
+  // Cancel the two refresh drivers and drop the graph event so no further
+  // refresh_cache() pass can clobber a test-injected entity cache. Safe to
+  // call right after construction (the constructor already ran one refresh).
+  if (graph_check_timer_) {
+    graph_check_timer_->cancel();
+  }
+  if (backstop_timer_) {
+    backstop_timer_->cancel();
+  }
+  graph_event_.reset();
```
bburda, line 1990: `graph_event_.reset()` here has none of the ordering or mutex the destructor uses; the destructor documents a TSan-flagged read/write race on this exact Event control block. It is safe only because the sole caller runs before the executor spins. A future test that calls it while a `MultiThreadedExecutor` is spinning hits the same race. Guard it like the destructor does, or assert that it runs before spin.

src/ros2_medkit_gateway/include/ros2_medkit_gateway/core/discovery/identity_merge.hpp
mfaferek93: Fixed in 0c022ed: source is stamped only when empty, so a provider that pre-sets a protocol source (e.g. opcua) survives and the precedence tags at the head of the list become reachable.

src/ros2_medkit_gateway/src/core/aggregation/entity_merger.cpp
mfaferek93: Fixed in 0c022ed: the collision gap-fill now preserves the remote field provenance instead of re-stamping peer:<name>, so the collision and no-collision paths keep the original source.

src/ros2_medkit_gateway/include/ros2_medkit_gateway/core/discovery/models/asset_identity.hpp
mfaferek93: The snake_case _provenance keys are intentional (AAS-aligned, documented in the header). Pinned the camelCase-field / snake_case-provenance mapping with a wire-level test in 0c022ed so it cannot silently drift.

src/ros2_medkit_gateway/src/gateway_node.cpp
mfaferek93: Hardened in 0c022ed: the helper now mirrors the destructor (cancels and resets all three timers before releasing graph_event_), which removes the window rather than shrinking it. One correction: the racing party is rclcpp graph-listener thread started in the constructor, not the executor, so it is live regardless of executor state - the same residual race the destructor already tolerated.

src/ros2_medkit_gateway/src/core/aggregation/entity_merger.cpp
mfaferek93: Intentional and now documented: a remote peer gap-fills identity at lowest authority so it cannot clobber local nameplate knowledge, while still owning the routed runtime state. The divergence is deliberate, not a correctness bug.

src/ros2_medkit_gateway/README.md
mfaferek93: Accurate, and a deliberately-scoped limitation: the nameplate identity-key correlator was dropped from this PR, so id-based correlation is the documented behavior. A key-based correlator is the proper follow-up.

mfaferek93:
On the no-producer point: model plus merge is the intended scope of this PR. x-medkit.identity is emitted once a producer lands (the OPC UA nameplate reader, a follow-up); until then the merge / precedence / provenance path is covered by the unit tests.

bburda (approved):

PR #491, opened 2026-07-02
Persist compact JSON freeze-frame on fault confirm

### LINKED ISSUE #476: Compact JSON freeze-frame in fault snapshots
On a fault transition to CONFIRMED, persist a compact JSON dict of the captured topic values to SQLite alongside the existing rosbag/mcap snapshot, keyed by fault_code, retained across clear_fault.

- Capture set is config-driven via `SnapshotCapture::resolve_topics(fault_code)` (per-fault-code topic list + regex patterns); an unconfigured fault_code yields an empty freeze-frame. Values come from the `/plc/*` topics published by the protocol plugins.
- Acceptance: a confirmed fault writes a `freeze_frames` row with the captured values; the row survives clear_fault; retrievable by fault_code.

mfaferek93:
On the transition to CONFIRMED, snapshot capture now also persists a compact JSON freeze-frame of the captured topic values to SQLite, keyed by fault_code and retained across clear_fault (unlike rosbag snapshots).

- Capture set is config-driven via the existing per-fault-code snapshot topic resolution; a fault code with no capture configuration writes no freeze-frame row (a lookup returns not-found).
- Rows survive clear_fault, startup reclassification, and a DB reopen; a genuine re-capture replaces the row.
- Note: timer-driven confirmations do not enqueue snapshot capture (existing capture-path scope), so they produce no freeze-frame either.

Tested: freeze-frame unit + storage suites on both backends and capture-pool end-to-end; full fault_manager suite passes.

Closes #476

### DIFF
--- a/src/ros2_medkit_fault_manager/include/ros2_medkit_fault_manager/fault_storage.hpp
+++ b/src/ros2_medkit_fault_manager/include/ros2_medkit_fault_manager/fault_storage.hpp
@@ -110,6 +110,16 @@ struct SnapshotData {
   int64_t captured_at_ns{0};
 };
 
+/// Compact freeze-frame captured when a fault confirms: a single JSON object mapping
+/// each captured topic to its latest value at confirmation time. Unlike per-topic
+/// snapshots, a freeze-frame is keyed by fault_code (one row per code) and is RETAINED
+/// across clear_fault, so the confirmed-state record persists after acknowledgement.
+struct FreezeFrameData {
+  std::string fault_code;
+  std::string data;  ///< Compact JSON object: {"<topic>": <value>, ...}
+  int64_t captured_at_ns{0};
+};
+
 /// Rosbag file metadata for time-window recording
 struct RosbagFileInfo {
   std::string fault_code;
@@ -189,6 +199,17 @@ class FaultStorage {
   virtual std::vector<SnapshotData> get_snapshots(const std::string & fault_code,
                                                   const std::string & topic_filter = "") const = 0;
 
+  /// Store the compact freeze-frame captured for a fault (JSON dict of topic values).
+  /// Keyed by fault_code: a later capture for the same code replaces the frame. The frame
+  /// is retained across clear_fault so the confirmed-state record survives acknowledgement.
+  /// @param frame The freeze-frame to store
+  virtual void store_freeze_frame(const FreezeFrameData & frame) = 0;
+
+  /// Get the freeze-frame captured for a fault, if any.
+  /// @param fault_code The fault code to look up
+  /// @return The freeze-frame if one was captured, nullopt otherwise
+  virtual std::optional<FreezeFrameData> get_freeze_frame(const std::string & fault_code) const = 0;
+
   /// Store rosbag file metadata for a fault
   /// @param info The rosbag file info to store (replaces any existing entry for fault_code)
   virtual void store_rosbag_file(const RosbagFileInfo & info) = 0;
@@ -267,6 +288,9 @@ class InMemoryFaultStorage : public FaultStorage {
   std::vector<SnapshotData> get_snapshots(const std::string & fault_code,
                                           const std::string & topic_filter = "") const override;
 
+  void store_freeze_frame(const FreezeFrameData & frame) override;
+  std::optional<FreezeFrameData> get_freeze_frame(const std::string & fault_code) const override;
+
   void store_rosbag_file(const RosbagFileInfo & info) override;
   std::optional<RosbagFileInfo> get_rosbag_file(const std::string & fault_code) const override;
   bool delete_rosbag_file(const std::string & fault_code) override;
@@ -283,7 +307,8 @@ class InMemoryFaultStorage : public FaultStorage {
   mutable std::mutex mutex_;
   std::map<std::string, FaultState> faults_;
   std::vector<SnapshotData> snapshots_;
-  std::map<std::string, RosbagFileInfo> rosbag_files_;  ///< fault_code -> rosbag info
+  std::map<std::string, FreezeFrameData> freeze_frames_;  ///< fault_code -> freeze-frame (retained across clear)
+  std::map<std::string, RosbagFileInfo> rosbag_files_;    ///< fault_code -> rosbag info
   DebounceConfig config_;
   size_t max_snapshots_per_fault_{0};  ///< 0 = unlimited
 };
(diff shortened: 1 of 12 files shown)


src/ros2_medkit_fault_manager/src/snapshot_capture.cpp
```diff
@@ -124,20 +124,24 @@ void SnapshotCapture::capture(const std::string & fault_code) {
 
   auto topics = resolve_topics(fault_code);
   if (topics.empty()) {
+    // Unconfigured fault code: no capture set resolved, so no freeze-frame is written and
+    // no subscriptions are created (no overhead). get_freeze_frame() stays empty for it.
     RCLCPP_DEBUG(node_->get_logger(), "No topics configured for fault '%s'", fault_code.c_str());
     return;
```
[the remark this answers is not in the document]
mfaferek93: Intended behavior, now documented explicitly in 4f14b8d: an unconfigured fault code writes NO freeze_frames row (lookup returns nullopt) - absence means "no capture configured". A configured set that captured nothing still writes {} to record that capture ran. The PR description was updated to match; covered by the existing UnconfiguredFaultWritesNoFreezeFrame test.

src/ros2_medkit_fault_manager/test/test_snapshot_capture.cpp
```diff
@@ -225,6 +231,75 @@ TEST_F(SnapshotCaptureTest, OnDemandCaptureHandlesNonExistentTopic) {
...
+  });
+
+  // Wait for the publisher to be visible to the capture node before capturing.
+  auto start = std::chrono::steady_clock::now();
+  while (node_->count_publishers("/plc/pressure") == 0 &&
+         std::chrono::steady_clock::now() - start < std::chrono::seconds(5)) {
+    std::this_thread::sleep_for(std::chrono::milliseconds(20));
+  }
+  ASSERT_GT(node_->count_publishers("/plc/pressure"), 0u);
+
```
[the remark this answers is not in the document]
mfaferek93: Fixed in 4f14b8d: the publisher thread is now managed by a ScopedPublisherThread RAII helper (stop-flag + join in the destructor), so an assertion failure can no longer destroy a joinable thread and abort. Checked the other threads in the package tests - no other instance of the pattern.

src/ros2_medkit_fault_manager/src/snapshot_capture.cpp
```diff
@@ -147,6 +153,16 @@ void SnapshotCapture::capture(const std::string & fault_code) {
...
+
+  // Persist the compact freeze-frame keyed by fault_code (retained across clear_fault).
+  // Only reached for faults with a configured capture set (unconfigured codes returned
+  // early above and get no row). If nothing was publishing at confirmation time the row
+  // holds an empty object, recording that a configured capture ran but sampled nothing.
+  FreezeFrameData frame;
+  frame.fault_code = fault_code;
+  frame.data = freeze_frame.dump();
+  frame.captured_at_ns = get_wall_clock_ns();
+  storage_->store_freeze_frame(frame);
```
bburda, line 176: `store_freeze_frame` runs unconditionally here, even when `captured_count == 0` (every capture failed), writing `data == "{}"`. Since the store is `INSERT OR REPLACE` keyed by fault_code, a re-confirm after the recapture cooldown while the source publishers are down overwrites a previously-good retained frame with `{}`, silently losing the confirmed-state record this feature exists to keep across clear_fault. Skip the store when nothing was captured (or keep the richer frame) so a failed re-capture cannot erase a good one.

src/ros2_medkit_fault_manager/include/ros2_medkit_fault_manager/fault_storage.hpp
```diff
@@ -189,6 +201,18 @@ class FaultStorage {
...
+  /// Keyed by fault_code: a later capture for the same code replaces the frame. The frame
+  /// is retained across clear_fault so the confirmed-state record survives acknowledgement.
+  /// @param frame The freeze-frame to store
+  virtual void store_freeze_frame(const FreezeFrameData & frame) = 0;
+
+  /// Get the freeze-frame captured for a fault, if any.
+  /// @param fault_code The fault code to look up
+  /// @return The freeze-frame if one was captured, nullopt otherwise (including fault
+  ///         codes with no capture configured, which never get a row)
+  virtual std::optional<FreezeFrameData> get_freeze_frame(const std::string & fault_code) const = 0;
```
bburda, line 217: `get_freeze_frame` has no production caller in this PR (only tests read it); the GetFault path builds its `TYPE_FREEZE_FRAME` data from `get_snapshots()`, i.e. the separate `snapshots` table, so the frame written to `freeze_frames` is currently read by nothing and its one distinguishing property (surviving clear_fault) is unobservable. The table also has no eviction or delete path (unlike `snapshots`' max-per-fault and `rosbag_files`' `delete_rosbag_file`), so it keeps one heavy JSON blob per distinct fault code for the DB lifetime, duplicating values already in `snapshots`. Is the read path a follow-up? If yes, worth stating in the PR; if not, this is write-only storage - consider wiring a reader plus a bound, or letting the existing snapshots survive clear instead of a parallel table.

bburda (commented):

- README not updated: `ros2_medkit_fault_manager/README.md` documents the snapshot subsystem, including that snapshots are deleted on clear, but not the new freeze-frame captured on the same confirm and retained across clear_fault - the exact contrasting behavior that section already describes. Per the docs gate, add it.

- Two test-coverage gaps on the new code: the `background_capture=true` cache path (`freeze_frame[topic] = json::parse(cached.data)` with the raw-string fallback) is never exercised - the new e2e test uses `background_capture=false` only; and the empty-`{}` row case (a configured fault whose topics all fail to sample writes a `{}` frame, so `get_freeze_frame()` is non-null) is not asserted, though `OnDemandCaptureHandlesNonExistentTopic` already drives that path and could assert it in one line.

src/ros2_medkit_fault_manager/src/snapshot_capture.cpp
mfaferek93: Fixed in 0198e4b: an empty capture no longer clobbers an existing non-empty frame (skip + warning); the documented first-capture {} behavior stays. Regression test EmptyRecaptureKeepsRetainedFreezeFrame covers the publishers-down re-confirm.

src/ros2_medkit_fault_manager/include/ros2_medkit_fault_manager/fault_storage.hpp
mfaferek93: Fixed in 88f54a7: get_fault now falls back to the persisted freeze_frames row when no snapshots remain (i.e. after clear_fault), so the retained frame is observable end to end; covered by GetFaultServesRetainedFreezeFrameAfterClear. On eviction: the table is bounded by distinct fault codes (INSERT OR REPLACE) and faults themselves are never deleted (clear only flips status), so there is no delete hook to tie eviction to - stated in the interface doc.

mfaferek93:
Both outside-the-diff points addressed in 88f54a7: the fault_manager README now documents the freeze-frame (retained across clear, contrasted with snapshots), and the two test gaps are covered - BackgroundCaptureCachesFreezeFrame exercises the background cache path incl. the json parse branch, and OnDemandCaptureHandlesNonExistentTopic now asserts the {} row.

bburda (approved):

PR #493, opened 2026-07-02
Manifest/CSV manual asset inventory

### LINKED ISSUE #490: Manifest/CSV manual asset inventory
Merge hand-authored or CSV-imported asset entries into the entity tree via the existing manifest discovery.

- CSV columns: id, manufacturer, model, serial, hardware_rev, firmware, endpoint, role, plus extensible extras; entries become Components with identity populated.
- Imported entries merge with protocol-discovered structure; a malformed row is skipped with a warning instead of aborting the import.
- Acceptance: a CSV/manifest asset appears in the tree and merges with discovered structure, with per-field source provenance.

mfaferek93:
Merges hand-authored or CSV-imported asset entries into the entity tree via the existing manifest discovery.

- CSV columns: id, manufacturer, model, serial, hardware_rev, firmware, endpoint, role, plus extensible extras; a malformed row is skipped with a warning instead of aborting the import.
- Entries populate the asset identity with per-field provenance `inventory`, ranked between `manifest` and `config` in the source precedence.
- An authoritative inventory entry with an empty scalar no longer clobbers a lower layer's real value (merge gap-fill fix), covered by a combined identity + placement merge test.

Tested: asset-inventory, merge-pipeline, identity, manifest and peer suites all pass.

Closes #490


src/ros2_medkit_gateway/src/gateway_node.cpp
```diff
@@ -279,6 +279,12 @@ GatewayNode::GatewayNode(const rclcpp::NodeOptions & options) : Node("ros2_medki
   // disabled (default). See ManifestManager::set_fragments_dir for the
   // merge semantics.
   declare_parameter("discovery.manifest.fragments_dir", "");
+  // CSV file listing manually inventoried assets (columns: id, manufacturer,
+  // model, serial, hardware_rev, firmware, endpoint, role, plus extras). Each
+  // row is appended to the manifest as a Component and merged into the tree by
+  // id. Requires a manifest-backed discovery mode (manifest_only / hybrid with
+  // discovery.manifest_path set). Empty = disabled (default).
+  declare_parameter("discovery.inventory.csv_path", "");
```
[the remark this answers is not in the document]
mfaferek93: Documented in a5a218d: CHANGELOG entry covering the manifest assets list + discovery.inventory.csv_path (columns, extras, quoting, size cap, missing-vs-malformed semantics), matching how fragments_dir is documented.

src/ros2_medkit_gateway/include/ros2_medkit_gateway/core/discovery/manifest/asset_inventory.hpp
```diff
@@ -0,0 +1,101 @@
...
+ *
+ * The component is tagged `source = "inventory"` so its provenance stays
+ * visible after the merge pipeline combines it (by id) with protocol-discovered
+ * structure. Identity is projected onto the existing Component fields until the
+ * INV1 structured identity model is available (see the AssetEntry note):
+ *   - name        <- "<manufacturer> <model>" (falls back to id at display time)
+ *   - variant     <- hardware_rev
+ *   - tags        <- role (when present)
+ *   - description <- human-readable identity summary (serial, firmware,
+ *                    endpoint, extras)
```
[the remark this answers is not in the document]
mfaferek93: Fixed in a5a218d: the comment now describes the structured Component identity population with per-field provenance; the stale variant/tags/description projection wording is gone.

bburda (commented):

- Doc-code mismatch: the gateway README source-precedence prose (`ros2_medkit_gateway/README.md`, the identity source-tag list and default order) still reads `... manifest > config > runtime` and never mentions `inventory`, but this PR inserted `inventory` between `manifest` and `config` (and updated the `identity_merge.hpp` doc comment). Update the README prose to match.

- Docs gate: the new `assets:` manifest list and the `discovery.inventory.csv_path` parameter with its CSV column schema (id/manufacturer/model/serial/hardware_rev/firmware/endpoint/role + aliases + extras + RFC-4180 quoting + size cap) are documented only in the CHANGELOG. Add them to `docs/config/manifest-schema.rst` and the annotated `config/gateway_params.yaml` where users find manifest/discovery config.

src/ros2_medkit_gateway/src/discovery/manifest/manifest_manager.cpp
```diff
@@ -476,6 +486,74 @@ std::string ManifestManager::get_fragments_dir() const {
...
+  std::stringstream buffer;
+  buffer << file.rdbuf();
+
+  try {
+    auto parsed = parse_asset_csv(buffer.str());
+    for (const auto & warn : parsed.warnings) {
+      log_warn("Inventory CSV: " + warn);
+    }
+    for (const auto & entry : parsed.entries) {
+      base.components.push_back(asset_entry_to_component(entry));
```
bburda: Inventory assets are appended to `base.components` with no dedup, then the merged manifest is validated - and `validate_unique_ids` raises a hard error R003 on any duplicate component id (manifest_validator.cpp:59), which `load_manifest` fails on unconditionally (manifest_manager.cpp:70-75, not gated on strict). So a single duplicate id aborts the ENTIRE load: two CSV rows sharing an id, or a CSV / `assets:` id equal to an existing manifest `components:` id, discards the whole hand-authored manifest (in hybrid, init/reload then silently falls back to runtime_only). This contradicts the per-row resilience the PR advertises - an empty-id row is skipped with a warning, but a duplicate-id row nukes everything - and it makes the obvious 'manifest declares the component, CSV supplies its nameplate for the same id' enrichment impossible (both land in the manifest layer and collide). Dedup by id in `parse_asset_csv` / `apply_inventory_csv` (skip-with-warning, keep-first or merge) so a duplicate is handled at row level like empty-id.

src/ros2_medkit_gateway/src/discovery/manifest/manifest_parser.cpp
```diff
@@ -257,6 +268,90 @@ Component ManifestParser::parse_component(const YAML::Node & node) const {
...
+  static const std::unordered_set<std::string> reserved = {
+      "id",          "manufacturer", "model",   "serial", "hardware_rev",        "firmware",
+      "endpoint",    "role",         "area",    "tags",   "parent_component_id", "name",
+      "description", "namespace",    "variant", "type",   "translation_id",      "depends_on"};
+
+  AssetEntry entry;
+  entry.id = get_string(node, "id");
+  entry.manufacturer = get_string(node, "manufacturer");
+  entry.model = get_string(node, "model");
+  entry.serial = get_string(node, "serial");
```
bburda: `parse_asset` reads only the canonical keys `serial` / `hardware_rev` / `firmware`, and the `reserved` set above lacks the aliases, so a manifest `assets:` entry using `serial_number:` / `hardware_revision:` / `firmware_version:` falls through to the extras loop and lands in `identity.extra["serial_number"]` instead of `identity.serial_number`. The CSV path's `canonical_column` accepts exactly those aliases and maps them to the typed fields, so the same key produces a different identity shape - and a different provenance key (`extra.serial_number` vs `serial_number`, so it merges under the wrong key) - depending on whether it came from CSV or the manifest. The `AssetEntry` doc calls this the single place the two import paths converge; share one alias mapping so they agree.

src/ros2_medkit_gateway/src/core/discovery/manifest/asset_inventory.cpp
```diff
@@ -0,0 +1,298 @@
...
+    end_row();
+  }
+
+  return rows;
+}
+
+// Canonical destination for a header column.
+enum class Column { kIgnore, kExtra, kId, kManufacturer, kModel, kSerial, kHardwareRev, kFirmware, kEndpoint, kRole };
+
+Column canonical_column(const std::string & header_lower) {
```
bburda: `canonical_column` has no `area` / `namespace` / `parent_component_id` column, and `asset_entry_to_component` leaves `fqn`/`namespace_path`/`area` empty, so a CSV-only asset that matches no runtime node is a placement-less Component: addressable at `/components/{id}` and in the flat component list, but `get_components_for_area` filters on `c.area`, so it never appears under any Area. The manifest `assets:` path can place via `namespace:`, but the CSV path (the headline workflow) cannot, so CSV-only assets fall out of the area tree. Is that intended (CSV only enriches runtime-discovered nodes), or should CSV support a placement column like the manifest path?

src/ros2_medkit_gateway/src/discovery/manifest/manifest_manager.cpp
mfaferek93: Fixed in 4058cde: inventory now dedups before validation - within CSV first row wins (later dups skipped with warning), a CSV id colliding with a manifest component folds in as identity gap-fill (manifest wins, inventory ranks below), and a CSV id colliding with a non-component entity is skipped with a warning. One duplicate no longer aborts the whole load; tested for all three collision cases.

src/ros2_medkit_gateway/src/discovery/manifest/manifest_parser.cpp
mfaferek93: Fixed in 4058cde: parse_asset now routes the aliases (serial_number/hardware_revision/firmware_version) through the same shared column/alias table as the CSV path, so they land in the typed identity fields, not extras.

src/ros2_medkit_gateway/src/core/discovery/manifest/asset_inventory.cpp
mfaferek93: Fixed in 4058cde: added an optional area column (CSV) / area: key (manifest) mapping to Component.area so a hand-authored asset can be placed; an unknown area is dropped with a warning (asset kept). Without it the asset stays in the flat component list only - now documented.

mfaferek93:
Both docs points addressed in 4058cde: the gateway README source-precedence prose now includes inventory (manifest > inventory > config > runtime), and manifest-schema.rst + gateway_params.yaml document the assets: list and discovery.inventory.csv_path (columns, aliases, quoting, size cap, collision policy). sphinx -W clean.

bburda (approved):

PR #495, opened 2026-07-02
Fix clang-tidy noise from bundled headers and flaky secured-test deadline

mfaferek93:
Fixes the red clang-tidy job on PRs touching the OPC UA plugin, and the flaky secured integration test. Four independent commits:

- Mark the bundled open62541 include dirs as SYSTEM in the OPC UA plugin CMake. Adding `#include <open62541/client_highlevel.h>` surfaced hard `-Werror=old-style-cast` errors from the third-party headers, because only open62541pp's dirs were SYSTEM while the bundled open62541 target's were plain `-I`. Verified with a probe TU: before = error flood from the vendor headers, after = only the TU's own deliberate violation is reported.
- NOLINT-wrap the vendored headers that the tidy header filter unavoidably matches (`tl/expected.hpp`, the dynmsg headers) - their install paths contain "ros2_medkit". Verified the suppression does not leak: a deliberate violation in the including TU is still reported.
- Resolve the six real clang-tidy findings in the plugin sources (special-member-functions on client/plugin/poller, two no-effect `std::move` on trivially-copyable handles, one range-for). Note: the event-field case was not a plain move-drop - the manual `UA_Variant_copy` + adopt was replaced with a direct deep-copy `opcua::Variant` construction to avoid leaking the copied buffer.
- Raise the fault-surface wait deadline in the secured integration test (40s -> 120s; CTest timeout 300 -> 420). The alarm fires on the server but can miss the old deadline on slow runners; passing runs exit the poll early and are unaffected. Verified 3x green.

Tested: clang-tidy with the CI filter reports 0 findings on the touched sources; OPC UA suite 222 tests pass; flake8 clean on the test script.

### DIFF
--- a/src/ros2_medkit_gateway/src/vendored/tl_expected/include/tl/expected.hpp
+++ b/src/ros2_medkit_gateway/src/vendored/tl_expected/include/tl/expected.hpp
@@ -1,3 +1,5 @@
+// Vendored third-party header, excluded from clang-tidy lint (installed path matches the CI header filter).
+// NOLINTBEGIN
 ///
 // expected - An implementation of std::expected with extensions
 // Written in 2017 by Sy Brand (tartanllama@gmail.com, @TartanLlama)
@@ -2473,3 +2475,4 @@ void swap(expected<T, E> &lhs,
 } // namespace tl
 
 #endif
+// NOLINTEND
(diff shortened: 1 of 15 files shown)


mfaferek93:
Folded into #492.

PR #501, opened 2026-07-04
fix(opcua): accept severity as an alias in event_alarms

### LINKED ISSUE #500: opcua: event_alarms severity is silently dropped when set as severity:
The `event_alarms` node-map parser only reads `severity_override:`. A `severity:` key on an `event_alarms` entry is silently ignored (no unknown-key validation), so the configured severity never reaches the fault.

This is inconsistent with the threshold `alarm:` block, which uses `severity:`. An operator who reuses `severity:` for an event alarm gets the wrong severity: the value is dropped and the fault falls back to the live OPC UA event Severity. For sources whose events carry a low default Severity, an important alarm then surfaces as INFO.

### Proposal
- In the `event_alarms` loader, accept `severity:` as an alias for `severity_override:` (source-level and per-mapping), validated like the threshold path.
- Warn on unknown keys in the `event_alarms` block so future typos are not silently dropped.
- Keep the existing precedence: an explicit configured severity wins over the live event Severity; with none configured, the event Severity band mapping applies.

### Acceptance
An `event_alarms` entry configured with `severity: ERROR` (no `severity_override:`) raises the fault at ERROR regardless of the live event Severity.

mfaferek93:
Fixes `event_alarms` silently dropping a `severity:` key.

- `node_map.cpp`: the `event_alarms` loader now accepts `severity:` as an alias for `severity_override:` (source-level and per-mapping), routed through `validate_severity` like the threshold `alarm:` path, and warns on unknown keys in the block.
- The final severity resolution (explicit configured severity wins over the live event Severity; otherwise the Severity band mapping) is extracted into a small testable helper.
- Tests: an `event_alarms` entry with `severity: ERROR` resolves to ERROR; the helper covers override-wins and the band boundaries.

Closes #500.

### DIFF
--- a/src/ros2_medkit_plugins/ros2_medkit_opcua/src/node_map.cpp
+++ b/src/ros2_medkit_plugins/ros2_medkit_opcua/src/node_map.cpp
@@ -578,11 +578,20 @@ bool NodeMap::load(const std::string & yaml_path) {
       // the threshold path; with neither key the result stays empty so the live
       // event Severity band mapping applies (see OpcuaPlugin::map_severity).
       auto read_severity_override = [&](const YAML::Node & node, const std::string & context) -> std::string {
-        const YAML::Node sev_node = node["severity_override"] ? node["severity_override"] : node["severity"];
-        if (!sev_node) {
+        // yaml-cpp reports a present-but-null key (``severity_override:`` with
+        // no value / ``~``) as defined/truthy; gate on a real value so a null
+        // override neither shadows a valid ``severity:`` alias nor emits a
+        // misleading "non-string severity_override" warning.
+        auto has_value = [](const YAML::Node & n) {
+          return n.IsDefined() && !n.IsNull();
+        };
+        const YAML::Node override_node = node["severity_override"];
+        const bool use_override = has_value(override_node);
+        const YAML::Node sev_node = use_override ? override_node : node["severity"];
+        if (!has_value(sev_node)) {
           return "";
         }
-        const char * field = node["severity_override"] ? "severity_override" : "severity";
+        const char * field = use_override ? "severity_override" : "severity";
         auto sev = parse_string(sev_node, field, context);
         if (!sev || sev->empty()) {
           return "";
--- a/src/ros2_medkit_plugins/ros2_medkit_opcua/src/node_map.cpp
+++ b/src/ros2_medkit_plugins/ros2_medkit_opcua/src/node_map.cpp
@@ -19,6 +19,7 @@
 #include <cctype>
 #include <cstdint>
 #include <fstream>
+#include <initializer_list>
 #include <iostream>
 #include <optional>
 #include <sstream>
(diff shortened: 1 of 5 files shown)


bburda (commented):
Additional notes outside the inline comments:

- Behaviour change for existing configs: routing `severity_override` through `validate_severity` means a config with a non-canonical string (e.g. `severity_override: high`) no longer reaches the fault verbatim. It becomes `ERROR` with a warning. This matches the threshold `alarm:` path and is an improvement, but it changes the emitted severity for such configs. Worth a line in the PR description or CHANGELOG.
- `README.md` `event_alarms` section documents `severity_override` but not the new `severity` alias. A one-line note that `event_alarms` also accepts `severity:` would match the fix.

src/ros2_medkit_plugins/ros2_medkit_opcua/src/node_map.cpp
```diff
@@ -569,6 +570,47 @@ bool NodeMap::load(const std::string & yaml_path) {
...
       }
+      // Resolve the configured severity for an event_alarms entry / mapping.
+      // ``severity`` is accepted as an alias for ``severity_override`` because
+      // the threshold ``alarm:`` block uses ``severity:``; an operator who
+      // reuses that key here must not have it silently dropped. When both are
+      // present ``severity_override`` wins. An explicit value is validated like
+      // the threshold path; with neither key the result stays empty so the live
+      // event Severity band mapping applies (see OpcuaPlugin::map_severity).
+      auto read_severity_override = [&](const YAML::Node & node, const std::string & context) -> std::string {
+        const YAML::Node sev_node = node["severity_override"] ? node["severity_override"] : node["severity"];
```
bburda: `node["severity_override"]` is truthy in yaml-cpp whenever the key is *defined*, which includes a present-but-null value (`severity_override:` with no value, or `severity_override: ~`). So a config that leaves `severity_override:` blank next to a real `severity: ERROR` picks the null override, ignores the valid `severity:`, and emits a misleading `non-string severity_override - skipping` warning (the null node fails `as<std::string>()`). Gating on value presence - pick `severity` when `severity_override` is absent *or* `IsNull()` - avoids the shadow. Minor: the config is contrived and the fallback is the live band, but the warning is confusing.

src/ros2_medkit_plugins/ros2_medkit_opcua/test/test_node_map.cpp
```diff
@@ -584,6 +584,77 @@ component_id: test
...
+    entity_id: plc_program
+    fault_code: PLC_PROGRAM_ALARM
+    severity: ERROR
+)";
+  f.close();
+
+  NodeMap map;
+  ASSERT_TRUE(map.load(path));
+  ASSERT_EQ(map.event_alarms().size(), 1u);
+  EXPECT_EQ(map.event_alarms()[0].severity_override, "ERROR");
```
bburda, line 609: These three alias tests only use valid buckets (`ERROR` / `CRITICAL` / `WARNING`), so they pass unchanged even if the `validate_severity(*sev, context)` call were removed - they prove the alias is *read*, not that it is *validated*. Since validation is what this PR adds on top of the old raw `.as<std::string>("")`, one case with an invalid value (e.g. `severity: notabucket` asserting `severity_override == "ERROR"`) would pin the actual new behaviour.

src/ros2_medkit_plugins/ros2_medkit_opcua/src/node_map.cpp
mfaferek93: Fixed in 83551b8c: gated read_severity_override on a real value (IsDefined() && !IsNull()) so a present-but-null severity_override: no longer shadows a valid severity: alias nor emits the spurious 'non-string severity_override' warning; severity_override still wins when it carries a value. Added regression test EventAlarmNullSeverityOverrideFallsBackToAlias.

src/ros2_medkit_plugins/ros2_medkit_opcua/test/test_node_map.cpp
mfaferek93: Fixed in 83551b8c: added EventAlarmSeverityAliasInvalidBucketDefaultsToError with severity: notabucket asserting severity_override == "ERROR" (validate_severity's real behaviour is warn + default to ERROR, not reject). This now fails if the validate_severity(*sev) call is dropped from the alias path.

bburda (approved):
LGTM! +1

PR #503, opened 2026-07-04
fix(gateway): surface external plugin-entity faults (list, detail, freeze-frame) via fault-scope ownership

### LINKED ISSUE #502: sovd: per-entity fault list and detail are empty for external protocol-plugin entities
For entities introspected by a protocol plugin as external assets (e.g. a PLC over OPC UA), the per-entity SOVD fault routes return nothing:
- `GET /components/<id>/faults` is empty even when the entity has reported faults.
- `GET /components/<id>/faults/<code>` returns "Fault not found", and would not carry the freeze-frame even if found.

Root cause: fault ownership is resolved from an entity's FQN. An external app has no ROS binding, so its `effective_fqn()` is empty and `resolve_entity_source_fqns` yields an empty scope set. The scope filter then drops every fault the entity reported (which carry `reporting_sources = [entity_id]`), on both the plugin `list_entity_faults` path and the gateway `fault_in_source_scope` guard. So the entity is never recognised as the owner of its own faults.

### Proposal
- Ownership: an external app (no ROS binding) contributes its own entity id to its fault-scope set, so it owns faults reported under that id. ROS-bound entities keep their existing FQN behavior; the change is limited to the empty-FQN external case, in the fault-scope resolution only.
- Detail freeze-frame: for a plugin entity that owns the fault, the per-entity detail returns the fault_manager's enriched record (including the freeze-frame), consistent with the non-plugin detail path, falling back to the plugin's own provider for faults the fault_manager does not hold (e.g. on-demand DTCs).

### Acceptance
For an external plugin entity that reported a fault: `GET /components/<id>/faults` lists it, and `GET /components/<id>/faults/<code>` returns it with its freeze-frame. Cross-entity disclosure and cross-entity clear stay blocked (scope guard unchanged for bound entities). A plugin-only fault not in the fault_manager is still served by the plugin provider.

mfaferek93:
External protocol-plugin entities (e.g. a PLC over OPC UA) report faults to the fault_manager under their entity id, but the per-entity SOVD fault routes dropped them: an external app has no ROS FQN, so fault-scope resolution produced an empty ownership set and the scope filter rejected every fault the entity reported.

- Fault-scope: an external app with no ROS binding now contributes its own entity id to its scope set, so it is recognised as the owner of faults reported under that id. ROS-bound entities are unchanged; the change is confined to the empty-FQN external case in the fault-scope resolution.
- Per-entity detail: for a plugin entity that owns the fault, the detail route returns the fault_manager's environment-enriched record (including the freeze-frame), consistent with the non-plugin detail path, and falls back to the plugin's own fault provider for faults the fault_manager does not hold.
- The per-entity list now surfaces the entity's faults via the same fixed ownership.
- Scope-guard behaviour (cross-entity disclosure on GET, cross-entity clear on DELETE) is unchanged for bound entities and covered by regression tests.

Closes #502.

### DIFF
--- a/src/ros2_medkit_gateway/src/http/handlers/fault_handlers.cpp
+++ b/src/ros2_medkit_gateway/src/http/handlers/fault_handlers.cpp
@@ -797,6 +797,34 @@ FaultHandlers::clear_fault(const http::TypedRequest & req) {
         return tl::make_unexpected(
             make_error(404, ERR_RESOURCE_NOT_FOUND, "No fault provider for plugin entity '" + entity_id + "'"));
       }
+
+      // Cross-entity clear guard: the plugin clear_fault forwards the code to
+      // the fault_manager with no scope check, so without this a
+      // DELETE /{A}/faults/{code} could clear a fault owned by entity B. When
+      // the fault_manager holds this fault, require it to be in this entity's
+      // source scope before delegating; reject out-of-scope. Faults the
+      // fault_manager does not hold (plugin-internal, e.g. on-demand UDS DTCs)
+      // fall through to the plugin provider unchanged. Mirrors the ownership
+      // check in the plugin get_fault branch and the non-plugin clear path.
+      if (auto * fault_mgr = ctx_.node()->get_fault_manager(); fault_mgr != nullptr) {
+        auto mgr_result = fault_mgr->get_fault_with_env(fault_code, "");
+        if (mgr_result.success) {
+          const auto & owned_fault_json = mgr_result.data.value("fault", json::object());
+          const auto & cache = ctx_.node()->get_thread_safe_cache();
+          auto source_fqns = HandlerContext::resolve_entity_source_fqns(cache, entity_info);
+          if (!FaultHandlers::fault_in_source_scope(owned_fault_json, source_fqns)) {
+            return tl::make_unexpected(
+                make_error(404, ERR_RESOURCE_NOT_FOUND, "Fault not found",
+                           json{{"details",
+                                 "Fault is not in scope for this entity: every reporting source must be one of the "
+                                 "entity's owned apps, and a mixed-source fault that includes any out-of-entity "
+                                 "reporter is rejected to prevent cross-entity clear"},
+                                {entity_info.id_field, entity_id},
+                                {"fault_code", fault_code}}));
+          }
+        }
+      }
+
       try {
         auto result = fault_prov->clear_fault(entity_id, fault_code);
         if (!result) {
--- a/src/ros2_medkit_gateway/src/http/handlers/fault_handlers.cpp
+++ b/src/ros2_medkit_gateway/src/http/handlers/fault_handlers.cpp
@@ -655,6 +655,27 @@ http::Result<dto::FaultDetailResult> FaultHandlers::get_fault(const http::TypedR
         return tl::make_unexpected(
             make_error(404, ERR_RESOURCE_NOT_FOUND, "No fault provider for plugin entity '" + entity_id + "'"));
       }
+
+      // Prefer the fault_manager's environment-enriched record when this entity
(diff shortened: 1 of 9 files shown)


src/ros2_medkit_gateway/test/test_fault_handlers.cpp
```diff
@@ -503,6 +503,77 @@ TEST(FaultInSourceScopeTest, RootFqnEdgeCase) {
...
 }
 
+// -----------------------------------------------------------------------------
+// External protocol-plugin entity ownership: bare-id scope sets.
+//
+// An external entity (e.g. a PLC over OPC UA) has no ROS FQN, so fault-scope
+// resolution grants it a scope set of exactly {its bare id}. These pins prove
+// the guard treats a bare id like any other scope token: it owns faults
+// reported under that id, and still blocks cross-entity disclosure (GET) and
+// cross-entity clear (DELETE), which both flow through this predicate.
```
bburda, line 513: This says cross-entity clear (DELETE) flows through this predicate, but for a plugin or external entity it does not. The single-fault DELETE takes the `is_plugin` branch and delegates to the plugin `clear_fault`, which for OPC UA forwards the code to `/fault_manager/clear_fault` with no scope check. So `DELETE /apps/A/faults/{code owned by B}` clears B's fault. Either apply the same `fault_in_source_scope` guard on the plugin DELETE branch, or correct this comment so it does not claim a property the plugin path does not enforce.

bburda (commented):
Additional notes outside the inline comment:

- The get_fault enriched-vs-fallback branch (the main change) has no end-to-end handler test. The four pieces are unit-tested, but nothing builds a node and calls `get_fault`, so the security selection is unpinned: an entity that does NOT own a fault code must fall through to the plugin provider (404), not return another entity's enriched freeze-frame record. If the new scope check is later changed, a cross-entity read leak would not be caught.
- The per-entity list and single-fault clear routes for an external entity are only covered at the resolver unit level, not end to end. The empty-list bug this PR fixes has no regression test on its own route.

src/ros2_medkit_gateway/src/http/handlers/fault_handlers.cpp
```diff
@@ -655,6 +655,27 @@ http::Result<dto::FaultDetailResult> FaultHandlers::get_fault(const http::TypedR
         return tl::make_unexpected(
             make_error(404, ERR_RESOURCE_NOT_FOUND, "No fault provider for plugin entity '" + entity_id + "'"));
       }
+
```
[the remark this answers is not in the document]
mfaferek93: Fixed in 0146cf18: hoisted the empty/length guard (fault_code.empty() || length > 256) ahead of the plugin branch in get_fault (and clear_fault for symmetry), so the plugin path rejects invalid/oversized codes before hitting get_fault_with_env or the plugin FaultProvider, matching the non-plugin path. The old post-branch check is removed (no duplication).

src/ros2_medkit_gateway/test/test_fault_handlers.cpp
mfaferek93: Fixed in eda690ae. The plugin single-fault DELETE branch now applies the same source-scope guard as the non-plugin path and the plugin get_fault branch: get_fault_with_env + fault_in_source_scope before delegating, rejecting out-of-scope fault_manager faults with 404; plugin-internal faults the fault_manager does not hold fall through unchanged. So DELETE /apps/A/faults/{code owned by B} now returns 404 instead of clearing B, and the property this comment describes holds for the plugin path too. The predicate cross-entity/mixed-source cases are covered by BareEntityIdScopeRejectsOtherEntityFault and RejectsMixedSourceFault.

bburda (approved):
LGTM! +1

PR #506, opened 2026-07-05
feat(opcua): route alarms to distinct faults by message substring

### LINKED ISSUE #505: opcua: route same-source alarms to distinct faults by message text
The opcua plugin's `event_alarms` mappings route events to distinct fault codes by `condition_name`, `source_node`, or `event_type`. Some PLCs emit many alarms that share all three and differ only by the Message text - notably a Siemens S7-1500 FB whose Program_Alarms all surface under one AlarmConditionType EventType and one SourceNode. Those alarms currently collapse into a single fault because no mapping field distinguishes them.

### Proposal
Add a `match_message` field to `event_alarms[].mappings`: a case-sensitive substring match on the event Message, AND-combined with the existing matchers. An alarm whose Message contains the substring routes to the mapping's `fault_code`; unmatched events fall back to the source-level `fault_code`.

### Acceptance
A config with two mappings differing only by `match_message` routes two alarms that share one EventType/SourceNode to two distinct fault codes by their text.

mfaferek93:
Adds a `match_message` field to `event_alarms[].mappings` - a case-sensitive substring match on the event Message, AND-combined with the existing `condition_name` / `source_node` / `event_type` matchers. This lets one OPC UA source route alarms that share an EventType and SourceNode (e.g. Program_Alarms from one Siemens S7-1500 FB) to distinct SOVD faults by their text.

- `AlarmMapping` gains `match_message`; `resolve_alarm` takes the event Message and skips a mapping whose non-empty `match_message` is not a substring of it. `condition_name` / `source_node` / `event_type` stay equality matches.
- Parsed from `mappings[].match_message`; the poller passes the live event Message (and the read-path snapshot message).
- README example + unit test (`MatchByMessageSubstring`) covering a match, the catch-all fallback, and an empty message.

Closes #505

### DIFF
--- a/src/ros2_medkit_plugins/ros2_medkit_opcua/include/ros2_medkit_opcua/node_map.hpp
+++ b/src/ros2_medkit_plugins/ros2_medkit_opcua/include/ros2_medkit_opcua/node_map.hpp
@@ -54,8 +54,11 @@ struct AssociatedValueRef {
 /// #389). A single OPC-UA source (e.g. the Server object as a catch-all, or a
 /// real owning Object) can emit many distinct conditions; each mapping routes
 /// a subset to its own SOVD fault. Match fields are AND-combined; an empty
-/// match field is a wildcard. The first mapping whose non-empty match fields
-/// all equal the observed event wins (declaration order = precedence).
+/// match field is a wildcard. ``match_condition_name`` / ``match_source_node``
+/// / ``match_event_type`` are equality matches; ``match_message`` is a
+/// case-sensitive substring match on the event Message. The first mapping whose
+/// non-empty match fields all match the observed event wins (declaration order
+/// = precedence).
 struct AlarmMapping {
   std::string match_condition_name;  // ConditionType.ConditionName (empty = any)
   std::string match_source_node;     // event SourceNode id string (empty = any)
--- a/src/ros2_medkit_plugins/ros2_medkit_opcua/include/ros2_medkit_opcua/node_map.hpp
+++ b/src/ros2_medkit_plugins/ros2_medkit_opcua/include/ros2_medkit_opcua/node_map.hpp
@@ -60,6 +60,7 @@ struct AlarmMapping {
   std::string match_condition_name;  // ConditionType.ConditionName (empty = any)
   std::string match_source_node;     // event SourceNode id string (empty = any)
   std::string match_event_type;      // event EventType id string (empty = any)
+  std::string match_message;         // Message substring (empty = any); case-sensitive contains
 
   std::string fault_code;
   std::string severity_override;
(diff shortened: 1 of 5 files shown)


src/ros2_medkit_plugins/ros2_medkit_opcua/include/ros2_medkit_opcua/node_map.hpp
```diff
@@ -60,6 +60,7 @@ struct AlarmMapping {
   std::string match_condition_name;  // ConditionType.ConditionName (empty = any)
   std::string match_source_node;     // event SourceNode id string (empty = any)
   std::string match_event_type;      // event EventType id string (empty = any)
+  std::string match_message;         // Message substring (empty = any); case-sensitive contains
```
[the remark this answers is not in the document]
mfaferek93: Fixed in 232bd4d3. The struct doc now spells out that condition_name/source_node/event_type are equality matches and match_message is a case-sensitive substring match on the event Message, so the "all equal" wording no longer misdescribes the substring field.

src/ros2_medkit_plugins/ros2_medkit_opcua/test/test_node_map.cpp
```diff
@@ -1527,12 +1530,42 @@ TEST(ResolveAlarmTest, MappingInheritsSourceLevelOverridesWhenUnset) {
...
+  EXPECT_EQ(r2.fault_code, "PLC_ALARM_SECOND");
+
+  // No substring match -> source-level catch-all.
+  auto r3 = NodeMap::resolve_alarm(cfg, "", "ns=3;i=1845", "ns=3;i=1805", "Unrelated text");
+  EXPECT_TRUE(r3.matched);
+  EXPECT_EQ(r3.fault_code, "PLC_PROGRAM_ALARM");
+
+  // Empty event message never matches a non-empty match_message -> catch-all.
+  auto r4 = NodeMap::resolve_alarm(cfg, "", "ns=3;i=1845", "ns=3;i=1805", "");
+  EXPECT_EQ(r4.fault_code, "PLC_PROGRAM_ALARM");
```
bburda, line 1566: This test does not exercise the case-sensitivity that the struct, the code comment, and the README all document: every case here uses an exact-case substring, so replacing `find` with a case-insensitive search would keep all assertions green. Consider one more case that pins it, e.g. `match_message` "Overtemp" against message "overtemp fault" should fall through to the catch-all (`PLC_PROGRAM_ALARM`).

bburda (commented):
Additional findings outside the diff (both in `ros2_medkit_opcua/README.md`):

- The `event_alarms` example uses `PLC_OVERPRESSURE` twice: once on the simple-form source (`entity_id: tank_process`) and once in the first mapping of the `line_1` source. The loader rejects a fault code reused across sources, which the comment just above this example states, so this example fails to load if copied as-is. It predates this PR, but since you are editing this block, renaming one `PLC_OVERPRESSURE` would make the documented example loadable.
- The multi-alarm form comment still says "the first whose non-empty match fields all `equal` the observed event wins". With `match_message` being a substring match, this is no longer accurate. You updated the precedence paragraph below (equal to match), so this inline comment can get the same fix.

src/ros2_medkit_plugins/ros2_medkit_opcua/test/test_node_map.cpp
mfaferek93: Fixed in 1abf5537. Added r5/r6 to `MatchByMessageSubstring`: `"overtemp fault"` against `match_message: "Overtemp"` must fall through to the catch-all `PLC_PROGRAM_ALARM`, and the exact-case `"Overtemp fault"` routes to the distinct fault. Confirmed with a negative control that flipping `find` to a case-insensitive search turns r5 red. `test_node_map` green.

bburda (approved):

PR #507, opened 2026-07-05
feat(opcua): raise PLC_COMMS_LOST fault on connection loss

### LINKED ISSUE #496: opcua: raise a comms-lost fault when the OPC UA connection drops
## Problem

When the OPC UA plugin loses its connection to the server (PLC powered off, cable/network fault, session drop), the poller sets `snapshot_.connected = false` and enters its reconnect-with-backoff loop (`opcua_poller.cpp` reconnect arm, ~L849), and the `connected` flag is surfaced on the `x-plc-status` view (`opcua_plugin.cpp`, `j["connected"]`). But no fault is raised on `/faults`.

So a controller going dark is invisible to `/faults` consumers: an operator watching the fault list sees nothing, even though the plugin has just gone blind to everything that PLC was reporting. The only signal is the `connected` boolean on a status endpoint that fault consumers do not watch. ISA-18.2 treats loss of communication with a controller as an alarm-worthy condition.

## Proposal

Raise a single, component-scoped comms-lost fault:

- Fault code `PLC_COMMS_LOST`, severity `ERROR` (configurable).
- Raised on the PLC component (the plugin's root PLC component/source), not per data point.
- Debounced: only raise once the connection has been continuously down for a configurable window (default ~5s), so a brief blip during a normal reconnect does not flap a fault. Idempotent (raise once, not every poll cycle).
- Cleared on successful reconnect.
- Config knobs: `comms_lost_fault_enabled` (default true), `comms_lost_debounce_ms` (default 5000).

This is a connection-state transition, so it belongs in the poller's reconnect arm using the existing `report_fault` / `clear_fault` path, not the value-evaluation path.

## Notes

- `report_fault` / `clear_fault` clients already exist (`opcua_plugin.cpp`), and the PLC component the fault should attach to already exists (`opcua_plugin.cpp`, "PLC runtime connected at ...").
- Keeps `/faults` semantics uniform across PLC sources.

mfaferek93:
Raises a single component-scoped `PLC_COMMS_LOST` fault (severity `ERROR`, configurable) when the OPC UA connection stays down, so a controller going dark surfaces on `/faults` instead of only the `x-plc-status` `connected` boolean. Brings the opcua plugin to parity with the modbus/s7/ads bridges.

- Debounced: raised only after the connection is continuously down for `comms_lost_debounce_ms` (default 5000), so a brief blip during a normal reconnect does not flap a fault. Idempotent (raised once), cleared on the next successful reconnect.
- Component-scoped to the node map's root component; message carries the endpoint.
- Config knobs: `comms_lost_fault_enabled` (default true), `comms_lost_debounce_ms` (default 5000), `comms_lost_severity` (default ERROR); env `OPCUA_COMMS_LOST_ENABLED` / `OPCUA_COMMS_LOST_DEBOUNCE_MS`.
- The debounce gate is a pure static helper with unit tests (`CommsLostShouldRaise`).

Closes #496

### DIFF
--- a/src/ros2_medkit_plugins/ros2_medkit_opcua/README.md
+++ b/src/ros2_medkit_plugins/ros2_medkit_opcua/README.md
@@ -491,6 +491,8 @@ Write operations use the `set_` prefix convention:
 | `OPCUA_USER_CERT` | X.509 user-token cert (DER) |
 | `OPCUA_CONDITION_REPLAY` | `method` / `read` / `auto` / `off` |
 | `OPCUA_REQUIRE_CONFIRM_FOR_CLEAR` | `0`/`false`/`no`/`off` to clear native alarms on Acknowledge alone (Confirm-less servers) |
+| `OPCUA_COMMS_LOST_ENABLED` | `0`/`false`/`no`/`off` to disable the `PLC_COMMS_LOST` fault |
+| `OPCUA_COMMS_LOST_DEBOUNCE_MS` | Continuous down time (ms) before `PLC_COMMS_LOST` is raised |
 
 ## Hardware Deployment
 
--- a/src/ros2_medkit_plugins/ros2_medkit_opcua/README.md
+++ b/src/ros2_medkit_plugins/ros2_medkit_opcua/README.md
@@ -367,6 +367,9 @@ ros2_medkit_gateway:
 | `subscription_interval_ms` | `500` | Publishing interval for OPC-UA subscriptions when `prefer_subscriptions: true` |
 | `condition_replay_strategy` | `auto` | Active-condition replay on reconnect: `method`, `read`, `auto`, `off` (see below) |
 | `require_confirm_for_clear` | `true` | Require both Acknowledge AND Confirm before a native alarm auto-clears. Set `false` for Confirm-less servers (e.g. Siemens S7-1500) so alarms clear on Acknowledge alone (see below) |
+| `comms_lost_fault_enabled` | `true` | Raise a component-scoped `PLC_COMMS_LOST` fault when the connection stays down (issue #496) |
+| `comms_lost_debounce_ms` | `5000` | Continuous down time before `PLC_COMMS_LOST` is raised (debounces reconnect blips) |
+| `comms_lost_severity` | `ERROR` | SOVD severity bucket for the `PLC_COMMS_LOST` fault |
 
 ### OPC-UA client security (SecurityPolicy, certificates, user auth)
 
--- a/src/ros2_medkit_plugins/ros2_medkit_opcua/README.md
+++ b/src/ros2_medkit_plugins/ros2_medkit_opcua/README.md
@@ -368,7 +368,7 @@ ros2_medkit_gateway:
 | `condition_replay_strategy` | `auto` | Active-condition replay on reconnect: `method`, `read`, `auto`, `off` (see below) |
 | `require_confirm_for_clear` | `true` | Require both Acknowledge AND Confirm before a native alarm auto-clears. Set `false` for Confirm-less servers (e.g. Siemens S7-1500) so alarms clear on Acknowledge alone (see below) |
 | `comms_lost_fault_enabled` | `true` | Raise a component-scoped `PLC_COMMS_LOST` fault when the connection stays down (issue #496) |
-| `comms_lost_debounce_ms` | `5000` | Continuous down time before `PLC_COMMS_LOST` is raised (debounces reconnect blips) |
+| `comms_lost_debounce_ms` | `5000` | Continuous down time before `PLC_COMMS_LOST` is raised (debounces reconnect blips; clamped to [0, 3600000] ms) |
 | `comms_lost_severity` | `ERROR` | SOVD severity bucket for the `PLC_COMMS_LOST` fault |
 
 ### OPC-UA client security (SecurityPolicy, certificates, user auth)
@@ -492,7 +492,7 @@ Write operations use the `set_` prefix convention:
 | `OPCUA_CONDITION_REPLAY` | `method` / `read` / `auto` / `off` |
 | `OPCUA_REQUIRE_CONFIRM_FOR_CLEAR` | `0`/`false`/`no`/`off` to clear native alarms on Acknowledge alone (Confirm-less servers) |
 | `OPCUA_COMMS_LOST_ENABLED` | `0`/`false`/`no`/`off` to disable the `PLC_COMMS_LOST` fault |
-| `OPCUA_COMMS_LOST_DEBOUNCE_MS` | Continuous down time (ms) before `PLC_COMMS_LOST` is raised |
+| `OPCUA_COMMS_LOST_DEBOUNCE_MS` | Continuous down time (ms) before `PLC_COMMS_LOST` is raised (clamped to [0, 3600000] ms; non-numeric / out-of-range keeps the existing value) |
 
 ## Hardware Deployment
 
--- a/src/ros2_medkit_plugins/ros2_medkit_opcua/src/opcua_plugin.cpp
+++ b/src/ros2_medkit_plugins/ros2_medkit_opcua/src/opcua_plugin.cpp
@@ -26,6 +26,7 @@
 
 #include <algorithm>
 #include <cctype>
+#include <cerrno>
 #include <chrono>
 #include <cmath>
 #include <cstdio>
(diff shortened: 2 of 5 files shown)


src/ros2_medkit_plugins/ros2_medkit_opcua/src/opcua_plugin.cpp
```diff
@@ -232,6 +232,24 @@ void OpcuaPlugin::configure(const nlohmann::json & config) {
...
+  if (auto * env = std::getenv("OPCUA_COMMS_LOST_ENABLED")) {
+    const std::string v = env;
+    poller_config_.comms_lost_fault_enabled = !(v == "0" || v == "false" || v == "no" || v == "off");
+  }
+  if (config.contains("comms_lost_debounce_ms")) {
+    poller_config_.comms_lost_debounce = std::chrono::milliseconds(config["comms_lost_debounce_ms"].get<int>());
+  }
+  if (auto * env = std::getenv("OPCUA_COMMS_LOST_DEBOUNCE_MS")) {
+    poller_config_.comms_lost_debounce = std::chrono::milliseconds(std::atoi(env));
+  }
```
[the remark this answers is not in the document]
mfaferek93: Fixed in 90c407d9. YAML `comms_lost_debounce_ms` is now clamped to >= 0, and `OPCUA_COMMS_LOST_DEBOUNCE_MS` is parsed with strtol + full-string/non-negative validation - an invalid or negative override is ignored with a warning and the existing value kept, instead of atoi silently yielding 0.

src/ros2_medkit_plugins/ros2_medkit_opcua/src/opcua_plugin.cpp
```diff
@@ -232,6 +232,24 @@ void OpcuaPlugin::configure(const nlohmann::json & config) {
...
+  if (config.contains("comms_lost_severity")) {
+    poller_config_.comms_lost_severity = config["comms_lost_severity"].get<std::string>();
+  }
```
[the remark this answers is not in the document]
mfaferek93: Fixed in 90c407d9. `comms_lost_severity` is now validated against the SOVD buckets (INFO/WARNING/ERROR/CRITICAL) and defaults to ERROR with a warning on an unknown value, matching the node-map severity guard, so a typo no longer falls through to INFO in send_report_fault.

src/ros2_medkit_plugins/ros2_medkit_opcua/README.md
```diff
@@ -359,6 +359,9 @@ ros2_medkit_gateway:
 | `subscription_interval_ms` | `500` | Publishing interval for OPC-UA subscriptions when `prefer_subscriptions: true` |
 | `condition_replay_strategy` | `auto` | Active-condition replay on reconnect: `method`, `read`, `auto`, `off` (see below) |
 | `require_confirm_for_clear` | `true` | Require both Acknowledge AND Confirm before a native alarm auto-clears. Set `false` for Confirm-less servers (e.g. Siemens S7-1500) so alarms clear on Acknowledge alone (see below) |
+| `comms_lost_fault_enabled` | `true` | Raise a component-scoped `PLC_COMMS_LOST` fault when the connection stays down (issue #496) |
+| `comms_lost_debounce_ms` | `5000` | Continuous down time before `PLC_COMMS_LOST` is raised (debounces reconnect blips) |
+| `comms_lost_severity` | `ERROR` | SOVD severity bucket for the `PLC_COMMS_LOST` fault |
```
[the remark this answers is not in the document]
mfaferek93: Fixed in 90c407d9. Added OPCUA_COMMS_LOST_ENABLED and OPCUA_COMMS_LOST_DEBOUNCE_MS to the Environment Variables table.

bburda (commented):
Additional notes outside the diff:

- `ros2_medkit_opcua/design/index.rst` documents the PLC condition/event to fault flow but not `PLC_COMMS_LOST`, which is different in kind (raised by the poll thread on connection loss, not sourced from a PLC condition node). A short note there would keep the design doc in step with the README.
- Only the pure `comms_lost_should_raise` gate is unit-tested. The clear path (`emit_comms_lost(false)` plus resetting `comms_lost_raised_` and `comms_down_since_` on reconnect) has no coverage and is the more state-heavy half. Extracting a pure reconnect-transition helper would let it be tested the same way as the raise gate.

src/ros2_medkit_plugins/ros2_medkit_opcua/src/opcua_plugin.cpp
```diff
@@ -232,6 +232,44 @@ void OpcuaPlugin::configure(const nlohmann::json & config) {
...
+  if (config.contains("comms_lost_debounce_ms")) {
+    const int ms = config["comms_lost_debounce_ms"].get<int>();
+    poller_config_.comms_lost_debounce = std::chrono::milliseconds(ms < 0 ? 0 : ms);
+  }
+  if (auto * env = std::getenv("OPCUA_COMMS_LOST_DEBOUNCE_MS")) {
+    // Keep the existing value on a non-numeric / negative override so a typo
+    // does not silently disable the debounce (atoi would parse it as 0).
+    char * end = nullptr;
+    const long ms = std::strtol(env, &end, 10);
+    if (end != env && *end == '\0' && ms >= 0) {
```
bburda: The `strtol` guard does not check `errno == ERANGE`, so an overflowing value (a long run of digits) returns `LONG_MAX`, still passes `end != env && *end == '\0' && ms >= 0`, and sets the debounce to about `LONG_MAX` ms. The gate then never fires, which silently disables the fault - the exact outcome the comment above says this guard prevents. Set `errno = 0` before `strtol` and treat `ERANGE` as invalid. The YAML path just above (`get<int>()`, clamped only for negatives) has the same overflow gap and no upper bound, while `poll_interval_ms` clamps to [100, 60000]; matching that would make the two paths consistent.

src/ros2_medkit_plugins/ros2_medkit_opcua/src/opcua_poller.cpp
```diff
@@ -872,6 +908,13 @@ void OpcuaPoller::poll_loop() {
           setup_event_subscriptions();
         }
       } else {
+        // Issue #496: still down - raise the component-scoped comms-lost fault
+        // once the debounce window has elapsed (idempotent via the raised flag).
+        if (comms_lost_should_raise(config_.comms_lost_fault_enabled, comms_lost_raised_, *comms_down_since_,
+                                    std::chrono::steady_clock::now(), config_.comms_lost_debounce)) {
+          emit_comms_lost(/*active=*/true);
+          comms_lost_raised_ = true;
```
bburda: `comms_lost_raised_` is set here regardless of whether the report was delivered. The report is fire-and-forget (`async_send_request`, no readiness check or retry) and nothing waits for the report service before the poller starts, so if `/fault_manager/report_fault` is not discovered yet when the raise fires, the report is dropped and the fault never re-fires until a full reconnect and re-down. Unlike native alarms, which re-fire on ConditionRefresh after reconnect, comms-lost is strictly one-shot. The 5s debounce makes this unlikely because the client is usually discovered by then, so it is low risk. Worth gating the flag on confirmed delivery, or accepting and noting it?

src/ros2_medkit_plugins/ros2_medkit_opcua/src/opcua_plugin.cpp
mfaferek93: Fixed in 520e63a9. Set `errno = 0` before `strtol` and reject `ERANGE`, so an overflowing `OPCUA_COMMS_LOST_DEBOUNCE_MS` is now ignored (existing value kept + warning) instead of saturating to `LONG_MAX` and slipping past the `>= 0` guard. Both the YAML `get<>` path and the env path go through one shared clamp to `[0, 3600000]` ms (1 h cap), documented in the README knob tables. Verified the parse edge cases (overflow / negative / non-numeric rejected; `7200000` -> `3600000`) and `test_opcua_plugin` stays green (38/38).

src/ros2_medkit_plugins/ros2_medkit_opcua/src/opcua_poller.cpp
mfaferek93: Fixed in 18bc2b7c. `comms_lost_raised_` is now latched only once the fault sink is ready (new `report_sink_ready` probe = ReportFault `service_is_ready`), checked in the poll arm. If the sink is not yet matched the arm leaves the flag unset and retries on the next poll, so the report is no longer sent-then-forgotten. Recovery stays best-effort as it runs once on reconnect.

bburda (dismissed):

bburda (approved):

PR #508, opened 2026-07-08
ci: arm64 multi-arch image (tag/dispatch gated)

mfaferek93:
Adds a native-runner multi-arch (linux/amd64 + linux/arm64) image publish for the gateway.

## What
- New workflow `.github/workflows/docker-publish-multiarch.yml`.
- Builds each arch on its own native runner (arm64 on `ubuntu-24.04-arm`, no QEMU), pushes by digest, then a per-distro merge job stitches the digests into one multi-arch manifest and tags it.
- Mirrors the `ros2_medkit_web_ui` digest + manifest-merge pattern, extended with the `ros_distro` build-arg dimension (jazzy, humble, lyrical).

## Gating
- Triggers on `push` to `v*` tags and `workflow_dispatch` ONLY. NOT on pull_request or push-to-main.
- The existing `docker-publish.yml` (amd64-only, push-to-main) is left untouched, so per-PR / push-to-main CI stays fast (no emulated arm64 on every change).

## Feasibility
- Verified `ros:{jazzy,humble,lyrical}-ros-base` all publish `linux/arm64/v8` manifests, so arm64 builds are feasible. No base-image blocker.

### DIFF
--- a/.github/workflows/docker-publish-multiarch.yml
+++ b/.github/workflows/docker-publish-multiarch.yml
@@ -138,8 +138,9 @@ jobs:
                       type=semver,pattern={{version}}
                       # major.minor (v1.2.3 -> 1.2)
                       type=semver,pattern={{major}}.{{minor}}
-                      # always publish latest for release / dispatch
-                      type=raw,value=latest
+                      # latest only on a stable release tag (vX.Y.Z with no
+                      # pre-release suffix); skips -rc/-alpha tags and dispatch
+                      type=raw,value=latest,enable=${{ startsWith(github.ref, 'refs/tags/v') && !contains(github.ref, '-') }}
                       # sha for traceability
                       type=sha,prefix=sha-,format=short
 
--- a/.github/workflows/docker-publish-multiarch.yml
+++ b/.github/workflows/docker-publish-multiarch.yml
@@ -1,0 +1,156 @@
+# Multi-arch (linux/amd64 + linux/arm64) image publish for the gateway.
+#
+# Gated on release tags and manual dispatch only - NOT on pull_request or
+# push-to-main. Per-PR / push-to-main CI still runs the amd64-only
+# docker-publish.yml, which stays fast (no emulated arm64 on every change).
+#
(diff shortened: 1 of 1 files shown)


.github/workflows/docker-publish-multiarch.yml
```diff
@@ -0,0 +1,156 @@
...
+              uses: docker/metadata-action@v5
+              with:
+                  images: ${{ env.REGISTRY }}/${{ github.repository_owner }}/ros2_medkit-${{ matrix.ros_distro }}
+                  tags: |
+                      # version tag (v1.2.3 -> 1.2.3)
+                      type=semver,pattern={{version}}
+                      # major.minor (v1.2.3 -> 1.2)
+                      type=semver,pattern={{major}}.{{minor}}
+                      # always publish latest for release / dispatch
+                      type=raw,value=latest
```
bburda: `type=raw,value=latest` has no `enable` guard, so `:latest` is set on every trigger, not only on stable release tags.

Two cases move `:latest` when they should not:
- The `v*` trigger also matches pre-release tags like `v1.2.3-rc1`. A release candidate would then become `:latest`.
- `workflow_dispatch` can run on any branch. There the semver tags are empty but `latest` is still set, so a manual run publishes `:latest` from that branch.

It also conflicts with `docker-publish.yml`. That workflow sets the same `:latest` tag as amd64-only on each push to main. So after a release `:latest` is multi-arch, and the next push to main makes it amd64-only again.

Fix: guard the tag, for example `type=raw,value=latest,enable=${{ github.ref_type == 'tag' }}`.

mfaferek93: Fixed in bf105acc: guarded the raw latest tag with enable=${{ startsWith(github.ref, 'refs/tags/v') && !contains(github.ref, '-') }}. This is stricter than ref_type == 'tag' alone: it also drops pre-release tags (any ref containing a hyphen, e.g. v1.2.3-rc1) and workflow_dispatch runs, so :latest only moves on a stable vX.Y.Z release.

bburda (approved):

PR #511, opened 2026-07-09
feat(opcua): zero-config native A&C alarms -> faults

### LINKED ISSUE #514: ros2_medkit_opcua: zero-config native A&C alarms -> faults
**Context.** Native OPC UA Alarms & Conditions require a per-alarm `event_alarms` mapping. A zero-config box should surface native PLC alarms as faults without hand-mapping each one.

**Scope.** Subscribe to the server EventNotifier for AlarmConditionType events; auto-derive a stable `fault_code` from condition identity (ConditionName/SourceName, with the alarm Message folded in - Siemens multiplexes multiple Program_Alarms of one FB through a single SourceNode); map Severity to a band; filter system messages (null ConditionId, e.g. "CPU not in RUN"); explicit `event_alarms` mappings take precedence. Config-gated, default off.

**Acceptance.** With only `auto_alarms` enabled (no `event_alarms`), a native PLC alarm surfaces as a fault with a sensible code + message and clears on inactive.

mfaferek93:
Closes #514

## Summary

Zero-config native OPC-UA Alarms & Conditions: subscribe to the server's alarm events and surface them as faults with no per-alarm `event_alarms` mapping. Completes "find a PLC (#509) -> tree (#510) -> native faults, zero config". Default off.

## What's included

- **Config** `auto_alarms`: bare `true` or a map (`source_node_id` default `i=2253`, `entity_id` default `<component_id>_alarms`, `auto_clear`, `severity_bands`, `include`/`exclude` substring filters). Unknown keys warn.
- **Fault derivation** (`derive_auto_fault_code`, pure): `PLC_ALARM_<slug(ConditionName)>` -> `slug(SourceName)` -> `hash(SourceNode+EventType+Message)`. Message is folded into the hash tier because Siemens multiplexes every Program_Alarm of one FB through a single SourceNode with empty names, distinguished only by message.
- **System-message filter**: events with a null `ConditionId` are dropped (Part 9 - only real Conditions carry one), so a "CPU not in RUN" system message on the Server EventNotifier raises no false fault.
- **Precedence**: explicit `event_alarms` mappings resolve first; `auto_alarms` adds one synthetic mapping-less source for the rest (no duplicate subscriptions).

## Tests

27 gtest cases (`test_node_map`, new `test_opcua_poller`): derivation tiers, message disambiguation, filters, severity bands, YAML parse, precedence, system-message filtering. Full suite green; `clang-format-18` clean.

## Hardware-tested

Real Siemens (`opc.tcp://192.168.1.10:4840`, zero `event_alarms`): zero-config subscription live, real "CPU not in RUN" system message correctly filtered (no false fault). Full raise->derive->clear proven against a real OPC-UA session (repo `test_alarm_server`): `Overpressure`/`SensorLost` -> `PLC_ALARM_OVERPRESSURE` (CRITICAL) / `PLC_ALARM_SENSORLOST` (INFO) with correct message/entity, cleared on clear. Fixed along the way: `auto_clear` must force both Ack+Confirm gates (auto alarms have no acknowledge op); default `entity_id` = `<component_id>_alarms` (bare component_id collided with the Component entity in the SOVD merge).

### DIFF
--- a/src/ros2_medkit_plugins/ros2_medkit_opcua/README.md
+++ b/src/ros2_medkit_plugins/ros2_medkit_opcua/README.md
@@ -467,13 +467,22 @@ This completes the zero-config chain: discovery (endpoint) ->
 no node-map file required.
 
 **Fault derivation** (no config, per observed event):
-- `fault_code`: `PLC_ALARM_<slug>` from `ConditionName` when present, else from
-  `SourceName`, else `PLC_ALARM_<hash>` of SourceNode + EventType + Message.
-  The hash tier folds in the Message deliberately - a real Siemens S7-1500
-  multiplexes every `Program_Alarm` of one FB through a single SourceNode
-  (e.g. `i=1845`) with no ConditionName/SourceName, distinguished only by
-  Message text ("pa" vs "pa2"); without the Message in the hash those two
-  alarms would collapse onto one fault_code.
+- `fault_code`: `PLC_ALARM_<slug>_<hash(SourceNode)>` from `ConditionName` when
+  present, else from `SourceName`, else `PLC_ALARM_<hash>` of
+  SourceNode + EventType + Message. The SourceNode is folded into every tier so
+  two distinct conditions sharing a ConditionName/SourceName but sitting on
+  different sources (e.g. two identical FB instances each raising
+  "Overpressure") never collapse onto one code - the fault manager keys/clears
+  by code alone, so a collision would let one condition's clear wipe the
+  other's still-active fault. Message is NOT folded into the slug tiers, so a
+  condition whose Message differs between its active and inactive notifications
+  still maps to one code. The hash tier additionally folds in the Message
+  deliberately - a real Siemens S7-1500 multiplexes every `Program_Alarm` of
+  one FB through a single SourceNode (e.g. `i=1845`) with no
+  ConditionName/SourceName, distinguished only by Message text ("pa" vs "pa2");
+  without the Message in the hash those two alarms would collapse onto one
+  fault_code. A given condition derives its code once (at first observation)
+  and keeps it for its whole lifecycle, so its raise and clear always agree.
 - `entity`: the node-map entity that owns the event's SourceNode when it
   matches a `nodes:` entry, else `auto_alarms.entity_id`.
 - `severity`: the event's raw `Severity` (1-1000) mapped through
--- a/src/ros2_medkit_plugins/ros2_medkit_opcua/README.md
+++ b/src/ros2_medkit_plugins/ros2_medkit_opcua/README.md
@@ -449,6 +449,23 @@ auto_alarms:
 
 Default `off`. Unknown keys are warned and ignored, not silently dropped.
 
+Instead of (or on top of) the node-map YAML, `auto_alarms` can be set purely
+via the plugin's ROS param / JSON config, which is what lets a discovered
+endpoint go straight to native faults with no node-map file at all:
+
+```yaml
+plugins.opcua.endpoint_url: "opc.tcp://192.168.1.10:4840"
+plugins.opcua.auto_alarms: true   # or the same map form as above
+```
+
+The JSON/ROS-param form takes precedence over whatever the node-map YAML's
+`auto_alarms:` block set (same precedence env vars use for the rest of the
+plugin config). Explicit `event_alarms:` mappings still win over
+auto-derivation at the poller regardless of how `auto_alarms` was enabled.
+This completes the zero-config chain: discovery (endpoint) ->
+`auto_browse` (SOVD tree) -> `auto_alarms` (native faults), each param-driven,
+no node-map file required.
+
 **Fault derivation** (no config, per observed event):
 - `fault_code`: `PLC_ALARM_<slug>` from `ConditionName` when present, else from
   `SourceName`, else `PLC_ALARM_<hash>` of SourceNode + EventType + Message.
@@ -499,13 +516,15 @@ ros2_medkit_gateway:
     plugins.opcua.poll_interval_ms: 1000
     plugins.opcua.prefer_subscriptions: false
     plugins.opcua.auto_browse: true
+    plugins.opcua.auto_alarms: true
 ```
 
 | Parameter | Default | Description |
 |-----------|---------|-------------|
 | `endpoint_url` | `opc.tcp://localhost:4840` | OPC-UA server endpoint |
-| `node_map_path` | (none) | Path to node map YAML (optional when `auto_browse` is enabled - see above) |
+| `node_map_path` | (none) | Path to node map YAML (optional when `auto_browse` or `auto_alarms` is enabled - see above) |
 | `auto_browse` | `false` | Boolean or object; recursively discover the SOVD tree from the live address space instead of (or in addition to) `node_map_path` - see above |
+| `auto_alarms` | `false` | Boolean or object; subscribe a server EventNotifier and auto-derive a fault per AlarmCondition with no per-alarm `event_alarms` mapping - the zero-config native A&C surface (see above) |
 | `poll_interval_ms` | `1000` | Polling interval in ms (clamped to [100, 60000]) |
 | `prefer_subscriptions` | `false` | Use OPC-UA subscriptions instead of polling |
 | `subscription_interval_ms` | `500` | Publishing interval for OPC-UA subscriptions when `prefer_subscriptions: true` |
--- a/src/ros2_medkit_plugins/ros2_medkit_opcua/README.md
+++ b/src/ros2_medkit_plugins/ros2_medkit_opcua/README.md
@@ -482,8 +482,11 @@ on `i=2253`) no duplicate subscription is created - unmatched events on an
 alarm as soon as it goes inactive, bypassing the Acknowledge/Confirm workflow
 entirely - there is no `acknowledge_fault` SOVD operation for a zero-config
 alarm to receive an Ack through, so without this it would latch forever on a
-server that requires one (as most do). Set `auto_clear: false` to require
-Acknowledge (via the `event_alarms` mapping path) before it clears.
+server that requires one (as most do). Set `auto_clear: false` to require the
+poller's normal Acknowledge/Confirm gating before it clears - since a
+zero-config alarm has no SOVD `acknowledge_fault` operation of its own,
+Acked/Confirmed must come from the OPC-UA server side instead (an HMI/SCADA
+system or another OPC-UA client acknowledging the condition directly).
 
 ### Gateway Parameters
 
--- a/src/ros2_medkit_plugins/ros2_medkit_opcua/README.md
+++ b/src/ros2_medkit_plugins/ros2_medkit_opcua/README.md
@@ -418,6 +418,73 @@ The JSON/ROS-param form takes precedence over whatever the node-map YAML's
 `auto_browse:` block set, mirroring how environment variables override the
 rest of the plugin's YAML config.
 
+### Zero-config native A&C (`auto_alarms`)
+
+`event_alarms:` above requires a hand-written mapping per alarm. `auto_alarms`
+needs none: enable it and every AlarmConditionType event on the subscribed
+source (default the Server object, `i=2253` - the same system-wide catch-all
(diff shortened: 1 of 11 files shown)


src/ros2_medkit_plugins/ros2_medkit_opcua/include/ros2_medkit_opcua/node_map.hpp
```diff
@@ -118,6 +118,69 @@ struct ResolvedAlarm {
...
+  /// CodeSys controllers surface there without the operator having to find
+  /// the owning Object first.
+  std::string source_node_id_str = "i=2253";
+  opcua::NodeId source_node_id;
+
+  /// SOVD entity that hosts an auto-derived fault when its SourceNode is not
+  /// a known node-map entry. Defaults to the node map's ``component_id`` at
+  /// load time (see ``NodeMap::load``). When the event's SourceNode matches
+  /// an existing ``nodes:`` entry, the fault is hosted on THAT entry's
+  /// entity instead (a more specific home than this fallback).
```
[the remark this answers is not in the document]
mfaferek93: Fixed in c8460f75: comment now says entity_id defaults to "<component_id>_alarms", not the bare component_id, matching NodeMap::load().

src/ros2_medkit_plugins/ros2_medkit_opcua/src/opcua_poller.cpp
```diff
@@ -720,13 +794,79 @@ void OpcuaPoller::on_event(const AlarmEventConfig & cfg, const std::vector<opcua
...
+    }
+    if (!NodeMap::auto_alarm_passes_filters(condition_name, source_name, message, auto_cfg.include_patterns,
+                                            auto_cfg.exclude_patterns)) {
+      return;
+    }
+    eff.fault_code = NodeMap::derive_auto_fault_code(condition_name, source_name, source_node.toString(),
+                                                     event_type.toString(), message);
+    // Host the fault on a known node-map entity when SourceNode resolves to
+    // one; otherwise fall back to auto_alarms.entity_id (default: the PLC
+    // root component).
```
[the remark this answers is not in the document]
mfaferek93: Fixed in c8460f75: same correction in the poller - default is "<component_id>_alarms" (a separate App), not the PLC root Component.

src/ros2_medkit_plugins/ros2_medkit_opcua/include/ros2_medkit_opcua/opcua_poller.hpp
```diff
@@ -297,10 +328,13 @@ class OpcuaPoller {
   /// Apply one condition state observation (from a live event or a read scan)
   /// to the tracked condition map + state machine, dispatching the resulting
   /// fault action. ``event_id`` is the live EventId for ack/confirm (null for
-  /// read-based observations).
+  /// read-based observations). ``require_confirm_for_clear`` overrides
+  /// ``config_.require_confirm_for_clear`` for this one observation -
+  /// auto-derived alarms pass ``!auto_alarms.auto_clear`` complemented
+  /// (see on_event) instead of the poller-wide setting.
```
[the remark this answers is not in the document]
mfaferek93: Fixed in c8460f75: rewrote the apply_condition_state() comment to describe what the caller actually does - forces require_confirm_for_clear=false (plus acked_state=true) when auto_clear is true, otherwise passes config_.require_confirm_for_clear through unchanged. It was never a value derived from !auto_clear.

src/ros2_medkit_plugins/ros2_medkit_opcua/README.md
```diff
@@ -346,6 +346,73 @@ curl -X POST http://localhost:8080/api/v1/apps/tank_process/operations/acknowled
...
+on `i=2253`) no duplicate subscription is created - unmatched events on an
+`event_alarms` source simply fall through to auto-derivation when
+`auto_alarms` is enabled for that source.
+
+**Acknowledge/Confirm:** `auto_clear` (default `true`) clears an auto-derived
+alarm as soon as it goes inactive, bypassing the Acknowledge/Confirm workflow
+entirely - there is no `acknowledge_fault` SOVD operation for a zero-config
+alarm to receive an Ack through, so without this it would latch forever on a
+server that requires one (as most do). Set `auto_clear: false` to require
+Acknowledge (via the `event_alarms` mapping path) before it clears.
```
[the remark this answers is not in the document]
mfaferek93: Fixed in c8460f75: reworded - auto_clear: false does not go through an event_alarms-provided acknowledge_fault operation (auto-derived alarms never get one); clearing is gated by the OPC-UA server's own Acked/Confirmed condition state, set by an HMI/SCADA system or another OPC-UA client.

src/ros2_medkit_plugins/ros2_medkit_opcua/src/opcua_poller.cpp
```diff
@@ -720,13 +794,80 @@ void OpcuaPoller::on_event(const AlarmEventConfig & cfg, const std::vector<opcua
...
+                          "on_event: dropping non-condition event (null ConditionId, likely a system "
+                          "message) message='"
+                              << message << "'");
+      return;
+    }
+    if (!NodeMap::auto_alarm_passes_filters(condition_name, source_name, message, auto_cfg.include_patterns,
+                                            auto_cfg.exclude_patterns)) {
+      return;
+    }
+    eff.fault_code = NodeMap::derive_auto_fault_code(condition_name, source_name, source_node.toString(),
```
bburda: The auto fault_code is re-derived from the current event here (and again in the read-replay path at line 585, with different inputs), and `apply_condition_state` delivers this per-event code (`delivery.fault_code = cfg.fault_code`) rather than the code stored for the ConditionId at raise time (`it->second.fault_code`). So the raise and the clear of one condition can carry different codes, and since fault_manager keys and clears by code alone, the original fault never clears and latches.

Two ways the code diverges within one condition's lifecycle:
- Tier 3 folds `message` into the hash (node_map.cpp line 1063). OPC-UA lets the Message differ between the active and inactive notifications, so the clear hashes to a different code. This is exactly the Siemens Program_Alarm (tier-3) case the feature targets.
- The read-replay path (line 585) derives with `source_name=""`, `event_type=""`, and the configured source id, while the live path (here) uses the real SourceName/EventType/SourceNode. A condition replayed on connect and then cleared live gets two different codes, with no message change needed.

The reconcile clears already use the stored code (`runtime.fault_code`, lines 657/681); only the live delivery re-derives.

Fix: derive the code (and entity_id) once at first observation of the ConditionId, store it, and deliver `it->second.fault_code`/`entity_id` for every event. That makes the live, replay, and reconcile paths agree and removes the message-drift hazard.

src/ros2_medkit_plugins/ros2_medkit_opcua/src/node_map.cpp
```diff
@@ -892,6 +994,110 @@ ResolvedAlarm NodeMap::resolve_alarm(const AlarmEventConfig & cfg, const std::st
...
+
+}  // namespace
+
+std::string NodeMap::derive_auto_fault_code(const std::string & condition_name, const std::string & source_name,
+                                            const std::string & source_node_str, const std::string & event_type_str,
+                                            const std::string & message) {
+  if (!condition_name.empty()) {
+    const std::string slug = slugify(condition_name);
+    if (!slug.empty()) {
+      return "PLC_ALARM_" + slug;
```
bburda: Tiers 1 and 2 derive the fault_code from ConditionName (or SourceName) alone and drop the SourceNode. Two distinct conditions (different ConditionId, different SourceNode) that share a ConditionName - e.g. two identical FB instances each raising "Overpressure" - derive the same `PLC_ALARM_OVERPRESSURE`.

fault_manager keys and clears by code alone (the file's own uniqueness comment at lines 838-854 relies on exactly this), and the load-time uniqueness check cannot see runtime-derived codes. So the two conditions collide: one's clear wipes the other's still-active fault, or they flap. This is likely under the default source `i=2253`, the whole-server notifier, which sees every source's alarms.

Fix: fold the SourceNode into tiers 1 and 2 as well (e.g. `PLC_ALARM_<slug>_<short_hash(source_node)>`), so same-named conditions on different sources stay distinct.

src/ros2_medkit_plugins/ros2_medkit_opcua/src/opcua_poller.cpp
```diff
@@ -286,6 +293,38 @@ void OpcuaPoller::setup_subscriptions() {
...
 }
 
+std::vector<AlarmEventConfig>
+OpcuaPoller::effective_alarm_sources(const std::vector<AlarmEventConfig> & explicit_sources,
+                                     const AutoAlarmsConfig & auto_cfg) {
+  std::vector<AlarmEventConfig> sources = explicit_sources;
+  if (!auto_cfg.enabled) {
+    return sources;
+  }
+  const bool already_covered = std::any_of(sources.begin(), sources.end(), [&](const AlarmEventConfig & c) {
```
bburda, line 322: `already_covered` (and the auto-branch guard at line 821) compares the raw configured `source_node_id_str`. This misses two overlap cases and produces duplicate faults for one alarm:

- Equivalent NodeId spellings: an explicit `event_alarms` source written as `ns=0;i=2253` does not string-match the auto default `i=2253`, so a synthetic monitored item is added on the same physical node - two monitored items, so every event fires both (one mapped fault + one auto fault).
- Notifier hierarchy: the default auto source `i=2253` is the Server object, the root of the notifier hierarchy, so it also receives events whose SourceNode is a specific sub-object that has its own explicit `event_alarms` subscription. In that hybrid config every explicitly-mapped alarm double-fires (explicit fault + auto fault), which defeats the documented "explicit event_alarms take precedence".

Fix: compare parsed/canonical NodeIds instead of raw strings, and in the auto branch drop any event whose actual `source_node` matches an explicit `event_alarms` source (that event is already handled by the explicit monitored item).

src/ros2_medkit_plugins/ros2_medkit_opcua/src/opcua_poller.cpp
mfaferek93: Fixed in 89ad593b: apply_condition_state now pins the fault_code/entity_id derived at the FIRST observation of each ConditionId and delivers it->second.fault_code/entity_id for every event, so the live, replay and reconcile paths carry one code for a condition's whole lifecycle (reconcile already used the stored code). Both divergences you flagged - tier-3 Message drift between the active/inactive notifications, and read-replay (source_name/event_type empty) vs live - can no longer produce a raise and clear with different codes, so the fault stops latching.

src/ros2_medkit_plugins/ros2_medkit_opcua/src/node_map.cpp
mfaferek93: Fixed in 89ad593b: derive_auto_fault_code now folds a canonical SourceNode hash into tiers 1 and 2 as well (PLC_ALARM_<slug>_<hash(SourceNode)>, mirroring the hash tier), so two distinct conditions that share a ConditionName/SourceName on different sources - e.g. two identical FB instances each raising "Overpressure" under i=2253 - no longer collapse onto one code and cross-clear. The SourceNode is canonicalized first (i=2253 == ns=0;i=2253), and Message is deliberately kept OUT of the slug tiers so a condition whose Message differs between its active and inactive notifications still maps to one code.

src/ros2_medkit_plugins/ros2_medkit_opcua/src/opcua_poller.cpp
mfaferek93: Fixed in 89ad593b: all source-id comparisons now go through a new canonical OpcuaPoller::node_ids_equivalent (parse to NodeId, compare, fall back to raw on unparseable), so already_covered and the auto-branch guard treat i=2253 and ns=0;i=2253 as one node - no second monitored item on the same physical source. For the notifier hierarchy, the auto branch now drops any event whose real SourceNode matches an explicit event_alarms source OTHER than the monitored item's own, so an explicitly-mapped alarm delivered on the root notifier is no longer auto-derived a second time (the shared-source fall-through still auto-derives its own unmatched events). read_fallback_replay's guard uses the same canonical compare.

bburda (dismissed):

bburda (approved):

PR #519, opened 2026-07-10
fix(opcua): thread-safety fix for the REST/poll fault-buffer race

mfaferek93:
## Problem
Under UI load the OPC-UA gateway crashed with heap-corruption signatures (GPF /
SIGILL / bad-pointer segfault) inside `libros2_medkit_opcua_plugin.so`. Root
cause is a data race on `pending_reports_`, the fire-and-forget fault
report/clear buffer: it is mutated by the poll thread (`publish_values`,
`on_alarm_change`/`on_event_alarm`) AND by the REST thread
(`FaultProvider::clear_fault` -> `send_clear_fault` -> `send_or_buffer`) with no
lock. A `push_back` that reallocates the vector while the other thread iterates
it is a use-after-free that corrupts the heap.

## Fix
Guard every access with a mutex. `flush_pending_reports` swaps the ready batch
out under the lock and dispatches it outside, so the vector is never reallocated
mid-iteration and the ROS service call never runs under the lock.

## Test
`OpcuaPluginConcurrency.ClearFaultBufferIsThreadSafe`: six threads drive the
REST `clear_fault` path against the running poll thread. It segfaults on the
unsynchronized code (verified) and passes with the fix. Full unit suite green.

### DIFF
--- a/src/ros2_medkit_plugins/ros2_medkit_opcua/test/test_opcua_plugin.cpp
+++ b/src/ros2_medkit_plugins/ros2_medkit_opcua/test/test_opcua_plugin.cpp
@@ -729,6 +729,25 @@ class RealNodePluginContext : public FakePluginContext {
   rclcpp::Node * node_;
 };
 
+// RAII: pair rclcpp init with a shutdown so global ROS state never leaks into
+// the rest of the process, even if an assertion returns from the test early.
+// Only tears down the init this guard performed.
+struct ScopedRclcpp {
+  const bool owned_;
+  ScopedRclcpp() : owned_(!rclcpp::ok()) {
+    if (owned_) {
+      rclcpp::init(0, nullptr);
+    }
+  }
+  ~ScopedRclcpp() {
+    if (owned_ && rclcpp::ok()) {
+      rclcpp::shutdown();
+    }
+  }
+  ScopedRclcpp(const ScopedRclcpp &) = delete;
+  ScopedRclcpp & operator=(const ScopedRclcpp &) = delete;
+};
+
 // The SOVD DELETE /faults/{code} route lands on FaultProvider::clear_fault(),
 // which buffers a dispatch into pending_reports_ - the SAME vector the poll
 // thread drains in publish_values()/flush_pending_reports(). Before the fix the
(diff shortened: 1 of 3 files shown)


src/ros2_medkit_plugins/ros2_medkit_opcua/test/test_opcua_plugin.cpp
```diff
@@ -565,4 +568,87 @@ TEST(MapSeverity, LiveSeverityBandBoundaries) {
...
+  }
+
+  std::this_thread::sleep_for(std::chrono::seconds(2));
+  stop.store(true, std::memory_order_relaxed);
+  for (auto & th : threads) {
+    th.join();
+  }
+
+  plugin.shutdown();
+  SUCCEED();
```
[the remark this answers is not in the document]
mfaferek93: Fixed in 71a37dd0: replaced the bare `rclcpp::init()` with a `ScopedRclcpp` RAII guard so `rclcpp::shutdown()` always runs at test-scope exit (even when an assertion returns early); it only tears down the init it performed. That was the only init-without-shutdown site in the file. Verified: `test_opcua_plugin` builds and its 39 tests pass (jazzy), clang-format-18 clean.

src/ros2_medkit_plugins/ros2_medkit_opcua/test/test_opcua_plugin.cpp
```diff
@@ -565,4 +568,104 @@ TEST(MapSeverity, LiveSeverityBandBoundaries) {
...
+    data_type: float
+)";
+  }
+
+  OpcuaPlugin plugin;
+  nlohmann::json config;
+  config["node_map_path"] = yaml_path;
+  // Nothing listening: connect fails fast (ECONNREFUSED) and the poll thread
+  // keeps retrying in the background while it drains the buffer each cycle.
+  config["endpoint_url"] = "opc.tcp://127.0.0.1:1";
```
bburda, line 842: This endpoint is dead and no fault_manager runs, so `service_is_ready()` stays false the whole test. That means `flush_pending_reports` returns at the top check and never reaches `batch.swap(pending_reports_)`. So the test covers the `send_or_buffer` push_back/erase race, but not the other new locked path: one thread draining the buffer (the swap) while another thread pushes into it. That drain-vs-push case is the reallocation-during-iteration the fix describes. Consider a stub client whose `service_is_ready()` returns true, so the swap runs against concurrent pushes.

mfaferek93: Good catch. The dead endpoint kept service_is_ready() false, so flush_pending_reports() early-returned and batch.swap() was never raced - only push_back/erase was covered. Fixed in 4549b2a: the test now stands up a real ReportFault/ClearFault sink so the gate opens, every clear_fault() pushes AND swaps concurrently, and it asserts the server actually received dispatches so the drain path can't silently regress. Verified in a ros:jazzy container (all opcua gtests pass, this test now runs the full 2s hammer).

bburda (approved):

PR #522, opened 2026-07-11
fix(gateway): preserve plugin App external flag through hybrid merge

### LINKED ISSUE #517: [BUG] Hybrid discovery merge drops plugin App external flag - external-app faults vanish from Function rollup
# Bug report

### Steps to reproduce
1. Run the gateway with `discovery.mode: hybrid` and a protocol plugin that introspects an external App with an empty `effective_fqn` (e.g. the OPC-UA plugin's process App).
2. Declare a manifest Function whose `hosted_by` lists that App id.
3. Report a fault to the fault_manager with `source_id` = the App id.
4. `GET /functions/<id>/faults`.

### Expected behavior

The fault rolls up into the Function (the external+empty-fqn rule in `collect_app_fqn()` grants the App its own id as fault-scope owner). This is exactly what happens under `discovery.mode: manifest_only`.

### Actual behavior

Empty `items` - under hybrid the merged entity cache holds the App with `external=false`, so the external rule never fires and the host contributes no fqn. The fault silently vanishes from the rollup.

Suspected mechanism: `src/ros2_medkit_gateway/src/discovery/merge_pipeline.cpp`, METADATA group - `external` merges with scalar semantics, so when the target layer wins the plugin's `external=true` is discarded (target default `false`). The comment says "external is a classification, not a status flag", but the hybrid merge does not preserve the classification.

### Environment
- ros2_medkit version: main (0d637f6f)
- ROS 2 distro: Jazzy
- OS: Ubuntu 24.04

### Additional information

Test-coverage gap that would have caught this: fa05ad86 ("surface external plugin-entity faults") added external-app fault-scope regressions for the APP and COMPONENT routes but not for FUNCTION-hosts-external-app. Worth adding that regression alongside the fix.

bburda:
# Pull Request

## Summary

Under `discovery.mode: hybrid`, an external App (a non-ROS asset such as a PLC bridged into SOVD by a protocol plugin: `external: true`, no ROS node) could lose its `external` classification and drop out of every fault rollup (#517). The root cause was that `App.external` was a plain `bool`: the manifest parser collapsed a missing `external:` key and an explicit `external: false` into the same `false`, so a manifest stub that merely declared the App id (to let a Function reference it) erased the classification a protocol plugin had introspected. External apps report faults under their own entity id, so without the flag their bare-id fault scope was lost and their faults vanished from Function, Component and Area rollups. `manifest_only` was unaffected because it bypasses the merge pipeline.

This makes `external` a tri-state `std::optional<bool>`:

- `nullopt` - no layer classified the app
- `true` - external (not a ROS node)
- `false` - explicitly a ROS node

A dedicated `merge_external` gap-fills the classification (mirroring the scalar merge): a layer that omits `external:` never shadows a classification another layer contributed, so a manifest stub keeps a plugin's `external: true`. An explicit value from a higher-priority layer still wins, so an authoritative manifest `external: false` corrects a wrong plugin `true` - something a plain bool could not express. This is deliberately not the generic optional merge: the manifest layer is the AUTHORITATIVE base for the metadata group, and that helper's target branch is a no-op, so it would drop the plugin's value.

Two follow-on correctness issues are fixed in the same change:

- **Fault scope, neighbor case.** An external app may still carry a manifest `ros_binding`, whose derived FQN would otherwise shadow the bare id and drop the app's faults (the same vanishing-faults symptom as #517). Fault-scope resolution now checks `external` before `effective_fqn()`, so an external app is always scoped by its entity id.
- **Validation R013.** An external app's binding is inert (ignored for linking and fault scoping), so it no longer enters the duplicate-binding bookkeeping - a real ROS app that later reuses the same node name no longer trips a phantom R010 conflict.

Docs (discovery options and manifest schema) describe the tri-state merge.

---

## Issue

- closes #517

---

## Type

- [x] Bug fix

---

## Testing

- Merge: reworked unit tests pin the tri-state branches - a manifest stub that omits `external:` keeps a plugin's `external: true`, and an authoritative manifest `external: false` overrides a plugin `true`.
- Fault scope: new unit tests cover an external app owning its faults by bare id on the App, Component, Function and Area routes, including the neighbor case (external app plus a stray `ros_binding`).
- Validation: a new test pins that an external app's inert binding does not cause a phantom R010 on a real app reusing the node name.
- Parser: new tests cover omitted -> unset and explicit `false` -> preserved.
- End-to-end: a new integration test reports a fault under an external app's bare id and asserts it surfaces on the app route and every rollup route (component, function, area). It fails against the pre-fix fault scope and passes only with the fix.
- Full gateway unit suite, linters and the affected integration test pass locally on Jazzy.

---

## Checklist

### DIFF
--- a/docs/config/discovery-options.rst
+++ b/docs/config/discovery-options.rst
@@ -177,12 +177,16 @@ Each layer declares a policy per field-group:
 
 .. note::
 
-   The app ``external`` flag is an exception to the policy table: it merges
-   monotonically. Once any layer classifies an app as external (e.g. a
-   protocol plugin introspecting a PLC), the classification is kept even if a
-   higher-priority layer carries the default ``false`` - a manifest entry that
-   omits ``external:`` cannot clear it. This keeps external apps owning their
-   faults (bare-id fault scope) in hybrid mode.
+   The app ``external`` classification is tri-state (unset / ``true`` /
+   ``false``), so it follows the policy table with one refinement: a layer that
+   does not classify the app (a manifest entry that omits ``external:``) never
+   clears or shadows a classification another layer contributed. A manifest stub
+   that omits ``external:`` therefore keeps a protocol plugin's introspected
+   ``external: true`` (so the app still owns its faults via bare-id fault scope
+   in hybrid mode). An *explicit* value is authoritative and resolves by normal
+   layer priority: an authoritative manifest ``external: false`` overrides a
+   plugin's ``external: true``. Only omission is a no-op; an explicit value is
+   never silently discarded.
 
 Override per-layer policies in ``gateway_params.yaml``. Empty string means
 "use layer default". Policy values are **case-sensitive** and must be lowercase
--- a/docs/config/discovery-options.rst
+++ b/docs/config/discovery-options.rst
@@ -175,6 +175,15 @@ Each layer declares a policy per field-group:
 - Manifest: ``authoritative`` for identity/hierarchy/metadata, ``enrichment`` for live_data, ``fallback`` for status
 - Runtime: ``authoritative`` for live_data/status, ``enrichment`` for metadata, ``fallback`` for identity/hierarchy
 
+.. note::
+
+   The app ``external`` flag is an exception to the policy table: it merges
+   monotonically. Once any layer classifies an app as external (e.g. a
+   protocol plugin introspecting a PLC), the classification is kept even if a
+   higher-priority layer carries the default ``false`` - a manifest entry that
+   omits ``external:`` cannot clear it. This keeps external apps owning their
+   faults (bare-id fault scope) in hybrid mode.
+
 Override per-layer policies in ``gateway_params.yaml``. Empty string means
 "use layer default". Policy values are **case-sensitive** and must be lowercase
 (``authoritative``, ``enrichment``, ``fallback``):
--- a/docs/config/manifest-schema.rst
+++ b/docs/config/manifest-schema.rst
@@ -553,6 +553,8 @@ Fields
      - True if not a ROS node (default: false). In hybrid mode the flag merges
        monotonically: omitting it does not clear an ``external`` classification
        contributed by another discovery layer (e.g. a protocol plugin).
+       Combining ``external: true`` with a ``ros_binding`` is contradictory and
+       raises validation warning R013 (the binding is ignored for linking).
 
 ros_binding Fields
 ~~~~~~~~~~~~~~~~~~
--- a/docs/config/manifest-schema.rst
+++ b/docs/config/manifest-schema.rst
@@ -550,11 +550,16 @@ Fields
    * - ``external``
      - boolean
      - No
-     - True if not a ROS node (default: false). In hybrid mode the flag merges
-       monotonically: omitting it does not clear an ``external`` classification
-       contributed by another discovery layer (e.g. a protocol plugin).
+     - True if not a ROS node (treated as not external when omitted). The
+       classification is tri-state internally: **omitting** ``external:`` leaves
+       it unset, so in hybrid mode it does not clear an ``external``
+       classification contributed by another discovery layer (e.g. a protocol
+       plugin) - the app keeps its bare-id fault scope. An **explicit** value is
+       authoritative and resolves by normal layer priority, so an authoritative
+       manifest ``external: false`` overrides a plugin's ``external: true``.
        Combining ``external: true`` with a ``ros_binding`` is contradictory and
-       raises validation warning R013 (the binding is ignored for linking).
+       raises validation warning R013 (the binding is ignored for linking and
+       fault scoping - the app is scoped by its entity id).
 
 ros_binding Fields
 ~~~~~~~~~~~~~~~~~~
--- a/docs/config/manifest-schema.rst
+++ b/docs/config/manifest-schema.rst
@@ -550,7 +550,9 @@ Fields
    * - ``external``
      - boolean
      - No
-     - True if not a ROS node (default: false)
+     - True if not a ROS node (default: false). In hybrid mode the flag merges
+       monotonically: omitting it does not clear an ``external`` classification
+       contributed by another discovery layer (e.g. a protocol plugin).
 
 ros_binding Fields
 ~~~~~~~~~~~~~~~~~~
--- a/src/ros2_medkit_gateway/src/core/discovery/manifest/manifest_validator.cpp
+++ b/src/ros2_medkit_gateway/src/core/discovery/manifest/manifest_validator.cpp
@@ -187,14 +187,18 @@ void ManifestValidator::validate_ros_bindings(const Manifest & manifest, Validat
   for (const auto & app : manifest.apps) {
     if (app.ros_binding.has_value() && !app.ros_binding->is_empty()) {
       // external means "not a ROS node": the runtime linker skips the binding
-      // and the hybrid merge keeps the external classification. Declaring both
-      // is contradictory - warn instead of silently ignoring one of the two.
-      if (app.external) {
+      // and fault scoping uses the app's entity id, so the binding is inert.
+      // Declaring both is contradictory - warn, and skip the R010 bookkeeping
+      // entirely: an inert binding must not reserve a node name and trip a
+      // phantom duplicate error on a real ROS app that later reuses it.
+      if (app.external.value_or(false)) {
         result.add_warning("R013",
                            "App '" + app.id +
-                               "' declares external: true together with a ros_binding; "
-                               "the binding is ignored for linking",
+                               "' declares external: true together with a ros_binding; the binding "
+                               "is ignored for linking and fault scoping (the app is scoped by its "
+                               "entity id)",
                            "apps/" + app.id + "/ros_binding");
+        continue;
       }
 
       std::string binding_key = app.ros_binding->node_name + "@" + app.ros_binding->namespace_pattern;
--- a/src/ros2_medkit_gateway/src/core/discovery/manifest/manifest_validator.cpp
+++ b/src/ros2_medkit_gateway/src/core/discovery/manifest/manifest_validator.cpp
@@ -186,6 +186,17 @@ void ManifestValidator::validate_ros_bindings(const Manifest & manifest, Validat
 
   for (const auto & app : manifest.apps) {
     if (app.ros_binding.has_value() && !app.ros_binding->is_empty()) {
+      // external means "not a ROS node": the runtime linker skips the binding
+      // and the hybrid merge keeps the external classification. Declaring both
+      // is contradictory - warn instead of silently ignoring one of the two.
+      if (app.external) {
+        result.add_warning("R013",
+                           "App '" + app.id +
+                               "' declares external: true together with a ros_binding; "
+                               "the binding is ignored for linking",
+                           "apps/" + app.id + "/ros_binding");
+      }
+
       std::string binding_key = app.ros_binding->node_name + "@" + app.ros_binding->namespace_pattern;
 
       // For topic namespace bindings
--- a/src/ros2_medkit_gateway/src/discovery/merge_pipeline.cpp
+++ b/src/ros2_medkit_gateway/src/discovery/merge_pipeline.cpp
@@ -169,6 +169,30 @@ void merge_optional(std::optional<T> & target, const std::optional<T> & source, 
   }
 }
 
+// Merge the tri-state `external` classification. Mirrors merge_scalar's
+// gap-filling policy (nullopt == "empty"): a higher-priority layer that does not
+// classify (nullopt) must never shadow a lower-priority layer's classification,
+// so once any layer sets it the value survives (a manifest stub cannot hide a
+// plugin's introspected external=true, #517). An explicit value from a strictly
+// higher-priority layer (SOURCE) still wins, so an authoritative manifest
+// `external: false` corrects a wrong plugin `true`. This is NOT merge_optional:
+// merge_optional's TARGET branch is a no-op and would drop the plugin's value.
+void merge_external(std::optional<bool> & target, const std::optional<bool> & source, MergeWinner winner) {
+  switch (winner) {
+    case MergeWinner::SOURCE:
+      if (source.has_value()) {
+        target = source;
+      }
+      break;
+    case MergeWinner::BOTH:
+    case MergeWinner::TARGET:
+      if (!target.has_value() && source.has_value()) {
+        target = source;
+      }
+      break;
+  }
+}
+
 void merge_topics(ComponentTopics & target, const ComponentTopics & source, MergeWinner winner) {
   merge_collection(target.publishes, source.publishes, winner);
   merge_collection(target.subscribes, source.subscribes, winner);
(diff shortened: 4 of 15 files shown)


src/ros2_medkit_gateway/src/discovery/merge_pipeline.cpp
```diff
@@ -278,11 +278,12 @@ void apply_field_group_merge(Entity & target, const Entity & source, FieldGroup
...
-        if (res.scalar == MergeWinner::SOURCE) {
-          target.external = source.external;
-        }
-        // TARGET and BOTH: keep target value (no OR semantics)
+        // `external` is a classification a layer either knows (true) or cannot
+        // express (bool default false == unset). No layer can un-classify, so
+        // once any layer marks the app external it stays - a manifest stub's
+        // default false must not erase a plugin's introspected classification,
+        // or the app silently drops out of every fault rollup (#517).
+        target.external = target.external || source.external;
```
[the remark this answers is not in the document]
bburda: This state was already representable before this PR: the manifest allows an app with both `external: true` and a `ros_binding`, and validation accepted that silently. The merge change adds one more path to the same state but does not change how it behaves: the linker skips ROS binding for external apps, and fault scoping uses the binding-derived FQN when one exists, same as before.

Clearing `ros_binding` when `external` becomes true would let an enrichment-level plugin layer erase authoritative manifest data, and it would silently move the app's fault scope from FQN to bare id when a plugin misclassifies a bound ROS app. Instead, d7413b26 adds validation warning R013: a manifest app that declares both `external: true` and a `ros_binding` is now flagged, so the contradiction is visible instead of one side being ignored.

mfaferek93: Fixes the bare-stub case but not the neighbor one: a manifest app with a real ros_binding (no `external:` key) plus a plugin app of the same id with external=true. The OR marks it external but merge keeps the manifest binding, so effective_fqn() (app.hpp:102) still derives a live FQN and the bare-id fault-scope branch (fault_scope.cpp:40) never runs - same vanishing-faults symptom as #517. Needs handling here or a follow-up.

src/ros2_medkit_gateway/src/discovery/merge_pipeline.cpp
```diff
@@ -278,11 +278,12 @@ void apply_field_group_merge(Entity & target, const Entity & source, FieldGroup
       case FieldGroup::METADATA:
         merge_scalar(target.source, source.source, res.scalar);
         merge_optional(target.ros_binding, source.ros_binding, res.scalar);
-        // Use scalar semantics (not OR) - external is a classification, not a status flag
-        if (res.scalar == MergeWinner::SOURCE) {
-          target.external = source.external;
-        }
-        // TARGET and BOTH: keep target value (no OR semantics)
+        // `external` is a classification a layer either knows (true) or cannot
+        // express (bool default false == unset). No layer can un-classify, so
```
mfaferek93: The durable fix is `optional<bool>`. This is the second flip of the same coin - 4e01daf removed the OR ("classification, not a status flag"), this restores it ("no layer can un-classify") - both right about different halves of a distinction a plain bool can't encode (manifest_parser:381 collapses explicit `false` and omission). Side effect of the OR: an authoritative manifest `external: false` can never correct a wrong plugin classification. Worth a follow-up issue.

src/ros2_medkit_gateway/src/core/discovery/manifest/manifest_validator.cpp
```diff
@@ -186,6 +186,17 @@ void ManifestValidator::validate_ros_bindings(const Manifest & manifest, Validat
...
   for (const auto & app : manifest.apps) {
     if (app.ros_binding.has_value() && !app.ros_binding->is_empty()) {
+      // external means "not a ROS node": the runtime linker skips the binding
+      // and the hybrid merge keeps the external classification. Declaring both
+      // is contradictory - warn instead of silently ignoring one of the two.
+      if (app.external) {
+        result.add_warning("R013",
+                           "App '" + app.id +
+                               "' declares external: true together with a ros_binding; "
+                               "the binding is ignored for linking",
```
mfaferek93: "ignored for linking" is true but understates it - effective_fqn() (app.hpp:102) still derives an FQN from the binding, so the app's fault scope points at a dead ROS FQN instead of its bare id (the exact bug this PR fixes). Suggest "...ignored for linking and the app loses bare-id fault ownership". Same wording in manifest-schema.rst:557.

src/ros2_medkit_gateway/src/core/discovery/manifest/manifest_validator.cpp
```diff
@@ -186,6 +186,17 @@ void ManifestValidator::validate_ros_bindings(const Manifest & manifest, Validat
...
+                           "apps/" + app.id + "/ros_binding");
+      }
```
mfaferek93, line 202: After the R013 warning it still inserts this external app's binding_key into seen_bindings (line 212), so a real ROS app reusing that node name later fails R010 (hard error) over a phantom conflict. A `continue` here (or skipping the bookkeeping for external apps) fixes it.

docs/config/manifest-schema.rst
```diff
@@ -550,7 +550,11 @@ Fields
    * - ``external``
      - boolean
      - No
-     - True if not a ROS node (default: false)
+     - True if not a ROS node (default: false). In hybrid mode the flag merges
+       monotonically: omitting it does not clear an ``external`` classification
```
mfaferek93: "omitting it does not clear" reads like an explicit `external: false` would win - it doesn't (parser collapses it, the OR discards it, no log). Suggest: "neither omitting `external:` nor setting it to `false` clears a classification contributed by another layer".

docs/config/discovery-options.rst
```diff
@@ -175,6 +175,15 @@ Each layer declares a policy per field-group:
 - Manifest: ``authoritative`` for identity/hierarchy/metadata, ``enrichment`` for live_data, ``fallback`` for status
 - Runtime: ``authoritative`` for live_data/status, ``enrichment`` for metadata, ``fallback`` for identity/hierarchy
 
+.. note::
+
+   The app ``external`` flag is an exception to the policy table: it merges
+   monotonically. Once any layer classifies an app as external (e.g. a
+   protocol plugin introspecting a PLC), the classification is kept even if a
+   higher-priority layer carries the default ``false`` - a manifest entry that
+   omits ``external:`` cannot clear it. This keeps external apps owning their
```
mfaferek93: Same as the manifest-schema note - with OR semantics no layer clears the flag, including an explicit `external: false`. Reword to "no layer can clear it - neither omitting `external:` nor setting it to `false` has any effect once another layer classified the app external".

src/ros2_medkit_gateway/test/test_handler_context.cpp
```diff
@@ -1699,6 +1699,25 @@ TEST(ResolveEntitySourceFqnsTest, ComponentHostingExternalAppOwnsItsFaults) {
   EXPECT_EQ(fqns, std::set<std::string>{"process"});
 }
 
+TEST(ResolveEntitySourceFqnsTest, FunctionHostingExternalAppOwnsItsFaults) {
```
mfaferek93, line 1702: AREA is the only rollup without an external-app test now (App/Component/Function have one). The AREA route walks the same gate (fault_scope.cpp:121 -> collect_area_app_fqns -> collect_app_fqn) and is production-real - the opcua plugin sets comp.area, so GET /areas/<area>/faults depends on the same branch. An AreaHostingExternalAppOwnsItsFaults sibling closes the set.

src/ros2_medkit_gateway/test/test_merge_pipeline.cpp
```diff
@@ -993,6 +993,47 @@ TEST_F(MergePipelineTest, ThreeLayerMerge_PerFieldGroupOwnerTracking) {
...
+// A protocol plugin introspects an external asset (e.g. a PLC over OPC UA) and
+// emits its App with external=true and no ROS binding. In hybrid mode the
+// manifest may declare the same App id as a bare stub (so a Function's
+// hosted_by can reference it) without an `external:` key, which parses to the
+// bool default false. `external` has no unset state, so the manifest's
+// AUTHORITATIVE METADATA must not let that default erase the plugin's
+// classification - collect_app_fqn() grants an empty-fqn App its bare-id fault
+// scope only when external is true, so dropping the flag silently empties the
+// App's fault rollup on every route that aggregates it.
+TEST_F(MergePipelineTest, PluginExternalClassificationSurvivesManifestMetadataMerge) {
```
mfaferek93, line 1008: Worth an end-to-end pin for #517: `external` appears nowhere in the integration tests, and the full chain (hybrid stub + plugin app -> merge -> fault under bare id -> visible in the rollup) only exists as two disjoint unit layers. This exact interaction regressed once via a semantics-only refactor (4e01daf) that every unit test of its era survived - an integration test is the layer that catches it.

mfaferek93 (commented):
A few minor comments

src/ros2_medkit_gateway/src/discovery/merge_pipeline.cpp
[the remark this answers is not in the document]
mfaferek93: Agreed with keeping the binding rather than clearing it. The contradiction is handled downstream instead: 7f52b568 scopes an external app by its bare id (not the derived FQN) and R013 skips the seen_bindings bookkeeping, so no vanishing faults and no phantom R010.

mfaferek93: Resolved in 7f52b568: collect_app_fqn now checks external before effective_fqn, so an external app with a stray ros_binding is scoped by its bare id instead of the dead FQN. Pinned by ExternalAppWithStrayRosBindingOwnsFaultsByBareId plus the end-to-end rollup test.

src/ros2_medkit_gateway/src/discovery/merge_pipeline.cpp
mfaferek93: Done in 7f52b568 - external is now std::optional<bool> and merge_external gap-fills: omission never shadows another layer, but an explicit higher-priority value wins, so an authoritative manifest external: false corrects a wrong plugin true. Closed the #523 follow-up as implemented.

src/ros2_medkit_gateway/src/core/discovery/manifest/manifest_validator.cpp
mfaferek93: 7f52b568 fixed the behavior rather than documenting the loss: external apps now keep bare-id fault ownership even with a binding, so R013 reads '...ignored for linking and fault scoping (the app is scoped by its entity id)'.

src/ros2_medkit_gateway/src/core/discovery/manifest/manifest_validator.cpp
mfaferek93: Good catch - fixed in 7f52b568: external apps continue before the seen_bindings bookkeeping, so their inert binding can't trip a phantom R010 on a real ROS app reusing that node name. Pinned by R013_ExternalAppBindingDoesNotReserveNodeName.

docs/config/manifest-schema.rst
mfaferek93: Reworded in 7f52b568: now spells out that omitting external: leaves it unset (does not clear a classification from another layer) while an explicit false is authoritative and overrides a plugin's true.

docs/config/discovery-options.rst
mfaferek93: Reworded in 7f52b568: only omission is a no-op now - an explicit value, including false, is authoritative and never silently discarded (tri-state, not monotone).

src/ros2_medkit_gateway/test/test_handler_context.cpp
mfaferek93: Added in 7f52b568: AreaHostingExternalAppOwnsItsFaults completes the App/Component/Function/Area set through the same collect_app_fqn gate.

src/ros2_medkit_gateway/test/test_merge_pipeline.cpp
mfaferek93: Added in 7f52b568: ros2_medkit_integration_tests/test/features/test_external_app_fault_rollup.test.py drives the full HTTP stack - reports a fault under the external app's bare id and asserts it on the app/component/function/area routes, with the stray-binding neighbor case baked into the fixture.

src/ros2_medkit_gateway/src/discovery/merge_pipeline.cpp
bburda: Handled here rather than deferred. `collect_app_fqn` in `fault_scope.cpp` now checks `external` before `effective_fqn()`, so an external app carrying a stray `ros_binding` is scoped by its bare entity id, not the derived FQN. Pinned by a unit test (`ExternalAppWithStrayRosBindingOwnsFaultsByBareId`) and an end-to-end integration test that fails against the pre-fix scope.

src/ros2_medkit_gateway/src/discovery/merge_pipeline.cpp
bburda: Done in this PR. `App.external` is now `std::optional<bool>`; the parser keeps omitted-vs-explicit-`false`, and an authoritative manifest `external: false` overrides a wrong plugin `true`. It deliberately does not reuse `merge_optional`: the manifest layer is the AUTHORITATIVE base for METADATA, so the winner is TARGET, and that helper's TARGET branch is a no-op and would drop the plugin's value. A dedicated `merge_external` gap-fills instead. This also resolves the tri-state follow-up (523).

src/ros2_medkit_gateway/src/core/discovery/manifest/manifest_validator.cpp
bburda: Reworded, though not to "loses bare-id fault ownership": with the fault-scope fix above the external app keeps its bare-id ownership, so the binding is simply inert. It now reads "ignored for linking and fault scoping (the app is scoped by its entity id)", and the manifest-schema doc matches.

src/ros2_medkit_gateway/src/core/discovery/manifest/manifest_validator.cpp
bburda: Fixed - the validator now `continue`s before the `seen_bindings` bookkeeping for an external app, so its inert binding never reserves a node name. New test `R013_ExternalAppBindingDoesNotReserveNodeName` pins that a real ROS app reusing the same node name does not trip a phantom R010.

docs/config/manifest-schema.rst
bburda: Reworded for the tri-state model rather than OR: with `optional<bool>` an explicit `external: false` is authoritative and does win by layer priority, so only an *omitted* key is a no-op. The doc now says omitting `external:` defers to other layers while an explicit `false` overrides a plugin's classification.

docs/config/discovery-options.rst
bburda: Same as the schema note - reworded for tri-state. Since an explicit `external: false` now wins by priority, it no longer claims "no layer can clear it"; it distinguishes omission (no-op) from an explicit value (authoritative).

src/ros2_medkit_gateway/test/test_handler_context.cpp
bburda: Added `AreaHostingExternalAppOwnsItsFaults` - a component with `area` set hosting an external app, asserting the area rollup includes the app's bare id. The set is now complete (App / Component / Function / Area).

src/ros2_medkit_gateway/test/test_merge_pipeline.cpp
bburda: Added `test_external_app_fault_rollup.test.py`: hybrid manifest with an external app, a fault reported under its bare id via the report_fault service, asserted on the app, component, function and area routes. It exercises the full merge -> scope -> rollup chain and fails against the pre-fix fault scope.

mfaferek93 (approved):

PR #527, opened 2026-07-12
Diag bridge code from key

### LINKED ISSUE #526: Allow using a diagnostic value as the fault code
## Summary

On our robots we have a preexisting set of codes in our diagnostics, embedded in the key value pairs. These aren't really fault codes, they are tokens used to map to end user facing translations of the diagnostics.

In order to phase medkit into our system it will be preferable to leverage some of the existing codes we have.

---

Adding a option in the diagnostic bridge to extract a diagnostic value as the fault code

---

nnarain:
# Pull Request

Hey guys. Finally getting around to putting medkit on our robots. Well hopefully haha.

## Summary

On our robots we have a preexisting set of codes in our diagnostics, embedded in the attributes. These aren't really fault codes, they are tokens used to map to end user facing translations of the diagnostics.

In order to phase medkit into our system it will be preferable leverage some of the existing code we have.

This MR adds an option to specify a diagnostic attribute that contains the desired fault code. 

---

## Issue

- closes #526 

---

## Type

- [ ] Bug fix
- [x] New feature or tests
- [ ] Breaking change
- [ ] Documentation only

---

## Testing

Unit test and local testing

image

---

## Checklist


src/ros2_medkit_diagnostic_bridge/src/diagnostic_bridge_node.cpp
```diff
@@ -86,13 +90,23 @@ void DiagnosticBridgeNode::process_diagnostic(const diagnostic_msgs::msg::Diagno
...
+  const auto it = name_to_code_.find(diagnostic_name);
   if (it != name_to_code_.end()) {
     return it->second;
   }
 
+  // Check if the diagnostic contains any attributes that contain a code
+  if (!keyvalue_codes_.empty()) {
+    for (const auto & key_value : status.values) {
+      if (keyvalue_codes_.count(key_value.key) > 0) {
+        return key_value.value;
```
mfaferek93: tiny edge case :D 
if a configured key is present but its value is empty, this returns "", and process_diagnostic then drops the fault entirely, right? before auto-generate can step in
A quick `if (!key_value.value.empty())` guard would let an empty attribute fall through to the name-based code instead of silently swallowing the fault

src/ros2_medkit_diagnostic_bridge/src/diagnostic_bridge_node.cpp
```diff
@@ -38,6 +38,10 @@ void DiagnosticBridgeNode::load_parameters() {
   diagnostics_topic_ = declare_parameter<std::string>("diagnostics_topic", "/diagnostics");
   auto_generate_codes_ = declare_parameter<bool>("auto_generate_codes", true);
 
+  std::vector<std::string> keyvalue_codes =
+      declare_parameter<std::vector<std::string>>("keyvalue_codes", std::vector<std::string>());
```
mfaferek93, line 42: Would you mind adding `keyvalue_codes` to the `Parameters` table in README.md (right after the `name_to_code` row), plus a line in the example config block below it?
Right now it's only discoverable from the source, and it's a genuinely handy knob for anyone layering medkit onto existing diagnostics.

src/ros2_medkit_diagnostic_bridge/src/diagnostic_bridge_node.cpp
```diff
@@ -86,13 +90,23 @@ void DiagnosticBridgeNode::process_diagnostic(const diagnostic_msgs::msg::Diagno
...
   if (it != name_to_code_.end()) {
     return it->second;
   }
 
+  // Check if the diagnostic contains any attributes that contain a code
+  if (!keyvalue_codes_.empty()) {
+    for (const auto & key_value : status.values) {
+      if (keyvalue_codes_.count(key_value.key) > 0) {
+        return key_value.value;
+      }
```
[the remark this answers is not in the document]
mfaferek93: Honestly this only matters if you ever configure more than one key: the match walks status.values in message order, so if a status happens to carry two of your keys, the winner is just whichever the publisher listed first. If you'd rather that be predictable, looping over your configured keys instead makes the first-declared one win. It can be something like this:
```
for (const auto & key : keyvalue_codes_) {
  for (const auto & kv : status.values) {
    if (kv.key == key) return kv.value;
  }
}
```
I leave it up to you @nnarain

mfaferek93 (commented):
Thanks for this! Great to see medkit going onto real robots, and reusing your existing diagnostic codes to phase it in is exactly what the bridge is for. A couple of small, non-blocking suggestions :)

src/ros2_medkit_diagnostic_bridge/src/diagnostic_bridge_node.cpp
nnarain: good call

src/ros2_medkit_diagnostic_bridge/src/diagnostic_bridge_node.cpp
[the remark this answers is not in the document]
nnarain: we only have the one code we use in the diagnostics. I extended it to multiples as it was easy enough to do. Opted to make the output based on the order specified in the parameter list.

mfaferek93 (approved):
Looks good to me!

PR #530, opened 2026-07-13
Fault scope: external Component owns its fault_manager faults (#516)

### LINKED ISSUE #516: Fault scope: external Component with no bound Apps cannot own its fault_manager faults
## Summary

`collect_app_fqn()` in `src/ros2_medkit_gateway/src/core/faults/fault_scope.cpp` has the external rule: an external App with an empty `effective_fqn()` contributes its own entity id as a fault-scope owner, so faults reported to the fault_manager under that id roll up correctly.

There is no equivalent for Components. `collect_component_app_fqns()` only aggregates child Apps, so a protocol-plugin device modeled as an external Component with no child Apps (a PLC, a fieldbus device, any non-ROS asset) can never own its fault_manager faults:

- `GET /components/<device>/faults` source filtering resolves an empty scope.
- `Function.hosts` -> Component: `collect_function_app_fqns()` already dispatches Component hosts, but the Component contributes nothing, so the Function rollup silently misses the device's faults.

The only workaround is injecting a synthetic child App per device whose sole purpose is carrying fault-scope ownership - a hollow entity that serves no data and pollutes the tree.

---

Mirror the App rule: an external Component whose Apps contribute no fqns (including the no-Apps case) contributes its own entity id to the scope. Non-external Components keep the current behavior for the same reason unbound ROS Apps do - an unbound entity must not claim faults it never reported.

---

With this rule a manifest Function can host device Components directly and the rollup spans them - no synthetic Apps needed.

bburda:
## Summary

An external Component is a non-ROS device (a PLC, a fieldbus node) that a protocol plugin bridges into SOVD. Before this change, such a Component could not own its fault_manager faults. `collect_component_app_fqns` only collected the FQNs of child Apps. A device modeled as a Component with no child App resolved to an empty fault scope. So `GET /components/<device>/faults` returned nothing, and a Function or Area that hosted the device also missed its faults. The only workaround was to add a fake child App per device, only to carry fault ownership.

This change gives Components the same `external` rule that Apps already have:

- `Component` gets an `external` field (tri-state `optional<bool>`, like `App::external`). It is read from the manifest `external:` key and merged across discovery layers with the same rule: a layer that does not set it never clears a value set by another layer (for example a plugin), and an explicit manifest value stays authoritative.
- In fault scope, an external Component whose child Apps contribute no FQNs (including the case of no Apps) owns its own entity id. A non-external Component with no Apps still resolves to an empty scope, so it cannot claim faults it never reported.
- The change is in one helper, so it fixes the Component, Area, and Function rollups at once, plus the plugin fault path.

A Function or Area can now host a device Component directly and roll up its faults. No fake App is needed.

The change also exposes the `external` flag on the REST wire as `x-medkit.external`, for both Components and Apps. Apps did not expose it before. The flag is written only when it is true, so entities that are not external keep the same wire shape as before.

This change is additive. A Component without an explicit `external: true` is treated exactly as before, so existing manifests do not change behavior.

---

## Issue

- closes #516

---

## Type

- [x] New feature or tests

---

## Testing

A new end-to-end launch test drives the real gateway and fault_manager over HTTP:

- an external Component reports a fault under its own id; the fault appears on the component, function, and area routes;
- an internal (ROS) Component reports a fault under its own id; the test first confirms the fault is on the server-wide `/faults` list, then checks that it does not appear on the component route (the guardrail);
- `x-medkit.external` is present on the external Component and App and absent on the internal one; the asset-identity nameplate is present on both.

Unit tests cover the model field, the manifest parse, the cross-layer merge, the fault-scope resolution for the component/area/function routes, the DTO output, and the handler wire output.

Run the end-to-end test:

```
colcon build --packages-up-to ros2_medkit_integration_tests
./scripts/test.sh test_external_component_fault_rollup
```

---

## Checklist

- [x] Breaking changes are clearly described (no breaking changes; the field is additive)

### DIFF
--- a/src/ros2_medkit_gateway/include/ros2_medkit_gateway/dto/x_medkit.hpp
+++ b/src/ros2_medkit_gateway/include/ros2_medkit_gateway/dto/x_medkit.hpp
@@ -103,6 +103,7 @@ inline constexpr std::string_view dto_name<XMedkitArea> = "XMedkitArea";
 //
 // Also used in sub-collection responses (depends-on, subcomponents, hosts, contains, etc.):
 //   missing           <- true when component reference cannot be resolved
+//   external          <- comp.external, non-ROS external asset classification (#516)
 //
 // Note: "parentComponentId" uses camelCase on the wire per discovery_handlers.cpp.
 // "dependsOn" uses camelCase on the wire per discovery_handlers.cpp.
@@ -122,7 +123,8 @@ struct XMedkitComponent {
   // "_provenance"). Free-form JSON so the DTO layer reuses the exact
   // serialization emitted by Component::to_json and consumed by peer parsing.
   std::optional<nlohmann::json> identity;
-  std::optional<bool> missing;  // broken reference sentinel
+  std::optional<bool> missing;   // broken reference sentinel
+  std::optional<bool> external;  // non-ROS external asset classification (#516)
 };
 
 template <>
@@ -132,7 +134,8 @@ inline constexpr auto dto_fields<XMedkitComponent> = std::make_tuple(
     field("dependsOn", &XMedkitComponent::depends_on), field("area", &XMedkitComponent::area),
     field("variant", &XMedkitComponent::variant), field("description", &XMedkitComponent::description),
     field("contributors", &XMedkitComponent::contributors), field("capabilities", &XMedkitComponent::capabilities),
-    field("identity", &XMedkitComponent::identity), field("missing", &XMedkitComponent::missing));
+    field("identity", &XMedkitComponent::identity), field("missing", &XMedkitComponent::missing),
+    field("external", &XMedkitComponent::external));
 
 template <>
 inline constexpr std::string_view dto_name<XMedkitComponent> = "XMedkitComponent";
@@ -149,6 +152,7 @@ inline constexpr std::string_view dto_name<XMedkitComponent> = "XMedkitComponent
 //
 // Also used in sub-collection responses (depends-on, hosts, function-hosts):
 //   missing       <- true when app reference cannot be resolved
+//   external      <- app.external, non-ROS external asset classification (#516/#517)
 // ---------------------------------------------------------------------------
 struct XMedkitApp {
   std::optional<XMedkitRos2> ros2;
@@ -156,14 +160,16 @@ struct XMedkitApp {
   std::optional<bool> is_online;
   std::optional<std::string> component_id;
   std::optional<std::vector<std::string>> contributors;
-  std::optional<bool> missing;  // broken reference sentinel
+  std::optional<bool> missing;   // broken reference sentinel
+  std::optional<bool> external;  // non-ROS external asset classification (#516/#517)
 };
 
 template <>
 inline constexpr auto dto_fields<XMedkitApp> =
     std::make_tuple(field("ros2", &XMedkitApp::ros2), field("source", &XMedkitApp::source),
                     field("is_online", &XMedkitApp::is_online), field("component_id", &XMedkitApp::component_id),
-                    field("contributors", &XMedkitApp::contributors), field("missing", &XMedkitApp::missing));
+                    field("contributors", &XMedkitApp::contributors), field("missing", &XMedkitApp::missing),
+                    field("external", &XMedkitApp::external));
 
 template <>
 inline constexpr std::string_view dto_name<XMedkitApp> = "XMedkitApp";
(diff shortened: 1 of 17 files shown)


src/ros2_medkit_gateway/include/ros2_medkit_gateway/dto/x_medkit.hpp
```diff
@@ -122,7 +123,8 @@ struct XMedkitComponent {
   // "_provenance"). Free-form JSON so the DTO layer reuses the exact
   // serialization emitted by Component::to_json and consumed by peer parsing.
   std::optional<nlohmann::json> identity;
-  std::optional<bool> missing;  // broken reference sentinel
+  std::optional<bool> missing;   // broken reference sentinel
+  std::optional<bool> external;  // non-ROS external asset classification (#516)
```
mfaferek93, line 128: The "explicit-false serializes as absent" contract isn't enforced here: JsonWriter emits any optional that has a value, so `external = false` goes on the wire as `false`. It holds today only because all six emit sites guard with `value_or(false)` and never assign the raw optional - but manifest_parser sets explicit false and merge_pipeline preserves it, so one natural `x.external = component.external;` breaks it. Consider true-or-absent semantics at the DTO layer instead of the scattered guards.

src/ros2_medkit_gateway/src/http/handlers/discovery_handlers.cpp
```diff
@@ -561,6 +561,9 @@ DiscoveryHandlers::get_components(const http::TypedRequest & req) {
       if (!component.identity.empty()) {
         x_medkit_comp.identity = component.identity.to_json();
       }
+      if (component.external.value_or(false)) {
+        x_medkit_comp.external = true;
```
mfaferek93: You set `external` here (and on the other three list/detail routes: 698/985/1126), but the relationship/sub-collection routes never get it: area-components:339, subcomponents:779, component-hosts:853, depends-on:913, contains:484, function-hosts:1574. On those `absence == not external` (component.hpp:109) is false, so a device browsed via a relationship route reads as a plain ROS component - and since #516 gives it its own fault scope, a consumer classifying ownership off a relationship item misclassifies. Add the same guard on each, or narrow the documented invariant to "sub-collections are navigation-only".

src/ros2_medkit_gateway/src/discovery/manifest/manifest_parser.cpp
```diff
@@ -255,6 +255,13 @@ Component ManifestParser::parse_component(const YAML::Node & node) const {
   comp.source = "manifest";
   comp.identity = parse_identity(node);
 
+  // Preserve the omitted-vs-explicit distinction: an absent `external:` key
+  // leaves nullopt so the hybrid merge cannot let a stub's default erase a
+  // plugin's introspected classification (#516, mirrors the App rule #517).
+  if (node["external"]) {
+    comp.external = node["external"].as<bool>();
```
mfaferek93, line 262: `parse_asset` doesn't do the same, and `external` is missing from its `structural` set (:287), so an `assets:` entry with `external: true` silently falls through to `identity.extra["external"]` and never classifies. Assets are the inventory path for the PLC/fieldbus hardware #516 exists for - mirror this in parse_asset and add `external` to `structural`.

src/ros2_medkit_gateway/include/ros2_medkit_gateway/dto/x_medkit.hpp
bburda: Fixed. All eleven sites that emit a Component or App x-medkit now go through one helper, set_x_medkit_external(dto, entity.external). It is the only place that writes the field: it sets true or leaves it absent, so no call site can assign the raw optional and put external: false on the wire.

src/ros2_medkit_gateway/src/http/handlers/discovery_handlers.cpp
bburda: Fixed with option (a). external is now emitted on the relationship and sub-collection routes too: area-components, area-contains, subcomponents, component depends-on, app depends-on, component-hosts, function-hosts. They use the same helper, so absence == not external holds on every route that carries the entity x-medkit. Non-external items keep the same shape. Added handler tests for the subcomponents and hosts projections.

src/ros2_medkit_gateway/src/discovery/manifest/manifest_parser.cpp
bburda: Fixed. parse_asset now reads external, and external is added to its structural key set, so an assets: entry with external: true classifies the Component instead of falling through to identity.extra. Tri-state is preserved, same as parse_component. Added a parser test for true, false, and omitted, and that the key does not land in the identity extras.

mfaferek93 (approved):

PR #532, opened 2026-07-13
Stop an unresponsive parameter node from hanging the REST API (#531)

### LINKED ISSUE #531: [BUG] Unbounded SyncParametersClient calls in ros2_parameter_transport can hang the whole gateway REST API
### Steps to reproduce
1. Run the gateway against a ROS 2 node that exposes a parameter service.
2. Put that node into a state where its parameter service is still visible in the ROS graph but does not answer requests. This can happen when a node is mid-recovery after a middleware router or discovery restart, when a node is hung, or when a node is shutting down.
3. Call a parameter endpoint that talks to that node, for example `GET /{entity}/configurations`, or read/set a single parameter on it.

### Expected behavior
The call is bounded by the configured service timeout and returns an error for that node. Other requests, and the rest of the REST API, keep working.

### Actual behavior
The parameter call blocks forever.

In `Ros2ParameterTransport` (`src/ros2_medkit_gateway/src/ros2/transports/ros2_parameter_transport.cpp`) the `client->wait_for_service(get_service_timeout())` check is bounded, but the follow-up calls are not:

- `client->list_parameters({}, 0)` / `client->list_parameters({name}, 1)`
- `client->get_parameters(...)`
- `client->set_parameters(...)`

These `rclcpp::SyncParametersClient` calls are made with no timeout argument, so they use the rclcpp default, which is to wait forever. `wait_for_service` only checks that the service is discoverable. Discovery can succeed while the request/response round-trip is not ready yet, so the code passes the bounded check and then blocks on the unbounded call.

This gets worse because of how the calls are scheduled. All parameter operations share one spin lock, and a Component or Area `/configurations` request fans out one thread per backing node. One stuck node blocks all parameter operations and holds one HTTP worker thread forever. A request that touches many nodes (for example reading the whole tree) can leave several workers stuck. Once the HTTP worker pool is exhausted, the whole REST API stops answering, including `/health`. It does not recover on its own; the gateway process has to be restarted.

### Suggested fix
Pass an explicit timeout to the `list_parameters` / `get_parameters` / `set_parameters` calls, the same `get_service_timeout()` already used for `wait_for_service`. Then a discoverable-but-unresponsive node fails fast for its own request instead of blocking forever.

Secondary hardening worth considering:
- One stuck node should not be able to block parameter operations for every other node. The shared spin lock plus the per-request thread fan-out let a single node exhaust the worker pool.
- The `invalidate()` path in the same file appears unused. Wiring it up would let a node that started failing be dropped from the cache.

### Environment
- ros2_medkit version: main (current)
- ROS 2 distro: Jazzy
- OS: Ubuntu 24.04

### Additional information
Reproduced live: a node whose parameter service was discoverable but not answering left HTTP worker threads blocked indefinitely (observed for more than 10 minutes with no recovery). The REST API only came back after restarting the gateway container.

bburda:
## Summary

`Ros2ParameterTransport` made parameter service round trips (`list_parameters`, `get_parameters`, `set_parameters`, `describe_parameters`) with no timeout, so a node whose parameter service is discoverable but does not answer (mid-recovery after a discovery restart, hung, or shutting down) blocked the calling thread forever. All parameter operations share one spin lock, and a Component or Area `/configurations` request fans out one thread per backing node, so one stuck node could hold HTTP worker threads forever, exhaust the worker pool, and stop the whole REST API (including `/health`) with no recovery until the gateway was restarted.

Bounding the calls alone is not enough. `SyncParametersClient` returns the same empty result for a real timeout and for a successful-but-empty response (only `list_parameters` throws on timeout; `get_parameters` and `set_parameters` return an empty vector), so a bounded call cannot tell "the node did not answer" from "the node answered with no value". This change therefore switches the transport to `AsyncParametersClient` and spins the client node with an explicit `spin_until_future_complete`, so the outcome of every round trip is an unambiguous `FutureReturnCode`.

What the change does:

- Every parameter round trip is bounded by the configured service timeout, and its outcome is an explicit `FutureReturnCode`:
  - `TIMEOUT`: the node did not answer, so return 503 (service unavailable) and negative-cache the node so the next request to it fails fast.
  - `SUCCESS`: use the result, which may legitimately be empty (an unset or not-found parameter). An empty result is handled as a normal empty response, never as a timeout, so a healthy node is never wrongly negative-cached.
  - `INTERRUPTED`: the gateway is shutting down, so return a shutdown error without marking the node.
- Read and write paths are both covered. A timed-out read or write now returns 503 (a write timeout was previously misclassified as 500). A client sending a bad value on a set request still cannot negative-cache a healthy node, because input parsing is outside the timeout scope.
- The default-values cache is never poisoned with an empty map on a timeout, so a node that timed out once is retried after the negative-cache TTL instead of staying unavailable until restart. A genuine empty result from a responsive node is still memoized.
- The shared spin lock is held for at most one round trip per default-values pass, so a stuck node cannot make a concurrent request to a different, healthy node time out on lock acquisition.
- Removes the dead `invalidate()` transport method (an interface method that was never called).

No behavior change for a healthy node: a responsive service answers well within the existing `parameter_service_timeout_sec` gateway parameter, which is the bound used. The client swap is internal to the transport.

---

## Issue

- closes #531

---

## Type

- [x] Bug fix

---

## Testing

Component tests (`test_ros2_parameter_transport`) drive the real transport against real nodes. One node blocks `list_parameters`; another responds to `list_parameters` but blocks `get`/`set`, which exercises the read and write timeout paths that a list-only stub cannot reach. The tests assert that a timeout returns 503 and negative-caches the node, that a responsive node returning an empty result is NOT marked (the false positive this change fixes), that a marked node recovers after the TTL, and that the default-values cache is not poisoned. Assertions check the error class rather than wall-clock time, so they do not flake under the slower sanitizer runs. Each timeout path was verified to hang against the pre-fix code (the call was killed by a watchdog) and to pass after the fix.

End-to-end launch test (`test_param_service_hang`) runs the real gateway plus a demo node with an unresponsive parameter service. It asserts that a single `GET /{entity}/configurations` on that node returns a bounded 503 (it hangs against the pre-fix code) and that `/health` stays responsive under concurrent stuck requests.

Run:

```
colcon build --packages-up-to ros2_medkit_integration_tests
./scripts/test.sh test_ros2_parameter_transport
./scripts/test.sh test_param_service_hang
```

---

## Checklist

- [x] Breaking changes are clearly described (none; the bound is the existing configured service timeout, and the client swap is internal to the transport)
- [x] Docs were updated if behavior or public API changed (no public API change; internal transport change, with header and design-doc comments updated)

### DIFF
--- a/src/ros2_medkit_gateway/src/ros2/transports/ros2_parameter_transport.cpp
+++ b/src/ros2_medkit_gateway/src/ros2/transports/ros2_parameter_transport.cpp
@@ -148,16 +148,29 @@ std::shared_ptr<rclcpp::AsyncParametersClient> Ros2ParameterTransport::get_param
 
 template <typename FutureT>
 rclcpp::FutureReturnCode Ros2ParameterTransport::spin_for(const FutureT & future) {
+  // Copy the node shared_ptr ONCE under clients_mutex_ and spin the local copy.
+  // spin_mutex_ (held by the caller) serialises spins in the normal case, but
+  // shutdown() has a degraded path: if it cannot acquire spin_mutex_ within
+  // kShutdownTimeout it resets param_node_ anyway. Reading the member directly here
+  // would then race that reset (concurrent access to the same shared_ptr instance =
+  // UB). shutdown() clears param_clients_ and resets param_node_ under clients_mutex_,
+  // so taking clients_mutex_ synchronises the read; the local copy also keeps the node
+  // alive for the whole spin even if shutdown resets the member mid-call (#531).
+  std::shared_ptr<rclcpp::Node> node;
+  {
+    std::lock_guard<std::mutex> lock(clients_mutex_);
+    node = param_node_;
+  }
   // Guard against spinning a torn-down node (shutdown races with an in-flight op).
-  if (shutdown_requested_.load() || !param_node_) {
+  if (shutdown_requested_.load() || !node) {
     return rclcpp::FutureReturnCode::INTERRUPTED;
   }
-  // Free-function overload spins a fresh SingleThreadedExecutor over param_node_
+  // Free-function overload spins a fresh SingleThreadedExecutor over the node
   // (collecting all current entities, including a just-created client) for up to
   // service_timeout_sec, then tears it down - no persistent executor to reconcile
   // with param_node_'s teardown. Serialised by spin_mutex_ (held by the caller),
-  // so only one spin runs on param_node_ at a time.
-  return rclcpp::spin_until_future_complete(param_node_, future, get_service_timeout());
+  // so only one spin runs on the node at a time.
+  return rclcpp::spin_until_future_complete(node, future, get_service_timeout());
 }
 
 ParameterResult Ros2ParameterTransport::list_own_parameters() {
@@ -1004,8 +1017,15 @@ bool Ros2ParameterTransport::cache_default_values(const std::string & node_name)
     }
 
     if (param_names.empty()) {
-      // Genuine zero-parameter node (responsive, SUCCESS with empty names). Nothing to
-      // cache; leave uncached and do NOT mark.
+      // Genuine zero-parameter node (responsive, SUCCESS with empty names). This is a
+      // STABLE, authoritative fact, so memoize an EMPTY defaults map: get_default then
+      // returns NOT_FOUND (not NO_DEFAULTS_CACHED) and subsequent list/get/reset calls
+      // do not re-round-trip under spin_mutex_ forever. Only a TIMEOUT must avoid caching
+      // (poison-avoidance); a SUCCESS-empty is safe to cache (#531).
+      std::lock_guard<std::mutex> lock(defaults_mutex_);
+      if (default_values_.find(node_name) == default_values_.end()) {
+        default_values_[node_name] = {};
+      }
       return false;
     }
 
@@ -1023,13 +1043,10 @@ bool Ros2ParameterTransport::cache_default_values(const std::string & node_name)
       parameters = get_future.get();
     }
 
-    if (parameters.empty()) {
-      // Responsive node but no values came back (unset / undeclared / raced). Do NOT
-      // cache an empty map and do NOT mark - leave uncached so a later call retries
-      // once the values exist (#531).
-      return false;
-    }
-
+    // SUCCESS with values (or a responsive node that returned an empty set - unset /
+    // undeclared / raced): both are authoritative, so cache the resulting map. An empty
+    // map here is a genuine SUCCESS-empty and is SAFE to memoize (unlike the TIMEOUT
+    // branches above, which must not cache to avoid permanent poison) (#531).
     std::map<std::string, rclcpp::Parameter> node_defaults;
     for (const auto & param : parameters) {
       node_defaults[param.get_name()] = param;
--- a/src/ros2_medkit_gateway/src/ros2/transports/ros2_parameter_transport.cpp
+++ b/src/ros2_medkit_gateway/src/ros2/transports/ros2_parameter_transport.cpp
@@ -71,21 +71,6 @@ bool Ros2ParameterTransport::is_node_available(const std::string & node_name) co
   return !is_node_unavailable(node_name);
 }
 
-void Ros2ParameterTransport::invalidate(const std::string & node_name) {
-  {
-    std::unique_lock<std::shared_mutex> lock(negative_cache_mutex_);
-    unavailable_nodes_.erase(node_name);
-  }
-  {
-    std::lock_guard<std::mutex> lock(defaults_mutex_);
-    default_values_.erase(node_name);
-  }
-  {
-    std::lock_guard<std::mutex> lock(clients_mutex_);
-    param_clients_.erase(node_name);
-  }
-}
-
 std::optional<std::unique_lock<std::timed_mutex>>
 Ros2ParameterTransport::try_acquire_spin_lock(ParameterResult & result) {
   auto timeout = std::chrono::duration_cast<std::chrono::milliseconds>(
@@ -271,13 +256,22 @@ ParameterResult Ros2ParameterTransport::list_parameters(const std::string & node
         return result;
       }
 
-      auto param_names = client->list_parameters({}, 0);
-      parameters = client->get_parameters(param_names.names);
+      std::vector<std::string> param_names;
+      try {
+        auto list_result = client->list_parameters({}, 0, get_service_timeout());
+        param_names = list_result.names;
+        parameters = client->get_parameters(param_names, get_service_timeout());
+      } catch (const std::exception &) {
+        // Discoverable-but-unresponsive node: negative-cache so the next
+        // request fails fast instead of waiting the full timeout again (#531).
+        mark_node_unavailable(node_name);
+        throw;
+      }
 
-      if (parameters.empty() && !param_names.names.empty()) {
-        for (const auto & name : param_names.names) {
+      if (parameters.empty() && !param_names.empty()) {
+        for (const auto & name : param_names) {
           try {
-            auto single_params = client->get_parameters({name});
+            auto single_params = client->get_parameters({name}, get_service_timeout());
             if (!single_params.empty()) {
               parameters.push_back(single_params[0]);
             }
@@ -353,7 +347,23 @@ ParameterResult Ros2ParameterTransport::get_parameter(const std::string & node_n
         return result;
       }
 
-      auto param_names = client->list_parameters({param_name}, 1);
+      rcl_interfaces::msg::ListParametersResult param_names;
+      std::vector<rclcpp::Parameter> parameters;
+      try {
+        param_names = client->list_parameters({param_name}, 1, get_service_timeout());
+        if (!param_names.names.empty()) {
+          parameters = client->get_parameters({param_name}, get_service_timeout());
+          if (!parameters.empty()) {
+            descriptors = client->describe_parameters({param_name}, get_service_timeout());
+          }
+        }
+      } catch (const std::exception &) {
+        // Discoverable-but-unresponsive node: negative-cache so the next
+        // request fails fast instead of waiting the full timeout again (#531).
+        mark_node_unavailable(node_name);
+        throw;
+      }
+
       if (param_names.names.empty()) {
         result.success = false;
         result.error_message = "Parameter not found: " + param_name;
@@ -361,7 +371,6 @@ ParameterResult Ros2ParameterTransport::get_parameter(const std::string & node_n
         return result;
       }
 
-      auto parameters = client->get_parameters({param_name});
       if (parameters.empty()) {
         result.success = false;
         result.error_message = "Failed to get parameter: " + param_name;
@@ -370,7 +379,6 @@ ParameterResult Ros2ParameterTransport::get_parameter(const std::string & node_n
       }
 
       param = parameters[0];
-      descriptors = client->describe_parameters({param_name});
     }  // spin_mutex_ released.
 
     json param_obj;
@@ -465,16 +473,34 @@ ParameterResult Ros2ParameterTransport::set_parameter(const std::string & node_n
       return result;
     }
 
-    auto current_params = client->get_parameters({param_name});
+    std::vector<rclcpp::Parameter> current_params;
+    try {
+      current_params = client->get_parameters({param_name}, get_service_timeout());
+    } catch (const std::exception &) {
+      // Discoverable-but-unresponsive node: negative-cache so the next
+      // request fails fast instead of waiting the full timeout again (#531).
+      mark_node_unavailable(node_name);
+      throw;
+    }
     rclcpp::ParameterType hint_type = rclcpp::ParameterType::PARAMETER_NOT_SET;
     if (!current_params.empty()) {
       hint_type = current_params[0].get_type();
     }
 
+    // json_to_parameter_value throws on bad CLIENT input (e.g. malformed
+    // value for the parameter's type) - deliberately NOT wrapped by the
+    // node-round-trip try/catch above/below so a bad client value never
+    // negative-caches a healthy node (#531).
     rclcpp::ParameterValue param_value = json_to_parameter_value(value, hint_type);
     rclcpp::Parameter param(param_name, param_value);
 
-    auto results = client->set_parameters({param});
+    std::vector<rcl_interfaces::msg::SetParametersResult> results;
+    try {
+      results = client->set_parameters({param}, get_service_timeout());
+    } catch (const std::exception &) {
+      mark_node_unavailable(node_name);
+      throw;
+    }
     if (results.empty() || !results[0].successful) {
       result.success = false;
       result.error_message = results.empty() ? "Failed to set parameter" : results[0].reason;
@@ -828,15 +854,23 @@ void Ros2ParameterTransport::cache_default_values(const std::string & node_name)
       return;
     }
 
-    auto param_names = client->list_parameters({}, 0);
-
+    std::vector<std::string> param_names;
     std::vector<rclcpp::Parameter> parameters;
-    parameters = client->get_parameters(param_names.names);
+    try {
+      auto list_result = client->list_parameters({}, 0, get_service_timeout());
+      param_names = list_result.names;
+      parameters = client->get_parameters(param_names, get_service_timeout());
+    } catch (const std::exception &) {
+      // Discoverable-but-unresponsive node: negative-cache so the next
+      // request fails fast instead of waiting the full timeout again (#531).
+      mark_node_unavailable(node_name);
+      throw;
+    }
 
-    if (parameters.empty() && !param_names.names.empty()) {
-      for (const auto & name : param_names.names) {
+    if (parameters.empty() && !param_names.empty()) {
+      for (const auto & name : param_names) {
         try {
-          auto single_params = client->get_parameters({name});
+          auto single_params = client->get_parameters({name}, get_service_timeout());
           if (!single_params.empty()) {
             parameters.push_back(single_params[0]);
           }
--- a/src/ros2_medkit_gateway/src/ros2/transports/ros2_parameter_transport.cpp
+++ b/src/ros2_medkit_gateway/src/ros2/transports/ros2_parameter_transport.cpp
@@ -288,9 +288,22 @@ ParameterResult Ros2ParameterTransport::list_parameters(const std::string & node
       }
 
       if (parameters.empty() && !param_names.empty()) {
+        // get_parameters came back empty even though list_parameters gave us names.
+        // Retry per-name (works around batch quirks), but under ONE shared deadline:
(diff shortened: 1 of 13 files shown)


src/ros2_medkit_gateway/src/ros2/transports/ros2_parameter_transport.cpp
```diff
@@ -271,13 +268,22 @@ ParameterResult Ros2ParameterTransport::list_parameters(const std::string & node
...
+      try {
+        auto list_result = client->list_parameters({}, 0, get_service_timeout());
+        param_names = list_result.names;
+        parameters = client->get_parameters(param_names, get_service_timeout());
+      } catch (const std::exception &) {
+        // Discoverable-but-unresponsive node: negative-cache so the next
+        // request fails fast instead of waiting the full timeout again (#531).
+        mark_node_unavailable(node_name);
+        throw;
+      }
```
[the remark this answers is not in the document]
bburda: Fixed. This catch no longer rethrows. It marks the node unavailable, sets ParameterErrorCode::TIMEOUT, and returns, so a bounded round-trip timeout is reported as node-unavailable and the REST layer returns 503 instead of 500.

src/ros2_medkit_gateway/src/ros2/transports/ros2_parameter_transport.cpp
```diff
@@ -353,15 +359,30 @@ ParameterResult Ros2ParameterTransport::get_parameter(const std::string & node_n
...
+          if (!parameters.empty()) {
+            descriptors = client->describe_parameters({param_name}, get_service_timeout());
+          }
+        }
+      } catch (const std::exception &) {
+        // Discoverable-but-unresponsive node: negative-cache so the next
+        // request fails fast instead of waiting the full timeout again (#531).
+        mark_node_unavailable(node_name);
+        throw;
+      }
```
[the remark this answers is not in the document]
bburda: Fixed. This path now sets ParameterErrorCode::TIMEOUT and returns instead of rethrowing, so a bounded round-trip timeout returns 503 instead of 500.

src/ros2_medkit_gateway/src/ros2/transports/ros2_parameter_transport.cpp
```diff
@@ -465,16 +485,34 @@ ParameterResult Ros2ParameterTransport::set_parameter(const std::string & node_n
...
-    auto current_params = client->get_parameters({param_name});
+    std::vector<rclcpp::Parameter> current_params;
+    try {
+      current_params = client->get_parameters({param_name}, get_service_timeout());
+    } catch (const std::exception &) {
+      // Discoverable-but-unresponsive node: negative-cache so the next
+      // request fails fast instead of waiting the full timeout again (#531).
+      mark_node_unavailable(node_name);
+      throw;
+    }
```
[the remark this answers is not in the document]
bburda: Fixed. Same change as the other round-trip catches: it sets ParameterErrorCode::TIMEOUT and returns, so a bounded get_parameters timeout returns 503 instead of 500.

src/ros2_medkit_gateway/src/ros2/transports/ros2_parameter_transport.cpp
```diff
@@ -465,16 +485,34 @@ ParameterResult Ros2ParameterTransport::set_parameter(const std::string & node_n
...
     rclcpp::Parameter param(param_name, param_value);
 
-    auto results = client->set_parameters({param});
+    std::vector<rcl_interfaces::msg::SetParametersResult> results;
+    try {
+      results = client->set_parameters({param}, get_service_timeout());
+    } catch (const std::exception &) {
+      mark_node_unavailable(node_name);
+      throw;
+    }
```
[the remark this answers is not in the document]
bburda: Fixed. This catch now sets ParameterErrorCode::TIMEOUT and returns, so a bounded set_parameters timeout returns 503 instead of 500. There is also a comment noting that a 503 here does not guarantee the set was not applied on the node, so clients should re-read.

src/ros2_medkit_integration_tests/demo_nodes/unresponsive_param_node.cpp
```diff
@@ -0,0 +1,181 @@
...
+        if (rc == 0) {
+          break;
+        }
+        // EINTR can theoretically fire; retry.
+      }
+      // Flip rclcpp::ok() to false first so the blocking list_parameters
+      // callback notices and returns; only then ask the executor to stop.
+      rclcpp::shutdown();
+      executor.cancel();
+    });
```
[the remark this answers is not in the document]
bburda: This order is intentional for this fixture. The ~/list_parameters callback blocks in a loop on rclcpp::ok(), so calling executor.cancel() first would deadlock: ok() stays true, the callback never returns, and the spin thread never joins. Calling rclcpp::shutdown() first flips ok() to false, the callback returns, then the executor is stopped and the node is destroyed. The node exits with an allowed code and the post-shutdown test confirms it. The code comment explains why the order differs from the standard demo-node pattern.

src/ros2_medkit_integration_tests/test/features/test_param_service_hang.test.py
```diff
@@ -0,0 +1,240 @@
...
+
+            start = time.monotonic()
+            response = requests.get(f'{self.BASE_URL}/health', timeout=30)
+            elapsed = time.monotonic() - start
+        except requests.exceptions.RequestException as e:
+            self.fail(
+                f'/health did not respond within 30s while '
+                f'{CONCURRENT_STUCK_REQUESTS} concurrent /configurations '
+                f'calls were stuck against the unresponsive node: {e}'
+            )
```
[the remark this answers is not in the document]
bburda: Fixed. The barrier wait now catches threading.BrokenBarrierError and calls self.fail with a clear message, so a barrier timeout fails the test cleanly instead of erroring with a traceback.

src/ros2_medkit_gateway/src/ros2/transports/ros2_parameter_transport.cpp
```diff
@@ -271,13 +271,26 @@ ParameterResult Ros2ParameterTransport::list_parameters(const std::string & node
         return result;
       }
 
-      auto param_names = client->list_parameters({}, 0);
-      parameters = client->get_parameters(param_names.names);
+      std::vector<std::string> param_names;
+      try {
+        auto list_result = client->list_parameters({}, 0, get_service_timeout());
+        param_names = list_result.names;
+        parameters = client->get_parameters(param_names, get_service_timeout());
```
mfaferek93: Core assumption issue: `SyncParametersClient::get_parameters` doesn't throw on timeout - it returns an empty vector (rclcpp parameter_client.cpp, Humble and Jazzy; only `list_parameters` throws). So this catch never fires for a timeout: control falls into the per-name fallback below, each per-name get also returns empty, and the function returns success=true with [] - REST answers 200+[] for a node that stopped answering, indistinguishable from zero parameters, and no negative-cache. After the fallback loop, treat `parameters.empty() && !param_names.empty()` as TIMEOUT + `mark_node_unavailable` (a responsive node returns one entry per name even for unset params), or switch to the async client where timeout is explicit.

src/ros2_medkit_gateway/src/ros2/transports/ros2_parameter_transport.cpp
```diff
@@ -465,16 +509,49 @@ ParameterResult Ros2ParameterTransport::set_parameter(const std::string & node_n
...
+    // value for the parameter's type) - deliberately NOT wrapped by the
+    // node-round-trip try/catch above/below so a bad client value never
+    // negative-caches a healthy node (#531).
     rclcpp::ParameterValue param_value = json_to_parameter_value(value, hint_type);
     rclcpp::Parameter param(param_name, param_value);
 
-    auto results = client->set_parameters({param});
+    std::vector<rcl_interfaces::msg::SetParametersResult> results;
+    try {
+      results = client->set_parameters({param}, get_service_timeout());
```
mfaferek93: Same dead-catch as the get paths: `set_parameters` returns an empty vector on timeout, it never throws (only `list_parameters` does). A timed-out set falls through to `results.empty()` at :555 and surfaces as 500 "Failed to set parameter" with no negative-cache - the exact #531 misclassification, still live on the write path. Map `results.empty()` to TIMEOUT + `mark_node_unavailable` and move the at-least-once note there. The pre-read get at :514 is equally dead: on timeout it comes back empty, hint_type stays NOT_SET, and the request burns a second full timeout.

src/ros2_medkit_gateway/src/ros2/transports/ros2_parameter_transport.cpp
```diff
@@ -353,15 +366,31 @@ ParameterResult Ros2ParameterTransport::get_parameter(const std::string & node_n
         return result;
       }
 
-      auto param_names = client->list_parameters({param_name}, 1);
+      rcl_interfaces::msg::ListParametersResult param_names;
+      std::vector<rclcpp::Parameter> parameters;
+      try {
+        param_names = client->list_parameters({param_name}, 1, get_service_timeout());
+        if (!param_names.names.empty()) {
+          parameters = client->get_parameters({param_name}, get_service_timeout());
```
mfaferek93: Also unreachable for timeouts: `get_parameters` returns empty instead of throwing, so after a successful `list_parameters` a timed-out get skips this catch, hits `parameters.empty()` at :394 and returns 500 with no negative-cache - and the next request pays the full timeout again. list just confirmed the name exists, so an empty result here is the timeout path (the only other cause is a delete race, and 503 beats 500 for that too).

src/ros2_medkit_gateway/src/ros2/transports/ros2_parameter_transport.cpp
```diff
@@ -825,18 +902,26 @@ void Ros2ParameterTransport::cache_default_values(const std::string & node_name)
...
 
-    auto param_names = client->list_parameters({}, 0);
-
+    std::vector<std::string> param_names;
     std::vector<rclcpp::Parameter> parameters;
-    parameters = client->get_parameters(param_names.names);
+    try {
+      auto list_result = client->list_parameters({}, 0, get_service_timeout());
+      param_names = list_result.names;
+      parameters = client->get_parameters(param_names, get_service_timeout());
```
mfaferek93: This one poisons permanently: a timed-out batch get returns empty (no throw), the catch is skipped, the per-name fallback burns N more timeouts silently, and lines 941-950 then cache an EMPTY defaults map. `default_values_` is never invalidated and :884 short-circuits on any cached entry, so the node returns NOT_FOUND / empty defaults forever, even after it recovers. Treat `parameters.empty() && !param_names.empty()` as a failed round-trip: `mark_node_unavailable` and don't commit the empty map.

src/ros2_medkit_gateway/src/ros2/transports/ros2_parameter_transport.cpp
```diff
@@ -271,13 +271,26 @@ ParameterResult Ros2ParameterTransport::list_parameters(const std::string & node
...
+        mark_node_unavailable(node_name);
+        result.success = false;
+        result.error_message = "Parameter service did not respond for node: " + node_name;
+        result.error_code = ParameterErrorCode::TIMEOUT;
+        return result;
+      }
 
-      if (parameters.empty() && !param_names.names.empty()) {
-        for (const auto & name : param_names.names) {
+      if (parameters.empty() && !param_names.empty()) {
```
mfaferek93: This fallback loop can hold spin_mutex_ for N x service_timeout (each per-name get burns the full timeout and returns empty, so nothing short-circuits), defeating the one-round-trip hold cap the comment at :244 establishes (same at :921-939). Meanwhile try_acquire_spin_lock waiters give up after timeout+1s, so concurrent requests to healthy nodes fail with spurious TIMEOUT - the starvation this PR exists to eliminate. Compute one deadline before the loop and pass the remaining budget to each call, bailing when exhausted.

src/ros2_medkit_gateway/test/test_ros2_parameter_transport.cpp
```diff
@@ -0,0 +1,277 @@
...
+          }
+        });
+
+    // The remaining parameter services just need to exist in the ROS graph
+    // so that wait_for_service() (which only probes service discovery, not
+    // responsiveness) succeeds regardless of which underlying client it
+    // checks. They are not expected to be invoked by the scenarios below
+    // since list_parameters is the first remote call in the round trip and
+    // it never returns within the bounded timeout.
+    get_parameters_service_ = unresponsive_node_->create_service<rcl_interfaces::srv::GetParameters>(
```
mfaferek93, line 93: Only the two list_parameters sites are actually exercised - the blocking stub is ListParameters-only and every path throws on its first list round-trip, so get_parameters (278/293, 374), describe_parameters (410) and the whole set path (514/539) have zero regression coverage. Deleting their timeout argument would pass this suite. Add blocking stubs for GetParameters/SetParameters and drive those paths.

src/ros2_medkit_gateway/test/test_ros2_parameter_transport.cpp
```diff
@@ -0,0 +1,277 @@
...
+         "service_timeout_sec, not block until the server replies (or forever, pre-fix)";
+  EXPECT_LT(elapsed, std::chrono::seconds(2))
+      << "list_parameters() must cap the spin_mutex_ hold at ~1x service_timeout_sec (single "
+         "round-trip) by short-circuiting once cache_default_values marks the node unavailable";
+}
+
+// A second, immediate call against the same node must short-circuit via the
+// negative cache (mark_node_unavailable() populated it on the first call's
+// round-trip failure) rather than paying the timeout again.
+TEST_F(TestRos2ParameterTransportUnresponsiveNode, SecondCallShortCircuitsViaNegativeCache) {
```
mfaferek93, line 218: This proves the negative-cache hit but not recovery: nothing tests that a node that timed out once becomes reachable after the TTL. A regression that keeps re-marking the node would pass green and 503 it until gateway restart. The fixture has what's needed: sub-second TTL, one timeout, release_blocking_callback_, sleep past TTL, assert list_parameters succeeds.

src/ros2_medkit_gateway/test/test_ros2_parameter_transport.cpp
```diff
@@ -0,0 +1,277 @@
...
+  constexpr double kNegativeCacheDisabled = 0.0;
+  auto ttl0_transport =
+      std::make_shared<ros2::Ros2ParameterTransport>(client_node_.get(), kServiceTimeoutSec, kNegativeCacheDisabled);
+
+  auto start = std::chrono::steady_clock::now();
+  auto result = ttl0_transport->list_parameters(kUnresponsiveNodeName);
+  auto elapsed = std::chrono::steady_clock::now() - start;
+
+  EXPECT_FALSE(result.success);
+  EXPECT_LT(elapsed, std::chrono::milliseconds(1800))
```
mfaferek93: This 1.8s bound leaves ~0.8s for client construction + endpoint matching + the exception path, and the test runs under TSan/ASan where that's 5-15x slower (quality.yml already triples ctest timeouts for this) - it will flake with no regression present. There's a timing-independent discriminator already: assert the returned error class (SERVICE_UNAVAILABLE short-circuit vs TIMEOUT) instead of wall-clock.

src/ros2_medkit_gateway/src/ros2/transports/ros2_parameter_transport.cpp
bburda: Confirmed and fixed at the root. You are right that only list_parameters throws on timeout while get_parameters and set_parameters return an empty vector. The transport was reworked to use AsyncParametersClient with an explicit spin_until_future_complete, so a timeout is now detected via FutureReturnCode::TIMEOUT rather than an empty response. A timed-out list op returns 503 and negative-caches the node; a successful but empty response is treated as a legitimate empty result, so a healthy node is no longer wrongly marked.

src/ros2_medkit_gateway/src/ros2/transports/ros2_parameter_transport.cpp
bburda: Fixed. With the async client the set round trip returns FutureReturnCode::TIMEOUT on a timeout, mapped to 503 plus negative-cache, so the write path is no longer misclassified as 500. A successful set that returns one result per parameter is handled as before (rejection vs success), and the at-least-once note is on the timeout path.

src/ros2_medkit_gateway/src/ros2/transports/ros2_parameter_transport.cpp
bburda: Fixed. The get round trip now reports FutureReturnCode::TIMEOUT explicitly (503 plus negative-cache). A successful but empty get is returned as NOT_FOUND without marking the node, so a raced or unset parameter no longer negative-caches a healthy node.

src/ros2_medkit_gateway/src/ros2/transports/ros2_parameter_transport.cpp
bburda: Fixed. cache_default_values does not cache anything on a timeout, so there is no permanent empty-map poison, and the node is negative-cached and retried after the TTL. A genuine empty result from a responsive node is memoized, so NOT_FOUND is returned and repeated calls do not re-round-trip.

src/ros2_medkit_gateway/src/ros2/transports/ros2_parameter_transport.cpp
bburda: Addressed. The per-name fallback that could hold the lock for N times the timeout was removed. With explicit timeout detection the batch result is authoritative, so the fallback is no longer needed; each round trip is a single bounded spin under the lock.

src/ros2_medkit_gateway/test/test_ros2_parameter_transport.cpp
bburda: Fixed. There is now a fixture whose node responds to list_parameters but blocks get and set, giving regression coverage of the get and set timeout paths that were previously not exercised. New tests assert those paths return TIMEOUT/503 plus negative-cache, and a test asserts that a responsive node returning an empty result is not marked.

src/ros2_medkit_gateway/test/test_ros2_parameter_transport.cpp
bburda: Fixed. There is a recovery test: a node times out once and is negative-cached, becomes responsive, and after the TTL a later call succeeds, proving a marked node recovers instead of staying 503 until restart.

src/ros2_medkit_gateway/test/test_ros2_parameter_transport.cpp
bburda: Fixed. The assertions now check the error class (SERVICE_UNAVAILABLE vs TIMEOUT) rather than wall-clock time, so they do not flake under the slower sanitizer runs. Any remaining time bound is a generous sanity check, not the pass/fail criterion.

mfaferek93 (approved):

PR #534, opened 2026-07-14
fix(gateway): bound parameter transport per-node caches with LRU eviction

### LINKED ISSUE #533: [BUG] Parameter transport per-node caches grow without bound (memory leak)
# Bug report

### Steps to reproduce
1. Run the gateway against a ROS 2 graph where nodes come and go over time, each with a distinct fully qualified name (for example a fleet or a test harness that starts and stops many uniquely named nodes).
2. Read parameters through the SOVD API (GET on a Component or Area `/configurations`, or on an App), so the gateway's `Ros2ParameterTransport` is asked about each node at least once.
3. Let this run for a long time, so many distinct node names are queried.

### Expected behavior
The transport's per-node caches should stay bounded. Memory use should not grow with the total number of distinct node names ever seen.

### Actual behavior
`Ros2ParameterTransport` keeps two caches keyed by node name that gain one entry per distinct node name and are never evicted at runtime:

- `param_clients_` (one `rclcpp::AsyncParametersClient` per node) is only cleared in `shutdown()`.
- `default_values_` (cached default parameter values per node) is never cleared at all.

So every distinct node name ever queried leaves a permanent entry. On a long-lived gateway facing a churning graph both caches grow without bound, which is a memory leak. The negative cache (`unavailable_nodes_`) is already bounded with a size cap and eviction; the other two are not.

### Environment
- ros2_medkit version: main
- ROS 2 distro: Jazzy (also applies to Humble and Lyrical)
- OS: Ubuntu 24.04

### Additional information
The two caches are internal to `Ros2ParameterTransport`. A single request still works; the problem is unbounded memory over the process lifetime. Fix: bound both caches the same way the negative cache is bounded (a size cap plus eviction).

bburda:
## Summary

`Ros2ParameterTransport` kept per-node caches that gained one entry per distinct node name and were never evicted at runtime: the `AsyncParametersClient` cache (only cleared at shutdown) and the default-value cache (never cleared). On a long-lived gateway facing a churning graph these grow without bound, which leaks memory. The negative cache was already bounded; this extends the same discipline to the others.

This change backs the caches with a new header-only `BoundedLruCache<Key, Value>` and adds the hardening needed to make eviction safe and correct:

- **Separate caps for separate weights.** The client cache uses a moderate cap (`kMaxParamClientCacheSize`) because each entry is heavy: an `AsyncParametersClient` owns several DDS service clients. The default-value caches use a much larger cap (`kMaxDefaultsCacheSize`) because a parameter map is light AND because evicting a default is not transparent.

- **Pre-write default preservation.** Evicting and re-seeding a node's defaults could rebase what "default" means to a current (already-overwritten) value, so reset-to-default would report success and change nothing. `set_parameter` now records a parameter's pre-write value (from the pre-read it already does, so no extra round trip) in a dedicated `pre_write_values_` cache that `get_default` and `list_defaults` prefer over the first-seen snapshot. A throttled WARN is logged when a default-value cache evicts an entry.

- **Client eviction is race-free.** Creating or evicting an `AsyncParametersClient` mutates `param_node_`'s entity set, which `spin_for()` reads under `spin_mutex_`. `get_param_client` now runs inside the spin lock in every operation (matching `get_parameter` and `set_parameter`), and `shutdown()` blocks for the in-flight spin before tearing the clients down, so an rcl entity is never freed while a spin still references it.

- `BoundedLruCache` tracks recency with a monotonic access counter (no clock, so eviction order is deterministic and tie-free) and is not internally locked: the transport holds its existing `clients_mutex_` / `defaults_mutex_` around every call.

No behavior change for a healthy gateway with a stable set of nodes: nothing is evicted until a cache passes its cap.

---

## Issue

- closes #533

---

## Type

- [x] Bug fix

---

## Testing

`BoundedLruCache` unit tests pin the eviction rules deterministically: size never passes the cap, `find()` refreshes recency (a hot entry survives) while `contains()` does not, replace on an existing key refreshes recency, insertion order breaks ties, a zero cap clamps to one, and the on-evict callback fires with the evicted entry.

Transport tests drive the real `Ros2ParameterTransport`:

- the client cache stays at its cap across many distinct node names;
- evicting a live node's client is transparent: the next query rebuilds it and still succeeds;
- an evicted live node's defaults are re-fetched, not reported missing;
- after a set, `get_default` returns the pre-write value while the live value is the new one.

Run:

```
colcon build --packages-select ros2_medkit_gateway
./scripts/test.sh test_bounded_lru_cache
./scripts/test.sh test_ros2_parameter_transport
```

---

## Checklist

- [x] Breaking changes are clearly described (none; internal caches only, no public API or wire change)
- [x] Docs were updated if behavior or public API changed (internal transport change; header and class doc comments updated, no public API change)

### DIFF
--- a/src/ros2_medkit_gateway/include/ros2_medkit_gateway/core/util/bounded_lru_cache.hpp
+++ b/src/ros2_medkit_gateway/include/ros2_medkit_gateway/core/util/bounded_lru_cache.hpp
@@ -16,6 +16,7 @@
 
 #include <cstddef>
 #include <cstdint>
+#include <functional>
 #include <iterator>
 #include <map>
 #include <utility>
@@ -48,10 +49,19 @@ namespace ros2_medkit_gateway {
 template <typename Key, typename Value>
 class BoundedLruCache {
  public:
+  /// Invoked with the (key, value) of an entry about to be evicted, just before
+  /// it is erased. Lets a caller observe evictions (for example to log a warning
+  /// when a cached default value is dropped). Runs while the caller's mutex is
+  /// held, so it must be cheap and must not call back into the cache.
+  using EvictCallback = std::function<void(const Key &, const Value &)>;
+
   /// @param max_size Maximum number of entries retained. Values below 1 are
   ///                 clamped to 1 (a zero-capacity LRU cache is meaningless and
   ///                 would evict the entry it just inserted).
-  explicit BoundedLruCache(std::size_t max_size) : max_size_(max_size == 0 ? 1 : max_size) {
+  /// @param on_evict Optional callback run for each least-recently-used entry
+  ///                 evicted on overflow. Empty by default (no observation).
+  explicit BoundedLruCache(std::size_t max_size, EvictCallback on_evict = {})
+    : max_size_(max_size == 0 ? 1 : max_size), on_evict_(std::move(on_evict)) {
   }
 
   /// Look up @p key, marking it most-recently-used on a hit.
@@ -118,11 +128,15 @@ class BoundedLruCache {
         victim = it;
       }
     }
+    if (on_evict_) {
+      on_evict_(victim->first, victim->second.value);
+    }
     entries_.erase(victim);
   }
 
   std::size_t max_size_;
   std::uint64_t counter_{0};
+  EvictCallback on_evict_;
   std::map<Key, Entry> entries_;
 };
 
--- a/src/ros2_medkit_gateway/include/ros2_medkit_gateway/core/util/bounded_lru_cache.hpp
+++ b/src/ros2_medkit_gateway/include/ros2_medkit_gateway/core/util/bounded_lru_cache.hpp
@@ -76,17 +76,17 @@ class BoundedLruCache {
   ///
   /// The just-stored entry always carries the largest stamp, so it is never the
   /// eviction victim; erasing a different entry from the underlying std::map
-  /// does not invalidate the reference returned here.
+  /// does not invalidate the iterator (and returned reference) kept here.
+  /// insert_or_assign is used (not operator[]) so Value need not be
+  /// default-constructible.
   ///
   /// @return reference to the stored value (valid until the next mutating call).
   Value & put(const Key & key, Value value) {
-    auto & entry = entries_[key];
-    entry.value = std::move(value);
-    entry.stamp = ++counter_;
+    auto it = entries_.insert_or_assign(key, Entry{std::move(value), ++counter_}).first;
     if (entries_.size() > max_size_) {
-      evict_lru();
+      evict_lru();  // never evicts `it`: it carries the largest stamp
     }
-    return entry.value;
+    return it->second.value;
   }
 
   /// Remove all entries. The access counter is intentionally not reset - stamps
--- a/src/ros2_medkit_gateway/include/ros2_medkit_gateway/core/util/bounded_lru_cache.hpp
+++ b/src/ros2_medkit_gateway/include/ros2_medkit_gateway/core/util/bounded_lru_cache.hpp
@@ -1,0 +1,129 @@
+// Copyright 2026 bburda
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+#pragma once
+
+#include <cstddef>
+#include <cstdint>
+#include <iterator>
+#include <map>
+#include <utility>
+
+namespace ros2_medkit_gateway {
+
+/// Fixed-capacity least-recently-used cache.
+///
+/// Backs per-key caches that would otherwise grow one entry per distinct key
+/// for a process lifetime - for example the parameter transport's per-node
+/// AsyncParametersClient cache and default-value cache, each of which gains one
+/// entry per node name ever queried and, unbounded, leaks memory over a long
+/// run against a churning graph. When an insert pushes the entry count past
+/// @c max_size, the least-recently-used entry is evicted.
+///
+/// Recency is tracked with a monotonic access counter, not a wall or steady
+/// clock: every find() hit and every put() stamps the entry with the next
+/// counter value, and eviction removes the entry with the smallest stamp. This
+/// gives a strict, tie-free ordering that is deterministic and test-friendly,
+/// and needs no clock. The counter is 64-bit; at one access per nanosecond it
+/// would take ~585 years to wrap, so overflow is not a practical concern.
+///
+/// NOT thread-safe. A shared instance must be guarded by the caller's own mutex
+/// on every call. Ros2ParameterTransport holds clients_mutex_ around its client
+/// cache and defaults_mutex_ around its default-value cache for exactly this
+/// reason.
+///
+/// @tparam Key   map key type (must be usable as a std::map key)
+/// @tparam Value cached value type
+template <typename Key, typename Value>
+class BoundedLruCache {
+ public:
+  /// @param max_size Maximum number of entries retained. Values below 1 are
+  ///                 clamped to 1 (a zero-capacity LRU cache is meaningless and
+  ///                 would evict the entry it just inserted).
+  explicit BoundedLruCache(std::size_t max_size) : max_size_(max_size == 0 ? 1 : max_size) {
+  }
+
+  /// Look up @p key, marking it most-recently-used on a hit.
+  /// @return pointer to the stored value, or nullptr if absent. The pointer is
+  ///         valid only until the next mutating call (put/clear) on this cache.
+  Value * find(const Key & key) {
+    auto it = entries_.find(key);
+    if (it == entries_.end()) {
+      return nullptr;
+    }
+    it->second.stamp = ++counter_;
+    return &it->second.value;
+  }
+
+  /// Report whether @p key is present WITHOUT changing its recency.
+  bool contains(const Key & key) const {
+    return entries_.find(key) != entries_.end();
+  }
+
+  /// Insert or replace the value for @p key, mark it most-recently-used, then
+  /// evict the least-recently-used entry if the cache is now over capacity.
+  ///
+  /// The just-stored entry always carries the largest stamp, so it is never the
+  /// eviction victim; erasing a different entry from the underlying std::map
+  /// does not invalidate the reference returned here.
+  ///
+  /// @return reference to the stored value (valid until the next mutating call).
+  Value & put(const Key & key, Value value) {
+    auto & entry = entries_[key];
+    entry.value = std::move(value);
+    entry.stamp = ++counter_;
+    if (entries_.size() > max_size_) {
+      evict_lru();
+    }
+    return entry.value;
+  }
+
+  /// Remove all entries. The access counter is intentionally not reset - stamps
+  /// only need to be monotonic, not dense.
+  void clear() {
+    entries_.clear();
+  }
+
+  std::size_t size() const {
+    return entries_.size();
+  }
+  std::size_t max_size() const {
+    return max_size_;
+  }
+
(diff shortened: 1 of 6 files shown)


src/ros2_medkit_gateway/src/ros2/transports/ros2_parameter_transport.cpp
```diff
@@ -1054,8 +1061,8 @@ bool Ros2ParameterTransport::cache_default_values(const std::string & node_name)
 
     {
       std::lock_guard<std::mutex> lock(defaults_mutex_);
-      if (default_values_.find(node_name) == default_values_.end()) {
-        default_values_[node_name] = std::move(node_defaults);
+      if (!default_values_.contains(node_name)) {
+        default_values_.put(node_name, std::move(node_defaults));
```
mfaferek93, line 1147: Bounding this cache silently rebases what "default" means. Defaults are captured as first-seen live values, and before this PR the entry lived for the process lifetime. With eviction, a node that goes LRU-cold gets re-seeded from its CURRENT values - including ones the gateway itself wrote via set_parameter - so reset-to-default reports success and changes nothing. That's a new failure mode in exactly the churning-graph deployment this PR targets. Cheap mitigations: give default_values_ its own much larger cap instead of sharing one knob with the heavyweight client cache (a Parameter map is orders of magnitude lighter than an AsyncParametersClient), pin entries for nodes the gateway has written to (or record the pre-write value at set time), and add a throttled WARN on defaults eviction.

src/ros2_medkit_gateway/test/test_ros2_parameter_transport.cpp
```diff
@@ -280,6 +281,35 @@ TEST_F(TestRos2ParameterTransportUnresponsiveNode, ListParametersCapHoldsWithNeg
...
+  constexpr std::size_t kCacheCap = 4;
+  constexpr double kShortTimeoutSec = 0.1;
+  constexpr double kNegativeCacheDisabled = 0.0;
+  auto bounded_transport = std::make_shared<ros2::Ros2ParameterTransport>(client_node_.get(), kShortTimeoutSec,
+                                                                          kNegativeCacheDisabled, kCacheCap);
+
+  constexpr std::size_t kDistinctNodes = kCacheCap * 3;
+  for (std::size_t i = 0; i < kDistinctNodes; ++i) {
+    const std::string node_name = "/nonexistent_param_node_" + std::to_string(i);
+    auto result = bounded_transport->get_parameter(node_name, "some_param");
```
mfaferek93, line 312: Every node queried here is nonexistent, so this pins the size bound but nothing about behavior after eviction. The contract the LRU must preserve is that evicting a live node's client is transparent: next query rebuilds the AsyncParametersClient and still succeeds (eviction destroys rcl entities on param_node_, so a regression is plausible and currently invisible). Add a happy-path case: fill past the cap, then re-query the evicted live node and assert success.

src/ros2_medkit_gateway/test/test_ros2_parameter_transport.cpp
```diff
@@ -280,6 +281,35 @@ TEST_F(TestRos2ParameterTransportUnresponsiveNode, ListParametersCapHoldsWithNeg
   EXPECT_LT(elapsed, std::chrono::seconds(5)) << "sanity: still bounded";
 }
 
+// The per-node AsyncParametersClient cache is bounded by LRU eviction, so a
```
mfaferek93, line 293: Both caches got bounded but only one is observable: param_client_cache_size() + the bound test cover param_clients_, while default_values_ has no size accessor and no eviction coverage - and that's exactly what changed. Missing: after a live node's defaults are evicted, get_default must re-fetch (not return NO_DEFAULTS_CACHED), and the contains-then-find across two defaults_mutex_ acquisitions (:729 vs :743) needs the eviction-between-locks case pinned.

src/ros2_medkit_gateway/test/test_bounded_lru_cache.cpp
```diff
@@ -0,0 +1,134 @@
...
+  cache.put(1, "one");
+  auto * value = cache.find(1);
+  ASSERT_NE(value, nullptr);
+  EXPECT_EQ(*value, "one");
+}
+
+TEST(BoundedLruCacheTest, PutReplaceUpdatesValueWithoutGrowing) {
+  BoundedLruCache<int, std::string> cache(4);
+  cache.put(1, "one");
+  cache.put(1, "uno");
```
mfaferek93, line 43: This checks value and size on replace, but not recency - the header documents that put() on an existing key re-stamps it MRU, and the suite pins the other two recency rules (find refreshes, contains doesn't), so a refactor that makes replace keep the old stamp stays green. Pin it: cap=2, put(1), put(2), put(1,"updated"), put(3) -> expect 2 evicted, 1 present.

src/ros2_medkit_gateway/include/ros2_medkit_gateway/ros2/transports/ros2_parameter_transport.hpp
```diff
@@ -166,13 +184,21 @@ class Ros2ParameterTransport : public ParameterTransport {
   /// Created once in constructor - must be in DDS graph early for fast service discovery.
   std::shared_ptr<rclcpp::Node> param_node_;
 
-  /// Cached AsyncParametersClient per target node.
+  /// Cached AsyncParametersClient per target node, bounded by LRU eviction.
   mutable std::mutex clients_mutex_;
-  std::map<std::string, std::shared_ptr<rclcpp::AsyncParametersClient>> param_clients_;
+  BoundedLruCache<std::string, std::shared_ptr<rclcpp::AsyncParametersClient>> param_clients_;
```
mfaferek93, line 219: One interaction with shutdown() worth closing (transport .cpp ~:66): when try_lock_for(kShutdownTimeout) fails, param_clients_.clear() on the degraded path destroys these cached clients' rcl entities while an in-flight operation is still spinning on param_node_. The in-flight caller's own client survives via its shared_ptr, but every other cached client whose last reference is this cache is destroyed mid-spin - the exact entity-set mutation the .cpp:270-276 comment classifies as unsafe (spin_for's local node copy only covers the node half). Prefer waiting unbounded on the degraded path, or leaking the clients there instead of destroying them under a live spinner.

src/ros2_medkit_gateway/include/ros2_medkit_gateway/core/util/bounded_lru_cache.hpp
```diff
@@ -0,0 +1,129 @@
...
+  /// @return reference to the stored value (valid until the next mutating call).
+  Value & put(const Key & key, Value value) {
+    auto & entry = entries_[key];
+    entry.value = std::move(value);
+    entry.stamp = ++counter_;
+    if (entries_.size() > max_size_) {
+      evict_lru();
+    }
+    return entry.value;
+  }
```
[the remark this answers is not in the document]
bburda: Good point. Changed put() to use insert_or_assign instead of operator[], so the cache no longer requires Value to be default-constructible. Same insert-or-replace semantics and the same recency and eviction behavior.

src/ros2_medkit_gateway/test/test_bounded_lru_cache.cpp
```diff
@@ -0,0 +1,134 @@
...
+#include "ros2_medkit_gateway/core/util/bounded_lru_cache.hpp"
+
+namespace ros2_medkit_gateway {
+namespace {
+
+TEST(BoundedLruCacheTest, FindMissReturnsNullptr) {
+  BoundedLruCache<int, std::string> cache(4);
+  EXPECT_EQ(cache.find(42), nullptr);
+  EXPECT_EQ(cache.size(), 0u);
+}
```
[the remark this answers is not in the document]
bburda: The @verifies tags in this repo track only SOVD interoperability requirements (REQ_INTEROP_*), and new internal requirement IDs are not created. These tests cover an internal memory-bounding utility, not a SOVD interop requirement, so there is no requirement ID to reference. The other parameter transport tests follow the same untagged pattern, so these stay untagged on purpose.

src/ros2_medkit_gateway/test/test_ros2_parameter_transport.cpp
```diff
@@ -280,6 +281,35 @@ TEST_F(TestRos2ParameterTransportUnresponsiveNode, ListParametersCapHoldsWithNeg
...
+// gateway that queries an ever-growing set of distinct node names cannot leak
+// one client per node for its whole lifetime (the pre-fix behaviour). A
+// transport built with a small cache cap and pointed at many distinct,
+// nonexistent node names must keep its client cache at the cap. get_param_client
+// runs (and inserts) before the failing wait_for_service, so every distinct
+// query would add an entry without the bound. A short service timeout is used
+// because every query hits a nonexistent service; the assertion is on the cache
+// size, not on timing.
+TEST_F(TestRos2ParameterTransportUnresponsiveNode, ClientCacheStaysBoundedAcrossManyNodes) {
```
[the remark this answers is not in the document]
bburda: Same as the other test file: @verifies here tracks only SOVD interop requirements (REQ_INTEROP_*), and this test covers internal client-cache bounding, not a SOVD interop requirement. No requirement ID applies and new internal IDs are not created, so it stays untagged like the other tests in this file.

src/ros2_medkit_gateway/src/ros2/transports/ros2_parameter_transport.cpp
bburda: Fixed. Separated the caps: the default-value caches use a much larger cap than the client cache (kMaxDefaultsCacheSize vs kMaxParamClientCacheSize), because a parameter map is far lighter than an AsyncParametersClient and evicting a default is not transparent. On top of that, set_parameter now records a parameter's pre-write value in a dedicated pre_write_values_ cache (from the pre-read it already does, so no extra round trip), and get_default / list_defaults prefer it over the first-seen snapshot. So reset-to-default restores the value from before the gateway changed it, even if the snapshot was captured after a write or was evicted and re-seeded. A throttled WARN is logged when a default-value cache evicts an entry.

src/ros2_medkit_gateway/test/test_ros2_parameter_transport.cpp
bburda: Added EvictedLiveClientIsTransparentlyRebuilt: it queries a live responsive node, fills the client cache past its cap with distinct nodes to evict that client, then re-queries the live node and asserts it still succeeds because the client is rebuilt. That covers the transparency contract the size-only test did not.

src/ros2_medkit_gateway/test/test_ros2_parameter_transport.cpp
bburda: Added default_values_cache_size() and EvictedDefaultsAreRefetchedNotReportedMissing: with a defaults cap of 1 and two responsive nodes, caching the second evicts the first, and a later get_default on the first re-fetches and succeeds instead of returning NO_DEFAULTS_CACHED; the size accessor is asserted to stay at the cap. On the contains-then-find window: a just-cached entry is the most-recently-used, so a single concurrent cache of another node evicts the least-recently-used entry, not it. Reaching it would need as many concurrent distinct-node caches as the cap (now large) inside the small gap between the two locks, and for written parameters the pre_write short-circuit skips that path entirely.

src/ros2_medkit_gateway/test/test_bounded_lru_cache.cpp
bburda: Added PutOnExistingKeyRefreshesRecency: cap 2, put(1), put(2), put(1,"updated"), put(3), and it asserts key 2 is evicted, not the re-stamped key 1. Pins that put() on an existing key refreshes recency.

src/ros2_medkit_gateway/include/ros2_medkit_gateway/ros2/transports/ros2_parameter_transport.hpp
bburda: Fixed. shutdown() no longer tears down under a live spinner: if the initial timed wait on spin_mutex_ lapses, it now blocks until the in-flight spin releases before clearing param_clients_ and resetting param_node_. Each spin is bounded by the service timeout, so this cannot hang in practice, and a bounded wait is safer than freeing an rcl entity a live spin still references. Every client-cache mutation (create, LRU-evict, shutdown-clear) now happens under spin_mutex_.

mfaferek93 (approved):

PR #536, opened 2026-07-16
Correct provisional action-fault attribution via supersedes_source_id

### LINKED ISSUE #467: Re-attribute action faults raised before DDS discovery resolves
## Summary

When an action aborts before the `action_status_bridge` can resolve the action server's node FQN (slow or flaky DDS discovery), the fault is reported immediately using the action interface name as the source. Reporting the fault on time is intentional (freeze-frame / snapshot timing is tied to the first CONFIRMED). But the fault then cannot be re-attributed to the server node entity: `reporting_sources` is append-only on the FaultManager side, and the per-entity `/faults` scope filter (`fault_in_source_scope`) is strict-AND, so the provisional action-name source permanently keeps the fault out of the server node entity.

Add a way to supersede a provisional reporting source. For example, an optional `supersedes_source_id` field on `ReportFault`, with the FaultManager removing the old source when the new one is reported. The bridge would then report immediately on the action-name fallback (fault and snapshot on time) and, once the FQN resolves, re-report under the FQN with `supersedes_source_id` set to the action name, ending with `reporting_sources = {FQN}` so the fault resolves to the server entity.

In the `action_status_bridge` the reporter source is fixed at the first report (server FQN if discovery resolved, otherwise the action name) and not re-attributed, precisely because of the append-only + strict-AND constraints above.

bburda:
# Pull Request

## Summary

An action fault raised before DDS discovery resolves the action server node's
FQN was reported under the action interface name (e.g. `/navigate_to_pose`).
Because `reporting_sources` is an append-only set and the per-entity `/faults`
scope filter is strict-AND (every source must be in the entity's scope), that
provisional source stayed in the set permanently and kept the fault hidden from
every node entity, including the real server node it belongs to.

This adds an optional `ReportFault.supersedes_source_id`: a re-report can drop a
previously-reported source before adding the new one. The `action_status_bridge`
records the provisional source and, once discovery resolves the FQN, re-reports
the already-delivered fault under the server FQN with the provisional source
superseded, so the FaultManager ends with `reporting_sources = {FQN}` and the
fault resolves to the correct entity.

Behavior is opt-in and backward compatible: an empty `supersedes_source_id`
(the default) preserves the existing add-only behavior.

Key pieces:

- **`ReportFault.srv`**: optional `supersedes_source_id` field.
- **`FaultStorage` / `SqliteFaultStorage`**: drop the superseded source before
  adding the new one, in both the FAILED-update and CLEARED-reactivation paths;
  a self-supersede (`supersedes == source_id`) is a no-op so a fault can never
  end up with an empty source set.
- **`FaultReporter::report`**: optional `supersedes_source_id`; a supersede
  bypasses local debounce filtering so the correcting report is never dropped.
- **`action_status_bridge`**: tracks the provisional source per action and
  corrects it on the next rescan once the FQN is known. A healed or
  not-yet-delivered fault is left untouched; re-attribution happens at most once
  per fault.

---

## Issue

- closes #467

---

## Type

- [x] Bug fix
- [x] New feature or tests
- [ ] Breaking change
- [x] Documentation only

Note: not a breaking change (the new field/argument default to empty), but
out-of-tree callers must rebuild against the regenerated `ros2_medkit_msgs`.

---

## Testing

- **Unit (storage)**: with/without supersede, superseding an absent source,
  self-supersede guard, supersede on a brand-new fault, and supersede across a
  CLEARED->FAILED reactivation - for both the in-memory and SQLite stores.
- **Unit (bridge)**: provisional source kept while the report is still deferred,
  resolved once delivered, and kept after a heal - driven deterministically
  through a test seam with a controllable server-FQN resolver.
- **End-to-end**: `test_fault_source_supersede.test.py` drives the real gateway
  scope filter and FaultManager - a fault reported under a provisional source is
  404 under the owning app, then 200 after a superseding re-report under the FQN.
  This exercises the exact strict-AND scope-filter interaction that caused #467.

The bridge's discovery-race timing itself cannot be reproduced deterministically
in a launch test, so the bridge decision logic is covered by the unit seam and
the gateway/FaultManager half by the e2e; together they cover the full path.

Local runs (Jazzy): fault_manager 457, fault_reporter 47, action_status_bridge
65 unit tests - 0 failures; clang-format, clang-tidy (changed files), flake8,
and ament-copyright all clean.

---

## Checklist


src/ros2_medkit_integration_tests/test/features/test_fault_source_supersede.test.py
```diff
@@ -0,0 +1,174 @@
+# Copyright 2026 bburda
+#
```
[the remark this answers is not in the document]
mfaferek93: Fixed in d87024d - the shebang line is now the first line of the file, matching the other *.test.py launch tests.

src/ros2_medkit_action_status_bridge/src/action_status_bridge_node.cpp
```diff
@@ -462,27 +471,111 @@ ros2_medkit_fault_reporter::FaultReporter * ActionStatusBridgeNode::reporter_for
...
   auto reporter = std::make_unique<ros2_medkit_fault_reporter::FaultReporter>(this->shared_from_this(), source_id);
   auto * raw = reporter.get();
   reporters_[action_name] = std::move(reporter);
   return raw;
 }
 
+void ActionStatusBridgeNode::reattribute_provisional() {
+  // Snapshot the provisional action names, then process each without holding a lock
+  // across the per-action state read + graph query + report.
+  std::vector<std::string> actions;
```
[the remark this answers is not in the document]
mfaferek93: Fixed in d87024d - the comment now describes the actual locking: the snapshot avoids iterating provisional_ while mutating it, and the correction itself runs under reporters_mutex_ (same lock as reporter_for()), with a note on why that cannot deadlock.

src/ros2_medkit_msgs/srv/ReportFault.srv
```diff
@@ -49,6 +49,14 @@ string description
...
 # This value is added to the fault's reporting_sources array for multi-source tracking.
 string source_id
+
+# Optional. A previously-reported source_id to drop from this fault's reporting_sources
+# before source_id is added. Empty = no supersede (default). Lets a reporter correct a
+# provisional attribution - e.g. an action bridge that first reports a fault under the
+# action interface name, then re-reports it under the resolved server node FQN once DDS
+# discovery settles - so the fault resolves to the right entity under the strict-AND
+# per-entity /faults scope filter. Only meaningful for FAILED events on an existing fault.
+string supersedes_source_id
```
mfaferek93, line 59: The PR body says this is not a breaking change. For a released interface package it is, and the failure mode on version skew is the worst one a fault system can have: silent.

Old reporter (pre-supersede typesupport) against a new fault_manager still matches at the DDS level: Fast DDS matches on type name + QoS, and the REP-2011 type hash is only advertised via USER_DATA (Jazzy; Humble has no hash at all), it does not gate matching. The old request then fails CDR deserialization on the missing trailing string: __rmw_take_request consumes the sample, catches the fastcdr exception, and returns RMW_RET_OK with taken=false; rcl returns RCL_RET_SERVICE_TAKE_FAILED without logging and the rclcpp executor logs it at DEBUG only. The reporter side sees service_is_ready() == true and fire-and-forgets (send_report never looks at the response), so it prints nothing either. Net effect: every fault report from a non-rebuilt node is dropped with zero output at default log levels, which reads as a healthy robot.

Measured, not just read: old ghcr image (msgs verified pre-supersede) against a fault_manager built from this branch, shared host network and domain - the service lists and types fine from the old side, the call passes the service-ready gate, then hangs to timeout (exit 124); the server prints nothing and the fault is never created, while a control call on the unchanged ListFaults type works on the same channel. The reverse pairing returns accepted=true and silently discards the supersede: the old manager ends with ['/fqn', '/prov'] where the new one ends with ['/fqn'] - the #467 hiding persists while looking healthy.

ros2_medkit_msgs is released into Jazzy and Humble and ships in the ghcr images, and the adoption path is users linking FaultReporter into their own nodes built against their installed msgs, so skew during an upgrade is a normal state, not an edge case. The changelog only says "out-of-tree callers must rebuild"; it should say what happens if they don't.

Asks: mark this as breaking, spell out the skew behavior (reports silently dropped) in the changelog/release notes, and version the release accordingly. If wire compat matters more, the additive alternative is a new service (e.g. SupersedeFaultSource) with ReportFault left untouched.

src/ros2_medkit_fault_manager/src/fault_storage.cpp
```diff
@@ -156,7 +167,7 @@ bool InMemoryFaultStorage::report_fault_event(const std::string & fault_code, ui
     state.first_occurred = timestamp;
     state.last_failed_time = timestamp;
     state.last_occurred = timestamp;
-    state.reporting_sources.insert(source_id);
+    add_source(state.reporting_sources);
```
mfaferek93, line 170: reattribute_provisional() assumes the fault is still active when it fires (its comment: "does not inflate occurrence_count... only swaps the source"), but an operator clear in the window breaks that, and the bridge cannot see one:

1. abort during slow discovery -> FAILED delivered under /navigate_to_pose, fault CONFIRMED
2. operator clears it (ClearFault / gateway clear-all keeps the row as STATUS_CLEARED)
3. bridge still has last_reported_state_ == kFailed, so once the FQN resolves and the new client is ready, it sends the synthetic FAILED with supersede
4. that lands in this reactivation branch: occurrence_count++, first_occurred reset, counter -1 -> CONFIRMED at the default threshold; handle_report_fault then publishes EVENT_CONFIRMED, appends occurred+confirmed audit records, and re-captures snapshot/rosbag once past the 60s cooldown. Same in the sqlite backend.

So a fault the operator explicitly dismissed comes back as CONFIRMED, with an inflated occurrence_count and audit entries for an occurrence that never happened, from a report carrying zero new evidence. Pre-PR the bridge was edge-triggered and never reported again after a clear. The window runs from delivery until reattribution completes: at least one rescan tick, unbounded under slow discovery, which is exactly the regime this PR targets; any clear-all sweep hits every fault pending reattribution.

The store cannot tell this synthetic re-report from a real refail. Suggest making the correction swap-only when the fault is not active (supersede on a CLEARED fault swaps sources and stays CLEARED), or giving the bridge a dedicated correction signal instead of a plain FAILED.

Related, same branch: this add_source is also the one supersede call site nothing tests. Verified by mutation - revert this line to reporting_sources.insert(source_id) and all 148 storage unit tests stay green (SupersedeAppliesAcrossReactivation covers sqlite, where reactivation falls through into the same merge block as a steady-state update; here it is a distinct branch, and neither launch test ever clears a fault). A memory twin of that test (clear_fault, supersede re-report, expect sources == {FQN}) fails on the mutant and passes on this branch.

src/ros2_medkit_fault_reporter/include/ros2_medkit_fault_reporter/fault_reporter.hpp
```diff
@@ -121,7 +121,12 @@ class FaultReporter {
   /// @param fault_code Global fault identifier (e.g., "MOTOR_OVERHEAT")
   /// @param severity Severity level (use Fault::SEVERITY_* constants)
   /// @param description Human-readable fault description
-  void report(const std::string & fault_code, uint8_t severity, const std::string & description);
+  /// @param supersedes_source_id Optional. A previously-reported source to drop from the fault's
+  ///        reporting_sources before this reporter's source is added, so a provisional attribution
+  ///        can be corrected (empty = no supersede). Bypasses local filtering when set, since a
+  ///        re-attribution must always reach the FaultManager.
+  void report(const std::string & fault_code, uint8_t severity, const std::string & description,
+              const std::string & supersedes_source_id = "");
```
mfaferek93, line 129: Adding a defaulted parameter changes the mangled symbol of report(), and fault_reporter_lib is an installed SHARED library with no SOVERSION, so an upgraded .so silently replaces the old one under existing binaries. Reproduced the upgrade with a minimal old-signature binary against the new lib: symbol lookup error, undefined symbol _ZN26ros2_medkit_fault_reporter13FaultReporter6reportERKNSt7__cxx1112basic_stringIcSt11char_traitsIcESaIcEEEhS8_ - at process start under BIND_NOW (Ubuntu's default link hardening), or with lazy binding at the first report() call, meaning the node runs fine until the first fault occurs and then dies. If the non-breaking claim should hold at the binary level, keep a real 3-arg report() overload that delegates to the 4-arg one instead of using a default argument: the old symbol survives, and a non-rebuilt node keeps working against a lockstep-upgraded stack because the request is built inside this library, not in the caller. Otherwise this package's changelog entry needs the same "must rebuild" note the msgs entry got, and the non-breaking claim in the PR text should go.

docs/api/messages.rst
```diff
@@ -179,13 +179,21 @@ Report a fault event to the FaultManager.
...
    string source_id    # Fully qualified node name (e.g., "/powertrain/temp_sensor")
+   string supersedes_source_id  # Optional: a prior source_id to drop before source_id is added
 
 **Response:**
 
 .. code-block:: text
 
    bool accepted       # True if event was accepted
 
+Setting ``supersedes_source_id`` drops that source from the fault's
```
mfaferek93, line 190: docs/api/rest.rst wasn't updated to match: the per-entity fault scope note (rest.rst:828-833) still promises REST consumers that reporting_sources "is the union of every app that has reported that code". Supersede breaks that invariant - a source can vanish between two GETs of the same fault, and a fault that 404'd under an entity can flip to 200 after it was already CONFIRMED (the exact flip test_fault_source_supersede.test.py asserts). rest.rst deserves the same paragraph this file got. While there, one sentence for event consumers would help: when the correction lands on an already-confirmed fault the only signal is a plain fault_updated (fault_manager_node.cpp:662), so a client building per-entity fault lists from fault_confirmed events alone never sees the fault move to the right entity.

src/ros2_medkit_action_status_bridge/src/action_status_bridge_node.cpp
```diff
@@ -462,27 +471,117 @@ ros2_medkit_fault_reporter::FaultReporter * ActionStatusBridgeNode::reporter_for
...
+
+    // Re-report the same fault (the code actually reported, per reported_canceled_)
+    // under the FQN, superseding the provisional source, so the FaultManager ends
+    // with reporting_sources = {FQN} and the fault resolves to the server node's
+    // SOVD entity under the strict-AND scope filter. The fault is still active, so this
+    // second FAILED report does not inflate occurrence_count (the store counts occurrence
+    // edges, not reports) - it only swaps the source. Happens at most once per fault.
+    const std::string code = fault_code_for(action_name, reported_canceled);
+    const std::string desc = std::string("Action ") + action_name + (reported_canceled ? " canceled" : " aborted");
+    pit->second.fqn_reporter->report(code, aborted_severity_, desc, pit->second.old_source);
```
mfaferek93, line 577: The one-shot correction is committed on an unacked send. report() here is fire-and-forget (async_send_request, no callback), send_report early-returns if service_is_ready() flipped after the check at 565, and the swap+erase below plus the "Re-attributed" INFO run either way. Realistic trigger: the FaultManager restarts and this tick lands in the down window, or it dies with the request in flight. The supersede is lost, provisional_ no longer has the entry, and every later FAILED goes out under the FQN with no supersedes_source_id, so reporting_sources keeps the provisional name and strict-AND hides the fault from the owning entity again - the pre-fix #467 state, permanent, logged as corrected.

Ordering is unguarded too: last_reported_state_ == kFailed only means the FAILED was sent (reconcile commits right after its async send), and the supersede travels on a different service client, with no cross-writer ordering guarantee. On heal-then-refail the rescan tick can fire ms after the re-raise FAILED (fqn_reporter is already cached and service-ready), so the manager can process the supersede first and the refail then re-adds the provisional source with nothing left to correct it. Narrower window, same permanent outcome.

Committing the swap+erase in the async_send_request response callback closes the lost-request case - reattribute already retries every rescan tick until then. It does not close the reorder case (the response proves the supersede was processed, not that it landed after the FAILED it names); that one needs the supersede sent on the same client as the FAILED, e.g. a source_id override in send_report, or a read-back before erasing.

src/ros2_medkit_action_status_bridge/src/action_status_bridge_node.cpp
```diff
@@ -462,27 +471,117 @@ ros2_medkit_fault_reporter::FaultReporter * ActionStatusBridgeNode::reporter_for
...
-  // needs a way to supersede a provisional source).
+  // A fallback (action-name) attribution is recorded as provisional and later
+  // corrected to the FQN by reattribute_provisional(): reporting_sources is
+  // append-only and the per-entity /faults scope filter is strict-AND, so the
+  // provisional source cannot merely be added-to - it is swapped for the FQN via
+  // ReportFault.supersedes_source_id once discovery resolves.
   const std::string fqn = server_fqn_for_action(action_name);
   const std::string & source_id = fqn.empty() ? action_name : fqn;
+  if (fqn.empty()) {
+    provisional_[action_name] = Provisional{action_name, nullptr};
```
mfaferek93, line 490: provisional_ is process memory guarding a correction to state that lives in the FaultManager (persistent with the SQLite backend), and prune_vanished erases it too (line 264). Once the entry is gone the fault has no fix path, and the next report makes the misattribution permanent: restart the bridge inside the provisional window, the latched status replays the abort (or the next abort arrives), reporter_for() now resolves the FQN and reports without supersede, and the store appends (fault_storage.cpp:204) -> reporting_sources = {/navigate_to_pose, /bt_navigator}, which the strict-AND scope filter keeps off every entity forever. Nothing ever drops the stale source after that; ReportWithoutSupersedeKeepsBothSources is exactly this pair.

Related corner: one provisional entry covers both codes but reported_canceled_ is last-writer-wins (line 369), so with canceled_is_fault a CANCELED raise + heal + ABORTED raise supersedes only _ABORTED; the healed _CANCELED fault keeps the provisional source, and a later re-raise of it appends the FQN - same permanent hide. So "at most once per fault" (line 574) is really at most once per action.

The provisional source is deterministic (it is the action name), so a stateless option: pass supersedes_source_id = action_name on every FAILED report. Self and absent supersedes are proven no-ops (SelfSupersedeKeepsTheSource, SupersedeAbsentSourceJustAddsNew), and the provisional_/reported_canceled_ bookkeeping goes away. Trade-off to check: report() skips the local filter whenever supersede is set; this bridge pre-declares filtering off, so it only bites if someone re-enables it. Failing that, the restart window belongs in the README limitations.

src/ros2_medkit_action_status_bridge/src/action_status_bridge_node.cpp
```diff
@@ -462,27 +471,117 @@ ros2_medkit_fault_reporter::FaultReporter * ActionStatusBridgeNode::reporter_for
...
+    if (!pit->second.fqn_reporter->is_service_ready()) {
+      continue;  // FaultManager service not discovered by the new client yet; retry
+    }
+
+    // Re-report the same fault (the code actually reported, per reported_canceled_)
+    // under the FQN, superseding the provisional source, so the FaultManager ends
+    // with reporting_sources = {FQN} and the fault resolves to the server node's
+    // SOVD entity under the strict-AND scope filter. The fault is still active, so this
+    // second FAILED report does not inflate occurrence_count (the store counts occurrence
+    // edges, not reports) - it only swaps the source. Happens at most once per fault.
```
mfaferek93, line 574: "It only swaps the source" is true for occurrence_count but not for the rest of the FAILED update this re-report runs through:

- last_occurred and last_failed_time move to reattribution time (fault_storage.cpp:189/192; the sqlite UPDATE binds both), so the fault reads as freshly active seconds after the real abort, and the EVENT_UPDATED emitted for the already-confirmed fault carries those refreshed timestamps.
- the debounce counter decrements. No-op at the default confirmation_threshold -1, but at -2 (the exact value test_fault_source_supersede.test.py configures) one real abort parks the fault PREFAILED at -1 and this bookkeeping report is what drives it to CONFIRMED - stamping confirmed_at and firing the confirmation-edge snapshot capture. The test masks this by sending 4 reports per step. Has to be a global -2: a per-entity override keyed on the FQN never reaches the provisional report.

None of the new tests (either e2e, storage units) assert last_occurred, debounce_counter, or occurrence_count across a supersede. Worth either scoping this comment down to the occurrence_count claim, or teaching the store to skip the timestamp/debounce mutation when a supersede re-report hits an already-active fault.

src/ros2_medkit_integration_tests/test/features/test_fault_source_supersede.test.py
```diff
@@ -0,0 +1,177 @@
...
+    def test_supersede_moves_fault_to_server_entity(self):
+        """Provisional attribution is invisible to the entity; supersede fixes it.
+
+        @verifies REQ_INTEROP_013
+        """
+        # 1. Report under the provisional (action-name) source. It maps to no
+        #    entity, so the strict-AND scope filter keeps it off the owning app.
+        self._report(PROVISIONAL_SOURCE)
+        self._poll_until(
+            lambda: self._fault_status_on_owner() == 404,
```
mfaferek93, line 131: This poll passes on the first probe whether or not step 1 landed: setUpClass already waits for gateway health and discovery, and the scoped detail is 404 both when the fault is absent and when it exists under an out-of-scope source. Replies are dropped too, so a rejected report is invisible. If a regression silently drops provisional-source FAILED reports, the test stays green: step 2 just creates the fault fresh under the FQN (absent supersede is a no-op per SupersedeAbsentSourceJustAddsNew) and the drop-before-add path through the gateway scope filter - the leg this test exists to pin - never runs here. Poll first until the fault exists under /navigate_to_pose (unscoped GET /faults returns all faults and its items carry reporting_sources, or a ListFaults client like test_reattribution_e2e), then assert the scoped 404.

src/ros2_medkit_action_status_bridge/include/ros2_medkit_action_status_bridge/action_status_bridge_node.hpp
```diff
@@ -172,6 +181,17 @@ class ActionStatusBridgeNode : public rclcpp::Node {
   std::map<std::string, std::unique_ptr<ros2_medkit_fault_reporter::FaultReporter>> reporters_;
   std::mutex reporters_mutex_;
 
+  // Actions whose fault was first reported under a provisional (action-name) source
```
mfaferek93, line 184: Two docs just above still state the invariant this PR removes: reporter_for (lines 132-136) says "It is never re-attributed afterwards (reporting_sources is append-only...)" and the reporters_ comment (lines 178-180) says the source_id is "never changed". reattribute_provisional() now swaps the map entry (reporters_[action_name] = std::move(...fqn_reporter), action_status_bridge_node.cpp:580) and supersede is exactly what makes reporting_sources non-append-only. reporter_identity's doc has the same problem (line 275, "never swapped out") now that the test is StickyCreatedOncePerCall. The equivalent comment in test_action_status_bridge.cpp got updated ("it never swaps the source on its own"); update these to match.

src/ros2_medkit_action_status_bridge/test/slow_discovery_bridge_main.cpp
```diff
@@ -0,0 +1,102 @@
...
+      if (n < kForcedUnresolvedCalls) {
+        ++n;
+        return "";  // simulate slow discovery: force the provisional fallback
+      }
+    }
+    // Resolve for real via the public graph API, mirroring the production
+    // server_fqn_for_action / server_fqn_from_endpoint logic.
+    constexpr const char * kUnknownName = "_NODE_NAME_UNKNOWN_";
+    constexpr const char * kUnknownNamespace = "_NODE_NAMESPACE_UNKNOWN_";
+    for (const auto & p : get_publishers_info_by_topic(action_name + "/_action/status")) {
```
mfaferek93, line 72: This reimplements server_fqn_from_endpoint's placeholder filter + namespace join inline (the comment above says as much). A copy drifts silently: production learns a new rcl placeholder or changes the join, this double keeps the old behavior, and the e2e stays green while exercising resolution logic the shipped bridge no longer has. server_fqn_from_endpoint is public static and the unit tests already call it directly, so the loop body can shrink to:

```cpp
const std::string fqn = server_fqn_from_endpoint(p.node_name(), p.node_namespace());
if (!fqn.empty()) {
  return fqn;
}
```

The loop itself has to stay (server_fqn_for_action is a private virtual, so the override can't delegate to the base), but the duplicated filter/join goes away.

src/ros2_medkit_fault_reporter/README.md
```diff
@@ -99,6 +99,22 @@ Report that a fault condition has cleared. Use this when FaultManager is configu
...
+provisional attribution - for example an action bridge that first reports a fault
+under an action interface name, then re-reports under the resolved server node FQN
+once discovery settles. A supersede bypasses local filtering so the correcting
+report always reaches the FaultManager.
+
+```cpp
+reporter_->report("MOTOR_OVERHEAT",
+                  ros2_medkit_msgs::msg::Fault::SEVERITY_ERROR,
+                  "Motor temperature exceeded safe limit",
+                  "/navigate_to_pose");  // provisional source to drop
```
mfaferek93, line 115: This is the reader's first contact with supersede, and the example pairs an unrelated fault and source: it drops /navigate_to_pose from MOTOR_OVERHEAT. That teaches the field as "drop whichever source you like" - and since the manager erases the named source with no ownership check, someone copying this pattern can silently strip another node's legitimate source from a shared fault code. Make the snippet act out the paragraph above it: the same fault code first reported under the provisional source, then re-reported by a reporter constructed with the resolved FQN, passing the provisional source as the fourth argument.

bburda:
Putting this on hold and closing for now, not dropping it.

Changing ReportFault is a breaking change for a released interface package, and
a version mismatch fails silently: a node built against the older msgs still
calls the service, the request does not deserialize, and the fault report is
dropped with no output at default log levels. Handling that properly needs
either a major ros2_medkit_msgs release with clear release notes, or a separate
service so ReportFault stays as it is.

The bug behind this work (an action fault raised before DDS discovery resolves
the server node FQN stays hidden from the owning app entity) does not block us
today, so we do not want to force that decision now.

The branch and all comments here stay. The open points from today are not
dismissed, they are the starting list when we come back. Reopen once we pick the
compatibility approach.

PR #538, opened 2026-07-16
feat: zero-config freeze-frame - capture the faulting entity's own data by default

### LINKED ISSUE #537: Freeze-frame without configuration: capture the faulting entity's own data by default
Today freeze-frame is config-driven only (`fault_specific` / `patterns` / `default_topics`); with no snapshot config a fault carries no freeze-frame at all. For PLC-backed entities (whose live values are not ROS topics) there is no working path even with config.

Proposal: an entity-default capture mode, on by default - when a fault is reported and no explicit snapshot config matches, capture the reporting entity's own current data points (for plugin entities: the plugin's latest polled values; for ROS apps: the entity's own topics) into the fault's environment_data. Zero configuration, every fault gets at-fault-time context out of the box; explicit config keeps overriding when present.

mfaferek93:
Today a fault carries freeze-frame context only when snapshot config matches (`fault_specific` / `patterns` / `default_topics`), and PLC-backed entities have no working path at all. This makes at-fault-time context the default: every confirmed fault carries the faulting entity's own current data, zero config.

Two complementary pieces, both default-ON with an opt-out; explicit snapshot config always wins:

- **ROS-backed entities** (`snapshots.entity_default`): when nothing else matches, capture the topics published by the fault's reporting node (excludes `/rosout`, `/parameter_events`, capped at 16). Empty run writes no row, preserving "no row = nothing captured".
- **Plugin / PLC-backed entities** (`entity_freeze_frame.enabled`): PLC values aren't ROS topics, so the gateway snapshots the entity's current values from the owning plugin (no plugin API change) and merges them into `environment_data.snapshots` as standard `freeze_frame` entries - skipped when a freeze-frame was already captured.

Precedence: `fault_specific` > `patterns` > `default_topics` > entity-default.

Tests: unit coverage for both paths (precedence, opt-out, plugin vs ROS source) plus an end-to-end test with no snapshot config proving a plugin fault carries entity values and a ROS fault carries its node's topics.

Closes #537

### DIFF
--- a/src/ros2_medkit_fault_manager/src/snapshot_capture.cpp
+++ b/src/ros2_medkit_fault_manager/src/snapshot_capture.cpp
@@ -19,6 +19,7 @@
 #include <rclcpp/serialized_message.hpp>
 #include <set>
 #include <thread>
+#include <utility>
 
 #include "ros2_medkit_fault_manager/time_utils.hpp"
 #include "ros2_medkit_serialization/json_serializer.hpp"
@@ -123,6 +124,13 @@ void SnapshotCapture::capture(const std::string & fault_code) {
   }
 
   auto topics = resolve_topics(fault_code);
+  // Zero-config fallback: no explicit config matched, so capture the reporting
+  // source node's own published topics (explicit config always wins above).
+  bool entity_scoped = false;
+  if (topics.empty() && config_.entity_default) {
+    topics = resolve_entity_topics(fault_code);
+    entity_scoped = !topics.empty();
+  }
   if (topics.empty()) {
     // Unconfigured fault code: resolve_topics() yielded nothing, so capture returns early.
     // No freeze_frames row is written (we never persist an empty {} row) and no
@@ -132,15 +140,17 @@ void SnapshotCapture::capture(const std::string & fault_code) {
     return;
   }
 
-  RCLCPP_INFO(node_->get_logger(), "Capturing snapshots for fault '%s' (%zu topics)", fault_code.c_str(),
-              topics.size());
+  RCLCPP_INFO(node_->get_logger(), "Capturing snapshots for fault '%s' (%zu topics%s)", fault_code.c_str(),
+              topics.size(), entity_scoped ? ", entity-default" : "");
 
   // Accumulate a compact dict of captured topic values alongside the per-topic snapshots.
   nlohmann::json freeze_frame = nlohmann::json::object();
   size_t captured_count = 0;
   for (const auto & topic : topics) {
     bool success = false;
-    if (config_.background_capture) {
+    // Entity-default topics are not known at construction, so the background
+    // cache never holds them - always sample those on demand.
+    if (config_.background_capture && !entity_scoped) {
       success = capture_topic_from_cache(fault_code, topic, freeze_frame);
     } else {
       success = capture_topic_on_demand(fault_code, topic, freeze_frame);
@@ -161,6 +171,14 @@ void SnapshotCapture::capture(const std::string & fault_code) {
   // An empty capture must never clobber an existing non-empty frame: a re-confirm while
   // the source publishers are down would otherwise replace the retained record with {}.
   if (captured_count == 0) {
+    // Entity-default runs are best-effort: an empty one must not create a row
+    // (absence still means "nothing configured captured anything") nor clobber
+    // a retained frame from an earlier explicit-config capture.
+    if (entity_scoped) {
+      RCLCPP_DEBUG(node_->get_logger(), "Entity-default capture for fault '%s' sampled nothing; no row written",
+                   fault_code.c_str());
+      return;
+    }
     auto existing = storage_->get_freeze_frame(fault_code);
     if (existing.has_value() && existing->data != "{}") {
       RCLCPP_WARN(node_->get_logger(), "Nothing captured for fault '%s'; keeping previously retained freeze-frame",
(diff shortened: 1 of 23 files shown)


src/ros2_medkit_gateway/src/entity_freeze_frame_capture.cpp
```diff
@@ -0,0 +1,120 @@
...
+  }
+  return values;
+}
+
+void EntityFreezeFrameCapture::on_fault_event(const ros2_medkit_msgs::msg::FaultEvent::ConstSharedPtr & msg) {
+  // Mirror the fault_manager's own capture trigger: freeze-frames are taken
+  // when a fault confirms, not on every update.
+  if (msg->event_type != ros2_medkit_msgs::msg::FaultEvent::EVENT_CONFIRMED) {
+    return;
+  }
```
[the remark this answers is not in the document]
mfaferek93: Retention across EVENT_CLEARED is deliberate, and evicting there would diverge from the fault_manager. Its freeze-frames are explicitly kept across clear_fault (sqlite_fault_storage.cpp: "NOT removed on clear_fault, so the confirmed-state record is retained after acknowledgement"; clear_fault deletes snapshots but not freeze_frames) and store_freeze_frame is INSERT OR REPLACE, i.e. overwrite on every confirm. This class already mirrors that exactly: frames survive clear so a cleared (historical) fault keeps its at-fault context, and every EVENT_CONFIRMED re-samples the plugin, so a re-occurrence never serves the previous incident's values past its own confirm. The window you describe (re-raised fault showing the last confirmed frame until it confirms again) exists identically on the fault_manager side - freeze-frames document the last confirmed state by design. Documented the semantics in the class docstring and pinned them with FramesRetainedAcrossClearAndOverwrittenOnReconfirm (clear retains, re-confirm with changed values overwrites). cab80bc0

src/ros2_medkit_fault_manager/src/snapshot_capture.cpp
```diff
@@ -201,6 +219,77 @@ std::vector<std::string> SnapshotCapture::resolve_topics(const std::string & fau
...
+      std::string name = source;
+      const auto slash = source.rfind('/');
+      if (slash != std::string::npos) {
+        name = source.substr(slash + 1);
+        ns = (slash == 0) ? "/" : source.substr(0, slash);
+      }
+      if (!name.empty()) {
+        wanted.emplace(name, ns);
+      }
+    }
```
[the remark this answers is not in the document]
mfaferek93: Right - a bare plugin id like tank_process parsed as (name, ns="/") and could match an unrelated root-namespace node of the same name, capturing its topics for a plugin fault. Non-FQN sources are now skipped (their zero-config frames come from the gateway-side capture), with an early return when no valid FQN remains. Regression test BarePluginIdNeverMatchesSameNamedNode reports a fault under a bare id equal to the live test node's own name and asserts no freeze-frame row is written. cab80bc0

src/ros2_medkit_gateway/src/entity_freeze_frame_capture.cpp
```diff
@@ -0,0 +1,185 @@
...
+                    msg->fault.fault_code.c_str(), source.c_str(), result.error().message.c_str());
+        continue;
+      }
+      Frame frame;
+      frame.entity_id = source;
+      frame.values = values_from_list_content(result->content);
+      frame.captured_at_ns =
+          std::chrono::duration_cast<std::chrono::nanoseconds>(std::chrono::system_clock::now().time_since_epoch())
+              .count();
+      frames.push_back(std::move(frame));
```
bburda: The DataProvider branch pushes the frame unconditionally - unlike the route path, which gates on `route_content_has_live_data` ("A PLC that is down must not freeze-frame a row of nulls"). So a plugin whose `list_data` returns `items:[]` (no data resources, or a cold poll cache at first-confirm), or items with `id` but no `value` (disconnected), or content with no `items` key at all (returned verbatim by `values_from_list_content`) all produce a `freeze_frame` row of `{}` / nulls / arbitrary content - violating the "no row = nothing captured" invariant that both the fault_manager path and the route path enforce. Gate this push with the same live-data check (reject empty/null values).

src/ros2_medkit_gateway/src/entity_freeze_frame_capture.cpp
```diff
@@ -0,0 +1,185 @@
...
+
+void EntityFreezeFrameCapture::on_fault_event(const ros2_medkit_msgs::msg::FaultEvent::ConstSharedPtr & msg) {
+  // Mirror the fault_manager's own capture trigger: freeze-frames are taken
+  // when a fault confirms, not on every update.
+  if (msg->event_type != ros2_medkit_msgs::msg::FaultEvent::EVENT_CONFIRMED) {
+    return;
+  }
+
+  std::vector<Frame> frames;
+  for (const auto & source : msg->fault.reporting_sources) {
```
bburda: `on_fault_event` runs on the single `Ros2SubscriptionExecutor` worker - the same serial thread that does ALL `/data` topic sampling and every subscribe/unsubscribe. Here it calls plugin code synchronously and with no time bound, looping over every reporting_source: `provider->list_data` (L148) and, on the fallback, the plugin's `x-plc-data` handler in-process (`route_fetcher_`, L92). Today's PLC bridges serve cached poll values so this is fast, but the contract isn't enforced: a plugin whose `list_data`/`x-plc-data` does a slow or live read (and the fallback exists precisely for live-value bridges) blocks the worker for the read's full duration, stalling all `/data` sampling + subscription create/destroy (`run_sync` hits its 5s deadline) and tripping the executor watchdog; it also lengthens shutdown (the worker join blocks on the read). If such a handler re-enters this same executor (`create_typed`/`run_sync`), it self-deadlocks until that 5s deadline. And capture fires at fault-confirm - often exactly when the PLC just became unreachable, when the read is most likely to hang. Consider bounding/offloading the capture (dedicated thread + timeout) or a hard non-blocking cached-snapshot contract for the dispatched handlers.

src/ros2_medkit_gateway/src/entity_freeze_frame_capture.cpp
```diff
@@ -0,0 +1,185 @@
...
+  if (!content) {
+    return std::nullopt;  // not plugin-owned, no x-plc-data route, or handler error
+  }
+  if (!route_content_has_live_data(*content)) {
+    log_fallback_failure_once(fault_code, "entity '" + entity_id + "' x-plc-data reported no live values");
+    return std::nullopt;
+  }
+  Frame frame;
+  frame.entity_id = entity_id;
+  frame.values = values_from_list_content(*content);
```
bburda: `values_from_list_content` is unguarded on the route path (only `route_fetcher_` is inside try/catch). It does `item.value("id", item.value("name", ""))`, and nlohmann's `value()` throws `type_error.302` when the key is present but not a string - so a plugin returning `items:[{"id": 42, ...}]` throws out of `on_fault_event`. The DataProvider branch wraps the equivalent call in try/catch (WARN + drop); here the throw is caught only by the executor's `worker_loop` at DEBUG (invisible by default), the frame is silently lost, and it aborts the current `spin_some` batch - skipping other subscriptions' ready callbacks in that slice. Guard the route path too, or make `values_from_list_content` total (skip non-string ids).

src/ros2_medkit_fault_manager/src/snapshot_capture.cpp
```diff
@@ -123,6 +124,13 @@ void SnapshotCapture::capture(const std::string & fault_code) {
   }
 
   auto topics = resolve_topics(fault_code);
+  // Zero-config fallback: no explicit config matched, so capture the reporting
+  // source node's own published topics (explicit config always wins above).
+  bool entity_scoped = false;
+  if (topics.empty() && config_.entity_default) {
```
bburda: An explicit-but-empty `fault_specific` / `patterns` entry now falls through to entity-default. `resolve_topics` returns `it->second` for a `fault_specific` hit even when the vector is empty (~L202), so `fault_specific: {OVERHEAT: []}` - an explicit "capture nothing for this code" - yields `topics.empty()` here and triggers entity_default, capturing the node's own topics. That inverts the stated "explicit config always wins": the per-fault opt-out becomes a capture. (`default_topics` is guarded by `!empty()`, but `fault_specific`/`patterns` aren't.) Treat a present `fault_specific`/`pattern` key as explicit - skip entity_default when the code matched one, regardless of list emptiness.

bburda (commented):

- The in-process x-plc-data dispatch now invokes a plugin's route handler from the subscription worker, concurrently with a real HTTP GET of the same route. Before this, only httplib threads called it. A handler with unguarded mutable state now races; worth a thread-safety line in the plugin route contract. The plugins_mutex_ re-entrancy itself is correctly handled by copy-then-release.
- fallback_logged_ is only inserted, never evicted, so it grows unbounded with distinct fault codes (churny or synthetic codes on a flaky bridge); and the DataProvider branch WARNs on every re-confirm with no once-guard (the route path dedups). Bound the set or add a once-guard.
- Two safety properties are implicit; pin them with a comment or test: (a) ~EntityFreezeFrameCapture is not a synchronous barrier (subscription_slot_.reset() waits a 30s deadline), so use-after-free safety relies on main.cpp resetting the executor (joining the worker) before the node; a reorder reintroduces a use-after-free. (b) frames_ is keyed by fault_code only; cross-entity isolation holds only because reporting_sources is monotonic and get_fault gates on source-scope; a future per-source clear would let a stale frame leak.
- Gateway entity frames are not evicted on EVENT_CLEARED (only CONFIRMED is handled), while the fault_manager deletes its own snapshots on clear. A cleared-but-retained fault could surface a gateway freeze-frame with no fault_manager counterpart. Confirm whether cleared faults stay queryable.
- Eviction, the gateway opt-out and gating, and the route-dispatch re-entrancy guard are untested (the C++ unit uses a fake fetcher, so the deadlock the design guards against is unproven). The design docs do not mention the new entity-default fallback or the EntityFreezeFrameCapture component; only the config docs and tutorial do.
- Doc: "explicit snapshot config always wins" is not exact; a rosbag-only capture does not suppress the entity freeze-frame (only a freeze_frame does).

src/ros2_medkit_gateway/src/entity_freeze_frame_capture.cpp
mfaferek93: Fixed in ba559d27. Both paths now go through the same frame_from_content gate: content_has_live_data (renamed from route_content_has_live_data since it guards list_data too) plus a values_have_data check that rejects {} / all-null rows on either path. Covered by DataProviderWithoutLiveValuesCapturesNothing (empty items, all-null values, connected:false).

src/ros2_medkit_gateway/src/entity_freeze_frame_capture.cpp
mfaferek93: Agreed, offloaded in ba559d27. on_fault_event now only filters and enqueues (bounded queue, oldest dropped past 64) and a dedicated capture thread makes the provider/route calls, so a slow or live read cannot stall /data sampling or subscription create/destroy, and a handler that re-enters the executor cannot self-deadlock. The destructor joins the thread, and gateway teardown still resets the capture before plugin shutdown, so a hung read delays shutdown instead of racing plugin unload. Also added the thread-safety line to the fetch_entity_data_via_route contract.

src/ros2_medkit_gateway/src/entity_freeze_frame_capture.cpp
mfaferek93: Fixed in ba559d27 - made values_from_list_content total: id/name are read via find() + is_string(), so a non-string id falls back to name or skips the item instead of throwing out of the callback. Test: ValuesFromListContent.NonStringIdOrNameNeverThrows.

src/ros2_medkit_fault_manager/src/snapshot_capture.cpp
mfaferek93: Right, fixed in b00761d8. resolve_topics now reports whether the code hit fault_specific / a pattern (or non-empty default_topics), and capture() skips entity-default whenever it did - so OVERHEAT: [] is an explicit opt-out again. Documented in snapshots.yaml; covered by ExplicitEmptyConfigEntryOptsOutOfEntityDefault for both fault_specific and pattern entries.

bburda (approved):

PR #540, opened 2026-07-16
fix(gateway): non-404 parameter errors win over NOT_FOUND across nodes

### LINKED ISSUE #539: Parameter error aggregation: NOT_FOUND from one node masks timeout/unavailable from another
In `GET /{entity-path}/configurations/{param_name}` the multi-node probe loop in `src/ros2_medkit_gateway/src/http/handlers/config_handlers.cpp` overwrites `last_result` on every per-node failure:

```cpp
  for (const auto & node_info : agg_configs.nodes) {
    auto result = config_mgr->get_parameter(node_info.node_fqn, parsed.param_name);
    if (result.success) {
      ...
    }

    last_result = result;
    auto err = classify_parameter_error(result);
    if (err.status_code != 404) {
      all_not_found = false;
    }
  }
```

So iteration order decides which error the client sees. Example: node A times out (`TIMEOUT`, maps to 503 `x-medkit-ros2-node-unavailable`), node B answers `NOT_FOUND`. `all_not_found` is correctly false, the handler takes the "surface the non-404" branch, but `last_result` holds B's `NOT_FOUND`, so `classify_parameter_error(last_result)` yields 404 and the client is told the parameter does not exist even though the node likely holding it was merely unreachable.

The comment above the loop already states the intent ("a non-404 from any node (e.g. unavailable) wins over a pure not found verdict"), but the code does not implement that precedence.

Expected: when any node fails with a non-404 error, that error is surfaced regardless of node iteration order; a `NOT_FOUND` must never replace a held non-404 failure.

Observed on main at 1269ca61.

mfaferek93:
GET /{entity-path}/configurations/{param} probes every backing node and overwrote last_result on each failure, so node iteration order decided the surfaced error: a NOT_FOUND from one node masked a timeout/unavailable from another and the client saw 404 instead of 503.

Failures now fold through ParameterErrorAccumulator: non-404 errors win and a 404 never replaces a held non-404. New gtest pins the 503 outcome for TIMEOUT plus NOT_FOUND in both iteration orders.

Tested: colcon build + colcon test for ros2_medkit_gateway in the jazzy image (2506 tests, 0 failures);

Closes #539

### DIFF
--- a/src/ros2_medkit_gateway/src/http/handlers/config_handlers.cpp
+++ b/src/ros2_medkit_gateway/src/http/handlers/config_handlers.cpp
@@ -114,16 +114,10 @@ const NodeConfigInfo * find_node_for_app(const std::vector<NodeConfigInfo> & nod
   return nullptr;
 }
 
-/// Error classification result for parameter operations.
-struct ErrorClassification {
-  int status_code;
-  std::string error_code;
-};
-
 /// Map a structured `ParameterErrorCode` to an HTTP status + SOVD error code.
 /// Identical mapping to the legacy `classify_error_code` helper.
-ErrorClassification classify_error_code(ParameterErrorCode error_code) {
-  ErrorClassification result;
+ParameterErrorClassification classify_error_code(ParameterErrorCode error_code) {
+  ParameterErrorClassification result;
 
   switch (error_code) {
     case ParameterErrorCode::NOT_FOUND:
@@ -165,14 +159,14 @@ ErrorClassification classify_error_code(ParameterErrorCode error_code) {
 /// Classify a ParameterResult into HTTP status + error code. Prefers the
 /// structured `error_code` field and falls back to string-matching the legacy
 /// error messages for backward compatibility with older transports.
-ErrorClassification classify_parameter_error(const ParameterResult & result) {
+ParameterErrorClassification classify_parameter_error(const ParameterResult & result) {
   // Prefer structured error code
   if (result.error_code != ParameterErrorCode::NONE) {
     return classify_error_code(result.error_code);
   }
 
   // Fallback to string parsing for backward compatibility
-  ErrorClassification classification;
+  ParameterErrorClassification classification;
   const auto & error_message = result.error_message;
 
   if (error_message.find("not found") != std::string::npos ||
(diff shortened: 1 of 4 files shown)


src/ros2_medkit_gateway/src/http/handlers/config_handlers.cpp
```diff
@@ -314,6 +308,19 @@ void apply_fan_out_observability(dto::ConfigListXMedkit & xm,
 
 }  // namespace
 
+void ParameterErrorAccumulator::add(const ParameterResult & result) {
+  auto err = classify_parameter_error(result);
+  const bool not_found = err.status_code == 404;
+  // Non-404 failures win; a 404 never replaces a held non-404.
+  if (!not_found || all_not_found_) {
+    worst_ = result;
```
bburda: `add` overwrites `worst_`/`classification_` on every non-404, so among two or more distinct non-404 codes the last one folded wins. The class docstring says the surfaced error "does not depend on node iteration order", but that only holds on the 404-vs-non-404 axis. On the read path the transport can return SERVICE_UNAVAILABLE/TIMEOUT (503) from one node and INTERNAL_ERROR/SHUT_DOWN (500) from another, so a partially-broken aggregated entity surfaces 503 or 500 depending on node iteration order - the same order-dependence the fix set out to remove, one level down. No test folds two distinct non-404 codes. Either rank non-404 results by severity (keep the most-retryable / highest, not the latest), or narrow the docstring to "a non-404 beats a 404" and rename `worst()`, which implies a ranking it doesn't do.

bburda (commented):

- The only new test drives ParameterErrorAccumulator in isolation; nothing exercises get_configuration's multi-node branch, so the four unit tests stay green even if the handler wiring regressed to the old overwrite. The precedence only fires on GET /{aggregated-entity}/configurations/{param} (no app: prefix) with two or more backing nodes where one is unavailable and another lacks the parameter. An integration test on that path in both node orders would prove the fix.
- Related: the accumulator and ParameterErrorClassification were lifted into the public handler header only so that test can drive them directly (no other translation unit uses them). Keeping the accumulator in the .cpp anonymous namespace and covering precedence through a handler-level test would tighten the public surface and close the proof gap.
- Semantics (deliberate): there is no signal distinguishing "absent on all reachable nodes, one unreachable" from a bare 503, so a client probing whether a knob exists on a partially-down aggregated entity gets a retryable 503 for a parameter that may be terminally absent.
- Pre-existing, out of scope: list_configurations returns 200 and silently drops a down backing node, while get /{param} for a parameter only on that node now returns 503. And the string fallback in classify_parameter_error (the error_code == NONE path) is dead on the shipped transport; if it were reached, the transport's real "did not respond" / "not currently set" messages do not contain its match substrings, so it would misclassify to 400.

src/ros2_medkit_gateway/src/http/handlers/config_handlers.cpp
mfaferek93: Reworked to a real ranking: 503 (unavailable/timeout) > other 5xx > non-404 4xx > 404, strictly-greater replacement so the first failure folded at the winning rank is kept. The surfaced status is now order-independent across distinct non-404 codes too, and the docstring describes the ranking so worst() means what it says. Added three-node tests in both orders (500+503+404 - the 404/503/500 order fails on the old last-wins fold) plus a 500-vs-400 pair. 591578dd

bburda (approved):

PR #543, opened 2026-07-18
fix(gateway): consistent parameter-config error surface (list vs get) and drop dead classify fallback

### LINKED ISSUE #541: Inconsistent error surface: list_configurations returns 200 when a backing node is down, get_configuration returns 503
GET /{entity}/configurations returns 200 and silently drops a down backing node: in list_configurations (src/ros2_medkit_gateway/src/http/handlers/config_handlers.cpp) per-node failures are only remembered as first_error, and unless every node fails the response is a plain 200 with the unreachable node's parameters missing. There is no partial indicator for local backing nodes (the x-medkit partial flag only covers peer fan-out).

After #540, GET /{entity}/configurations/{param} on the same entity surfaces 503 ERR_X_MEDKIT_ROS2_NODE_UNAVAILABLE when one backing node is down and the others lack the parameter. So the list endpoint reports success in exactly the situation where the single-parameter endpoint reports the node as unavailable - an inconsistent surface between list and get.

Raised in the PR #540 review discussion: https://github.com/selfpatch/ros2_medkit/pull/540

### LINKED ISSUE #542: Dead string-matching fallback in classify_parameter_error misclassifies to 400 if ever reached
classify_parameter_error in src/ros2_medkit_gateway/src/http/handlers/config_handlers.cpp keeps a string-matching fallback for results with error_code == ParameterErrorCode::NONE, documented as backward compatibility with older transports. The shipped Ros2ParameterTransport sets a structured error_code on every failure path, and no other transport exists, so the fallback is dead code.

It is also wrong if ever reached: its substrings do not match the transport's real messages - "Parameter service did not respond for node: ..." matches none of the 503 patterns ("not available", "timed out", "timeout"), and "Parameter not currently set: ..." does not match the 404 patterns, so both would fall through to 400 ERR_INVALID_REQUEST. Either delete the fallback or align it with the real transport messages.

Raised in the PR #540 review discussion: https://github.com/selfpatch/ros2_medkit/pull/540

mfaferek93:
Two follow-ups from the #540 review, both in the parameter-error path of `config_handlers.cpp`.

## #541 - consistent list vs get error surface
`GET /{entity}/configurations` silently dropped a down backing node and returned a plain 200, while `GET /{entity}/configurations/{param}` returns 503 for the same outage. The list side now flags the response `partial` and names the down nodes in `unavailable_nodes` (mirroring the existing peer fan-out `partial`/`failed_peers` convention) instead of returning a shrunken list with no signal. It stays 200 (partial success is legitimate for a collection); the all-nodes-down case already returns 503. Documented in the REST API guide.

## #542 - drop dead string-matching classify fallback
Every failure path on the shipped `Ros2ParameterTransport` sets a structured `error_code`, so the `error_code == NONE` string-matching fallback was dead - and wrong if ever reached (its substrings never matched the real "did not respond" / "not currently set" messages, so failures fell through to 400 invalid-request). The fallback is removed; the classifier is extracted into its own translation unit and a NONE-coded failure now maps to 500 internal-error, never a 400.

## Tests
- New `test_parameter_error_classification` unit test pins the structured mapping and asserts a NONE failure cannot reach a 400 misclassification.
- `test_parameter_error_precedence` gains handler-level cases: one backing node down surfaces `partial` + `unavailable_nodes` in both node orders, with a control asserting no partial flag when every node is reachable.
- colcon build + full gateway unit suite green in the jazzy image

Closes #541
Closes #542


src/ros2_medkit_gateway/src/http/handlers/config_handlers.cpp
```diff
@@ -435,9 +353,14 @@ ConfigHandlers::list_configurations(const http::TypedRequest & req) {
...
           }
         }
-      } else if (first_error.empty()) {
-        first_error = result.error_message;
-        first_error_node = node_info.node_fqn;
+      } else {
+        // Record every failed backing node so a partial list can name what is
+        // missing, and keep the first failure for the all-nodes-down 503.
+        unavailable_nodes.push_back(node_info.node_fqn);
+        if (first_error.empty()) {
```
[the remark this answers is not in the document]
mfaferek93: Good catch, fixed in c782a5bb. The list now classifies each failed node and only puts genuine 503-class failures (SERVICE_UNAVAILABLE/TIMEOUT) into `unavailable_nodes` + `partial`. A non-503 failure (SHUT_DOWN/INTERNAL_ERROR) is a hard error, not an availability gap, so it fails the whole request with its real status (500) rather than getting mislabeled unavailable behind a partial 200 - which keeps the list consistent with GET, which returns 500 for the same codes. Added a test (`ListWithHardErrorFailsRequestNotPartial`) that shuts the manager down so every node returns SHUT_DOWN and asserts a 500.

src/ros2_medkit_gateway/CMakeLists.txt
```diff
@@ -175,6 +175,7 @@ add_library(gateway_ros2 STATIC
   src/fault_manager_paths.cpp
   src/gateway_node.cpp
   src/http/fan_out_helpers.cpp
+  src/http/parameter_error_classification.cpp
   src/http/handlers/auth_handlers.cpp
   src/http/handlers/bulkdata_handlers.cpp
```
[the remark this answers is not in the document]
mfaferek93: Agreed, fixed in c782a5bb. Moved the translation unit to `src/core/http/parameter_error_classification.cpp` so it is picked up by the `src/core/*.cpp` glob and compiled into `gateway_core` - matching its neutral header under `include/.../core/http/` and the sibling core/http TUs (`entity_path_utils.cpp`, `rate_limiter.cpp`). The symbol is now provided by `gateway_core`, so a `gateway_core`-only consumer no longer risks an undefined reference; `gateway_ros2` still gets it via its PUBLIC link to core. Removed the explicit `gateway_ros2` source line. Built + tested green in the Jazzy CI image.

src/ros2_medkit_gateway/src/http/handlers/config_handlers.cpp
```diff
@@ -435,12 +354,32 @@ ConfigHandlers::list_configurations(const http::TypedRequest & req) {
...
+        // unavailable and mask it behind a partial 200.
+        auto cls = classify_parameter_error(result);
+        if (cls.status_code == 503) {
+          unavailable_nodes.push_back(node_info.node_fqn);
+          if (first_error.empty()) {
+            first_error = result.error_message;
+            first_error_node = node_info.node_fqn;
+          }
+        } else if (!hard_error) {
+          hard_error = make_error(
```
bburda: One node failing with a non-503 code now fails the whole list, and the parameters of the nodes that answered are thrown away.

Entity with nodes A and B. A answers. B returns `INTERNAL_ERROR` or `SHUT_DOWN`. Before this change the caller got 200 with A's parameters. Now the caller gets 500 with no data at all. `INTERNAL_ERROR` is set on 11 paths in `Ros2ParameterTransport`, per node, and `SHUT_DOWN` is a normal lifecycle event, so this is not a rare case.

The goal of the change is to stop dropping one node silently. As written it drops every node loudly instead.

`delete_all_configurations` in this same file handles the same shape with a 207 multi-status. Suggest the same idea here: record the failed node next to `unavailable_nodes`, keep the 200, and fail the request only when no node succeeded.

src/ros2_medkit_gateway/src/http/handlers/config_handlers.cpp
```diff
@@ -435,12 +354,32 @@ ConfigHandlers::list_configurations(const http::TypedRequest & req) {
...
+              cls.status_code, cls.error_code, "Failed to list parameters",
+              json{{"details", result.error_message}, {"entity_id", entity_id}, {"failed_node", node_info.node_fqn}});
+        }
       }
     }
 
+    if (hard_error) {
```
bburda: This runs before the `!any_success` check below, so when no node succeeds a 500 wins over a 503.

Entity with node A down (`SERVICE_UNAVAILABLE`, 503) and node B failing internally (500). Nothing succeeds. This returns 500 `ERR_INTERNAL_ERROR`. The old code always returned 503 here.

`GET /{entity}/configurations/{param}` returns 503 for the same outage, because `ParameterErrorAccumulator` uses `severity_rank`, where 503 is rank 3 and other 5xx is rank 2. So the two endpoints now disagree on the same failure, in the opposite direction to the goal of this PR.

Also note the `unavailable_nodes` list built in the loop is discarded on this path, so the 503 node is not named anywhere in the 500 response.

Moving this check after `!any_success` fixes both points.

src/ros2_medkit_gateway/src/http/handlers/config_handlers.cpp
```diff
@@ -454,6 +393,16 @@ ConfigHandlers::list_configurations(const http::TypedRequest & req) {
...
       xm.queried_nodes = queried_nodes;
     }
+    // Some (but not all) backing nodes were unreachable. Mirror the peer
+    // fan-out honesty convention (partial + failed_peers) for local nodes:
+    // flag the response partial and name the unavailable nodes instead of
+    // silently returning a shrunken 200. This keeps the list surface
+    // consistent with GET /{entity}/configurations/{param}, which returns 503
+    // ERR_X_MEDKIT_ROS2_NODE_UNAVAILABLE for the same backing-node outage.
+    if (!unavailable_nodes.empty()) {
+      xm.partial = true;
```
bburda, line 402: This signal is lost when one gateway aggregates another. That is the setup where it matters most.

Gateway A fans out to peer B. B has a down backing node and correctly answers 200 with `partial: true` and `unavailable_nodes`. In `fan_out_get`, a 200 from a peer is treated as success and only `response_json["items"]` is copied, so the peer's `x-medkit` block is dropped. `fan_out.partial` stays false, and A returns a shrunken list with no `partial` and no `unavailable_nodes`.

Note the asymmetry. If all of B's nodes are down, B answers 503, the peer call fails, and A does flag `partial` and `failed_peers`. So the total outage propagates and the partial one stays silent.

Carrying the peer's `partial` and `unavailable_nodes` through the merge would close this.

src/ros2_medkit_gateway/src/http/handlers/config_handlers.cpp
```diff
@@ -435,12 +354,32 @@ ConfigHandlers::list_configurations(const http::TypedRequest & req) {
...
+      } else {
+        // Only a genuine availability failure (503: node down/unresponsive)
+        // makes the node "unavailable" and the list merely partial, mirroring
+        // the 503 the single-parameter GET returns for the same outage. A
+        // non-503 failure (shutdown/internal) is a hard error, not an
+        // availability gap: surface it rather than mislabel the node
+        // unavailable and mask it behind a partial 200.
+        auto cls = classify_parameter_error(result);
+        if (cls.status_code == 503) {
+          unavailable_nodes.push_back(node_info.node_fqn);
```
bburda, line 363: `unavailable_nodes` can name a node that is healthy.

`cache_default_values` catches any exception and returns `true`. `list_parameters` maps that return value unconditionally to `SERVICE_UNAVAILABLE` with the message "Parameter service not available for node". So a gateway-side bug while caching defaults is reported to the client as "this node is down".

That is the same mislabelling the comment below this block argues against for the non-503 case.

Worth separating the "defaults caching failed" case from the real availability case in the transport, so the new list only names nodes that really did not answer.

src/ros2_medkit_gateway/src/http/handlers/config_handlers.cpp
```diff
@@ -454,6 +393,16 @@ ConfigHandlers::list_configurations(const http::TypedRequest & req) {
     if (!queried_nodes.empty()) {
       xm.queried_nodes = queried_nodes;
     }
+    // Some (but not all) backing nodes were unreachable. Mirror the peer
+    // fan-out honesty convention (partial + failed_peers) for local nodes:
+    // flag the response partial and name the unavailable nodes instead of
+    // silently returning a shrunken 200. This keeps the list surface
+    // consistent with GET /{entity}/configurations/{param}, which returns 503
+    // ERR_X_MEDKIT_ROS2_NODE_UNAVAILABLE for the same backing-node outage.
+    if (!unavailable_nodes.empty()) {
```
bburda: This block is only reachable for entities backed by more than one node.

`get_app_configurations` produces at most one `NodeConfigInfo`. With one node, a failure means `any_success` is false and the request returns 503 from the branch above, so `partial` is never set. The same holds for any Component, Area or Function that resolves to a single app.

So on `/apps/{id}/configurations` the new signal can never appear. The docs added in this PR say `GET /{entity}/configurations` without that qualification, and the new tests cover components only.

Either say "entities with more than one backing node" in the docs, or add a single-node case so the behaviour is pinned.

docs/api/rest.rst
```diff
@@ -2530,6 +2530,26 @@ response), the response includes vendor metadata indicating partial results:
...
+     "items": [],
+     "x-medkit": {
+       "partial": true,
+       "unavailable_nodes": ["/down_node"]
+     }
+   }
+
+This matches ``GET /{entity}/configurations/{param}``, which returns ``503``
+``x-medkit-ros2-node-unavailable`` for the same backing-node outage. When every
+backing node responds, both fields are omitted.
```
bburda: This is not correct for `partial`. Only `unavailable_nodes` is omitted when every backing node responds.

`partial` is also set by the peer fan-out path in `apply_fan_out_observability`, which runs after the local node block and does not depend on it. So `partial: true` with no unavailable backing nodes is a normal response.

Suggest: "When every backing node responds, `unavailable_nodes` is omitted."

Related: `partial` now appears without `failed_peers`, while the DTO comment still defines `partial` as "true when a fan-out peer request failed" and every other producer sets the two together. A client that reads `partial` and then enumerates `failed_peers` will find nothing. The DTO comment needs updating too.

docs/api/rest.rst
```diff
@@ -2530,6 +2530,26 @@ response), the response includes vendor metadata indicating partial results:
 When all peers respond successfully, these fields are omitted. See the
 :doc:`aggregation configuration guide </config/aggregation>` for setup details.
 
+``GET /{entity}/configurations`` applies the same honesty to its *local*
+backing nodes. When an entity is backed by several ROS 2 nodes and some are
+reachable while others are down, the response stays ``200`` but flags itself
+``partial`` and names the unreachable nodes in ``unavailable_nodes`` rather than
```
bburda: The new text documents only the 503 case and says the list stays 200. Two other behaviour changes from this PR are not documented anywhere.

First, a backing node failing with a non-availability code now fails the whole request with the classified status, so this route can return 500. It could not do that before.

Second, when the configuration manager is shut down, every node returns `SHUT_DOWN` and the route now returns 500 instead of the 503 it returned before. The new test pins this, so it is intended, but it is a status change for clients. 503 is also the more useful status for a shutdown, because it tells the client the call is worth retrying.

The route entry earlier in this file lists no status codes at all, so there is no other place a reader would find this.

docs/api/rest.rst
```diff
@@ -2530,6 +2530,26 @@ response), the response includes vendor metadata indicating partial results:
...
+silently returning a shrunken list:
+
+.. code-block:: json
+
+   {
+     "items": [],
```
bburda: The example shows an empty `items` array, but the text above describes a case where some nodes are reachable and their parameters are returned. A reachable node contributes its parameters, so `items` is not empty here.

This looks copied from the peer fan-out example above. Suggest showing one parameter entry so the example matches the paragraph.

src/ros2_medkit_gateway/test/test_parameter_error_precedence.cpp
```diff
@@ -271,3 +297,56 @@ TEST_F(ParameterErrorPrecedenceTest, AllNodesMissingParameterReturns404) {
...
+  EXPECT_FALSE(xm.partial.has_value() && *xm.partial);
+  EXPECT_FALSE(xm.unavailable_nodes.has_value());
+}
+
+// A non-availability hard failure (manager shut down -> SHUT_DOWN on every
+// node) must fail the list with a 500 internal-error, not mislabel the nodes
+// "unavailable" behind a partial 200. Guards the list-vs-GET consistency:
+// GET surfaces 500 for the same code, so the list must not swallow it.
+TEST_F(ParameterErrorPrecedenceTest, ListWithHardErrorFailsRequestNotPartial) {
+  gateway_node_->get_configuration_manager()->shutdown();
```
bburda, line 360: This test cannot pin the behaviour it is named for. `shutdown()` sets a manager-wide flag, so every node returns `SHUT_DOWN` and `any_success` is false.

That means the test passes both with the current `if (hard_error)` and with `if (hard_error && !any_success)`. Both give 500.

The behaviour this PR actually introduces is that one node failing hard while another one succeeds turns a 200 into a 500. That case has no test. It is reachable on the shipped transport, because `list_parameters` sets `INTERNAL_ERROR` per node.

A component with one succeeding node and one hard-failing node would cover it.

src/ros2_medkit_gateway/test/test_parameter_error_precedence.cpp
```diff
@@ -271,3 +297,56 @@ TEST_F(ParameterErrorPrecedenceTest, AllNodesMissingParameterReturns404) {
...
+TEST_F(ParameterErrorPrecedenceTest, ListWithUnavailableNodeFirstIsPartial) {
+  auto xm = list_xmedkit("stack_unavail_first");
+
+  ASSERT_TRUE(xm.partial.has_value());
+  EXPECT_TRUE(*xm.partial);
+  ASSERT_TRUE(xm.unavailable_nodes.has_value());
+  EXPECT_NE(std::find(xm.unavailable_nodes->begin(), xm.unavailable_nodes->end(), kGhostFqn),
+            xm.unavailable_nodes->end())
+      << "unavailable node list must name the down node";
+  // The reachable node's parameters are still listed (queried_nodes non-empty).
```
bburda: The comment says the reachable node's parameters are still listed, but the assertion only proves that some node succeeded. `queried_nodes.push_back` happens on `result.success`, before and independently of the parameter merge, and `list_xmedkit()` drops the items, so no test here looks at the merged list.

This assertion is also missing from `ListWithUnavailableNodeLastIsPartial` below, and that is the order that actually stresses accumulation, since the successful node comes first.

Assert on the returned items instead, and add the same check to both orders.

src/ros2_medkit_gateway/src/core/http/parameter_error_classification.cpp
```diff
@@ -0,0 +1,76 @@
...
+    case ParameterErrorCode::INVALID_VALUE:
+      result.status_code = 400;
+      result.error_code = ERR_INVALID_PARAMETER;
+      break;
+    case ParameterErrorCode::NONE:
+      // A failed ParameterResult that still carries NONE cannot occur on the
+      // shipped transport (every failure path sets a structured code), so a
+      // NONE failure is a gateway-side defect: surface it as 500 internal-error.
+      // The removed legacy string-matching fallback misrouted these to 400
+      // invalid-request because its substrings never matched the real transport
```
bburda: "never matched the real transport messages" is too broad. The transport also emits "Parameter service not available for node: " and "Parameter not found: ", and the removed fallback matched both and classified them correctly, to 503 and 404.

Across the transport's error messages the fallback would have agreed with the structured mapping for those, and disagreed for the rest. The two messages named here are real examples of the disagreement, so the conclusion is fine, but the reason as written is not accurate.

Suggest: "the fallback was unreachable, and would have misrouted messages such as these to 400 if it had been reached".

src/ros2_medkit_gateway/test/test_parameter_error_classification.cpp
```diff
@@ -0,0 +1,111 @@
...
+
+// Regression pin for issue #542: the removed legacy string-matching fallback
+// routed a failure carrying ParameterErrorCode::NONE to 400 ERR_INVALID_REQUEST
+// whenever its substrings failed to match the real transport messages (which is
+// always - the shipped transport uses "did not respond" / "not currently set",
+// none of which the fallback recognised). Such a NONE failure cannot occur on
+// the shipped transport, so if one ever does it is a gateway-side defect and
+// must surface as 500 internal-error, never as a 400 that blames the client.
+TEST(ParameterErrorClassification, NoneFailureIsInternalErrorNot400) {
+  // A message the old fallback would have misrouted to 503, still NONE-coded.
```
bburda: This message would not have gone to 503. The old 503 branch matched only "not available", "service not available", "timed out" and "timeout". "Parameter service did not respond for node" contains none of them, so it fell to the `else` and produced 400.

That makes this case identical to the one below it, which is already labelled as the 400 case, so the two cases do not cover two different misroutings as the comments suggest.

Fix the comment, and if you want a real second case, use a message the fallback did match, for example "Parameter service not available for node".

src/ros2_medkit_gateway/include/ros2_medkit_gateway/dto/config.hpp
```diff
@@ -101,6 +101,7 @@ struct ConfigListXMedkit {
   std::optional<nlohmann::json> parameters;  // free-form: array of raw param JSON
   std::optional<std::vector<std::string>> source_ids;
   std::optional<std::vector<std::string>> queried_nodes;
+  std::optional<std::vector<std::string>> unavailable_nodes;
```
bburda, line 107: The field list comment above this block was not updated. `unavailable_nodes` is missing from it, and it is also missing from the second enumeration further down in this file.

The same comment still says `partial` means "true when a fan-out peer request failed". After this PR `partial` is also set for a purely local backing node outage, with no peers involved, so that line is now wrong.

src/ros2_medkit_gateway/src/http/handlers/config_handlers.cpp
mfaferek93: Fixed in a00898e. The list no longer fails on a non-503 node: every failure folds through the ParameterErrorAccumulator, the answering nodes' parameters are kept, the response stays 200 partial, and the failed node is named - 503-class in `unavailable_nodes`, anything else in a new `failed_nodes` field. Only when no node answers does the request fail. Same shape as `delete_all_configurations`' partial handling, as you suggested.

src/ros2_medkit_gateway/src/http/handlers/config_handlers.cpp
mfaferek93: Fixed in a00898e. The `if (hard_error)` early-return is gone; failures now fold through `ParameterErrorAccumulator` and are surfaced only in the `!any_success` branch via its `severity_rank`, so an all-down mix of 503 + 500 returns 503, matching GET. The error detail now also carries `unavailable_nodes`/`failed_nodes`, so the 503 node is named on that path.

src/ros2_medkit_gateway/src/http/handlers/config_handlers.cpp
mfaferek93: Declining here as a separate follow-up (not disputing the gap). This is the generic peer-merge dropping the peer's `x-medkit` sub-block - it predates this PR and hits every collection endpoint (faults/logs/data all merge only `items`, never carry a peer's `partial`/`unavailable_nodes`). Closing it means threading peer `x-medkit` through `fan_out_collection`/`FanOutResult` in the shared aggregation layer with its own tests, which is out of scope for this local-node error-surface change. Worth its own issue.

src/ros2_medkit_gateway/src/http/handlers/config_handlers.cpp
mfaferek93: Declining here as a separate transport follow-up. The mislabel is real, but its root cause is in `Ros2ParameterTransport` (`cache_default_values` returns `bool`; `list_parameters` maps `true` -> `SERVICE_UNAVAILABLE`), reached only via the defensive catch documented as 'no round trip is expected to throw now'. Separating 'defaults caching failed' from genuine unavailability is a transport-level code change across all `cache_default_values` callers (get/set/reset), orthogonal to this error-surface PR. Filing as a follow-up rather than widening scope.

src/ros2_medkit_gateway/src/http/handlers/config_handlers.cpp
mfaferek93: Fixed in a00898e (docs). The REST guide now says this only arises for entities backed by more than one ROS 2 node, and that a single-node entity has no partial state - its one node either answers or the whole request fails. Left the tests on the multi-node path since single-node cannot reach partial by construction.

docs/api/rest.rst
mfaferek93: Fixed in a00898e. Docs now say only `unavailable_nodes`/`failed_nodes` are omitted when every backing node responds, and note `partial` may appear on its own from the peer fan-out. The DTO comment in `config.hpp` is updated too: `partial` is now 'true when a fan-out peer request failed OR a local backing node did not answer', and `unavailable_nodes`/`failed_nodes` are added to both field enumerations.

docs/api/rest.rst
mfaferek93: Fixed in a00898e (docs). The guide now documents that this route can return 500 (a non-availability per-node failure surfaces its classified status), and that an all-nodes-shutdown outage returns 500 where it previously returned 503 - calling out the status change for clients.

docs/api/rest.rst
mfaferek93: Fixed in a00898e. The example now shows a reachable node's parameter (`use_sim_time`) in `items`, matching the paragraph, instead of the empty array copied from the peer example.

src/ros2_medkit_gateway/test/test_parameter_error_precedence.cpp
mfaferek93: Addressed in a00898e, though not the way the test name implied. The redesign removes the 200->500 conversion entirely: one node failing while another answers now stays a 200 partial and keeps the answering node's parameters (now asserted via `collection.items` in both `ListWith*NodeIsPartial` cases). The shutdown test is kept as the 'no node answered -> surface the hard error, not a partial 200' pin, comment rewritten to say so. The specific mixed hard-fail-plus-success case can't be injected in this live-node harness - shutdown is manager-wide and the ghost/unspun nodes only yield 503, so there is no per-node non-503 fault hook; I noted that limitation in the test comment.

src/ros2_medkit_gateway/test/test_parameter_error_precedence.cpp
mfaferek93: Fixed in a00898e. Added a `list_collection()` helper; both `ListWithUnavailableNode{First,Last}IsPartial` now assert `collection.items` is non-empty, proving the answering node's parameters survive the partial - in both orders, including the last-order case that stresses accumulation.

src/ros2_medkit_gateway/src/core/http/parameter_error_classification.cpp
mfaferek93: Fixed in a00898e. Reworded to 'the fallback was unreachable, and would have misrouted messages it did not recognise (e.g. "did not respond", "not currently set") to 400 had it ever run' - no longer claiming it never matched any real transport message.

src/ros2_medkit_gateway/test/test_parameter_error_classification.cpp
mfaferek93: Fixed in a00898e. Swapped the first case to "Parameter service not available for node" (which the old fallback DID match -> 503), so the two cases now cover two genuinely different misroutings (503 vs 400), and corrected the inline comments. Also tightened the block comment above to drop the 'never matched (which is always)' overclaim.

src/ros2_medkit_gateway/include/ros2_medkit_gateway/dto/config.hpp
mfaferek93: Fixed in a00898e. Both field-list comments in `config.hpp` now include `unavailable_nodes` and `failed_nodes`, and the `partial` definition is corrected to 'true when a fan-out peer request failed OR a local backing node did not answer' instead of the peer-only wording.

bburda (approved):

PR #546, opened 2026-07-21
feat(graph_provider): rework x-medkit-graph health model (#545)

### LINKED ISSUE #545: [BUG] x-medkit-graph reports wrong pipeline health
# Bug report

The `x-medkit-graph` resource reports wrong pipeline health. A healthy system can be reported as broken, and a dead one as healthy. Several other defects in the same plugin are listed below, because they share one root cause and one fix.

### Steps to reproduce

1. Build and run the gateway with the `graph_provider` plugin loaded (this is the default in `gateway.launch.py`).
2. Run any node that publishes `/diagnostics`. For example `greenwave_monitor`, configured to watch some but not all topics of one SOVD Function.
3. `GET /api/v1/functions/{id}/x-medkit-graph`

### Expected behavior

Topics that no producer watches are reported as not measured. They should not change the health of the whole Function.

### Actual behavior

`pipeline_status` is `broken`. Every edge without matching metrics reports `metrics_status: "error"` with `error_reason: "no_data_source"`.

The cause is a global flag. `diagnostics_seen_` is set to `true` when any `/diagnostics` message arrives, even one with no readable status. After that, every edge that has no metrics is treated as an error.

### Environment

- ros2_medkit version: 0.6.0
- ROS 2 distro: all supported (humble, jazzy, lyrical)
- OS: Ubuntu 24.04

### Additional information

**Why no test caught this.** The demo nodes contain only two subscriptions, both on `command` topics that nothing publishes. An edge needs a publisher and a subscriber that share a topic inside one Function. So `edges` is always empty in every test, and `pipeline_status` is always `healthy`. The integration test only checks that `edges` is a list. The feature has never been exercised.

**Other defects found in the same plugin.**

Health and metrics:

- Metrics carry no timestamp. Old data cannot be told apart from current data, so a producer that stops is never noticed.
- For messages with no header, `greenwave_monitor` reports `frame_rate_msg: 0.000000` and puts the real rate in `frame_rate_node`. The plugin reads only `frame_rate_msg`, so it reports 0 Hz as `active` and forces `degraded`.
- `frame_rate_msg` is not checked for sign. A negative value, which some producers use to mean "not measured yet", gives a negative ratio. That ratio always wins the bottleneck comparison and hides the real bottleneck.
- One sample that cannot be parsed overwrites good metrics for that topic, because the callback replaces instead of merging.
- Metrics with latency or drop rate but no frequency are reported as `no_data_source`, while the same object still carries the values.
- A Function whose apps have all disappeared returns zero nodes and zero edges and reports `healthy`.

Config:

- Per-function overrides never apply. `load_parameters` expects flat keys like `function_overrides.<id>.<field>`, but the gateway delivers nested objects. Both guard conditions reject them.
- Config values are not validated. Setting `expected_frequency_hz_default: 0` silently disables all degradation detection.

Resources and dead code:

- The cyclic subscription sampler registers a lambda that captures the plugin pointer, and the sampler registry has no way to remove it. If the plugin is unloaded, the registry keeps a dangling callback.
- `last_seen_by_app_` never drops entries and is copied in full on every request and every sampler tick.
- `graph_cache_` is written but never read. `introspect()` builds a full document for every Function on every discovery cycle into a map that nobody reads.
- Stale topic detection never runs. Both callers of `build_state_snapshot` pass `false`, so `collect_stale_topics` and the fault correlation behind it are unreachable.
- `metrics.source` is a fixed string and does not depend on which node published the data.

Docs that do not match the code:

- README gives the wrong `.so` name. The built artifact is `libros2_medkit_graph_provider_plugin.so`.
- README gives the config key `expected_frequency_hz`. The code reads `expected_frequency_hz_default`.
- README says the plugin detects stale topics. That code never runs.
- The design doc says the metrics cache uses LRU eviction. It is FIFO.
- The design doc says topic IDs are stable. They are positional and renumber when the topic set changes.

## Proposed fix

Add a timestamp to the cached metrics and build the status model on freshness:

| Status | Meaning |
|---|---|
| `pending` | no metrics ever seen for this topic |
| `active` | metrics present and current |
| `error` with `metrics_stale` | metrics present but older than the freshness window |

The freshness window comes from the resolved expected frequency, with headroom and a floor of a few seconds.

`no_data_source` and the global `diagnostics_seen_` flag are removed. `error_reason: node_offline` is also removed from edges. It cannot happen today, because an app that is offline has an empty topic list and so produces no edges at all. Freshness covers that case instead. Node level `node_status: unreachable` is correct and stays.

Work also needed before any of this can be verified: the demo node set needs a subscriber, so that at least one edge exists.

This changes the response format, so `schema_version` will be raised and the meaning of the field will be documented.

### LINKED ISSUE #547: Robustness and operator controls for x-medkit-graph health
## Summary

Fixing the wrong-health report surfaced that the `x-medkit-graph` health signal
is not robust and gives the operator no controls. This tracks the enhancement
work that goes beyond the bug fix itself.

Gaps:

- Freshness age was computed on a wall clock, so a backward clock step (for
  example an NTP correction on a robot that boots before NTP sync) could hide a
  dead topic as fresh.
- A single late `/diagnostics` sample (DDS or executor jitter) could flip the
  pipeline to `broken` and back, with no debounce.
- `frame_rate_msg` is a topic-level rate summed over all publishers on a topic.
  A duplicate or leftover publisher inflates it, so a slow producer can read as
  healthy, and nothing told the operator the rate was suspect.
- There was no way to tune any of this per deployment.

## Proposed solution

- Compute freshness age on a monotonic clock; keep the human-readable timestamp
  on wall time.
- Add `stale_grace_sec` (per-plugin and per-function): report `metrics_stale`
  only after the age has been past the window continuously for the grace, so a
  single late sample does not flap. `0` keeps the old behaviour.
- Report `publisher_count` and `rate_ambiguous` on each edge, and add
  `multi_publisher_rate` (`annotate` / `suppress`) so a deployment can choose to
  keep the flagged rate or hide the untrustworthy number.
- Cover the behaviour with integration tests that drive a real diagnostics
  producer through the gateway.

## Additional context

Builds on the wrong-health bug fix (#545). Both are addressed in the same pull
request.

bburda:
## Summary

The `x-medkit-graph` resource reported wrong pipeline health. It could show a
healthy system as `broken`, and a broken one as `healthy`. The feature had
never produced a real graph edge in tests, because no demo publisher and
subscriber shared a topic, so many defects were never caught.

This PR replaces the status model with one based on data freshness, and fixes
the defects found once a real edge existed.

Key changes:

- Status model is now `pending` / `active` / `error` (`metrics_stale`), driven
  by whether metrics are present and fresh. The old global "diagnostics seen"
  flag that turned any unmatched edge into `broken` is gone.
- Freshness age uses a monotonic clock, so a backward clock step cannot mask a
  dead topic as fresh. A configurable `stale_grace_sec` debounces a single
  late `/diagnostics` sample so the pipeline does not flap.
- `frame_rate_msg` from `greenwave_monitor` is the arrival rate summed across
  all publishers on a topic. A duplicate or leftover publisher inflates it and
  could hide a slow producer. Each edge now reports `publisher_count` and
  `rate_ambiguous`, and a new config `multi_publisher_rate` (`annotate` /
  `suppress`) controls how the plugin reacts.
- `metrics.source` is the resolved publisher node name; it is cleared when the
  latest sample cannot be attributed, instead of showing a stale or hardcoded
  value.
- Per-function config overrides now apply (they were silently dropped), config
  values are validated, the topic filter matches path segments not substrings,
  `last_seen_by_app` is bounded, and a dead stale-topic code path plus an
  unused cache are removed.
- A cyclic-subscription sampler could be called after its plugin was torn
  down. The sampler registry now removes a plugin's samplers on teardown and
  drains any in-flight call.

`schema_version` is `2.0.0`.

## Issue

- closes #545 (the wrong-health bug)
- closes #547 (robustness and operator controls)

## Type

- [x] Bug fix
- [x] New feature or tests
- [x] Breaking change
- [ ] Documentation only

Breaking: `error_reason` values changed (only `metrics_stale` remains reachable
on an edge), `metrics.source` can now be omitted, new optional fields were
added, and `pipeline_status` reporting changed. Deployments that previously saw
`broken` from the old bug will now report `healthy` or `degraded`.

## Testing

- Unit tests for the freshness math, config validation, the topic filter, GID
  based source resolution, and the rate policies.
- A new integration suite drives a real `greenwave_monitor` node feeding real
  `/diagnostics` through the gateway HTTP endpoint: metrics path, staleness,
  cyclic subscriptions over SSE, and function scoping.
- Reviewers can run the metrics suite with the graph provider plugin loaded and
  a `greenwave_monitor` node, then GET
  `/api/v1/functions/{id}/x-medkit-graph` and check `pipeline_status` follows
  the producer state.
- Docs build clean under `-W`.

## Checklist

## Known limitations / follow-ups

- Two separate `/diagnostics` producers reporting the same topic can mix field
  attribution on one edge. Uncommon setup; documented.
- Per-function overrides are set via a params file (`config_file`), not a bare
  `-p` CLI argument, because the ROS parameter CLI cannot express a nested key
  with a hyphen in the function id. Documented in the config reference.
- A standard `diagnostic_updater` producer uses different diagnostic keys and
  names the status after the task, not the topic, so it will not match. The
  producer contract (greenwave-style keys, `DiagnosticStatus.name` = the
  fully-qualified topic name) is documented.
- A separate follow-up is needed for the client tool that consumes this
  endpoint; it currently parses a document shape this plugin does not emit.

- A pre-existing TSan data race in the trigger subscription teardown path
  (unrelated to this change) surfaces on the sanitizer job. It is fixed in
  #548, which lands first; this PR then rebases onto that fix.


bburda:
@eclipse0922 you wrote the original x-medkit-graph provider, so your review here would be valuable, especially on the greenwave_monitor interop.

Short summary of what this adds to the plugin:

- Pipeline status is now based on data freshness. Before, any `/diagnostics` traffic could turn an edge without matching metrics into an error and make the whole function read as `broken`. Now an edge is broken only when its own metrics actually went stale.
- `greenwave_monitor` stays the reference producer. The plugin now handles a header-less topic (it falls back to `frame_rate_node`) and resolves `metrics.source` to the real publisher node name instead of a fixed string.
- New operator controls: `stale_grace_sec` debounces a single late sample so the pipeline does not flap, and `multi_publisher_rate` (`annotate` / `suppress`) handles the case where more than one publisher shares a topic. `frame_rate_msg` is a topic-level rate summed over all publishers, so a duplicate or leftover publisher inflates it.
- New per-edge fields `publisher_count` and `rate_ambiguous` so the operator can see when the rate may be inflated.
- An integration suite now drives a real `greenwave_monitor` node through the gateway.

The producer contract (greenwave-style keys, `DiagnosticStatus.name` set to the fully-qualified topic name) is documented in the tutorial. If anything about the greenwave side looks wrong to you, please flag it.

eclipse0922:
Thanks for catching the major blind spot in my original implementation, @bburda! Missing real graph edges in the initial test suite was definitely a miss on my end, and refactoring the status model around per-edge data freshness is a much cleaner and more robust approach.

A few quick notes on the review:

Freshness & Debouncing: Moving away from the global diagnostic flag to per-edge freshness with stale_grace_sec makes total sense. Debouncing late samples will stop a lot of status flapping in real deployments.

Publisher Resolution: Handling header-less topics with the frame_rate_node fallback and dynamically resolving metrics.source nicely fixes the stale attribution issue.

Multi-publisher Handling: Exposing publisher_count and rate_ambiguous alongside multi_publisher_rate controls provides great operational visibility for aggregated rates.

Producer Contract: The contract requirements (greenwave-style keys, FQTN in DiagnosticStatus.name) line up well with what greenwave producers expect.

Overall, this redesign fixes the core architectural flaws while making the interop much more reliable. Looks great!

src/ros2_medkit_plugins/ros2_medkit_graph_provider/src/graph_provider_plugin.cpp
```diff
@@ -286,8 +444,8 @@ nlohmann::json build_graph_document_for_apps(const std::string & function_id,
         const auto metrics_it = state.topic_metrics.find(topic_name);
         const GraphProviderPlugin::TopicMetrics * metrics =
             metrics_it == state.topic_metrics.end() ? nullptr : &metrics_it->second;
-        auto edge =
-            build_edge_json(edge_id, *publisher, *subscriber, topic_name, topic_it->second, metrics, state, config);
+        auto edge = build_edge_json(edge_id, *publisher, *subscriber, topic_it->second, metrics, publisher_count, state,
+                                    config);
 
         has_errors = has_errors || edge.is_error;
```
mfaferek93, line 453: `pipeline_status` is rolled up from edges only (`has_errors`/`has_degraded`), and an offline app carries an empty topic list, so it produces an `unreachable` node but no edge. If the node that dies is a subscriber, every edge into it disappears and the remaining edges can all be `active` -> `pipeline_status: "healthy"` while a critical node reads `unreachable`. The rework rightly stopped conflating "never observed" with "broken", but this swings the other way: a Function with a dead consumer reports healthy. Worth pinning the intended contract - either fold "any scoped node unreachable" into a non-healthy status, or document that `pipeline_status` reflects only observed data flow and node reachability must be read separately.

src/ros2_medkit_plugins/ros2_medkit_graph_provider/src/graph_provider_plugin.cpp
```diff
@@ -151,73 +227,145 @@ double resolve_expected_frequency(const GraphProviderPlugin::TopicMetrics * metr
...
+    if (rate_ambiguous) {
+      metrics_json["rate_ambiguous"] = true;
+    }
+  }
+
   if (metrics) {
-    if (metrics->frequency_hz.has_value()) {
+    // Under kSuppress with more than one live publisher, the measured rate is
+    // untrustworthy (it may be inflated by a leftover/duplicate publisher) -
+    // omit it rather than let a genuinely broken pipeline read as healthy.
```
mfaferek93: The comment says the rate is omitted "rather than let a genuinely broken pipeline read as healthy", but omitting it doesn't change the verdict. With the frequency branch skipped, the edge is still `metrics_status: active` and contributes neither `degraded` nor `error` (unless drop_rate trips on its own), so `pipeline_status` stays "healthy" regardless. `suppress` removes the misleading number but not the misleading health - an operator who picks it to avoid a broken pipeline masquerading as healthy still gets healthy. Either downgrade an ambiguous suppressed edge in the verdict so the policy actually affects `pipeline_status`, or drop the "rather than let it read as healthy" claim from the comment.

src/ros2_medkit_plugins/ros2_medkit_graph_provider/src/graph_provider_plugin.cpp
bburda: Pinned it to a non-healthy status. `pipeline_status` now folds node reachability in: any unreachable scoped node makes it at least `degraded` (a stale edge still outranks it as `broken`). A dead node carries no topics so it contributes no edge, but `node_status` names which one. Added a unit test, an HTTP-handler test, and an end-to-end launch test that kills a manifest-scoped node and asserts the served graph goes from healthy to degraded.

src/ros2_medkit_plugins/ros2_medkit_graph_provider/src/graph_provider_plugin.cpp
bburda: Dropped the claim - the comment now says suppress removes the misleading number, not the health verdict. Multiple publishers are not by themselves a failure and `rate_ambiguous` already flags the ambiguity, so I kept the verdict unchanged rather than downgrading a suppressed edge. Fixed the same overclaim in the config and design docs, which said suppress was for avoiding a false healthy reading.

mfaferek93 (approved):

PR #549, opened 2026-07-22
fix(gateway): route trigger subscriptions through the shared subscription executor

### LINKED ISSUE #548: [BUG] TSan data race: trigger topic subscription callback races shutdown teardown
# Bug report

### Steps to reproduce
1. Build with `-DSANITIZER=tsan` (RelWithDebInfo) and run the integration tests.
2. `test_triggers_data` intermittently fails: the gateway process exits with code 66 (TSan detected a data race). It is nondeterministic and has been seen on `main` as well.

### Expected behavior

No data race; the gateway shuts down cleanly.

### Actual behavior

TSan reports a data race during shutdown. A trigger topic subscription callback runs on a main-executor thread (deserializing a message via `JsonSerializer::deserialize` -> `TypeCache::parse_type_string`) while the same subscription is being torn down on the main thread, destroying the callback lambda's captured strings and unloading type-support libraries.

Two racing stacks (abridged):
```
Write (main thread): ~lambda  trigger_topic_subscriber.cpp:92   (destroy captured strings)
Read  (executor):    trigger_topic_subscriber.cpp:111 -> JsonSerializer::deserialize -> TypeCache::parse_type_string
```

Root cause: `TriggerTopicSubscriber` assumes `rclcpp::GenericSubscription` destruction drains in-flight callbacks (see the comments in `unsubscribe()` and `shutdown()`). That assumption is false: destroying a `GenericSubscription` does not wait for a callback already running on an executor thread. Trigger subscriptions run on the shared main `MultiThreadedExecutor`, so at shutdown a callback can still be running when the subscription (and the subscriber's serializer and captured strings) is destroyed.

### Environment
- ros2_medkit version: current main
- ROS 2 distro: Jazzy (TSan sanitizer job)
- OS: Ubuntu 24.04

### Additional information

Impact: shutdown-only race. The process is already exiting, so operational impact is low, but TSan fails the sanitizer job whenever it triggers.

Proposed fix: route trigger subscriptions through the isolated subscription executor (the same one `/data` uses), so their callbacks dispatch on a dedicated worker thread that is drained and joined before teardown, instead of the shared main executor. Two things to get right:
- Do not call the executor's synchronous run-on-worker helper while holding the subscriber mutex that a callback also takes, or shutdown can deadlock.
- Order teardown so trigger subscriptions are released before the subscription executor is reset.

A narrow TSan suppression for this race is in `tsan_suppressions.txt` (referencing this issue) so it does not block unrelated work. Remove it when the fix lands.

bburda:
## Summary

`TriggerTopicSubscriber` created its per-trigger subscriptions on the shared main
executor and destroyed them believing that `rclcpp::GenericSubscription`
destruction drains in-flight callbacks. rclcpp gives no such guarantee, so at
gateway shutdown a subscription callback could still be running (deserializing a
sample) while the subscription, its captured strings, and the subscriber's
serializer were being destroyed on the main thread. TSan flags this as a data
race and the gateway child exits 66. It is nondeterministic and pre-existing
(also on `main`).

This routes trigger subscriptions through the shared `Ros2SubscriptionExecutor`
(the same isolated single-worker path `/data` already uses), via
`Ros2SubscriptionSlot`. Their callbacks now dispatch on - and are drained by -
that one worker thread before teardown, so a callback can no longer fire on a
partially destroyed subscriber.

Key points:
- Every slot create/destroy posts a `run_sync` task to that worker, and all of
  them happen strictly outside the subscriber mutex, so a create/destroy can
  never deadlock against a callback that takes the same mutex.
- `main.cpp` shuts the trigger subscriber down (destroying its slots on the
  drained worker) before `sub_exec.reset()`.
- The executor pointer is set once after construction and read on HTTP threads,
  so it is `std::atomic` (acquire/release).
- Behavior note: a persistent "data" trigger whose topic type is already
  resolvable during `GatewayNode` construction (before the executor is wired)
  now subscribes within one retry tick instead of immediately. No crash, no lost
  trigger.

## Issue

- closes #548

## Type

- [x] Bug fix
- [ ] New feature or tests
- [ ] Breaking change
- [ ] Documentation only

## Testing

- `test_triggers_data` passes repeatedly under the default RMW with a clean
  gateway shutdown (no hang = no deadlock, exit 0 = no crash); the trigger unit
  tests pass; clang-tidy is clean.
- The race fix itself is confirmed by the TSan sanitizer job in CI. TSan cannot
  run in the local sandbox (it aborts on an "unexpected memory mapping" because
  ASLR cannot be disabled there), so a local fastrtps run only proves the absence
  of deadlock/crash, not race-freedom.

## Checklist

### DIFF
--- a/src/ros2_medkit_gateway/include/ros2_medkit_gateway/ros2/trigger_topic_subscriber.hpp
+++ b/src/ros2_medkit_gateway/include/ros2_medkit_gateway/ros2/trigger_topic_subscriber.hpp
@@ -24,7 +24,10 @@
 #include <unordered_map>
 
 #include <rclcpp/rclcpp.hpp>
+#include <rclcpp/serialized_message.hpp>
+#include <tl/expected.hpp>
 
+#include "ros2_medkit_gateway/ros2_common/ros2_subscription_slot.hpp"
 #include "ros2_medkit_serialization/json_serializer.hpp"
 
 namespace ros2_medkit_gateway {
@@ -34,22 +37,36 @@ namespace ros2_medkit_gateway {
  *
  * Each trigger that observes a "data" resource registers exactly one
  * subscription here, identified by a caller-provided handle key. The
- * subscriber owns the rclcpp::GenericSubscription, performs the CDR -> JSON
+ * subscriber owns a subscription slot on the shared subscription executor,
+ * performs the CDR -> JSON
  * deserialization, and delivers each sample to the handle's callback. When
  * the topic's type cannot be resolved at subscribe time (publisher not yet
  * up), the entry is queued as "pending" and retried every 5 seconds. Pending
  * entries time out after 60 seconds.
  *
  * Threading: subscribe()/unsubscribe()/shutdown() are mutex-guarded.
- * Per-handle callbacks are invoked from rclcpp executor threads. The
- * subscription-destructor pattern (callbacks cannot fire on a partially
- * destroyed subscriber) is enforced by the shutdown_flag_ guard followed by
- * subscription map clearing inside shutdown().
+ * Every subscription is created, destroyed, AND dispatched on the single
+ * worker thread of the shared Ros2SubscriptionExecutor (via
+ * Ros2SubscriptionSlot). Because create, destroy, and callback dispatch all
+ * run on the same worker, a destroy can never race an in-flight callback, and
+ * executor teardown drains any queued callback before the subscriber's state
+ * (serializer, captured strings) is torn down. This closes the shutdown data
+ * race in issue #548, where subscriptions created directly on the main node
+ * were dispatched by the multi-threaded main executor and could fire on a
+ * partially destroyed subscriber.
+ *
+ * @warning Ros2SubscriptionSlot creation and destruction post a task to the
+ * executor worker and block until it runs (run_sync). The per-handle callback
+ * acquires mutex_. Therefore slot create/destroy must NEVER happen while
+ * holding mutex_: the worker could be inside a callback blocked on mutex_ while
+ * the caller waits for that same worker, deadlocking. All map bookkeeping
+ * happens under mutex_; all slot create/destroy happens strictly outside it.
  *
  * The lifetime of an individual subscription is tied to the corresponding
  * handle key: unsubscribe(handle_key) atomically removes both the user
- * callback and the underlying rclcpp::Subscription, so any in-flight
- * dispatch sees the empty map and short-circuits.
+ * callback and the subscription slot from the map, so any in-flight dispatch
+ * sees the empty map and short-circuits; the slot is then destroyed outside
+ * the lock, draining the worker.
  *
  * The retry callback (set via set_retry_callback) is fired on every retry
  * tick. The trigger manager wires this to retry_unresolved_triggers() so
@@ -77,9 +94,23 @@ class TriggerTopicSubscriber {
   TriggerTopicSubscriber & operator=(TriggerTopicSubscriber &&) = delete;
 
   /**
+   * @brief Wire the shared subscription executor used to create, destroy, and
+   *        dispatch every per-handle subscription.
+   *
+   * Non-owning: the executor must outlive every subscription (the gateway
+   * calls shutdown() before destroying the executor). Must be set before the
+   * first subscribe() that resolves to a real topic; if it is null when a
+   * subscription would be created, subscribe() throws and retry logs+skips
+   * rather than silently falling back to a racy path.
+   *
+   * @param exec Shared subscription executor, or nullptr to detach.
+   */
+  void set_subscription_executor(ros2_common::Ros2SubscriptionExecutor * exec);
+
+  /**
    * @brief Subscribe to a topic under a unique handle key.
    *
-   * Each call yields one rclcpp::GenericSubscription dedicated to the
+   * Each call yields one subscription slot dedicated to the
    * caller's handle. If @p msg_type is empty the type is resolved from the
    * ROS graph; if no publisher is yet advertising the topic, the entry is
    * queued for retry.
@@ -97,7 +128,7 @@ class TriggerTopicSubscriber {
                  SampleCallback callback);
 
   /// Drop the subscription for @p handle_key. Both the user callback and the
-  /// rclcpp::GenericSubscription are released; in-flight dispatch is
+  /// subscription slot are released; in-flight dispatch is
   /// guaranteed not to fire after this returns.
   void unsubscribe(const std::string & handle_key);
 
@@ -110,9 +141,13 @@ class TriggerTopicSubscriber {
   void set_retry_callback(RetryCallback cb);
 
  private:
-  /// Per-handle live subscription state.
+  /// rclcpp generic-subscription callback delivering one serialized sample.
+  using MessageCallback = std::function<void(std::shared_ptr<const rclcpp::SerializedMessage>)>;
+
+  /// Per-handle live subscription state. The slot creates its subscription on
+  /// the shared executor worker and destroys it there when reset.
   struct LiveEntry {
-    rclcpp::GenericSubscription::SharedPtr subscription;
+    std::unique_ptr<ros2_common::Ros2SubscriptionSlot> slot;
     std::string topic_name;
     std::string msg_type;
     SampleCallback callback;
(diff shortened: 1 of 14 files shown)


src/ros2_medkit_gateway/src/ros2/trigger_topic_subscriber.cpp
```diff
@@ -123,50 +97,146 @@ void TriggerTopicSubscriber::create_subscription_internal(const std::string & ha
...
   };
+}
+
+tl::expected<std::unique_ptr<ros2_common::Ros2SubscriptionSlot>, std::string>
+TriggerTopicSubscriber::create_slot_unlocked(const std::string & handle_key, const std::string & topic_name,
+                                             const std::string & msg_type) {
+  auto * exec = sub_exec_.load(std::memory_order_acquire);
+  if (!exec) {
+    // Fail loudly instead of silently creating on the racy main node/executor
+    // path. In the wired gateway this never happens (main.cpp calls
```
mfaferek93: This contradicts the `sub_exec_` member comment in the header, which says the REST server - started in the GatewayNode constructor - can already be running subscribe() on an httplib thread before main.cpp wires the executor. So this path *can* fire in the wired gateway: a trigger POST landing in that window throws, the transport returns nullptr, and the client gets a 5xx for an otherwise valid request.

Either align this comment with reality (accepted narrow window, fail-loud), or queue the handle as pending when the executor is unwired - the retry timer only starts firing once spin begins, which is after wiring, so it would resolve on the first tick. That would also match the persistent-restore behavior the PR body already documents.

src/ros2_medkit_gateway/src/ros2/trigger_topic_subscriber.cpp
```diff
@@ -209,39 +287,94 @@ void TriggerTopicSubscriber::retry_pending_subscriptions() {
...
+  std::vector<Resolvable> resolvable;
+  {
+    std::lock_guard<std::mutex> lock(mutex_);
+    if (pending_.empty()) {
+      return;
+    }
 
-  auto topic_names_and_types = node_->get_topic_names_and_types();
+    const auto now = std::chrono::steady_clock::now();
+    auto topic_names_and_types = node_->get_topic_names_and_types();
```
mfaferek93: `get_topic_names_and_types()` builds the full graph map under the rmw graph locks while holding `mutex_` - and `mutex_` is what every sample dispatch on the executor worker takes per message. With pending handles, each 5s tick stalls dispatch across *all* live handles for the duration of a graph query, which gets ugly on large graphs and under TSan. `subscribe()` Phase A has the same shape via `resolve_topic_type`.

Not a deadlock (dispatch never takes graph locks while holding `mutex_`), just latency - but the fix is nearly free: snapshot the pending keys/topics under the lock, resolve types outside it, and let the existing commit-phase re-check handle the world changing in between. Phase B/C already tolerates exactly that.

src/ros2_medkit_integration_tests/ros2_medkit_test_utils/constants.py
```diff
@@ -70,6 +70,16 @@ def get_test_domain_id(offset=0):
...
 
+# Parameter service readiness
+# A node's ROS 2 parameter service can lag its graph discovery, and the
+# configurations endpoint returns 503 until it responds. This lag is small
+# without instrumentation but grows under the TSan job, where the service was
+# still 503 more than 15s after discovery finished. Generous on purpose: a
+# larger timeout costs nothing on the passing path (the poll returns the moment
+# the endpoint answers 200) and only bounds how long a genuinely dead service
+# waits before failing, well inside the sanitizer jobs' 360s ctest budget.
+PARAM_SERVICE_TIMEOUT = 90.0
```
mfaferek93, line 81: The 360s claim holds for one poll, but this constant feeds a per-test `setUp`: test_configuration_api alone has 9 tests, so a genuinely dead param service costs up to 9 x 90s = 810s and the job dies on a ctest timeout kill instead of the readable assertion this comment promises.

Readiness is monotonic - cache it at class level so the first success skips later polls, and/or fail the class fast after the first timeout.

src/ros2_medkit_gateway/src/ros2/trigger_topic_subscriber.cpp
```diff
@@ -123,50 +97,146 @@ void TriggerTopicSubscriber::create_subscription_internal(const std::string & ha
...
+  }
+  if (!sub_exec_.load(std::memory_order_acquire)) {
+    // Fail loudly rather than falling back to the racy main-node path. The
+    // transport wraps subscribe() in try/catch and returns nullptr on throw,
+    // so the trigger creation is rejected instead of registering a dead handle.
+    RCLCPP_ERROR(node_->get_logger(),
+                 "TriggerTopicSubscriber: no subscription executor wired; cannot subscribe handle '%s' to '%s'",
+                 handle_key.c_str(), topic_name.c_str());
+    throw std::runtime_error("TriggerTopicSubscriber: subscription executor not wired");
+  }
```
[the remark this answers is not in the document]
bburda: Changed subscribe() to defer instead of throwing. When the executor is not wired yet, the handle is queued in pending_ (which now also carries the caller's msg_type) and subscribe() returns. retry_pending_subscriptions() then creates the slot on its first tick, which only runs after spin, so after the executor is wired. No ERROR log and no exception on the persistent-restore path now. Done in 77a47bd8.

src/ros2_medkit_gateway/src/ros2/trigger_topic_subscriber.cpp
bburda: Agreed, the throw could fire in the wired gateway during that startup window. subscribe() now defers to pending when the executor is unwired instead of throwing, so a POST that lands in the window is queued and resolved on the next retry tick rather than returning 5xx. The header comment now matches this behaviour. Done in 77a47bd8.

src/ros2_medkit_gateway/src/ros2/trigger_topic_subscriber.cpp
bburda: Moved type resolution out of mutex_ in both subscribe() and retry_pending_subscriptions(). Phase A now only snapshots the pending handles under the lock; get_topic_names_and_types() runs off the lock and the commit phase re-checks the world, as you suggested. A graph query no longer stalls sample dispatch on the worker. Done in 77a47bd8.

src/ros2_medkit_integration_tests/ros2_medkit_test_utils/constants.py
bburda: Cached readiness on the class. Only the first test polls (up to PARAM_SERVICE_TIMEOUT); the rest return immediately. The first timeout marks the class dead so the remaining tests fail fast, instead of each one paying the full timeout and blowing the ctest budget. Done in 7f65be5d.

src/ros2_medkit_gateway/include/ros2_medkit_gateway/ros2/trigger_topic_subscriber.hpp
```diff
@@ -76,10 +93,24 @@ class TriggerTopicSubscriber {
...
+   *        dispatch every per-handle subscription.
+   *
+   * Non-owning: the executor must outlive every subscription (the gateway
+   * calls shutdown() before destroying the executor). Must be set before the
+   * first subscribe() that resolves to a real topic; if it is null when a
+   * subscription would be created, subscribe() throws and retry logs+skips
+   * rather than silently falling back to a racy path.
+   *
+   * @param exec Shared subscription executor, or nullptr to detach.
+   */
```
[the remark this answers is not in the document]
bburda: Fixed the docstring, it still described the old behaviour. The executor is set once and never cleared, and subscribe() now queues the handle as pending when it is unwired instead of throwing; the retry timer creates the slot once the executor is wired. Removed the "nullptr to detach" note, detach is not a supported operation. Done in d2e537d3.

src/ros2_medkit_gateway/src/ros2/trigger_topic_subscriber.cpp
```diff
@@ -42,54 +42,28 @@ TriggerTopicSubscriber::~TriggerTopicSubscriber() {
...
-
-  // Idempotent: drop any prior registration for this handle so the new one
-  // takes over cleanly (both live and pending paths).
-  live_.erase(handle_key);
-  pending_.erase(handle_key);
+void TriggerTopicSubscriber::set_subscription_executor(ros2_common::Ros2SubscriptionExecutor * exec) {
+  // Release so a subscribe() on an httplib thread that observes this executor
+  // also observes everything the setter's caller published before it.
+  sub_exec_.store(exec, std::memory_order_release);
+}
```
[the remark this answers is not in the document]
bburda: Fixed. set_subscription_executor now wires once with compare_exchange: it rejects null and ignores a later re-set to a different executor (with a warning), so existing slots are never orphaned. The doc comment is updated to match. Done in be20278e.

src/ros2_medkit_gateway/src/ros2/trigger_topic_subscriber.cpp
```diff
@@ -123,50 +97,173 @@ void TriggerTopicSubscriber::create_subscription_internal(const std::string & ha
...
+                                       const std::string & handle_key, SampleCallback callback) {
+  if (shutdown_flag_.load()) {
+    return;
+  }
+
+  // If the subscription executor is not wired yet, defer to pending instead of
+  // failing. This window is real: persistent triggers are restored in the
+  // GatewayNode constructor, and the REST server (also started there) can take a
+  // POST /triggers, both before main.cpp wires the executor. The retry timer
+  // only fires once the executor is spinning (after wiring), so a deferred
```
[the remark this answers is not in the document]
bburda: Added a gateway_ros2 GTest (test_trigger_topic_subscriber). It wires a real Ros2SubscriptionExecutor and a publisher and checks that subscribe() before wiring does not throw and parks the handle in pending, and that after wiring one retry tick creates the live slot which then delivers a published sample. It also checks the set-once executor behaviour. The unsubscribe/shutdown drain-without-deadlock path is already exercised by the test_triggers_data integration test (it deletes triggers while data flows, then shuts down) and by the TSan sanitizer job. Done in bf22ab0b.

src/ros2_medkit_gateway/src/ros2/trigger_topic_subscriber.cpp
```diff
@@ -123,50 +108,173 @@ void TriggerTopicSubscriber::create_subscription_internal(const std::string & ha
...
-                 "TriggerTopicSubscriber: failed to create subscription for '%s' (handle '%s'): %s", topic_name.c_str(),
-                 handle_key.c_str(), e.what());
-    // Propagate. The transport adapter wraps each subscribe in a try/catch
-    // and returns nullptr so the trigger manager rejects the trigger
-    // creation with a 5xx instead of returning 201 for a dead handle.
-    throw;
+tl::expected<std::unique_ptr<ros2_common::Ros2SubscriptionSlot>, std::string>
+TriggerTopicSubscriber::create_slot_unlocked(const std::string & handle_key, const std::string & topic_name,
+                                             const std::string & msg_type) {
+  auto * exec = sub_exec_.load(std::memory_order_acquire);
```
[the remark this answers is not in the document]
bburda: handle_key is used: it is passed to make_message_callback() a few lines below (create_generic call), which captures it so the sample callback can look the handle up in live_ under mutex_. So it is not an unused parameter. The build is clean under -Wall -Wextra (which includes -Wunused-parameter), no warning here.

mfaferek93 (approved):
LGTM!

PR #551, opened 2026-07-23
test: isolate every integration launch test on its own DDS domain

### LINKED ISSUE #550: [BUG] Integration launch tests run on the default DDS domain and cross-contaminate under parallel ctest
# Bug report

### Steps to reproduce
1. Run the test suite so ctest schedules `launch_testing` tests in parallel (for example the conda-based jazzy unit job: `colcon test --ctest-args -LE 'linter|integration'`).
2. The `ros2_medkit_log_bridge` and `ros2_medkit_action_status_bridge` integration tests run at the same time.
3. `test_01_error_log_creates_attributed_fault` in `ros2_medkit_log_bridge` fails: `AssertionError: unexpectedly None : expected a LOG_PLANNER_SERVER_* fault`.

### Expected behavior
Each integration test runs on its own DDS domain and only discovers its own nodes.

### Actual behavior
The `add_launch_test` integration tests in `ros2_medkit_fault_manager`, `ros2_medkit_diagnostic_bridge`, `ros2_medkit_fault_reporter`, `ros2_medkit_action_status_bridge` and `ros2_medkit_log_bridge` do not set `ROS_DOMAIN_ID`, so they all run on the default domain `0`. Only `ros2_medkit_integration_tests` assigns a unique domain per launch test.

Under parallel ctest, tests on domain 0 discover each other's nodes. In the observed failure the `ros2_medkit_log_bridge` fault manager picked up a fault from another test (`ACTION_TEST_ACTION_ABORTED` from `/test_action_status_client`, a node owned by the `ros2_medkit_action_status_bridge` integration test). That extra traffic shifted timing so the `planner_server` ERROR log was not turned into a fault in time, and the assertion failed.

It is non-deterministic: it only triggers when ctest schedules two colliding tests at the same time and DDS discovery is fast enough. Most runs pass, which is why the same commit passes the Docker jazzy job and the conda humble job but fails the conda jazzy job.

Related: the `ros2_medkit_integration_tests` domain pool declares an upper bound of 229, which overruns the gtest ranges assigned to `ros2_medkit_log_bridge` (210-214) and `ros2_medkit_action_status_bridge` (215-219) once it has more than 70 tests (it currently has 74+).

### Environment
- ros2_medkit version: main
- ROS 2 distro: Jazzy (also latent on Humble and Lyrical)
- OS: Ubuntu 24.04

### Additional information
`add_launch_test` targets are not covered by the existing `medkit_set_test_domain` calls, which are applied only to the gtest targets in those packages.

bburda:
# Pull Request

## Summary

Give every `launch_testing` integration test a unique `ROS_DOMAIN_ID` so parallel ctest runs cannot cross-contaminate, and make that invariant hard to break.

Before this change, only `ros2_medkit_integration_tests` isolated its launch tests. The `add_launch_test` targets in `ros2_medkit_fault_manager`, `ros2_medkit_diagnostic_bridge`, `ros2_medkit_fault_reporter`, `ros2_medkit_action_status_bridge` and `ros2_medkit_log_bridge` set no domain, so they all ran on the default domain 0. Under parallel ctest they discovered each other's nodes. That is how `test_01_error_log_creates_attributed_fault` in `ros2_medkit_log_bridge` failed: its fault manager picked up `ACTION_TEST_ACTION_ABORTED` from the `ros2_medkit_action_status_bridge` test node.

Changes (commit 1):
- Add a `medkit_add_launch_test` wrapper that registers a launch test and assigns its domain in one call, and route every launch test through it, so a launch test can no longer be added without isolation. `fault_reporter` had no range, so it gets one.
- Reallocate the domain table so the pools are disjoint and the glob-grown `integration_tests` pool gets headroom: reclaim the slack in the fixed beacon pools (`param_beacon`, `topic_beacon` each used 1 of 10, now 5 each), move `fault_reporter`/`log_bridge`/`action_status_bridge` into 110-119, and grow `integration_tests` to 130-219 (90 slots, 74 used). `graph_provider` (120-129) and `opcua` (220-229) are unchanged.
- Drop three hardcoded `ROS_DOMAIN_ID` overrides that bypassed the table: the `fault_manager` rosbag launch tests forced 42/43 (which collided with the gateway gtests assigned 42/43), and the gateway `test_fault_manager` gtest forced 99 as a manual workaround for the launch tests running on domain 0. Each test now inherits its assigned domain.

Commit 2 fixes a separate flake in the same test that CI surfaced: the `log_bridge` integration test used a fixed 1s sleep to let the bridge connect, then published a single ERROR log. On a slow machine the bridge had not matched the `/rosout` publisher within 1s, so the first log was dropped and its fault never appeared. It now waits on the matched subscription count instead, the same pattern the other bridge tests already use.

---

## Issue

- closes #550

---

## Type

- [x] Bug fix
- [ ] New feature or tests
- [ ] Breaking change
- [ ] Documentation only

---

## Testing

Built the affected packages and confirmed each launch test's ctest `ENVIRONMENT` carries a unique `ROS_DOMAIN_ID` (fault_manager 15-18, diagnostic_bridge 91, fault_reporter 110, log_bridge 114, action_status_bridge 117, integration_tests 130-203). `medkit_add_launch_test` verified to assign the domain (fault_reporter test -> 110).

Ran the two tests that were cross-contaminating (`log_bridge` and `action_status_bridge`) at the same time; both pass together now. Ran the two rosbag tests and the gateway `test_fault_manager` on their assigned domains, concurrently with a real `fault_manager_node`, to confirm the removed hardcodes were not needed.

CMake configure passes with no domain-range exhaustion.

---

## Checklist

### DIFF
--- a/src/ros2_medkit_cmake/cmake/ROS2MedkitTestDomain.cmake
+++ b/src/ros2_medkit_cmake/cmake/ROS2MedkitTestDomain.cmake
@@ -19,19 +19,33 @@
 # non-overlapping domain ID ranges to prevent DDS cross-contamination.
 #
 # Allocated ranges:
-#   ros2_medkit_sovd_service_iface:   1 -   9 (9 slots)
-#   ros2_medkit_fault_manager:       10 -  29 (20 slots)
-#   ros2_medkit_gateway:             30 -  89 (60 slots)
-#   ros2_medkit_diagnostic_bridge:   90 -  99 (10 slots)
-#   ros2_medkit_param_beacon:       100 - 109 (10 slots)
-#   ros2_medkit_topic_beacon:       110 - 119 (10 slots)
-#   ros2_medkit_graph_provider:     120 - 129 (10 slots)
-#   ros2_medkit_linux_introspection: 130 - 139 (10 slots)
-#   ros2_medkit_integration_tests:  140 - 209 (70 slots)
-#   ros2_medkit_log_bridge:         210 - 214 (5 slots, carved from integration_tests)
-#   ros2_medkit_action_status_bridge: 215 - 219 (5 slots, carved from integration_tests)
-#   ros2_medkit_opcua:              220 - 229 (10 slots, carved from integration_tests)
-#   multi-domain tests (secondary): 230 - 232 (3 slots, reserved for peer_aggregation etc.)
+#   ros2_medkit_sovd_service_interface: 1 -   9 (9 slots)
+#   ros2_medkit_fault_manager:         10 -  29 (20 slots)
+#   ros2_medkit_gateway:               30 -  89 (60 slots)
+#   ros2_medkit_diagnostic_bridge:     90 -  99 (10 slots)
+#   ros2_medkit_param_beacon:         100 - 104 (5 slots)
+#   ros2_medkit_topic_beacon:         105 - 109 (5 slots)
+#   ros2_medkit_fault_reporter:       110 - 112 (3 slots)
+#   ros2_medkit_log_bridge:           113 - 115 (3 slots)
+#   ros2_medkit_action_status_bridge: 116 - 119 (4 slots)
+#   ros2_medkit_graph_provider:       120 - 129 (10 slots)
+#   ros2_medkit_integration_tests:    130 - 219 (90 slots)
+#   ros2_medkit_opcua:                220 - 229 (10 slots)
+#   multi-domain tests (secondary):   230 - 232 (3 slots, reserved for peer_aggregation etc.)
+#
+# Launch_testing tests (add_launch_test) create real ROS nodes and MUST consume
+# a domain, exactly like a gtest. Two concurrently scheduled tests on the same
+# domain (the default, 0) discover each other's nodes and cross-contaminate.
+# Use medkit_add_launch_test (below), which registers the test and assigns its
+# domain in one call, so a launch test can never be added without isolation.
+#
+# ros2_medkit_linux_introspection reads /proc and cgroups only (no ROS nodes),
+# so it needs no domain isolation and reserves no range.
+#
+# integration_tests (it grows by glob) and opcua are the tightest pools. When
+# either fills, reclaim slack from the fixed 5-slot beacon pools above (each
+# uses one of five) and re-carve - the 1-232 DDS space is otherwise fully
+# allocated.
 #
 # To add a new package: pick the next free range and update this comment.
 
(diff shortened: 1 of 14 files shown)


src/ros2_medkit_cmake/cmake/ROS2MedkitTestDomain.cmake
```diff
@@ -19,19 +19,27 @@
...
+#   ros2_medkit_topic_beacon:         110 - 119 (10 slots)
+#   ros2_medkit_graph_provider:       120 - 129 (10 slots)
+#   ros2_medkit_fault_reporter:       130 - 132 (3 slots)
+#   ros2_medkit_log_bridge:           133 - 135 (3 slots)
+#   ros2_medkit_action_status_bridge: 136 - 139 (4 slots)
+#   ros2_medkit_integration_tests:    140 - 219 (80 slots)
+#   ros2_medkit_opcua:                220 - 229 (10 slots)
+#   multi-domain tests (secondary):   230 - 232 (3 slots, reserved for peer_aggregation etc.)
+#
+# Every launch_testing test (add_launch_test) must consume a domain via
```
mfaferek93: This states the invariant as a MUST, but nothing enforces it - and it fails open. If someone adds an `add_launch_test` and forgets the matching `medkit_set_test_domain`, the test keeps `ROS_DOMAIN_ID` unset; `constants.py` resolves `int(os.environ.get('ROS_DOMAIN_ID', '0'))` to the shared default domain 0, and it silently cross-contaminates with every other domain-0 test - exactly the failure this PR removes, reintroduced with no configure-time or runtime signal. Worth making it fail closed: a CMake check that errors when a launch test has no assigned domain, or a `constants.py` fallback that raises instead of defaulting to 0 in the isolated-test path. As-is the whole scheme rests on author discipline.

src/ros2_medkit_integration_tests/CMakeLists.txt
```diff
@@ -120,11 +120,12 @@ if(BUILD_TESTING)
...
-  # integration_tests allocation is 140-229, with 230-232 reserved for the
-  # multi-gateway secondary pool (see get_test_domain_id below). One domain is
-  # consumed per feature + scenario test.
-  medkit_init_test_domains(START 140 END 229)
+  # Primary per-test domain pool. Upper bound is 219: the documented
+  # integration_tests allocation is 140-219, with 220-229 owned by opcua and
+  # 230-232 reserved for the multi-gateway secondary pool (see
+  # get_test_domain_id below). One domain is consumed per feature + scenario
+  # test.
+  medkit_init_test_domains(START 140 END 219)
```
mfaferek93: Heads-up on headroom: this pool is at 74/80, and it's the only one that grows by glob (`file(GLOB test/features/*.test.py)`) - a new feature-test file consumes a domain with no CMake change. The whole 1-232 space is now carved (220 = opcua), so ~6 more feature files hit the final `FATAL_ERROR` and block the entire package build, with no contiguous range left to extend into. Meanwhile the fixed 10-slot pools (param_beacon, topic_beacon, graph_provider) each use 1 of 10. The tightest headroom is on the fastest-growing pool; opcua is also tight at 9/10. Consider reclaiming from the underused fixed pools and handing that slack to this one, or at least noting in the comment that integration_tests/opcua are near capacity so the next person knows a re-carve is coming.

src/ros2_medkit_cmake/cmake/ROS2MedkitTestDomain.cmake
bburda: It now fails closed. Added a `medkit_add_launch_test` wrapper in `ROS2MedkitTestDomain.cmake` that registers the launch test and assigns its domain in one call, and routed all five packages through it, so a launch test can no longer be added without a domain. `integration_tests` still calls `add_launch_test` directly because each of its tests needs a combined `GATEWAY_TEST_PORT` + `ROS_DOMAIN_ID` set inline in the loop, and that path always assigns a domain.

src/ros2_medkit_integration_tests/CMakeLists.txt
bburda: Reclaimed the slack instead of only noting it: `param_beacon` and `topic_beacon` each used 1 of 10, now 5 slots each. That freed room to move `fault_reporter`/`log_bridge`/`action_status_bridge` into 110-119 and grow `integration_tests` to 130-219 (90 slots, 74 used, 16 free). `graph_provider` (120-129) and `opcua` (220-229) are unchanged. The allocation table comment now flags `integration_tests` and `opcua` as the tightest pools and points to the beacon pools for the next re-carve.

mfaferek93 (approved):

PR #552, opened 2026-07-23
fix(opcua): drop non-condition events from the alarm path

mfaferek93:
The server-wide EventNotifier (`ns=0;i=2253`, the Server object) emits `BaseEventType`/`AuditEventType` housekeeping events - session-state changes, audit/system events (e.g. "Session state changed to Created") - alongside genuine A&C Condition events. The catch-all alarm mapping routed all of them to a single `PLC_ALARM` fault code, and because the fault store dedups by code, a housekeeping event could overwrite and mask a real process alarm sharing that code.

Only `ConditionType`-derived events carry a non-null `ConditionId` (the Part 9 auto-prepended select clause). Guarding `on_event` on a non-null `ConditionId` drops the non-condition housekeeping events cleanly while leaving the real-alarm path untouched - genuine conditions always carry a non-null ConditionId (the plugin already keys per-condition tracking and Ack/Confirm on it).

Verified: `colcon test ros2_medkit_opcua` green - 265 tests including the live open62541 A&C integration tests that drive real `AlarmConditionType` events (real conditions still flow through).

### DIFF
--- a/src/ros2_medkit_plugins/ros2_medkit_opcua/src/opcua_poller.cpp
+++ b/src/ros2_medkit_plugins/ros2_medkit_opcua/src/opcua_poller.cpp
@@ -752,6 +752,19 @@ void OpcuaPoller::on_event(const AlarmEventConfig & cfg, const std::vector<opcua
     }
   }
 
+  // Drop non-condition events. The server-wide EventNotifier (ns=0;i=2253)
+  // also emits BaseEventType / AuditEventType housekeeping (e.g. "Session
+  // state changed to Created ..."), which is NOT an A&C Condition and carries
+  // a null ConditionId. Such events must never enter the alarm path: they
+  // would resolve to the catch-all fault_code (PLC_ALARM) and mask a genuine
+  // process alarm that the fault store dedups on the same code+source.
+  if (condition_id.isNull()) {
+    RCLCPP_DEBUG_STREAM(opcua_poller_logger(),
+                        "on_event: non-condition event (null ConditionId, type=" << event_type.toString()
+                                                                                 << ") - ignoring");
+    return;
+  }
+
   // A condition (re)reported inside a refresh burst is still active; remember it
   // so RefreshEnd does not clear it. Recorded before the field-count bail below
   // so even a partial-field replay counts as "seen".
--- a/src/ros2_medkit_plugins/ros2_medkit_opcua/src/opcua_poller.cpp
+++ b/src/ros2_medkit_plugins/ros2_medkit_opcua/src/opcua_poller.cpp
@@ -774,6 +774,24 @@ void OpcuaPoller::on_event(const AlarmEventConfig & cfg, const std::vector<opcua
   //       null ConditionId (call_method on runtime->condition_id).
   // Dropping the event up front stops the entry from ever being created.
   if (!is_condition_event(condition_id)) {
+    // On an EXPLICIT source (operator supplied a fault_code and/or mappings) a
+    // real alarm can legitimately arrive with a null ConditionId - an edge
+    // open62541 server built without UA_ENABLE_SUBSCRIPTIONS_ALARMS_CONDITIONS,
+    // or an AlarmConditionType fired via UA_Server_triggerEvent rather than
+    // createCondition. Dropping it with only the DEBUG line below would regress
+    // that path silently (pre-PR it still raised), so warn the operator once per
+    // source. The synthetic auto source carries no fault_code/mappings
+    // (effective_alarm_sources leaves them empty), so this predicate excludes
+    // it and the auto/synthetic path keeps only the DEBUG line.
+    if ((!cfg.fault_code.empty() || !cfg.mappings.empty()) &&
+        noncondition_drop_warned_sources_.insert(cfg.source_node_id_str).second) {
+      const std::string code_desc =
+          cfg.fault_code.empty() ? std::string("its mapped fault codes") : "fault_code '" + cfg.fault_code + "'";
+      warn_operator("OPC-UA non-condition event on configured alarm source '" + cfg.source_node_id_str + "' (" +
+                    code_desc +
+                    ") carries a null ConditionId and was dropped; if this source's alarms carry no ConditionId they "
+                    "will not surface as faults.");
+    }
     RCLCPP_DEBUG_STREAM(opcua_poller_logger(), "on_event: non-condition event (null ConditionId, type="
                                                    << event_type.toString() << ") - ignoring");
     return;
--- a/src/ros2_medkit_plugins/ros2_medkit_opcua/src/opcua_poller.cpp
+++ b/src/ros2_medkit_plugins/ros2_medkit_opcua/src/opcua_poller.cpp
@@ -752,13 +752,18 @@ void OpcuaPoller::on_event(const AlarmEventConfig & cfg, const std::vector<opcua
     }
   }
 
-  // Drop non-condition events. The server-wide EventNotifier (ns=0;i=2253)
-  // also emits BaseEventType / AuditEventType housekeeping (e.g. "Session
-  // state changed to Created ..."), which is NOT an A&C Condition and carries
-  // a null ConditionId. Such events must never enter the alarm path: they
-  // would resolve to the catch-all fault_code (PLC_ALARM) and mask a genuine
-  // process alarm that the fault store dedups on the same code+source.
-  if (condition_id.isNull()) {
+  // Drop non-condition events for EVERY alarm source, explicit or auto. The
+  // server-wide EventNotifier (ns=0;i=2253) also emits BaseEventType /
+  // AuditEventType housekeeping (e.g. "Session state changed to Created ...")
+  // that is not an A&C Condition and carries a null ConditionId (Part 9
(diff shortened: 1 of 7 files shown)


src/ros2_medkit_plugins/ros2_medkit_opcua/src/opcua_poller.cpp
```diff
@@ -752,6 +752,19 @@ void OpcuaPoller::on_event(const AlarmEventConfig & cfg, const std::vector<opcua
...
+  // also emits BaseEventType / AuditEventType housekeeping (e.g. "Session
+  // state changed to Created ..."), which is NOT an A&C Condition and carries
+  // a null ConditionId. Such events must never enter the alarm path: they
+  // would resolve to the catch-all fault_code (PLC_ALARM) and mask a genuine
+  // process alarm that the fault store dedups on the same code+source.
+  if (condition_id.isNull()) {
+    RCLCPP_DEBUG_STREAM(opcua_poller_logger(),
+                        "on_event: non-condition event (null ConditionId, type=" << event_type.toString()
+                                                                                 << ") - ignoring");
+    return;
```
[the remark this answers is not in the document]
mfaferek93: Good catch. Reworked in 87c07226: the guard no longer uses a raw `condition_id.isNull()` - it now delegates to the shared `OpcuaPoller::is_condition_event()` helper, which has focused unit tests asserting exactly this decision: `IsConditionEventTest.NullConditionIdIsRejected` (a null ConditionId - i.e. a BaseEvent/AuditEvent housekeeping notification - is rejected) and `IsConditionEventTest.RealConditionIdIsAccepted` (a real Condition still passes). The guard is now a one-line call into that tested helper, and I removed the duplicate copy of the same check that lived in the auto-alarm branch (it was dead once the guard was hoisted above the explicit/auto split). A full end-to-end assertion (fire a bare BaseEvent at the Server object under a catch-all event_alarms mapping, verify no fault is delivered) belongs in the `test_opcua_secured` integration suite and I can add it there if you'd prefer the behavioural coverage on top of the unit-level decision test.

src/ros2_medkit_plugins/ros2_medkit_opcua/src/opcua_poller.cpp
```diff
@@ -752,6 +752,23 @@ void OpcuaPoller::on_event(const AlarmEventConfig & cfg, const std::vector<opcua
     }
   }
 
+  // Drop non-condition events for EVERY alarm source, explicit or auto. The
+  // server-wide EventNotifier (ns=0;i=2253) also emits BaseEventType /
+  // AuditEventType housekeeping (e.g. "Session state changed to Created ...")
+  // that is not an A&C Condition and carries a null ConditionId (Part 9
+  // §5.5.2.13: only a Condition resolves the ConditionId select clause). A
+  // catch-all explicit event_alarms entry pointed at i=2253 would otherwise
+  // map such an event to its configured fault_code (e.g. PLC_ALARM) and mask a
```
bburda: This explanation does not match what the code does, and the comment stays in the file, so it should be correct.

For a non-Condition event all condition select clauses resolve to Null, so `input.active_state` is false. `AlarmStateMachine::compute` then returns `ReportHealed`, and `OpcuaPlugin::on_event_alarm` handles `ReportHealed` as an intentional no-op. No fault is reported. Nothing is written to the fault store, so nothing is overwritten there.

The real problem is different, and it is a stronger reason to keep the guard. All housekeeping events end up in one `conditions_` entry, keyed by the null ConditionId string, holding the catch-all `fault_code` and latched at `Healed`. `Healed` counts as active in `should_clear_after_refresh`, so the next ConditionRefresh sends a `ClearFault` for that code and clears the real alarm. The same entry can also win the linear search in `lookup_condition`, so an Acknowledge or Confirm is sent to a null ConditionId.

src/ros2_medkit_plugins/ros2_medkit_opcua/src/opcua_poller.cpp
```diff
@@ -752,6 +752,23 @@ void OpcuaPoller::on_event(const AlarmEventConfig & cfg, const std::vector<opcua
...
+  // genuine process alarm the fault store dedups on the same code. The auto
+  // path had this guard already (see the shared is_condition_event() helper);
+  // hoisting it here covers the explicit path too, so it runs once, before the
+  // explicit/auto split below.
+  if (!is_condition_event(condition_id)) {
+    RCLCPP_DEBUG_STREAM(opcua_poller_logger(), "on_event: non-condition event (null ConditionId, type="
```
bburda, line 795: The drop is logged at DEBUG only. On an explicit `event_alarms` entry the user gave both a source and a fault_code, so it is hard to find out why nothing happens: `x-plc-status` reports connected with no errors, and `/faults` stays empty.

It also makes some servers stop working completely. If a server cannot resolve the ConditionId select clause, real alarms arrive with a null ConditionId as well. In open62541 this happens in two cases. First, a server built without `UA_ENABLE_SUBSCRIPTIONS_ALARMS_CONDITIONS` returns BadNotSupported for that clause. Second, `UA_getConditionId` only knows conditions registered with `UA_Server_createCondition`, so an AlarmConditionType modelled in the address space and fired with `UA_Server_triggerEvent` returns BadNotFound. A failed select clause is delivered as an empty field, so `condition_id` is null. Before this change such an alarm still raised its configured fault. Now it is dropped and the operator sees nothing.

Suggestion: call `warn_operator()` once per source when the dropped event comes from a source that has `fault_code` or `mappings` set. The file already uses this pattern with `condition_refresh_warned_`.

bburda (commented):

- `README.md` still describes this drop as auto_alarms only. The paragraph starting "**System messages:**" is inside the `auto_alarms` chapter and says `auto_alarms` drops those events before deriving a fault. After this change it applies to every alarm source. The same paragraph says to use `exclude` for the rest, but `auto_alarm_passes_filters` runs only in the auto branch, so an explicit `event_alarms` entry has no text filter at all.
- `README.md` also says "An event matching no mapping uses the source-level `fault_code`; if that is also empty the event is ignored." This is what the README promises for the catch-all `ns=0;i=2253` example with `fault_code: PLC_GENERIC_ALARM`. Non-Condition events now never reach `NodeMap::resolve_alarm`, so this sentence needs the exception.
- `design/index.rst` lists the decision order without the ConditionId condition. It is now a global rule that runs before that table.
- The doc comment on `is_condition_event` in `opcua_poller.hpp` and the comment above `IsConditionEventTest` still say the filter is for auto_alarms.
- Test coverage: `IsConditionEventTest` only checks the predicate. It passes the same way before and after the move, so it does not cover the routing this PR changes. The live fixture can emit Conditions only: every trigger goes through `UA_Server_triggerConditionEvent` on nodes built by `UA_Server_createCondition`, and there is no command for a plain event. To cover this, add a fixture command that fires a `BaseEventType` on `i=2253` with `UA_Server_triggerEvent`. Then add an integration case with an explicit `event_alarms` entry on that source: the plain event must produce no fault, and a real condition on the same source must still produce its fault. The second check matters too, it catches dropping more than intended.
- Pre-existing, not caused by this PR: `reconcile_after_refresh` does not filter by source, while `reconcile_after_read` filters on `runtime.source_id`. In open62541 `ConditionRefresh` loops over every monitored item in the subscription and sends a separate RefreshStart/RefreshEnd pair for each one. The plugin creates one monitored item per alarm source on a single subscription. With two sources the first RefreshEnd reconciles against a `refresh_seen_` set that holds only the conditions of the first item. Conditions from the other source are then cleared and raised again by the next burst. Worth a separate issue.

src/ros2_medkit_plugins/ros2_medkit_opcua/src/opcua_poller.cpp
mfaferek93: Fixed in 0b231c8d: the comment now states the true mechanism. An unfiltered non-condition event does NOT map to its fault_code - every alarm-state select clause resolves to null, so `AlarmStateMachine::compute` returns `ReportHealed` (a no-op) and nothing is reported or masked. The real harm is the latched null-ConditionId entry `apply_condition_state` leaves in `conditions_`: it drives a spurious `ClearFault` on the next `ConditionRefresh` (`should_clear_after_refresh` treats `Healed` as active) and can misroute a later Acknowledge/Confirm via `lookup_condition`. Comment-only.

src/ros2_medkit_plugins/ros2_medkit_opcua/src/opcua_poller.cpp
mfaferek93: Fixed in 30d32a23: added a per-source warn-once (`noncondition_drop_warned_sources_`, mirroring `read_unsupported_warned_sources_`) gated on the source having a `fault_code` or `mappings`. An explicit `event_alarms` source whose alarm arrives with a null ConditionId (edge open62541 build, or an AlarmConditionType fired via `triggerEvent`) now warns the operator once instead of dropping silently at DEBUG; the synthetic auto source has no fault_code/mappings so it stays on the quiet DEBUG path. 0988954f adds a routing regression test and scopes the guard as global in the docs.

bburda (approved):

PR #554, opened 2026-07-23
test: route inline gateways and fault managers through the shared helpers (coverage SIGKILL fix)

### LINKED ISSUE #553: [BUG] Integration test gateways SIGKILLed (-9) under coverage; shutdown-grace and coverage-env applied inconsistently
# Bug report

### Steps to reproduce
1. Build with coverage enabled and run the integration tests under gcov.
2. `test_scripts_api` launches a gateway that, at shutdown, must flush gcov data. It does not finish within the launch default SIGTERM window (~10s), so launch_testing escalates to SIGKILL and the process exits with code -9.
3. The post-shutdown check fails: -9 is not in `ALLOWED_EXIT_CODES` (`{0, -2, -15}`).

### Expected behavior
Every launched gateway and fault-manager gets a shutdown window long enough to flush coverage data before SIGKILL, and its coverage env is set, consistently - the same way the shared `create_gateway_node` helper already does.

### Actual behavior
Only `create_gateway_node` sets `sigterm_timeout='30'` / `sigkill_timeout='15'`. Several tests build the gateway (or fault manager) inline and miss the grace, so they SIGKILL to -9 under coverage:
- `test_scripts_api` (the one that failed), `test_updates`, `test_update_status_subscription`, `test_triggers_persistent`, `test_tls` - inline gateways, no shutdown grace.
- `create_fault_manager_node` (shared helper) sets the coverage env but no shutdown grace, so the ~51 tests that launch a fault manager via `create_test_launch` share the same latent risk.

Related inconsistencies found while tracing this:
- Coverage env is dropped on the aggregation peer gateways (`test_cross_ecu_fanout`, `test_peer_aggregation`, `test_daisy_chain_aggregation`, `test_leaf_collision_aggregation`), the two `test_cross_ecu_fanout` fault managers, and the `subprocess.Popen` demo spawns in `test_graph_event_discovery` and `test_graph_provider_unreachable`.
- `test_tls` has an inline gateway with no grace AND its post-shutdown class does not assert exit codes, so a -9 there would be masked instead of failing.
- Several fault-manager tests use a fixed `time.sleep` plus a single service call where the sibling test already polls: `test_integration.py` snapshot test (leaves a `_wait_for_topic_subscribers` helper defined but unused), `test_rosbag_integration.py` (single `GetRosbag` after a 1s sleep vs the polling loop in `test_rosbag_entity_scope.py`), and `test_update_status_subscription.py` (SSE connect gated by a fixed sleep vs a `threading.Event` in the trigger tests).
- `test_opcua_secured` hardcodes `ROS_DOMAIN_ID=228`, which is the domain CMake already assigns to `test_opcua_identity` in the same build block; the two collide if the secure suite runs.

### Environment
- ros2_medkit version: main
- ROS 2 distro: Jazzy (the coverage job); latent on all distros
- OS: Ubuntu 24.04

### Additional information
Root cause is that inline node construction bypasses the shared helpers, and `create_gateway_node` could not be used by multi-gateway / multi-domain / distinct-name tests because it lacked a `name` and an `extra_env` parameter. Extending the two helpers lets those tests route through them and get the grace + coverage env in one place.

bburda:
# Pull Request

## Summary

Make integration-test gateways and fault managers survive shutdown under a coverage (gcov) build, route the inline copies through the shared helpers, and replace fixed sleeps with active waits.

Under coverage a gateway or fault manager flushes coverage data at shutdown, which can exceed the launch default SIGTERM window (~10s). The process is then SIGKILLed and exits with `-9`, which fails the post-shutdown exit-code check (`ALLOWED_EXIT_CODES = {0, -2, -15}`). Only `create_gateway_node` set a wide enough window; several tests built the node inline and missed it, and `create_fault_manager_node` had no window at all.

Shutdown / coverage:
- Extend `create_gateway_node` with `name=` and `extra_env=`, and give `create_fault_manager_node` the same `name=`/`extra_env=` plus `sigterm_timeout='30'` / `sigkill_timeout='15'`.
- Route the inline gateways and fault managers through the two helpers (`test_scripts_api`, `test_updates`, `test_update_status_subscription`, `test_triggers_persistent`, and the four aggregation tests). This fixes the SIGKILL failure, restores the coverage env the aggregation peers and fault managers had dropped, and removes four hand-copied shutdown-window constants.
- `test_tls` keeps its inline gateway (it loads from a params file) but gets the windows directly and a post-shutdown exit-code check it was missing.
- The fault-manager-side tests build the node inline in their own packages (they cannot import `create_fault_manager_node`, which lives downstream), so give the inline `fault_manager_node` the same windows directly in `test_rosbag_integration`, `test_integration`, `test_rosbag_entity_scope`, `test_entity_thresholds_integration` (fault_manager) and the `diagnostic_bridge`, `log_bridge` and `action_status_bridge` integration tests.
- Pass the coverage env to the two `subprocess.Popen` demo spawns.
- Route each subprocess's coverage to its own package: `create_fault_manager_node` and `create_demo_nodes` passed the default gateway package to `get_coverage_env`, so their `.gcda` were written into the gateway build dir. Pass `ros2_medkit_fault_manager` and `ros2_medkit_integration_tests` respectively.
- Give the same windows to every other coverage-instrumented C++ node that a post-shutdown exit-code check guards, not only the gateway/fault_manager: the three bridge nodes that are the subject under test (`diagnostic_bridge_node`, `log_bridge_node`, `action_status_bridge_node`), the demo nodes launched through `create_demo_nodes`, and the inline demo-node launches in the discovery, graph-provider and beacon feature tests. Each is small, but the shutdown checks reject `-9`, so a slow-runner gcov flush past the default SIGTERM window would SIGKILL them the same way.

Timing / isolation:
- Poll `GetRosbag` and `GetSnapshots` until the async capture is written instead of a fixed sleep plus a single call, and poll for the rosbag buffer publisher and the cleanup delete. The snapshot test relies on `_wait_for_snapshot` (which polls `GetSnapshots` for the real capture) as its gate instead of a pre-wait on the short-lived on-demand subscription.
- Republish the ERROR log in the `log_bridge` test until the fault appears: a single publish raced the bridge's report client discovering the fault manager (the second discovery hop), which the `/rosout` subscription wait does not cover.
- Gate the update SSE test on the reader thread opening the stream (a `threading.Event`) instead of a fixed sleep.
- `test_opcua_secured` hardcoded `ROS_DOMAIN_ID=228`, which collides with the domain CMake assigns `test_opcua_identity`; move it to the free 229.

ThreadSanitizer:
- `test_ros2_parameter_transport` TearDown called `executor->cancel()` while the spin thread was still inside `spin_some()`, racing the executor's notify-waitable (`ExecutorNotifyWaitable::is_ready` vs `trigger_entity_recollect`). Clear the run flag and join the spin thread before touching the executor; the `spin_some(10ms)` loop exits on its own, so `cancel()` is not needed.

CI:
- Give the sanitizer jobs timeout headroom so a slow runner does not cancel them mid-run: a full ASan run is ~55 min (bump 60 -> 90) and a full TSan run is ~40 min (bump 45 -> 60). Both were cancelling intermittently at the old limits.

---

## Issue

- closes #553

---

## Type

- [x] Bug fix
- [ ] New feature or tests
- [ ] Breaking change
- [ ] Documentation only

---

## Testing

Built the affected packages and ran the rerouted / retimed tests locally: `test_scripts_api` (the one that failed), `test_updates`, `test_update_status_subscription`, `test_cross_ecu_fanout`, `test_graph_event_discovery`, `test_graph_provider_unreachable`, the two fault-manager suites `test_integration` (snapshot poll) and `test_rosbag_integration` (rosbag poll), `log_bridge` `test_integration` (3 runs, to check the republish fix), and `test_ros2_parameter_transport` - all pass. `run-clang-tidy` on the changed C++ file returns 0.

---

## Checklist

### DIFF
--- a/src/ros2_medkit_action_status_bridge/test/test_integration.test.py
+++ b/src/ros2_medkit_action_status_bridge/test/test_integration.test.py
@@ -56,6 +56,9 @@ def generate_test_description():
             # fault heals on that single event only with healing_threshold 0.
             'healing_threshold': 0,
         }],
+        # Give the node room to flush coverage data at shutdown before SIGKILL.
+        sigterm_timeout='30',
+        sigkill_timeout='15',
     )
 
     # Load the packaged config first so assertExitCodes guards against the
@@ -72,6 +75,9 @@ def generate_test_description():
         parameters=[config_file, {
             'rescan_period_sec': 1.0,  # fast discovery for the test
         }],
+        # Give the node room to flush coverage data at shutdown before SIGKILL.
+        sigterm_timeout='30',
+        sigkill_timeout='15',
     )
 
     return (
--- a/src/ros2_medkit_diagnostic_bridge/test/test_integration.test.py
+++ b/src/ros2_medkit_diagnostic_bridge/test/test_integration.test.py
@@ -42,6 +42,9 @@ def generate_test_description():
             'healing_enabled': True,
             'healing_threshold': 1,
         }],
+        # Give the node room to flush coverage data at shutdown before SIGKILL.
+        sigterm_timeout='30',
+        sigkill_timeout='15',
     )
 
     diagnostic_bridge_node = launch_ros.actions.Node(
@@ -53,6 +56,9 @@ def generate_test_description():
             'diagnostics_topic': '/diagnostics',
             'auto_generate_codes': True,
         }],
+        # Give the node room to flush coverage data at shutdown before SIGKILL.
+        sigterm_timeout='30',
+        sigkill_timeout='15',
     )
 
     return (
--- a/src/ros2_medkit_fault_manager/test/test_integration.test.py
+++ b/src/ros2_medkit_fault_manager/test/test_integration.test.py
@@ -97,6 +97,9 @@ def generate_test_description():
             'snapshots.capture_queue_full_policy': 'reject_newest',
             'correlation.config_file': correlation_config,  # Enable correlation
         }],
+        # Give the node room to flush coverage data at shutdown before SIGKILL.
+        sigterm_timeout='30',
+        sigkill_timeout='15',
     )
 
     return (
@@ -549,17 +552,26 @@ class TestFaultManagerIntegration(unittest.TestCase):
             rclpy.spin_once(self.node, timeout_sec=0.01)
             time.sleep(interval)
 
-    def _wait_for_topic_subscribers(self, topic_name, timeout_sec=5.0):
-        """Wait for topic to have subscribers (indicates discovery complete)."""
-        start = time.time()
-        while time.time() - start < timeout_sec:
-            count = self.node.count_subscribers(topic_name)
-            if count > 0:
-                return True
-            # Keep publishing to maintain topic presence
-            rclpy.spin_once(self.node, timeout_sec=0.01)
-            time.sleep(0.1)
-        return False
+    def _wait_for_snapshot(self, fault_code, timeout=10.0):
+        """
+        Poll GetSnapshots until the async on-demand capture completes.
+
+        The snapshot subscription is created on demand when the fault confirms
+        and fills over its own timeout window, so a fixed sleep plus a single
+        query flakes under load - poll until data arrives.
+        """
+        request = GetSnapshots.Request()
+        request.fault_code = fault_code
+        request.topic = ''
+        deadline = time.time() + timeout
+        response = self._call_service(self.get_snapshots_client, request)
+        while time.time() < deadline:
+            if (response is not None and response.success
+                    and len(response.data) > 0):
+                return response
+            time.sleep(0.5)
+            response = self._call_service(self.get_snapshots_client, request)
+        return response
 
     def test_16_snapshot_capture_on_fault_confirmation(self):
         """
@@ -602,18 +614,14 @@ class TestFaultManagerIntegration(unittest.TestCase):
             response = self._call_service(self.report_fault_client, request)
             self.assertTrue(response.accepted)
 
-            # Wait for snapshot capture to complete (snapshot timeout is 3s)
-            time.sleep(4.0)
+            # Poll for the snapshot; the publisher thread keeps feeding data.
+            # _wait_for_snapshot is the ground-truth gate (it polls GetSnapshots
+            # for the actual capture), so no fragile pre-wait on the short-lived
+            # on-demand subscription is needed.
+            snap_response = self._wait_for_snapshot(fault_code)
         finally:
             stop_publishing.set()
             pub_thread.join()
-
-        # Query snapshots
-        snap_request = GetSnapshots.Request()
-        snap_request.fault_code = fault_code
-        snap_request.topic = ''  # All topics
-
-        snap_response = self._call_service(self.get_snapshots_client, snap_request)
 
         self.assertTrue(snap_response.success)
         self.assertGreater(len(snap_response.data), 0)
--- a/src/ros2_medkit_integration_tests/ros2_medkit_test_utils/launch_helpers.py
+++ b/src/ros2_medkit_integration_tests/ros2_medkit_test_utils/launch_helpers.py
@@ -89,17 +89,25 @@ LIDAR_FAULTY_PARAMS = {
 # Factory: gateway node
 # ---------------------------------------------------------------------------
 
-def create_gateway_node(*, port=DEFAULT_PORT, extra_params=None, coverage=True):
+def create_gateway_node(*, port=DEFAULT_PORT, name='ros2_medkit_gateway',
+                        extra_params=None, coverage=True, extra_env=None):
     """Create a ``gateway_node`` launch action with standard config.
 
     Parameters
     ----------
     port : int
         HTTP server port (default: 8080).
+    name : str
+        ROS node name. Override when a test launches more than one gateway
+        so their names do not collide (e.g. ``gateway_with_scripts``).
     extra_params : dict or None
         Additional ROS parameters merged into the node config.
     coverage : bool
         If True, set GCOV_PREFIX env vars for code coverage collection.
+    extra_env : dict or None
+        Additional environment variables merged into ``additional_env`` on
+        top of the coverage env. Useful for setting ``ROS_DOMAIN_ID`` to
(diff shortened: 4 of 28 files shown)


src/ros2_medkit_integration_tests/ros2_medkit_test_utils/launch_helpers.py
```diff
@@ -187,13 +208,22 @@ def create_fault_manager_node(
     if extra_params:
         params.update(extra_params)
 
+    env = dict(get_coverage_env() if coverage else {})
+    if extra_env:
+        env.update(extra_env)
```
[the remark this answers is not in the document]
bburda: Fixed. create_fault_manager_node and create_demo_nodes now pass their own package to get_coverage_env (ros2_medkit_fault_manager and ros2_medkit_integration_tests), so each process writes its .gcda to its own build dir instead of the gateway one.

src/ros2_medkit_fault_manager/test/test_integration.test.py
```diff
@@ -602,19 +626,15 @@ def keep_publishing():
             response = self._call_service(self.report_fault_client, request)
             self.assertTrue(response.accepted)
 
-            # Wait for snapshot capture to complete (snapshot timeout is 3s)
-            time.sleep(4.0)
+            # Confirm the fault manager's on-demand snapshot subscription
+            # matched our publisher, then poll for the snapshot (the publisher
+            # thread keeps feeding data) instead of a fixed sleep + single query.
+            self._wait_for_topic_subscribers('/test/temperature')
```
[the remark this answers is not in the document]
bburda: I removed the pre-wait instead of asserting it. The on-demand snapshot subscription is torn down right after the first sample arrives, so a 0.1s poll can miss it and fail a capture that actually worked. _wait_for_snapshot already polls GetSnapshots for the real snapshot data, so it is the gate now and the pre-wait plus the unused helper are gone.

src/ros2_medkit_action_status_bridge/test/test_integration.test.py
```diff
@@ -56,6 +56,9 @@ def generate_test_description():
             # fault heals on that single event only with healing_threshold 0.
             'healing_threshold': 0,
         }],
+        # Give the node room to flush coverage data at shutdown before SIGKILL.
+        sigterm_timeout='30',
+        sigkill_timeout='15',
```
mfaferek93, line 61: You widen the fault_manager's teardown here so its gcov flush survives on a slow CI runner, but `action_status_bridge_node` right below gets no `sigterm_timeout`/`sigkill_timeout`. It's the same class of process - a coverage-instrumented C++ binary flushing gcov at shutdown - and it's the one actually under test. `ALLOWED_EXIT_CODES` excludes -9 and `TestShutdown.test_exit_codes` checks every proc, so if the bridge's shutdown+flush ever crosses the default 5s SIGTERM window on a slow runner it SIGKILLs to -9 - exactly the failure this widening prevents for the fault_manager beside it. Same gap in the diagnostic and log bridge tests.

src/ros2_medkit_diagnostic_bridge/test/test_integration.test.py
```diff
@@ -42,6 +42,9 @@ def generate_test_description():
             'healing_enabled': True,
             'healing_threshold': 1,
         }],
+        # Give the node room to flush coverage data at shutdown before SIGKILL.
+        sigterm_timeout='30',
+        sigkill_timeout='15',
     )
 
     diagnostic_bridge_node = launch_ros.actions.Node(
```
mfaferek93, line 50: `diagnostic_bridge_node` has no `sigterm_timeout`/`sigkill_timeout`, unlike the `fault_manager_node` above that this PR just widened for gcov-flush-under-coverage. It's a coverage-instrumented C++ binary flushing gcov at shutdown too, and it's the subject under test; with `ALLOWED_EXIT_CODES` rejecting -9 and `TestShutdown` checking all procs, a slow-runner flush past the 5s SIGTERM window SIGKILLs it to -9 - the same failure the widening fixes for its sibling. (Same in the action_status and log bridge tests.)

src/ros2_medkit_log_bridge/test/test_integration.test.py
```diff
@@ -48,6 +48,9 @@ def generate_test_description():
             'healing_enabled': True,
             'healing_threshold': 1,
         }],
+        # Give the node room to flush coverage data at shutdown before SIGKILL.
+        sigterm_timeout='30',
+        sigkill_timeout='15',
     )
 
     log_bridge_node = launch_ros.actions.Node(
```
mfaferek93, line 56: `log_bridge_node` gets no `sigterm_timeout`/`sigkill_timeout` while the `fault_manager_node` above was just widened for the gcov-flush SIGKILL this PR targets. The bridge node is the coverage-instrumented C++ node under test; if its shutdown flush crosses the default 5s SIGTERM window on a slow runner it exits -9, which `TestShutdown` rejects (`ALLOWED_EXIT_CODES` has no -9). Same pattern in action_status/diagnostic.

src/ros2_medkit_diagnostic_bridge/test/test_integration.test.py
bburda: Fixed. diagnostic_bridge_node now sets sigterm_timeout='30' and sigkill_timeout='15', same as the fault_manager_node beside it.

src/ros2_medkit_log_bridge/test/test_integration.test.py
bburda: Fixed. log_bridge_node now sets the same sigterm_timeout='30' and sigkill_timeout='15' windows.

src/ros2_medkit_action_status_bridge/test/test_integration.test.py
bburda: Fixed. action_status_bridge_node now sets sigterm_timeout='30' and sigkill_timeout='15'. I also widened the demo nodes, which are the same case (coverage-instrumented C++ binaries under exit-code checks that reject -9): create_demo_nodes and the inline demo launches in the discovery, graph-provider and beacon feature tests now carry the windows.

mfaferek93 (approved):

PR #560, opened 2026-07-26
fix(gateway): plugin-provided entities own their bulk data and logs

### LINKED ISSUE #558: gateway: bulk-data and logs 404 for plugin-provided entities
## Summary

A fault reported by a plugin-provided (external) entity advertises a rosbag in its
fault detail, the `.mcap` exists on the volume, and `GET
/apps/<entity>/bulk-data/rosbags/<code>` answers `404 Bulk-data not found for this
entity`. Component logs for the same entity are empty for the same reason.

## Cause

Fault scoping and bulk-data scoping resolve an entity's reporting sources
differently.

`collect_app_fqn` (`core/faults/fault_scope.cpp`) checks `external` first and
returns the app's bare entity id, because an external app reports faults under
that id - it has no ROS binding, so `effective_fqn()` is empty.

`HandlerContext::resolve_app_host_fqns` (`http/handlers/handler_context.cpp`)
resolves by `effective_fqn()` only, with no external branch. For an external app
it yields nothing, so `compute_bulkdata_source_filters` returns an empty filter
set and the download ownership loop in `bulkdata_handlers.cpp` never matches the
fault. `LogManager::get_logs` goes through the same resolver.

`compute_bulkdata_source_filters` has two further gaps of the same kind:

- The `APP` branch falls through to `entity.fqn` / `entity.namespace_path`, both
  empty for a plugin app, so an app-scoped bulk-data request resolves to no
  source at all.
- The `COMPONENT` branch never adds the component's own id, unlike
  `collect_component_app_fqns`, so a fault a bridge reports under the component
  id (`PLC_COMMS_LOST`) has no matching filter either.

## Reproduce

1. Run a gateway with a plugin that introspects an external app and reports a
   fault under that entity id.
2. Enable the rosbag black box (`snapshots.rosbag.enabled: true`).
3. Trigger the fault, wait for the bag to be written, then read the fault detail
   and follow its `bulk_data_uri`.

Expected: the bag downloads. Actual: `404`, with the file present under the
snapshot storage path.

## Fix direction

One rule for "what source does this entity own", shared by both paths: external
entities own their bare id, everything else its effective FQN.

mfaferek93:
Closes #558.

Fault scoping and bulk-data scoping disagreed about what source an entity owns,
so a plugin-provided entity could not download the rosbag its own fault detail
advertised.

- The per-app rule (external -> bare entity id, checked before `effective_fqn()`)
  moves out of the anonymous namespace into `faults::resolve_app_source_fqn`, and
  both `collect_app_fqn` and `HandlerContext::resolve_app_host_fqns` use it, so
  there is one definition of the rule rather than two that drift.
- `compute_bulkdata_source_filters` gains the two branches it was missing: an APP
  entity resolves through the same helper instead of falling through to an empty
  `fqn`, and an external COMPONENT owns its own id alongside its hosted apps -
  the same rule `collect_component_app_fqns` already applies for faults.

Logs go through the same resolver, so `LogManager::get_logs` is fixed with it.

Tests: `ExternalAppResolvesToItsBareEntityId`,
`ExternalAppWithStrayBindingStillResolvesToItsId`,
`ComponentWithExternalAppsReturnsBareEntityIds`, `ExternalComponentAlsoOwnsItsOwnId`.
Each fails on the pre-fix resolver. Full gateway suite: 4289 tests, 0 failures,
484 skipped; clang-format clean.

One behaviour change worth naming: a component that hosts external apps no longer
falls through to the `fqn` / `namespace_path` prefix filter, because that
fall-through only fires on an empty source set. That affects a manifest component
whose bags are sourced from a ROS namespace and which also hosts external apps.

### DIFF
--- a/src/ros2_medkit_gateway/include/ros2_medkit_gateway/core/faults/fault_scope.hpp
+++ b/src/ros2_medkit_gateway/include/ros2_medkit_gateway/core/faults/fault_scope.hpp
@@ -33,12 +33,15 @@ namespace faults {
 /// it claim faults it never reported.
 std::string resolve_app_source_fqn(const ThreadSafeEntityCache & cache, const std::string & app_id);
 
-/// Resolve the set of App effective-FQNs that fall within an entity's scope by
-/// walking the entity cache:
-///   - APP: the app's own effective FQN
-///   - COMPONENT: every hosted app's FQN
-///   - AREA: every app under the area and its (recursive) subareas
-///   - FUNCTION: every app hosted directly or via a hosted component
+/// Resolve the set of fault-reporting sources that fall within an entity's
+/// scope by walking the entity cache. Each hosted app contributes what
+/// `resolve_app_source_fqn` yields (bare entity id for an external app,
+/// effective FQN otherwise):
+///   - APP: the app's own source, if it owns one
+///   - COMPONENT: every hosted app's source, plus the component's own id when
+///     the component is external (protocol bridges report under that id)
+///   - AREA: every source under the area and its (recursive) subareas
+///   - FUNCTION: every source hosted directly or via a hosted component
 /// Returns an empty set for SERVER / UNKNOWN.
 ///
 /// Shared by the HTTP fault handlers (`GET /{entity}/faults`) and the ROS 2
--- a/src/ros2_medkit_gateway/include/ros2_medkit_gateway/core/faults/fault_scope.hpp
+++ b/src/ros2_medkit_gateway/include/ros2_medkit_gateway/core/faults/fault_scope.hpp
@@ -26,6 +26,13 @@ class ThreadSafeEntityCache;
 
 namespace faults {
 
+/// Resolve the single fault-reporting source an app owns, or "" when it owns
+/// none. An external app (a plugin-introspected asset with no ROS binding)
+/// reports under its bare entity id; every other app owns its `effective_fqn()`.
+/// An unbound non-external app owns nothing - granting it its bare id would let
+/// it claim faults it never reported.
+std::string resolve_app_source_fqn(const ThreadSafeEntityCache & cache, const std::string & app_id);
+
 /// Resolve the set of App effective-FQNs that fall within an entity's scope by
 /// walking the entity cache:
 ///   - APP: the app's own effective FQN
--- a/src/ros2_medkit_gateway/src/http/handlers/bulkdata_handlers.cpp
+++ b/src/ros2_medkit_gateway/src/http/handlers/bulkdata_handlers.cpp
@@ -14,6 +14,8 @@
 
 #include "ros2_medkit_gateway/core/http/handlers/bulkdata_handlers.hpp"
 
+#include "ros2_medkit_gateway/core/faults/fault_scope.hpp"
+
 #include <algorithm>
 #include <cstdint>
 #include <filesystem>
@@ -109,6 +111,19 @@ std::vector<std::string> compute_bulkdata_source_filters(const ThreadSafeEntityC
     return HandlerContext::resolve_app_host_fqns(cache, cache.get_apps_for_function(entity.id));
   }
 
+  if (entity.type == EntityType::APP) {
+    // A plugin-introspected app has no ROS binding, so fqn and namespace_path
+    // are both empty and the bare-fqn path below would return no filter at all -
+    // its rosbags and logs then belong to nobody and every download 404s. Its
+    // reporting source is its bare id, exactly as fault scoping resolves it.
+    std::string source = faults::resolve_app_source_fqn(cache, entity.id);
+    if (!source.empty()) {
+      std::vector<std::string> filters;
+      filters.push_back(std::move(source));
+      return filters;
+    }
+  }
+
   if (entity.type == EntityType::COMPONENT) {
     // Synthetic / runtime-discovered components have an empty fqn / namespace_path,
     // so the bare-fqn path used to silently return zero source filters and produce
@@ -116,6 +131,13 @@ std::vector<std::string> compute_bulkdata_source_filters(const ThreadSafeEntityC
     // apps first; manifest deployments where the component groups topics rather than
     // nodes still need the namespace prefix path, so fall through if no apps host it.
     auto filters = HandlerContext::resolve_app_host_fqns(cache, cache.get_apps_for_component(entity.id));
+    // An external component reports faults under its own id too (a bridge raises
+    // PLC_COMMS_LOST there), so it owns that source as well - the same rule
+    // faults::collect_component_app_fqns already applies.
+    auto comp = cache.get_component(entity.id);
+    if (comp && comp->external.value_or(false)) {
+      filters.push_back(entity.id);
+    }
     if (!filters.empty()) {
       return filters;
     }
--- a/src/ros2_medkit_gateway/src/http/handlers/bulkdata_handlers.cpp
+++ b/src/ros2_medkit_gateway/src/http/handlers/bulkdata_handlers.cpp
@@ -21,6 +21,7 @@
 #include <filesystem>
 #include <fstream>
 #include <memory>
+#include <set>
 #include <string>
 #include <type_traits>
 #include <unordered_map>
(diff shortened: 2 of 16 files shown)


src/ros2_medkit_gateway/src/http/handlers/bulkdata_handlers.cpp
```diff
@@ -109,13 +111,31 @@ std::vector<std::string> compute_bulkdata_source_filters(const ThreadSafeEntityC
...
 
+  if (entity.type == EntityType::APP) {
+    // A plugin-introspected app has no ROS binding, so fqn and namespace_path
+    // are both empty and the bare-fqn path below would return no filter at all -
+    // its rosbags and logs then belong to nobody and every download 404s. Its
+    // reporting source is its bare id, exactly as fault scoping resolves it.
+    std::string source = faults::resolve_app_source_fqn(cache, entity.id);
+    if (!source.empty()) {
+      return {std::move(source)};
+    }
```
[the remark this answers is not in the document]
mfaferek93: Fixed in d571e98b: the APP branch now builds the vector and `push_back(std::move(source))` instead of returning a braced initializer list, so the string is really moved rather than copied through `initializer_list`.

src/ros2_medkit_gateway/src/http/handlers/bulkdata_handlers.cpp
```diff
@@ -109,13 +111,33 @@ std::vector<std::string> compute_bulkdata_source_filters(const ThreadSafeEntityC
...
 
+  if (entity.type == EntityType::APP) {
+    // A plugin-introspected app has no ROS binding, so fqn and namespace_path
+    // are both empty and the bare-fqn path below would return no filter at all -
+    // its rosbags and logs then belong to nobody and every download 404s. Its
+    // reporting source is its bare id, exactly as fault scoping resolves it.
+    std::string source = faults::resolve_app_source_fqn(cache, entity.id);
+    if (!source.empty()) {
+      std::vector<std::string> filters;
+      filters.push_back(std::move(source));
```
bburda: This lets bare plugin ids reach the download ownership check. That check compares by prefix and has no separator guard. `download` calls `fault_mgr->get_fault(fault_code, sf)`, and `Ros2FaultServiceTransport::get_fault_with_env` verifies the source with `src.rfind(source_id, 0) == 0`. Before this change external entities had an empty filter set, so the check was never reached for them.

Example: two plugin apps named `plc` and `plc_line1`. `GET /apps/plc/bulk-data/rosbags/<code reported by plc_line1>` returns 200 and sends the other asset's bag, because `"plc_line1"` starts with `"plc"`. Entity ids allow letters, digits, `_` and `-`, and there is no separator rule. The ids in the new tests (`plc_hw`, `plc_hw2`, `plc_hw2_device`) already have this prefix relation.

Suggestion: do not use `get_fault(code, source)` as the ownership check. Read the fault once and test it with `faults::fault_in_source_scope`. That function requires an exact match or a `/` boundary. Then the download check and the fault scope use the same rule.

src/ros2_medkit_gateway/src/http/handlers/handler_context.cpp
```diff
@@ -390,11 +390,11 @@ std::vector<std::string> HandlerContext::resolve_app_host_fqns(const ThreadSafeE
...
-    auto app = cache.get_app(app_id);
-    if (!app) {
-      continue;
-    }
-    auto fqn = app->effective_fqn();
+    // Same app -> source resolution as the fault-scope path, so a plugin app
+    // (external, no ROS binding) keeps its bare entity id here too. Resolving
+    // by effective_fqn() alone dropped it, and the empty filter set made every
+    // rosbag download and log query for its component match nothing.
+    auto fqn = faults::resolve_app_source_fqn(cache, app_id);
```
bburda, line 397: This now returns a non-empty list for a component whose only apps are external. That disables the namespace fallback for component logs. In `log_handlers.cpp` the COMPONENT branch enters `if (!host_fqns.empty())` and runs an exact-match query. The `else if (!entity.fqn.empty())` prefix query is never reached.

Example: a manifest component with `fqn = /plant` that hosts one external app. Before, it returned the logs of every `/plant/*` node. Now it returns only the entries stored under the external app id. The PR description names the bulk-data version of this side effect. The log version is not named, and it is the one that loses data.

Suggestion: in the COMPONENT log branch, run both queries when `entity.fqn` is not empty and merge the results.

bburda (commented):

- Could you also look at the APP branch in `log_handlers.cpp` (the last `else`)? It queries `{entity.fqn}`, which is `""` for an external app, so `GET /apps/<external_app>/logs` returns nothing even though the entries are stored under the entity id. The logs claim in the description holds for functions, areas and components only.
- Could you add the external component's own id in the COMPONENT branch there as well? Right now such a component owns its rosbags but not its logs.
- Please check the cyclic subscription samplers in `gateway_node.cpp`. They still resolve with a raw `effective_fqn()` and use a private `filter_faults_by_fqns`, so a subscription on an external app returns `"Entity no longer available"` every cycle, and an external component gets an empty fault list.
- Could you update the docs in this PR? `docs/api/rest.rst` says component logs query child apps "by exact FQN", and `bulkdata_handlers.hpp` still documents the old APP and COMPONENT behaviour of the function you changed.
- Would you mind adding an integration test for the reported 404? `test_external_app_fault_rollup.test.py` and `external_app_fault_manifest.yaml` already set up an external app, so reporting a fault there and downloading the advertised `bulk_data_uri` would cover it.

src/ros2_medkit_gateway/include/ros2_medkit_gateway/core/faults/fault_scope.hpp
```diff
@@ -26,6 +26,13 @@ class ThreadSafeEntityCache;
 
 namespace faults {
 
+/// Resolve the single fault-reporting source an app owns, or "" when it owns
+/// none. An external app (a plugin-introspected asset with no ROS binding)
+/// reports under its bare entity id; every other app owns its `effective_fqn()`.
+/// An unbound non-external app owns nothing - granting it its bare id would let
+/// it claim faults it never reported.
+std::string resolve_app_source_fqn(const ThreadSafeEntityCache & cache, const std::string & app_id);
```
bburda, line 34: The doc block right below this declaration still describes the old rule. It says the function resolves "App effective-FQNs", "APP: the app's own effective FQN" and "COMPONENT: every hosted app's FQN". Neither is true now: an external app resolves to its bare id, and an external component also owns its own id.

Fix: update that block to point at the rule the new helper defines, so the two comments do not disagree.

src/ros2_medkit_gateway/src/http/handlers/bulkdata_handlers.cpp
```diff
@@ -109,13 +111,33 @@ std::vector<std::string> compute_bulkdata_source_filters(const ThreadSafeEntityC
     return HandlerContext::resolve_app_host_fqns(cache, cache.get_apps_for_function(entity.id));
   }
 
+  if (entity.type == EntityType::APP) {
+    // A plugin-introspected app has no ROS binding, so fqn and namespace_path
+    // are both empty and the bare-fqn path below would return no filter at all -
+    // its rosbags and logs then belong to nobody and every download 404s. Its
+    // reporting source is its bare id, exactly as fault scoping resolves it.
+    std::string source = faults::resolve_app_source_fqn(cache, entity.id);
```
bburda: APP and COMPONENT now resolve through the shared rule, but AREA still falls through to the `entity.fqn` / `namespace_path` filter at the bottom of this function, and `list_rosbags_for_entity` matches reporting sources exactly. Example with the shipped `external_app_fault_manifest.yaml`: `GET /areas/plc-cell/faults/<code>` returns a fault whose `bulk_data_uri` points at `/areas/plc-cell/bulk-data/rosbags/<code>`. A GET on that URI returns 404.

Fix: give AREA the same treatment as COMPONENT, resolving hosted apps over `get_components_for_area` and its subareas the way `faults::collect_area_app_fqns` does.

src/ros2_medkit_gateway/src/http/handlers/bulkdata_handlers.cpp
```diff
@@ -109,13 +111,33 @@ std::vector<std::string> compute_bulkdata_source_filters(const ThreadSafeEntityC
...
     // empty descriptor lists plus failed ownership checks on download. Resolve hosted
     // apps first; manifest deployments where the component groups topics rather than
     // nodes still need the namespace prefix path, so fall through if no apps host it.
     auto filters = HandlerContext::resolve_app_host_fqns(cache, cache.get_apps_for_component(entity.id));
+    // An external component reports faults under its own id too (a bridge raises
+    // PLC_COMMS_LOST there), so it owns that source as well - the same rule
+    // faults::collect_component_app_fqns already applies.
+    auto comp = cache.get_component(entity.id);
+    if (comp && comp->external.value_or(false)) {
+      filters.push_back(entity.id);
```
bburda: The FUNCTION branch above (line 111) uses `cache.get_apps_for_function`, and the cache index behind it drops host ids that are not apps, while `faults::collect_function_app_fqns` walks `func->hosts` and also handles Component hosts. So a function that hosts a component lists the fault but cannot download its bag, which is the same mismatch this PR fixes for apps.

Fix: resolve the FUNCTION branch through the fault-scope helper instead of the app-only cache index.

src/ros2_medkit_gateway/src/http/handlers/bulkdata_handlers.cpp
mfaferek93: Fixed in fcf8bc73. download() reads the fault once and checks ownership with faults::fault_in_source_scope over the computed filters, so exact or '/'-boundary only. DownloadOwnershipScopeRejectsPrefixSiblingApp pins the plc vs plc_line1 case.

src/ros2_medkit_gateway/src/http/handlers/handler_context.cpp
mfaferek93: Fixed in 7669b295. The COMPONENT branch (and AREA, which had the identical regression) now runs the exact query over resolved sources plus the entity.fqn prefix query and merges deduped, via the shared logs::query_entity_logs. ComponentMergesExternalAppAndNamespaceEntries covers the /plant case.

src/ros2_medkit_gateway/include/ros2_medkit_gateway/core/faults/fault_scope.hpp
mfaferek93: Fixed in fcf8bc73 - block rewritten to the resolve_app_source_fqn rule (external app owns its bare id, external component also owns its own id). The duplicate stale mapping on HandlerContext::resolve_entity_source_fqns got the same treatment.

src/ros2_medkit_gateway/src/http/handlers/bulkdata_handlers.cpp
mfaferek93: Fixed in fcf8bc73. compute_bulkdata_source_filters routes every entity type through resolve_entity_source_fqns, which reuses collect_area_app_fqns (subarea recursion, external ids). The extended rollup launch test asserts the advertised /areas/plc-cell URI downloads with bytes - it fails on the pre-fix binary with an empty area listing.

src/ros2_medkit_gateway/src/http/handlers/bulkdata_handlers.cpp
mfaferek93: Fixed in fcf8bc73 - same shared resolver, so component hosts come via collect_function_app_fqns. Unit test FunctionWithComponentHostResolvesComponentApps plus the function-level listing assert in the launch test.

mfaferek93:
The outside-diff findings, all on the branch now:

- App logs resolve through the shared rule (bare id for external apps), with the exact-fqn fallback kept when the resolver yields nothing - 7669b295.
- The external component's own id is in the log scope, and component/area logs merge the namespace query instead of losing it.
- Both cyclic samplers use the shared resolvers plus faults::filter_faults_by_sources; the private filter_faults_by_fqns copy is deleted. A cyclic subscription on an external app now returns the fault list instead of 'Entity no longer available'.
- rest.rst, bulkdata_handlers.hpp and fault_scope.hpp describe the one rule.
- test_external_app_fault_rollup.test.py now captures a real bag and downloads every advertised URI (app, area, function) - the area assert fails on the pre-fix binary.

Agreed on the last note: the namespace filter never matched an app source in the listing, so the visible change is only the stricter download check.

bburda (approved):

PR #561, opened 2026-07-26
fix(fault_manager): keep a black box for faults confirmed mid post-roll

### LINKED ISSUE #559: fault_manager: a fault confirmed during another fault's post-roll never gets a recording
## Summary

When a fault confirms while another fault's post-roll recording is still running,
it gets no rosbag at all - not then, and not later. The log line is:

```
Already recording post-fault data, skipping confirmation for '<CODE>'
```

Faults arrive in bursts from a single root cause, so this drops the black box for
precisely the correlated faults, which are the ones worth replaying. With the
shipped default `snapshots.rosbag.duration_after_sec: 1.0`, anything confirming
within a second of the first fault is affected.

## Cause

`RosbagCapture::on_fault_confirmed` returns early while `recording_post_fault_`
is set. The guard exists because one writer is open at a time, but the
confirmation is dropped rather than deferred or attached, so nothing ever writes
a bag for that fault.

## Observed

Two faults from one source within the same second: the first has three bags
across its flaps, the second has a freeze frame and zero bags.

## Fix direction

A burst is one root cause in one window - one recording can serve several faults.
Attaching the in-flight recording to the additional fault (one metadata row per
fault code, same `file_path`) fits the existing schema. Deferring the flush
instead does not work: the ring buffer is not filled during a post-roll, so a
queued confirmation would flush an empty buffer.

Bag lifetime then needs refcounting, so deleting one fault of a burst does not
unlink a file another fault still references, and the storage quota counts a
shared bag once.

mfaferek93:
Closes #559.

A fault that confirmed while another fault's post-roll was running was dropped
("Already recording post-fault data, skipping confirmation"), so the correlated
faults - the ones a burst is actually about - ended up with no black box at all.
With the shipped `duration_after_sec: 1.0` that is anything confirming within a
second of the first fault.

A burst is one root cause in one window, so it gets one recording with several
lookup keys: `attach_to_active_recording()` registers the additional fault
against the in-flight bag (capped at 32 codes) and one metadata row per code is
written when the bag closes. The `rosbag_files` schema already allowed several
codes to name the same `file_path`.

Deferring the flush instead does not work: the ring buffer is not filled during a
post-roll (messages go straight to the writer), so a queued confirmation would
flush an empty buffer.

Shared bags need refcounted cleanup, so `delete_rosbag_file` / `store_rosbag_file`
unlink a bag directory only when no other fault still references the path, and
`get_total_rosbag_storage_bytes` counts a shared bag once instead of once per
fault - otherwise the quota evicts bags that fit. Both storage backends.

The recording state (`current_fault_code_`, `current_bag_path_`,
`recording_post_fault_`) is now published under `post_fault_timer_mutex_`, which
also removes an unlocked read of `current_bag_path_`.

Tests: `FaultConfirmedDuringPostRollGetsTheSameBag`,
`ClearingOneFaultOfABurstKeepsTheSharedBag`,
`SharedRosbagSurvivesUntilTheLastFaultIsDeleted`,
`SharedRosbagCountsOnceTowardsStorageTotal`. Full fault_manager suite green;
clang-format clean.

Known edge: with `duration_after_sec: 0` there is no post-roll, and a burst there
still loses bags because the first flush drains the ring buffer. Different
mechanism, left alone.


src/ros2_medkit_fault_manager/src/sqlite_fault_storage.cpp
```diff
@@ -979,10 +984,23 @@ bool SqliteFaultStorage::delete_rosbag_file(const std::string & fault_code) {
...
+
 size_t SqliteFaultStorage::get_total_rosbag_storage_bytes() const {
   std::lock_guard<std::mutex> lock(mutex_);
 
-  SqliteStatement stmt(db_, "SELECT COALESCE(SUM(size_bytes), 0) FROM rosbag_files");
+  // Sum per bag, not per fault: one recording can back a burst of correlated
+  // faults, and double-counting it would evict bags that still fit the quota.
+  SqliteStatement stmt(
+      db_,
+      "SELECT COALESCE(SUM(size_bytes), 0) FROM (SELECT MAX(size_bytes) AS size_bytes FROM rosbag_files "
```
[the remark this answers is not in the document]
mfaferek93: Confirmed and fixed in ccaeff09. Eviction now groups the per-fault rows back into bags and frees a shared bag as one unit (`evict_bags_over_quota` in rosbag_capture.cpp), so its bytes are subtracted once and all referencing rows are deleted together, which is what actually removes the directory; the subtraction is also saturating so the quota can never be satisfied by underflow. New tests `EvictionFreesASharedBagOnceAndKeepsTheQuotaHonest` (shared 3-row 1000B bag + solo 600B bag, quota 1000) and `EvictionKeepsGoingWhenTheSharedBagIsNotEnough` cover it; on the old per-row loop they fail with the shared directory still on disk and the total stuck at 1600.

src/ros2_medkit_fault_manager/src/fault_storage.cpp
```diff
@@ -415,21 +416,38 @@ bool InMemoryFaultStorage::delete_rosbag_file(const std::string & fault_code) {
...
   std::lock_guard<std::mutex> lock(mutex_);
 
-  size_t total = 0;
+  // Sum per bag, not per fault: one recording can back a burst of correlated
+  // faults, and double-counting it would evict bags that still fit the quota.
+  std::map<std::string, size_t> bytes_per_path;
   for (const auto & [code, info] : rosbag_files_) {
-    total += info.size_bytes;
+    bytes_per_path[info.file_path] = info.size_bytes;
+  }
```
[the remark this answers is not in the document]
mfaferek93: Confirmed and fixed in ccaeff09. `InMemoryFaultStorage::get_total_rosbag_storage_bytes()` now takes `std::max` per path instead of last-write-wins, matching the SQLite backend MAX(size_bytes). New test `SharedBagTotalTakesTheLargestRowPerPath` uses two shared bags with opposite fault-code orderings, since rows iterate in fault_code order and a single-ordering test would pass by luck; it reports 2560 instead of 6144 on the old code.

src/ros2_medkit_fault_manager/src/rosbag_capture.cpp
```diff
@@ -203,6 +203,7 @@ void RosbagCapture::stop() {
       post_fault_timer_->cancel();
       post_fault_timer_.reset();
     }
```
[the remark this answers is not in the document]
mfaferek93: Confirmed against the current branch, not already handled: `stop()` only cancelled the timer and cleared `attached_fault_codes_`, leaving `recording_post_fault_` true and `active_writer_` open, so the bag lost its metadata row entirely and the next confirmation after a restart attached to a recording whose timer was gone. Fixed in ccaeff09 by extracting `finalize_post_fault_recording()` (idempotent, tests the guard under `post_fault_timer_mutex_` so a timer firing concurrently with `stop()` is settled) and calling it from both `post_fault_timer_callback()` and `stop()`. New test `StopDuringPostRollFinalisesTheRecording` stops mid post-roll then restarts and confirms again; on the old code the first bag is never stored.

src/ros2_medkit_fault_manager/src/rosbag_capture.cpp
```diff
@@ -196,14 +197,12 @@ void RosbagCapture::stop() {
...
-      post_fault_timer_->cancel();
-      post_fault_timer_.reset();
-    }
-  }
+  // A post-roll in flight owns an open writer and a bag that has no metadata row
+  // yet. Finalise it instead of just cancelling the timer: the faults of that
+  // burst keep their black box, the writer is closed, and the recording state is
+  // cleared so a confirmation after a restart opens its own bag rather than
+  // attaching to a recording whose timer is gone.
+  finalize_post_fault_recording();
```
bburda, line 205: `finalize_post_fault_recording()` writes to storage with no `try` in the chain, and `store_rosbag_file` / `delete_rosbag_file` throw on the SQLite backend (`sqlite_fault_storage.cpp:923`, `:929`, `:981`). `stop()` is only reached from destructors (`~RosbagCapture()` line 179, `~FaultManagerNode()` at `fault_manager_node.cpp:379`), which are `noexcept`, so a busy or full database during shutdown ends in `std::terminate`.

Fix: wrap the storage part of `finalize_post_fault_recording()` in try/catch and log the failure, the same way `fault_manager_node.cpp:385` guards `audit_log_->append()`.

src/ros2_medkit_fault_manager/src/rosbag_capture.cpp
```diff
@@ -240,15 +239,39 @@ void RosbagCapture::on_fault_prefailed(const std::string & fault_code) {
...
 
-  // Don't start a new recording if we're already recording post-fault
-  if (recording_post_fault_.load()) {
-    RCLCPP_WARN(node_->get_logger(), "Already recording post-fault data, skipping confirmation for '%s'",
-                fault_code.c_str());
+  // Faults arrive in bursts from one root cause, so a confirmation landing inside
+  // an in-flight post-roll is exactly the correlated fault whose black box matters
+  // most. It shares the recording window, so attach it to the running bag instead
+  // of leaving it with no recording at all.
+  if (attach_to_active_recording(fault_code)) {
```
bburda, line 304: An attached fault returns here before `resolve_entity_topics()` on line 282, so in the default `entity` mode the capture filter stays scoped to the first fault's node while line 1024 still writes a row for the attached fault. `GET /apps/B/bulk-data/rosbags/FAULT_B` then returns a bag with none of B's topics, which is the opposite of what `docs/tutorials/snapshots.rst` promises for this mode.

Fix: add the attached fault's entity topics to `active_capture_topics_` on attach, or do not attach when the two topic sets do not overlap.

src/ros2_medkit_fault_manager/src/rosbag_capture.cpp
```diff
@@ -945,24 +1006,28 @@ void RosbagCapture::post_fault_timer_callback() {
...
   info.duration_sec = config_.duration_sec + config_.duration_after_sec;
   info.size_bytes = bag_size;
   info.created_at_ns = get_wall_clock_ns();
 
   storage_->store_rosbag_file(info);
+  // One recording, one row per fault it covers: the correlated faults that
+  // confirmed inside this window each need their own lookup key to serve it.
+  for (const auto & code : attached) {
+    info.fault_code = code;
+    storage_->store_rosbag_file(info);
```
bburda: A fault cleared during the post-roll still gets a row here, because `on_fault_cleared` (line 345) only calls `delete_rosbag_file`, which does nothing while the row does not exist yet, and it never removes the code from `attached_fault_codes_`. That leftover row keeps `path_shared_with_other_fault` true, so clearing the rest of the burst never unlinks the bag and `auto_cleanup` stops working for it.

Fix: in `on_fault_cleared`, take `post_fault_timer_mutex_` and erase the code from `attached_fault_codes_`.

bburda (commented):

- Could you update the docs? `docs/tutorials/snapshots.rst` and `docs/config/fault-manager.rst` still say rosbag "records one fault at a time", as do the package README and `config/fault_manager.yaml`. The `auto_cleanup` row says bags are deleted when faults are cleared, the `entity` mode row says only the faulting node's topics are written, and "Bags are named `fault_{code}_{timestamp}/`" now uses another fault's code. It would also help to say that a shared bag counts once against `max_total_storage_mb` and that eviction removes a whole burst.
- Could you change `test_05_multiple_faults_separate_bags` to wait for the first bag instead of sleeping? It asserts the two faults get different bags, and it passes only because its 1.0 s sleep is longer than the 0.5 s post-roll. Under load the second fault attaches and the assert fails.
- Would you add a launch test for this? Reporting two faults with no sleep between them in `test_rosbag_integration.test.py`, and asserting they return the same `file_path`, would cover the fix. That test fails on main.
- Could you also put a shared recording id in the `x-medkit` block? `GET /{entity}/bulk-data/rosbags` now lists one descriptor per fault for a shared bag, each with the full size, and a client cannot tell they are one file.

src/ros2_medkit_fault_manager/src/rosbag_capture.cpp
```diff
@@ -240,15 +239,39 @@ void RosbagCapture::on_fault_prefailed(const std::string & fault_code) {
   }
 }
 
+bool RosbagCapture::attach_to_active_recording(const std::string & fault_code) {
+  std::lock_guard<std::mutex> lock(post_fault_timer_mutex_);
+  if (!recording_post_fault_.load()) {
+    return false;
+  }
+  if (fault_code == current_fault_code_) {
```
bburda, line 247: Past the cap this returns true, so `on_fault_confirmed` returns and fault 33 of a burst gets no bag and no metadata row at all. That is the same symptom the PR fixes below the cap. The only trace is this warning.

Fix: document the cap next to `duration_after_sec` and add a test that holds the window open and confirms more than 32 faults, so the drop is a pinned decision rather than a silent limit.

src/ros2_medkit_fault_manager/src/rosbag_capture.cpp
bburda: Each row here is its own SQLite transaction, so a crash or a kill in the middle of the loop leaves some faults of the burst with a row and some without. `evict_bags_over_quota` has a sharper version of the same problem. It deletes the rows of one bag one at a time after the directory is already unlinked, so an interruption leaves rows pointing at a bag that is gone.

Fix: wrap the row loop, and the per-path delete loop in eviction, in one `BEGIN IMMEDIATE` / `COMMIT`.

src/ros2_medkit_fault_manager/src/fault_storage.cpp
```diff
@@ -387,7 +387,8 @@ void InMemoryFaultStorage::store_rosbag_file(const RosbagFileInfo & info) {
   // Delete existing bag file if present (prevent orphaned files on re-confirm)
   auto it = rosbag_files_.find(info.fault_code);
   if (it != rosbag_files_.end()) {
-    if (it->second.file_path != info.file_path) {
+    if (it->second.file_path != info.file_path &&
```
bburda, line 390: This new sharing guard has no test in either backend. No test in the repo re-stores an existing fault code with a different `file_path`. That branch decides between leaving a bag on disk forever and deleting one a sibling fault still points at.

Fix: two cases per backend. Store `X -> A` solo then `X -> B` and assert A is gone, then store `X -> A`, `Y -> A`, `X -> B` and assert A survives and `Y` still resolves.

src/ros2_medkit_fault_manager/test/test_sqlite_storage.cpp
```diff
@@ -1184,6 +1184,63 @@ TEST_F(SqliteFaultStorageTest, GetAllRosbagFilesReturnsSortedByCreatedAt) {
...
+
+TEST_F(SqliteFaultStorageTest, SharedRosbagCountsOnceTowardsStorageTotal) {
+  using ros2_medkit_fault_manager::RosbagFileInfo;
+
+  RosbagFileInfo shared;
+  shared.fault_code = "ROOT_CAUSE";
+  shared.file_path = "/tmp/shared.mcap";
+  shared.format = "mcap";
+  shared.duration_sec = 5.0;
+  shared.size_bytes = 1000;
```
bburda: Both shared rows use `size_bytes = 1000`, so `MAX`, `MIN` or picking any row all give 1500. The new `SELECT MAX(size_bytes) ... GROUP BY file_path` is therefore not verified by this test, and neither is its parity with the in-memory `std::max`.

Fix: give the two shared rows different sizes, the way `SharedBagTotalTakesTheLargestRowPerPath` does for the in-memory backend.

src/ros2_medkit_fault_manager/test/test_rosbag_capture.cpp
```diff
@@ -439,6 +442,115 @@ TEST_F(RosbagCaptureTest, ZeroMaxTotalStorageHandled) {
...
+  EXPECT_FALSE(storage_->get_rosbag_file("ROOT_CAUSE").has_value());
+  EXPECT_FALSE(storage_->get_rosbag_file("CORRELATED_A").has_value());
+  EXPECT_FALSE(storage_->get_rosbag_file("CORRELATED_B").has_value());
+
+  // The newer bag fits under the quota and must survive.
+  EXPECT_TRUE(storage_->get_rosbag_file("UNRELATED").has_value());
+  EXPECT_TRUE(std::filesystem::exists(solo_path));
+}
+
+TEST_F(RosbagCaptureTest, EvictionKeepsGoingWhenTheSharedBagIsNotEnough) {
```
bburda, line 525: With 1000 + 600 against a quota of 500, every implementation evicts both bags, including a naive per-row one, so this test does not discriminate. The underflow it names in its comment is also unreachable, because `current_bytes` is seeded from the same per-path maxima that are subtracted. It never creates the directories either, so nothing on disk is checked.

Fix: use shared rows of differing sizes and a quota that a per-row credit would satisfy early, or drop the test and leave the property to `EvictionFreesASharedBagOnceAndKeepsTheQuotaHonest`.

src/ros2_medkit_fault_manager/src/rosbag_capture.cpp
mfaferek93: Fixed in 0aacfd20 - the storage half of finalize_post_fault_recording runs in try/catch with a WARN, same shape as the audit_log_->append guard.

src/ros2_medkit_fault_manager/src/rosbag_capture.cpp
mfaferek93: Fixed in 0aacfd20, extend option: attach widens the live filter with the attached fault's entity topics (rosbag2 allows create_topic mid-flight); an unresolvable scope clears the filter. AttachedFaultTopicsJoinTheEntityScopedRecording asserts the attached node's topic lands in the shared bag's metadata and fails on the pre-fix code. One caveat documented in snapshots.rst: the attached fault gets no pre-roll of its own topics, the buffer was flushed under the first fault's filter.

src/ros2_medkit_fault_manager/src/rosbag_capture.cpp
mfaferek93: Fixed in 0aacfd20. on_fault_cleared erases the code from attached_fault_codes_ under post_fault_timer_mutex_, and the same defect for a cleared primary is handled too (row skipped at finalize, bag discarded when the whole burst cleared). Three tests, each verified failing pre-fix.

src/ros2_medkit_fault_manager/src/rosbag_capture.cpp
mfaferek93: Fixed in 0aacfd20 + ae6b21da: cap documented next to duration_after_sec in fault-manager.rst and the yaml comment, and AttachmentCapDropsFaultsPastIt pins it - 34 faults in one window, rows for 33 sharing one path, no row and no second bag for the last.

src/ros2_medkit_fault_manager/src/rosbag_capture.cpp
mfaferek93: Fixed in 0aacfd20: store_rosbag_files and delete_rosbag_files run one BEGIN IMMEDIATE per burst, delete removes rows first and unlinks only after COMMIT. c8314ec2 makes store defer replaced-bag unlinks the same way, so a ROLLBACK never leaves rows pointing at unlinked bags. In-memory backend keeps the plain loop.

src/ros2_medkit_fault_manager/src/fault_storage.cpp
mfaferek93: Fixed in 0aacfd20 - both cases in both backends: RestoreWithNewPathUnlinksTheOldExclusiveBag / RestoreWithNewPathKeepsTheBagASiblingStillReferences in test_sqlite_storage and the same pair in-memory in test_fault_manager.

src/ros2_medkit_fault_manager/test/test_sqlite_storage.cpp
mfaferek93: Fixed in 0aacfd20 - the shared rows now differ per path in both insert orders (1000/300 and 200/800), mirroring SharedBagTotalTakesTheLargestRowPerPath, so neither first- nor last-write-wins can fake the MAX.

src/ros2_medkit_fault_manager/test/test_rosbag_capture.cpp
mfaferek93: Reworked in 0aacfd20 rather than dropped: shared rows 500/400 plus a solo bag, quota 300, real directories. A per-row credit believes itself under quota after the shared bag and stops with the solo bag still over budget; the test demands both evictions, total 0 and both dirs gone. That covers the over-credit early-stop mode the other eviction test does not, so keeping it adds value.

mfaferek93:
The outside-diff items:

- Docs updated everywhere the one-bag story lived (snapshots.rst, fault-manager.rst, package README, fault_manager.yaml): burst semantics, auto_cleanup on the last fault of the burst, entity-mode topic widening, attached faults under the first fault's directory name, once-per-bag accounting and whole-burst eviction - ae6b21da.
- test_05 no longer sleeps: it waits for the previous fault's bag to be finalized before the next report - 0aacfd20.
- New test_07_burst_faults_share_one_recording reports two faults back to back and asserts one shared file_path; passes on the branch, run twice.
- Descriptors carry x-medkit.recording_id (bag directory basename) so clients can group a burst - 3238f182, unit tests plus a live assert in test_bulk_data_api.test.py.

bburda (approved):

PR #563, opened 2026-07-26
fix(gateway): freeze-frame the faults that were already raised at startup

### LINKED ISSUE #562: gateway: faults already raised at startup never get a freeze frame
## Summary

A fault that is already active when the gateway starts never gets an entity
freeze frame. Plugins begin polling - and reporting - while
`EntityFreezeFrameCapture` is still being wired, so the `EVENT_CONFIRMED` that
would trigger the capture is published before the subscription exists.

## Evidence

Gateway startup, same process, 15 ms apart:

```
[INFO] [plugin.<protocol>]: Fault active: PLC_MINOR_FAULT on <entity>_device
[INFO] [fault_manager]: New fault reported: PLC_MINOR_FAULT (source=<entity>_device)
[INFO] [ros2_medkit_gateway]: EntityFreezeFrameCapture initialized, subscribed to /fault_manager/events
```

The fault detail then carries a rosbag but no freeze frame, and never gains one
while it stays raised - the capture only fires on the confirm edge.

## Why it matters

This is the normal case, not an edge case: a device that is in fault stays in
fault, so every restart of the gateway loses the freeze frames for everything
standing at that moment. `entity_freeze_frame.enabled` defaults to true, so the
feature silently under-delivers exactly when a customer restarts to look at a
problem.

## Cause

`GatewayNode` loads and starts plugins during construction, while
`init_entity_freeze_frame_capture()` runs from `main.cpp` after the shared
subscription executor exists (the executor is built after the node). The
subscription is `QoS(100).reliable()` with volatile durability, so nothing that
was published before it joined is delivered.

## Fix direction

Frames are process-local (`frames_` in the capture), so a catch-up restores them
rather than duplicating anything: once the capture is live, list the confirmed
faults and capture a frame for each that has none yet. Deferring plugin start
until after the capture is wired would also work but reorders the plugin
lifecycle against entity introspection.

A frame captured this way holds the values at catch-up time, not at confirm
time, which is worth stating in the field description.

mfaferek93:
Closes #562.

Plugins start reporting during `GatewayNode` construction, while the capture is
wired later from `main.cpp` (it needs the shared subscription executor, which is
built after the node). A device that is already in fault confirms in that window,
its `EVENT_CONFIRMED` is published to nobody, and since the capture only fires on
the confirm edge the fault stays frameless for as long as it is raised.

`capture_standing_fault_frames()` runs once the capture is live: it lists the
confirmed faults and enqueues a capture for each fault that has no frames yet.
Frames are process-local, so this restores what the missed edge would have
produced rather than duplicating anything, and `capture_missed_fault()` is a
no-op when the live event already captured the fault.

It runs detached, after `wait_for_services` with a bounded timeout, so startup
never blocks on the fault services being up. The synthetic event goes through the
same queue and worker as a real one, so the backlog bound and the plugin-call
threading rules are unchanged.

Worth naming: a frame captured this way holds the values at catch-up time, a
second or so after the confirm, not the values at the confirm itself. For a fault
that is standing at boot that is the only honest option - the earlier values no
longer exist anywhere.

### DIFF
--- a/src/ros2_medkit_gateway/src/entity_freeze_frame_capture.cpp
+++ b/src/ros2_medkit_gateway/src/entity_freeze_frame_capture.cpp
@@ -15,6 +15,7 @@
 #include "ros2_medkit_gateway/entity_freeze_frame_capture.hpp"
 
 #include <chrono>
+#include <thread>
 
 #include "ros2_medkit_gateway/fault_manager_paths.hpp"
 
@@ -28,6 +29,10 @@ constexpr size_t kMaxQueuedEvents = 64;
 /// fallback_logged_ bound: past this the set resets (re-arming one warn per
 /// code) instead of growing forever on churny/synthetic fault codes.
 constexpr size_t kMaxLoggedFaultCodes = 1024;
+
+/// Bound on waiting for the events subscription to match a publisher before
+/// the standing-fault snapshot; past it the catch-up proceeds best-effort.
+constexpr std::chrono::seconds kEventsMatchTimeout{10};
 
 /// Read a string field totally: json::value() throws type_error.302 when the
 /// key is present but not a string, and plugin content is untrusted.
@@ -50,6 +55,10 @@ EntityFreezeFrameCapture::EntityFreezeFrameCapture(rclcpp::Node * node, ros2_com
   // the subscription itself is created on the executor's dedicated _sub node so
   // it never races rcl's hash-map on the main node (issue #375).
   const auto fault_events_topic = build_fault_manager_events_topic(node);
+  // Volatile on purpose: a transient_local reader does not match the fault
+  // manager's volatile publisher, so upgrading the QoS here would break
+  // matching against older fault managers. The startup catch-up covers the
+  // pre-match window instead.
   auto slot = ros2_common::Ros2SubscriptionSlot::create_typed<ros2_medkit_msgs::msg::FaultEvent>(
       exec, fault_events_topic, rclcpp::QoS(100).reliable(),
       [this](std::shared_ptr<const ros2_medkit_msgs::msg::FaultEvent> msg) {
@@ -69,21 +78,23 @@ EntityFreezeFrameCapture::EntityFreezeFrameCapture(rclcpp::Node * node, ros2_com
 }
 
 EntityFreezeFrameCapture::~EntityFreezeFrameCapture() {
-  // Not a synchronous barrier: the slot posts an async destroy with a bounded
-  // deadline, so use-after-free safety also relies on the owner tearing down
-  // the subscription executor (joining its worker) before this object's node
-  // dies - keep that ordering in main.cpp / gateway_node shutdown.
-  subscription_slot_.reset();
   {
     std::lock_guard<std::mutex> lock(queue_mutex_);
     stop_ = true;
   }
   queue_cv_.notify_all();
   // Joins through any in-flight plugin call: a hung read delays shutdown
-  // rather than leaving a capture racing plugin unload.
+  // rather than leaving a capture racing plugin unload. Must precede the slot
+  // reset - the capture thread reads the slot's publisher count.
   if (capture_thread_.joinable()) {
     capture_thread_.join();
   }
+  // Not a synchronous barrier: the slot posts an async destroy with a bounded
+  // deadline, so use-after-free safety also relies on the owner tearing down
+  // the subscription executor (joining its worker) before this object's node
+  // dies - keep that ordering in main.cpp / gateway_node shutdown. Late
+  // callbacks before that teardown see stop_ and drop their event.
+  subscription_slot_.reset();
 }
 
 std::vector<EntityFreezeFrameCapture::Frame>
@@ -134,6 +145,38 @@ bool EntityFreezeFrameCapture::values_have_data(const nlohmann::json & values) {
     return false;
   }
   return !values.is_null();
+}
+
+std::optional<std::vector<EntityFreezeFrameCapture::StandingFault>>
+EntityFreezeFrameCapture::standing_faults_from_list_reply(const nlohmann::json & data) {
+  if (!data.is_object()) {
+    return std::nullopt;
+  }
+  // ListFaults answers under "faults" (see Ros2FaultServiceTransport).
+  const auto faults = data.find("faults");
+  if (faults == data.end() || !faults->is_array()) {
+    return std::nullopt;
+  }
+  std::vector<StandingFault> standing;
+  for (const auto & item : *faults) {
+    if (!item.is_object()) {
+      continue;
+    }
+    const auto code = item.find("fault_code");
+    const auto sources = item.find("reporting_sources");
+    if (code == item.end() || !code->is_string() || sources == item.end() || !sources->is_array()) {
+      continue;
+    }
+    StandingFault fault;
+    fault.fault_code = code->get<std::string>();
+    for (const auto & src : *sources) {
+      if (src.is_string()) {
+        fault.reporting_sources.push_back(src.get<std::string>());
+      }
+    }
+    standing.push_back(std::move(fault));
+  }
+  return standing;
 }
 
 std::optional<EntityFreezeFrameCapture::Frame>
@@ -210,37 +253,92 @@ void EntityFreezeFrameCapture::on_fault_event(const ros2_medkit_msgs::msg::Fault
   queue_cv_.notify_one();
 }
 
+void EntityFreezeFrameCapture::wait_for_events_publisher(const std::function<bool()> & should_abort) const {
+  if (!subscription_slot_) {
+    return;
+  }
+  const auto deadline = std::chrono::steady_clock::now() + kEventsMatchTimeout;
+  while (std::chrono::steady_clock::now() < deadline && !should_abort()) {
+    try {
+      if (subscription_slot_->publisher_count() > 0) {
+        return;
+      }
+    } catch (const std::exception &) {
+      return;  // graph query failed; proceed best-effort
+    }
+    std::this_thread::sleep_for(std::chrono::milliseconds(50));
+  }
+}
+
 void EntityFreezeFrameCapture::capture_standing_faults() {
   if (!standing_lister_) {
     return;
   }
+  const auto should_abort = [this] {
+    std::lock_guard<std::mutex> lock(queue_mutex_);
+    return stop_;
+  };
+  // Snapshot only after the events subscription has matched: a confirm
+  // published after the lister's snapshot but before the reader matched would
+  // be missed by both paths.
+  wait_for_events_publisher(should_abort);
+  if (should_abort()) {
+    return;
+  }
   // The lister blocks until the fault services answer, so it runs here rather
-  // than on a thread of its own: this thread is joined by the destructor, which
-  // is what keeps that wait from outliving the node it waits on.
-  const auto standing = standing_lister_();
+  // than on a thread of its own: this thread is joined by the destructor, and
+  // should_abort is what lets that join interrupt the wait.
+  std::vector<StandingFault> standing;
+  try {
+    standing = standing_lister_(should_abort);
+  } catch (const std::exception & e) {
+    RCLCPP_WARN(logger_, "Entity freeze-frame startup catch-up failed: standing-fault lister threw: %s", e.what());
+    return;
+  } catch (...) {
+    RCLCPP_WARN(logger_, "Entity freeze-frame startup catch-up failed: standing-fault lister threw");
+    return;
+  }
+  // Codes with a confirm already queued belong to the drain loop: capturing
+  // them here too would read the plugin twice for one confirm.
+  std::unordered_set<std::string> queued_codes;
+  {
+    std::lock_guard<std::mutex> lock(queue_mutex_);
+    if (stop_) {
+      return;
+    }
+    for (const auto & queued : queue_) {
+      queued_codes.insert(queued->fault.fault_code);
+    }
+  }
   size_t framed = 0;
+  size_t over_cap = 0;
   for (const auto & fault : standing) {
-    {
-      std::lock_guard<std::mutex> lock(queue_mutex_);
-      if (stop_) {
-        return;
-      }
+    if (should_abort()) {
+      return;
     }
     if (fault.fault_code.empty() || fault.reporting_sources.empty()) {
       continue;
     }
-    {
-      std::lock_guard<std::mutex> lock(mutex_);
-      if (frames_.count(fault.fault_code) != 0) {
-        continue;  // the live event beat us to it
-      }
+    if (queued_codes.count(fault.fault_code) != 0) {
+      continue;
+    }
+    if (framed >= max_faults_) {
+      ++over_cap;  // storing more would FIFO-evict this catch-up's own frames
+      continue;
     }
     ros2_medkit_msgs::msg::FaultEvent event;
     event.event_type = ros2_medkit_msgs::msg::FaultEvent::EVENT_CONFIRMED;
     event.fault.fault_code = fault.fault_code;
     event.fault.reporting_sources = fault.reporting_sources;
-    capture_for_event(event);
-    ++framed;
+    if (capture_for_event(event, /*startup_catchup=*/true)) {
+      ++framed;
+    }
+  }
+  if (over_cap > 0) {
+    RCLCPP_WARN(logger_,
+                "Entity freeze-frame startup catch-up truncated: %zu standing fault(s) beyond the retained-frame "
+                "bound of %zu",
+                over_cap, max_faults_);
   }
   if (framed > 0) {
     RCLCPP_INFO(logger_, "Entity freeze-frame: captured %zu fault(s) that were already confirmed at startup", framed);
@@ -268,7 +366,8 @@ void EntityFreezeFrameCapture::capture_worker() {
   }
 }
 
-void EntityFreezeFrameCapture::capture_for_event(const ros2_medkit_msgs::msg::FaultEvent & event) {
+bool EntityFreezeFrameCapture::capture_for_event(const ros2_medkit_msgs::msg::FaultEvent & event,
+                                                 bool startup_catchup) {
   const std::string & fault_code = event.fault.fault_code;
 
   std::vector<Frame> frames;
@@ -304,7 +403,12 @@ void EntityFreezeFrameCapture::capture_for_event(const ros2_medkit_msgs::msg::Fa
   }
 
   if (frames.empty()) {
-    return;
+    return false;
+  }
+  if (startup_catchup) {
+    for (auto & frame : frames) {
+      frame.startup_catchup = true;
+    }
   }
 
   std::lock_guard<std::mutex> lock(mutex_);
@@ -318,6 +422,7 @@ void EntityFreezeFrameCapture::capture_for_event(const ros2_medkit_msgs::msg::Fa
   frames_[fault_code] = std::move(frames);
 
   RCLCPP_DEBUG(logger_, "Captured entity freeze-frame(s) for fault '%s'", fault_code.c_str());
+  return true;
 }
 
 }  // namespace ros2_medkit_gateway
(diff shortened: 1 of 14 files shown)


src/ros2_medkit_gateway/src/gateway_node.cpp
```diff
@@ -1817,6 +1817,49 @@ void GatewayNode::init_entity_freeze_frame_capture(ros2_common::Ros2Subscription
...
+  // without a catch-up every fault the box comes up with is frameless for as
+  // long as it stays raised. Runs detached: the fault services may not be up
+  // yet, and startup must not block on them.
+  if (!entity_freeze_frame_capture_ || !fault_service_transport_) {
+    return;
+  }
+  std::thread([this] {
+    if (!fault_service_transport_->wait_for_services(std::chrono::seconds(10))) {
+      return;
+    }
```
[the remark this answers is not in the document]
mfaferek93: Rewritten since this comment: there is no detached thread any more. The gateway now only passes a `StandingFaultLister` callback, and the catch-up runs on the capture's own `capture_thread_` (`entity_freeze_frame_capture.cpp:251`, first statement of `capture_worker()`). That thread is created in the constructor and joined in `~EntityFreezeFrameCapture` (`entity_freeze_frame_capture.cpp:84-86`), so the `wait_for_services` block cannot outlive the node it waits on. `grep -rn 'detach()' src/ros2_medkit_gateway` finds nothing. Current head c146a1cf.

src/ros2_medkit_gateway/src/entity_freeze_frame_capture.cpp
```diff
@@ -209,6 +209,31 @@ void EntityFreezeFrameCapture::on_fault_event(const ros2_medkit_msgs::msg::Fault
...
+  {
+    std::lock_guard<std::mutex> lock(mutex_);
+    if (frames_.count(fault_code) != 0) {
+      return;  // the live event already captured it
+    }
+  }
+  auto event = std::make_shared<ros2_medkit_msgs::msg::FaultEvent>();
+  event->event_type = ros2_medkit_msgs::msg::FaultEvent::EVENT_CONFIRMED;
+  event->fault.fault_code = fault_code;
+  event->fault.reporting_sources = reporting_sources;
```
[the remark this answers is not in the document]
mfaferek93: There is no public `capture_missed_fault()` any more; the catch-up is `capture_standing_faults()`, called as the first statement of `capture_worker()` (`entity_freeze_frame_capture.cpp:251`) before the drain loop is entered. `on_fault_event` only enqueues (`entity_freeze_frame_capture.cpp:199-209`), and one thread owns both stages, so a live EVENT_CONFIRMED that arrives during the catch-up is drained strictly after it and its `frames_[fault_code] = ...` assignment (`entity_freeze_frame_capture.cpp:318`) overwrites the synthetic frame. The ordering is live-wins, not catch-up-wins - the `frames_.count` check at line 234 is only a cheap belt-and-braces skip, not the mechanism. Added `LiveConfirmFrameWinsOverStandingCatchUp` to pin exactly this: the lister blocks until the confirm is in flight, and the test asserts the surviving frame is the live entity. Commit c146a1cf.

src/ros2_medkit_gateway/src/entity_freeze_frame_capture.cpp
```diff
@@ -209,6 +209,31 @@ void EntityFreezeFrameCapture::on_fault_event(const ros2_medkit_msgs::msg::Fault
...
+  {
+    std::lock_guard<std::mutex> lock(queue_mutex_);
+    if (stop_ || queue_.size() >= kMaxQueuedEvents) {
+      return;
+    }
+    queue_.push_back(std::move(event));
+  }
+  queue_cv_.notify_one();
+}
```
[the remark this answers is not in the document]
mfaferek93: Half of this was already covered by `StandingFaultsAreFramedWithoutAnyEvent` (frames appear from the catch-up with no event published at all); the no-overwrite half was not, so I added `LiveConfirmFrameWinsOverStandingCatchUp` in c146a1cf (`test/test_entity_freeze_frame_capture.cpp:278`). Both paths carry the same fault code with distinct entity ids, the lister holds the catch-up until the live confirm is in flight, and the test asserts the catch-up sampled first yet the surviving frame is `route_live_app` at 99.0 - re-checked after a settle window. Verified it has teeth: with the catch-up moved onto its own thread and the skip guard removed, it fails on `settled[0].entity_id Which is: "route_standing_app"`. On the real code 20/20 pass in `test_entity_freeze_frame_capture`.

src/ros2_medkit_gateway/include/ros2_medkit_gateway/gateway_node.hpp
```diff
@@ -189,6 +189,10 @@ class GatewayNode : public rclcpp::Node {
    */
   void init_entity_freeze_frame_capture(ros2_common::Ros2SubscriptionExecutor & exec);
 
+  /// Freeze-frame the faults that were already confirmed when the capture came
+  /// up (see the definition for why they are missed otherwise). Detached.
+  void capture_standing_fault_frames();
+
```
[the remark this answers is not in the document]
mfaferek93: Moot at the current head - `gateway_node.hpp` gained no member at all. `grep -n 'capture_standing_fault_frames\|standing' src/ros2_medkit_gateway/include/ros2_medkit_gateway/gateway_node.hpp` returns nothing; the gateway just passes a lambda into the `EntityFreezeFrameCapture` constructor from `init_entity_freeze_frame_capture` (`gateway_node.cpp:1809`, lister lambda at `:1826`). The catch-up itself is `EntityFreezeFrameCapture::capture_standing_faults()`, declared private (`entity_freeze_frame_capture.hpp:148`, under `private:` at `:136`). Current head c146a1cf.

src/ros2_medkit_gateway/src/entity_freeze_frame_capture.cpp
```diff
@@ -209,7 +210,45 @@ void EntityFreezeFrameCapture::on_fault_event(const ros2_medkit_msgs::msg::Fault
...
 }
 
+void EntityFreezeFrameCapture::capture_standing_faults() {
+  if (!standing_lister_) {
+    return;
+  }
+  // The lister blocks until the fault services answer, so it runs here rather
+  // than on a thread of its own: this thread is joined by the destructor, which
+  // is what keeps that wait from outliving the node it waits on.
+  const auto standing = standing_lister_();
```
bburda: `standing_lister_()` is called with no `try` around it, and `capture_standing_faults()` runs from `capture_worker()`, which is the thread entry function, so an escaping exception calls `std::terminate`. The shipped lister does a blocking RPC and then walks `nlohmann::json`, which throws on a present-but-wrong-typed key, and every other injected callback in this file is guarded (`route_fetcher_` at line 163, `provider->list_data` at line 292).

Fix: wrap the call in `try`/`catch (const std::exception &)` plus `catch (...)` and log, the same way `capture_via_route` does.

src/ros2_medkit_gateway/src/gateway_node.cpp
```diff
@@ -1816,6 +1816,46 @@ void GatewayNode::init_entity_freeze_frame_capture(ros2_common::Ros2Subscription
...
+      },
+      256,  // retained-frame bound (constructor default)
+      // Plugins start polling - and reporting - while this object is still
+      // being wired, and a device that is in fault when the box boots confirms
+      // immediately, to nobody. The capture asks for those on its own thread,
+      // which it joins on destruction, so the service wait cannot outlive the
+      // node it waits on.
+      [this]() {
+        std::vector<EntityFreezeFrameCapture::StandingFault> standing;
+        if (!fault_service_transport_ || !fault_service_transport_->wait_for_services(std::chrono::seconds(10))) {
```
bburda: `wait_for_services` waits on four clients in sequence and gives each the full timeout (`ros2_fault_service_transport.cpp:130-133`), so this is a budget of up to 40 s, not 10 s, and all of it runs before the drain loop starts. Meanwhile `on_fault_event` drops the OLDEST event once `queue_` passes `kMaxQueuedEvents` (64), so live confirms in that window are lost, and the destructor joins through the whole wait because `stop_` is not checked until the lister returns.

Fix: share one deadline across the four waits, pass a stop token into the lister, and check `stop_` before calling it.

src/ros2_medkit_gateway/src/gateway_node.cpp
```diff
@@ -1816,6 +1816,46 @@ void GatewayNode::init_entity_freeze_frame_capture(ros2_common::Ros2Subscription
...
+          return standing;
+        }
+        auto result = fault_service_transport_->list_faults("", false, true, false, false, false, false);
```
bburda, line 1850: Nothing orders this snapshot against the fault-events subscription being matched, and `~/events` is `QoS(100).reliable()` with no `transient_local` on either end (`fault_manager_node.cpp:165`, `entity_freeze_frame_capture.cpp:54`), so a sample published before the reader matches is dropped and never replayed. A fault that confirms after this call returns but before that matching completes is therefore in neither the snapshot nor the event stream, and stays frameless, which is the case this PR is meant to fix.

Fix: wait for the subscription to see a publisher before taking the snapshot (`get_publisher_count()` on the slot), or make `~/events` `transient_local` on both ends so a late joiner replays the confirm.

src/ros2_medkit_gateway/src/entity_freeze_frame_capture.cpp
```diff
@@ -209,7 +210,45 @@ void EntityFreezeFrameCapture::on_fault_event(const ros2_medkit_msgs::msg::Fault
...
+      std::lock_guard<std::mutex> lock(mutex_);
+      if (frames_.count(fault.fault_code) != 0) {
+        continue;  // the live event beat us to it
+      }
+    }
+    ros2_medkit_msgs::msg::FaultEvent event;
+    event.event_type = ros2_medkit_msgs::msg::FaultEvent::EVENT_CONFIRMED;
+    event.fault.fault_code = fault.fault_code;
+    event.fault.reporting_sources = fault.reporting_sources;
+    capture_for_event(event);
```
bburda: This stamps `captured_at` with `now()` for every fault in the list, and the list comes from the fault manager's persisted store (`storage_type` defaults to `sqlite`), so restarting only the gateway writes a `freeze_frame` holding current values for faults that confirmed hours or days ago. A consumer of `GET /{entity}/faults/{code}` cannot tell such a frame from a real confirm-time one, because the snapshot entry carries only type, name, data and `captured_at`.

Fix: mark catch-up frames in the `x-medkit` block so consumers can tell them apart, or carry the fault's occurrence timestamp in `StandingFault` and skip faults too old to be the boot race.

bburda (commented):

- Would you add a launch test for this? `create_test_launch` always starts the gateway first, so `test_entity_freeze_frame.test.py` cannot reach the new path and the lister always sees an empty list. Starting the fault manager first, holding a confirmed fault, would prove the fix, and that test fails on main.
- Could you update the docs? `docs/tutorials/snapshots.rst` and `docs/config/server.rst` describe freeze frames as captured "when the fault confirms", and `docs/tutorials/snapshots.rst` and `docs/api/rest.rst` describe `freeze_frame` as "captured at fault confirmation". None of them mention the startup catch-up, and the value-age caveat from your description is not written down anywhere a user would see it.
- Could you extend `REQ_INTEROP_088` in `docs/requirements/specs/faults.rst`? Its text ties capture to the confirm event, but both new tests verify the startup trigger.

src/ros2_medkit_gateway/src/entity_freeze_frame_capture.cpp
```diff
@@ -209,7 +210,45 @@ void EntityFreezeFrameCapture::on_fault_event(const ros2_medkit_msgs::msg::Fault
...
+      if (stop_) {
+        return;
+      }
+    }
+    if (fault.fault_code.empty() || fault.reporting_sources.empty()) {
+      continue;
+    }
+    {
+      std::lock_guard<std::mutex> lock(mutex_);
+      if (frames_.count(fault.fault_code) != 0) {
```
bburda: This guard can never fire, so the comment is wrong. `frames_` is written only by `capture_for_event`, which runs only on `capture_thread_`, and this catch-up is the first statement of that thread. A live confirm at this moment is still in `queue_`, not in `frames_`. The fault is then captured twice, once here and once when the queue drains, so the plugin or route is read twice for one fault.

Fix: snapshot the queued fault codes under `queue_mutex_` and skip those, which both removes the double read and gives the guard a real meaning.

src/ros2_medkit_gateway/src/entity_freeze_frame_capture.cpp
```diff
@@ -209,7 +210,45 @@ void EntityFreezeFrameCapture::on_fault_event(const ros2_medkit_msgs::msg::Fault
...
+        continue;  // the live event beat us to it
+      }
+    }
+    ros2_medkit_msgs::msg::FaultEvent event;
+    event.event_type = ros2_medkit_msgs::msg::FaultEvent::EVENT_CONFIRMED;
+    event.fault.fault_code = fault.fault_code;
+    event.fault.reporting_sources = fault.reporting_sources;
+    capture_for_event(event);
+    ++framed;
```
bburda: `framed` counts every standing fault, but `capture_for_event` stores nothing when no data comes back, and `list_faults("")` returns every confirmed fault including plain ROS ones that have no plugin data. A gateway with 30 confirmed ROS faults and no plugin fault logs "captured 30 fault(s) that were already confirmed at startup" after capturing none.

Fix: have `capture_for_event` report whether it stored a frame and increment only then.

src/ros2_medkit_gateway/src/gateway_node.cpp
```diff
@@ -1816,6 +1816,46 @@ void GatewayNode::init_entity_freeze_frame_capture(ros2_common::Ros2Subscription
...
+        if (!result.success || !result.data.is_object()) {
```
bburda: This and the four other early returns in the lambda give up silently, and because `framed` then stays 0 the success line is suppressed too, so the gateway logs nothing at all. "There were no standing faults" and "we never reached the fault manager" look identical in the journal. A wrong `fault_manager.namespace`, or a fault manager that starts late, then disables the feature for the whole process lifetime with no signal.

Fix: `RCLCPP_WARN` on the failure branches, including `result.error_message` here, and log the legitimately empty case separately.

src/ros2_medkit_gateway/src/gateway_node.cpp
```diff
@@ -1816,6 +1816,46 @@ void GatewayNode::init_entity_freeze_frame_capture(ros2_common::Ros2Subscription
...
+          }
+          EntityFreezeFrameCapture::StandingFault fault;
+          fault.fault_code = code->get<std::string>();
+          for (const auto & src : *sources) {
+            if (src.is_string()) {
+              fault.reporting_sources.push_back(src.get<std::string>());
+            }
+          }
+          standing.push_back(std::move(fault));
+        }
```
bburda, line 1863: Nothing bounds this list, but `frames_` keeps only `max_faults_` (256) entries and evicts FIFO. More than 256 standing faults means the catch-up evicts frames it created moments earlier, and which 256 survive depends on the order the fault manager returned them.

Fix: cap the catch-up at `max_faults_` and log when the list was truncated.

src/ros2_medkit_gateway/test/test_entity_freeze_frame_capture.cpp
```diff
@@ -246,6 +246,102 @@ TEST_F(EntityFreezeFrameCaptureTest, NonConfirmedEventsAreIgnored) {
   EXPECT_TRUE(capture.frames_for("PLC_UPDATED_ONLY").empty());
 }
 
+/// @verifies REQ_INTEROP_088
+TEST_F(EntityFreezeFrameCaptureTest, StandingFaultsAreFramedWithoutAnyEvent) {
+  // The startup case: the fault confirmed before this object subscribed, so no
```
bburda, line 254: Both new tests inject a hand-written lister, so the real lambda in `gateway_node.cpp` is never exercised. None of its JSON handling is covered: a failed call, a reply that is not an object, a missing `faults` key, a non-object item, a non-string `fault_code`, or a `reporting_sources` array of non-strings. A rename on either side would still compile and would silently disable the feature.

Fix: move the parsing into a static helper on `EntityFreezeFrameCapture` and test it directly, the way `values_from_list_content` already is.

src/ros2_medkit_gateway/src/entity_freeze_frame_capture.cpp
mfaferek93: Fixed in 8415ad79 - try/catch(const std::exception&) plus catch(...) with a WARN, same style as capture_via_route. ThrowingListerDoesNotKillCaptureThread proves the capture thread survives and still drains live confirms.

src/ros2_medkit_gateway/src/gateway_node.cpp
mfaferek93: Fixed in 8415ad79. wait_for_services shares one deadline across the four clients (0 = non-blocking probe); the gateway lister polls it in 0.5 s slices under a single 10 s deadline and honours a should_abort hook; the capture checks stop_ before calling the lister and the destructor joins first. DestructionAbortsBlockedLister bounds the join.

src/ros2_medkit_gateway/src/gateway_node.cpp
mfaferek93: Fixed in 8415ad79 with the publisher-count option: a bounded abortable wait on the events slot's publisher_count() before the snapshot. Stayed volatile on purpose - a transient_local subscriber does not match a volatile publisher, so a rolling upgrade against an older fault manager would lose the stream entirely; there is a terse comment at the QoS site. ListerRunsOnlyAfterEventsPublisherMatched discriminates the immediate-call bug.

src/ros2_medkit_gateway/src/entity_freeze_frame_capture.cpp
mfaferek93: Fixed in 8415ad79: catch-up frames carry capture_origin "startup" in the x-medkit block (absent = confirm edge), propagated through the merge and build_sovd_fault_response, asserted in unit tests and the new e2e, documented with the captured_at caveat. No age cutoff on purpose - a slow boot would skip legitimate boot-race faults.

src/ros2_medkit_gateway/src/entity_freeze_frame_capture.cpp
mfaferek93: Fixed in 8415ad79 - the catch-up snapshots the queued codes under queue_mutex_ and skips them, and the dead frames_ guard is gone. QueuedLiveConfirmDedupesStandingCatchUp asserts the standing entity is never sampled twice and fails without the fix.

src/ros2_medkit_gateway/src/entity_freeze_frame_capture.cpp
mfaferek93: Fixed in 8415ad79 - capture_for_event returns whether it stored a frame and framed counts only those, so the INFO line reports real captures.

src/ros2_medkit_gateway/src/gateway_node.cpp
mfaferek93: Fixed in 8415ad79 - WARN on every failure branch (transport not initialised, service wait timeout, ListFaults error including error_message, malformed reply) and a separate INFO for the legitimately empty list.

src/ros2_medkit_gateway/src/gateway_node.cpp
mfaferek93: Fixed in 8415ad79 - the catch-up stops storing at max_faults_ (only stored frames count toward the cap, so it cannot evict itself) and WARNs how many standing faults stayed unframed.

src/ros2_medkit_gateway/test/test_entity_freeze_frame_capture.cpp
mfaferek93: Fixed in 8415ad79 - parsing moved to static EntityFreezeFrameCapture::standing_faults_from_list_reply, nullopt for a reply not shaped like a ListFaults answer so the caller warns instead of going silent. Three suites cover well-formed, nullopt shapes and malformed items.

mfaferek93:
The outside-diff items:

- New e2e test_standing_fault_freeze_frame.test.py: the fault manager starts first, a seeder confirms a fault and exits, the gateway starts after and must serve the freeze_frame with capture_origin startup and the seeded values - 76af95e2, three green runs.
- snapshots.rst, server.rst and rest.rst document the catch-up, the captured_at semantics and the origin marker; the REQ_INTEROP_088 text now covers the startup case - 3a1857c1.

bburda (approved):

PR #565, opened 2026-07-26
fix(gateway): freeze-frame an entity that is serving last known values

### LINKED ISSUE #564: gateway: a comms-loss fault can never get a freeze frame
## Summary

A loss-of-communication fault never gets an entity freeze frame, because the
capture refuses any content whose entity reports `connected: false`:

```
[WARN] Entity freeze-frame for fault 'PLC_COMMS_LOST': entity '<id>' reported no live values
```

That is the one fault where the last known values matter most - what the device
was doing right before it went away - and it is exactly when they are refused.

## Cause

`EntityFreezeFrameCapture::content_has_live_data()` rejects the payload outright
when `connected` is false. The intent is sound (a dead link must not freeze a row
of nulls), but the test is on the link flag rather than on whether the payload
carries usable values, and `values_have_data()` already rejects the all-null row
one step later.

## Effect

Any bridge that keeps a last-known value cache serves real numbers while
disconnected, and they are dropped. The fault detail then shows a recording but
no freeze frame, so the operator sees that something was recorded and nothing
about what the device held.

## Fix direction

Decide on the values, not on the flag: capture when the payload carries at least
one non-null value, whatever `connected` says. A frame taken this way holds the
last values the bridge read, and `captured_at_ns` already states when it was
taken, so nothing is presented as fresher than it is.

mfaferek93:
Closes #564.

`content_has_live_data()` rejected any payload whose entity reported
`connected: false`, so a loss-of-comms fault - the one fault whose freeze frame
is the whole point, showing what the device held right before it went away -
could never have one. The guard was aimed at the row of nulls a cold cache
yields, but that case is already caught by `values_have_data()` one step later.

The gate now judges the payload: a non-empty items array captures, whatever the
link flag says. `captured_at_ns` continues to state when the frame was taken, so
nothing is presented as fresher than it is, and a bridge that serves per-value
staleness keeps saying so on its own route.

Tests: `DisconnectedEntityWithLastKnownValuesIsCaptured` (new, fails before),
`ContentHasLiveData.GatesOnItemsNotOnTheLinkFlag`, and the two "nothing usable ->
no frame" tests re-pointed at the case they were really guarding (down *and*
empty). Full gateway suite: 3793 tests, 0 failures, 448 skipped.

### DIFF
--- a/src/ros2_medkit_gateway/src/entity_freeze_frame_capture.cpp
+++ b/src/ros2_medkit_gateway/src/entity_freeze_frame_capture.cpp
@@ -129,9 +129,11 @@ bool EntityFreezeFrameCapture::content_has_live_data(const nlohmann::json & cont
   if (!content.is_object()) {
     return false;
   }
-  if (content.contains("connected") && content["connected"].is_boolean() && !content["connected"].get<bool>()) {
-    return false;
-  }
+  // Judged on the values, not on the link flag. A bridge that serves its last
+  // known values while disconnected is exactly what a loss-of-comms fault wants
+  // frozen - refusing on `connected: false` denied a frame to the one fault
+  // where the values before the link died are the whole story. The all-null row
+  // a genuinely cold cache yields is still rejected, by values_have_data().
   return content.contains("items") && content["items"].is_array() && !content["items"].empty();
 }
 
--- a/src/ros2_medkit_gateway/src/entity_freeze_frame_capture.cpp
+++ b/src/ros2_medkit_gateway/src/entity_freeze_frame_capture.cpp
@@ -137,6 +137,11 @@ bool EntityFreezeFrameCapture::content_has_live_data(const nlohmann::json & cont
   return content.contains("items") && content["items"].is_array() && !content["items"].empty();
 }
 
+bool EntityFreezeFrameCapture::content_reports_disconnected(const nlohmann::json & content) {
+  return content.is_object() && content.contains("connected") && content["connected"].is_boolean() &&
+         !content["connected"].get<bool>();
+}
+
 bool EntityFreezeFrameCapture::values_have_data(const nlohmann::json & values) {
   if (values.is_object()) {
     for (const auto & entry : values.items()) {
@@ -185,7 +190,7 @@ std::optional<EntityFreezeFrameCapture::Frame>
 EntityFreezeFrameCapture::frame_from_content(const std::string & entity_id, const std::string & fault_code,
                                              const nlohmann::json & content) {
   if (!content_has_live_data(content)) {
-    log_fallback_failure_once(fault_code, "entity '" + entity_id + "' reported no live values");
+    log_fallback_failure_once(fault_code, "entity '" + entity_id + "' returned no data items");
     return std::nullopt;
   }
   Frame frame;
@@ -196,6 +201,15 @@ EntityFreezeFrameCapture::frame_from_content(const std::string & entity_id, cons
     // ids): still a dead/cold row - no frame, same invariant as above.
     log_fallback_failure_once(fault_code, "entity '" + entity_id + "' values are all null");
     return std::nullopt;
+  }
+  // Provenance for the snapshot's x-medkit block: the payload's link flag and
+  // its own timestamp, when the plugin reports them. captured_at_ns dates the
+  // capture; a disconnected entity's values may be much older.
+  if (content.contains("connected") && content["connected"].is_boolean()) {
+    frame.connected = content["connected"].get<bool>();
+  }
+  if (content.contains("timestamp")) {
+    frame.source_timestamp = content["timestamp"];
   }
   frame.captured_at_ns =
       std::chrono::duration_cast<std::chrono::nanoseconds>(std::chrono::system_clock::now().time_since_epoch()).count();
--- a/src/ros2_medkit_gateway/test/test_entity_freeze_frame_capture.cpp
+++ b/src/ros2_medkit_gateway/test/test_entity_freeze_frame_capture.cpp
@@ -552,7 +552,8 @@ TEST_F(EntityFreezeFrameCaptureTest, RouteFallbackSkipsDisconnectedPlcWithNoValu
 /// @verifies REQ_INTEROP_088
 TEST_F(EntityFreezeFrameCaptureTest, DisconnectedEntityWithLastKnownValuesIsCaptured) {
   // The loss-of-comms case: the bridge reports the link down and still serves
-  // the values it last read. That is what the fault is about, so it is frozen.
+  // the values it last read. That is what the fault is about, so it is frozen,
+  // with the payload's link flag and own timestamp carried as provenance.
   EntityFreezeFrameCapture capture(
       node_.get(), *sub_exec_,
       [](const std::string &) -> DataProvider * {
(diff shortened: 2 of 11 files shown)


src/ros2_medkit_gateway/src/entity_freeze_frame_capture.cpp
```diff
@@ -117,9 +117,11 @@ bool EntityFreezeFrameCapture::content_has_live_data(const nlohmann::json & cont
   if (!content.is_object()) {
     return false;
   }
-  if (content.contains("connected") && content["connected"].is_boolean() && !content["connected"].get<bool>()) {
-    return false;
-  }
+  // Judged on the values, not on the link flag. A bridge that serves its last
```
bburda, line 132: Could you keep a separate link check for the fault-trigger path? This predicate also gates the trigger value fetcher at `gateway_node.cpp:1883`, and there `connected` really is set, because `OpcuaPlugin::build_data_response` emits it.

Without the check, a down link makes that fetcher return the frozen last known value instead of `nullopt`. `fault_trigger_engine.cpp` holds rule state only on `nullopt`, so threshold rules keep firing on a stale number for the whole outage.

src/ros2_medkit_gateway/src/entity_freeze_frame_capture.cpp
```diff
@@ -117,9 +117,11 @@ bool EntityFreezeFrameCapture::content_has_live_data(const nlohmann::json & cont
...
+  // known values while disconnected is exactly what a loss-of-comms fault wants
+  // frozen - refusing on `connected: false` denied a frame to the one fault
```
bburda, line 134: Could you reword the log line at line 144? The rejection moved into `values_have_data`, so a disconnected entity with a null row now logs "values are all null". The issue quotes the old "reported no live values" as the symptom.

bburda (commented):
Two things outside this diff that I think also need a change:

- Would you add an integration test for the loss-of-comms case? The route test plugin hardcodes `{"connected", true}`, so no launch test can reach the new path today. `test_entity_freeze_frame.test.py` already loads that plugin, so a second entity returning `{"connected": false, "items": [...]}` would cover it.
- Could you update the docs in this PR? `docs/tutorials/snapshots.rst` and `docs/config/server.rst` say the gateway snapshots the entity's "current data values" when the fault confirms, and `docs/api/rest.rst` calls `freeze_frame` "captured at fault confirmation". A frame may now hold values of unbounded age.

src/ros2_medkit_gateway/src/entity_freeze_frame_capture.cpp
```diff
@@ -117,9 +117,11 @@ bool EntityFreezeFrameCapture::content_has_live_data(const nlohmann::json & cont
...
+  // where the values before the link died are the whole story. The all-null row
+  // a genuinely cold cache yields is still rejected, by values_have_data().
```
bburda, line 136: Could you carry `connected` and the source timestamp on `Frame` and put them in the `x-medkit` block? `captured_at_ns` is stamped with `now()`, so it dates the frame and not the values, and `values_from_list_content` drops `connected`, the route `timestamp` and the per-item `stale` flag.

src/ros2_medkit_gateway/test/test_entity_freeze_frame_capture.cpp
```diff
@@ -331,6 +333,27 @@ TEST_F(EntityFreezeFrameCaptureTest, RouteFallbackSkipsDisconnectedPlc) {
...
+  // The loss-of-comms case: the bridge reports the link down and still serves
+  // the values it last read. That is what the fault is about, so it is frozen.
+  EntityFreezeFrameCapture capture(
+      node_.get(), *sub_exec_,
+      [](const std::string &) -> DataProvider * {
+        return nullptr;
+      },
+      [](const std::string &) -> std::optional<json> {
+        return json{{"connected", false},
+                    {"items", json::array({{{"name", "level"}, {"value", 42.0}, {"stale", true}}})}};
```
bburda: Could you drop `"stale": true` here? `values_from_list_content` keeps only `value`, so the flag never reaches the assertion below and the payload suggests coverage that does not exist.

src/ros2_medkit_gateway/test/test_entity_freeze_frame_capture.cpp
```diff
@@ -355,12 +378,14 @@ TEST_F(EntityFreezeFrameCaptureTest, DataProviderWinsOverRouteFallback) {
...
-  // disconnected flag must not freeze-frame a row of {} / nulls.
+  // items array (cold poll cache) or all-null values must not freeze-frame a
+  // row of {} / nulls, connected or not.
   StaticContentDataProvider cold_provider("cold_app", json{{"items", json::array()}});
   StaticContentDataProvider null_provider("null_app", json{{"items", json::array({{{"id", "level"}}})}});
-  StaticContentDataProvider down_provider(
-      "down_app", json{{"connected", false}, {"items", json::array({{{"id", "level"}, {"value", 1.0}}})}});
+  // Down AND empty: nothing to freeze. (Down with a real last known value is a
+  // capture, covered separately.)
+  StaticContentDataProvider down_provider("down_app",
```
bburda: Could you give `down_provider` a real value again and assert it IS captured, in its own test? With the `connected` check gone it is now the same as `null_provider` on line 384, so the DataProvider branch has no positive test for the new behaviour.

src/ros2_medkit_gateway/src/entity_freeze_frame_capture.cpp
mfaferek93: Fixed in 844f1953. The trigger fetcher now rejects a payload whose connected flag reports the link down (new content_reports_disconnected, consulted only there - the freeze-frame paths stay content-based). Verified in fault_trigger_engine that rule state holds only on nullopt. New e2e test_07_disconnected_app_holds_rule_state creates a threshold rule against the disconnected app and asserts it never fires on the frozen 42.0; it fails with the guard removed.

src/ros2_medkit_gateway/src/entity_freeze_frame_capture.cpp
mfaferek93: Fixed in 844f1953 - the line now says the entity returned no data items, which is the only case it fires on; the all-null row logs through values_have_data's own message.

src/ros2_medkit_gateway/src/entity_freeze_frame_capture.cpp
mfaferek93: Fixed in 844f1953. Frame carries the payload's connected flag and its own timestamp field verbatim; both land in the x-medkit block (merge and build_sovd_fault_response) only when known, and captured_at is documented as dating the capture, not the values. Covered at merge and wire level by CarriesLinkStateAndSourceTimestampWhenKnown and asserted end to end in the new launch case.

src/ros2_medkit_gateway/test/test_entity_freeze_frame_capture.cpp
mfaferek93: Fixed in 844f1953 - the flag is gone; the test now asserts what actually flows through: frame.connected == false and the payload timestamp.

src/ros2_medkit_gateway/test/test_entity_freeze_frame_capture.cpp
mfaferek93: Fixed in 844f1953 - DisconnectedDataProviderWithLastKnownValuesIsCaptured serves a real 5.5 through the down provider and asserts the capture (plus connected false, no timestamp), as its own test; the negative test keeps the cold and down-with-all-null providers. It fails with the pre-PR connected rejection reinstated.

mfaferek93:
Both outside-diff asks are in:

- test_route_data_plugin gained a second, external app serving {"connected": false, "items": [last known values], "timestamp": ...}; the new launch case asserts its fault carries those values plus the x-medkit connected / source_timestamp markers - 844f1953.
- snapshots.rst, server.rst and rest.rst now say freeze-frame values may be last known when the entity is disconnected (age bounded only by the outage) and document the new x-medkit fields; captured_at is defined as dating the capture - daf53239.

bburda (approved):

PR #567, opened 2026-07-27
fix(gateway): release the parameter transport's node before the context dies

### LINKED ISSUE #566: [BUG] Gateway aborts with SIGABRT on shutdown via the parameter transport's internal node
# Bug report

The gateway process sometimes dies with SIGABRT (exit code -6) on shutdown instead of exiting cleanly. It is a race, so it does not happen on every run.

### Steps to reproduce
1. Run `gateway_node` in a setup where something reads node parameters, so `Ros2ParameterTransport` gets created.
2. Stop the gateway with Ctrl+C (SIGINT).
3. Repeat. On our runs the abort happened 6 times out of 33 shutdowns.

### Expected behavior

The process exits cleanly (0, or the signal exit code) on every shutdown.

### Actual behavior

The process aborts:

```
terminate called after throwing an instance of 'rclcpp::graph_listener::NodeNotFoundError'
  what():  node not found
process has died [exit code -6]
```

Cause: `Ros2ParameterTransport` creates its own `rclcpp::Node` (`_param_client_node`) and releases it only in its destructor. On the Ctrl+C path the signal handler runs `rclcpp::shutdown()`, and that is what makes `executor.spin()` return. Only after that does `main.cpp` destroy the gateway node. So the transport is destroyed when the context is already invalid and the `GraphListener` has cleared its node list. rclcpp's `~NodeGraph` then calls `GraphListener::remove_node()`, which throws `NodeNotFoundError`.

`~NodeGraph` is implicitly `noexcept`, so the throw goes straight to `std::terminate`. A `try/catch` around the destruction cannot catch it. Whether the listener has already cleared its list at that moment is a race, which is why the abort is intermittent.

Any owner that outlives `rclcpp::shutdown()` hits this, including the gateway's own managers.

### Environment
- ros2_medkit version: main
- ROS 2 distro: Jazzy
- OS: Ubuntu 24.04

### Additional information

The fix has to change the order, not catch the exception: the node must be released while the context is still valid.

bburda:
# Pull Request

## Summary

`Ros2ParameterTransport` creates its own `rclcpp::Node` (`_param_client_node`) and released it only in its destructor. If the owner outlives `rclcpp::shutdown()`, the node is dropped after the `GraphListener` has already cleared its node list. rclcpp's `~NodeGraph` then calls `GraphListener::remove_node()`, which throws `NodeNotFoundError`. `~NodeGraph` is implicitly `noexcept`, so the throw goes straight to `std::terminate` and the process dies with SIGABRT. A `try/catch` at the call site cannot stop it, so the fix has to change the order, not catch the exception.

This is reachable on a normal Ctrl+C. The signal handler runs `rclcpp::shutdown()`, and that is what makes `executor.spin()` return. Only after that does `main.cpp` destroy the gateway node, so `~GatewayNode` and everything it owns run when the context is already invalid. Whether the listener has cleared its list by then is a race, which is why the abort is intermittent and not every run.

The transport now registers a pre-shutdown callback that runs its `shutdown()` while the context is still valid. `shutdown()` was already idempotent. The destructor removes the callback and still calls `shutdown()`, so the normal case (transport destroyed first) is unchanged.

---

## Issue

- closes #566

---

## Type

- [x] Bug fix
- [ ] New feature or tests
- [ ] Breaking change
- [ ] Documentation only

---

## Testing

Measured on a gateway that reproduced the abort: 6 aborts in 33 shutdowns before the change, 0 in 40 after, with everything else identical.

Two unit tests were added in a new file. They check the teardown order in both directions: transport destroyed after context shutdown, and before it. They do NOT reproduce the abort, because a process with only one node tears down cleanly either way. So they are contract coverage, not the proof. The measurement above is the proof, and the test file says so.

Full `ros2_medkit_gateway` suite passes: 111 gtests plus all linters.

---

## Checklist

### DIFF
--- a/src/ros2_medkit_gateway/include/ros2_medkit_gateway/ros2/transports/ros2_parameter_transport.hpp
+++ b/src/ros2_medkit_gateway/include/ros2_medkit_gateway/ros2/transports/ros2_parameter_transport.hpp
@@ -215,14 +215,30 @@ class Ros2ParameterTransport : public ParameterTransport {
   std::shared_ptr<rclcpp::Node> param_node_;
 
   /// Context of param_node_, plus the handle of the pre-shutdown callback registered on
-  /// it. param_node_ MUST be destroyed while the context is still valid: rclcpp's
-  /// ~NodeGraph calls GraphListener::remove_node(), which throws NodeNotFoundError once
-  /// the context has been shut down and the listener has dropped its node list. That
-  /// throw escapes an implicitly-noexcept destructor, so it is an immediate
-  /// std::terminate (SIGABRT) that no try/catch at the call site can intercept - the
-  /// only fix is to not be holding the node by then. The callback runs shutdown() ahead
-  /// of context invalidation; the destructor deregisters it and calls the (idempotent)
-  /// shutdown() for the ordinary case where the transport dies first.
+  /// it. param_node_ must be released while the GraphListener is still running, and the
+  /// callback is what guarantees that: Context::shutdown() runs pre-shutdown callbacks,
+  /// then rcl_shutdown, then the on_shutdown callbacks - and the listener is stopped from
+  /// one of the latter.
+  ///
+  /// The crash this prevents is a HALF-REGISTERED node, not a late destruction.
+  /// NodeGraph::get_graph_event() consumes should_add_to_graph_listener_ before calling
+  /// GraphListener::add_node(), and add_node() throws GraphListenerShutdownError once the
+  /// listener is down. The flag is spent while the node was never listed, so ~NodeGraph
+  /// takes its remove_node() branch, the node is absent from node_graph_interfaces_, and
+  /// NodeNotFoundError escapes an implicitly-noexcept destructor: immediate std::terminate
+  /// (SIGABRT), which no try/catch at the call site can intercept. Only a first-ever graph
+  /// use racing the listener's shutdown can produce it, which is why the crash was
+  /// intermittent rather than every run.
+  ///
+  /// A node that IS registered is fine either way: GraphListener::__shutdown() only joins
+  /// its thread and finalizes the wait set, it never clears node_graph_interfaces_, and
+  /// remove_node() erases directly once is_shutdown() is set. Do not "fix" a suspected
+  /// late-destruction bug here - that path provably works.
+  ///
+  /// Every wait_for_service in this class runs under spin_mutex_ and the callback blocks
+  /// on the same mutex, so no first graph use can straddle the listener's shutdown. The
+  /// destructor deregisters the callback and calls the (idempotent) shutdown() for the
+  /// ordinary case where the transport dies first.
   std::shared_ptr<rclcpp::Context> context_;
   rclcpp::PreShutdownCallbackHandle pre_shutdown_handle_;
 
--- a/src/ros2_medkit_gateway/include/ros2_medkit_gateway/ros2/transports/ros2_parameter_transport.hpp
+++ b/src/ros2_medkit_gateway/include/ros2_medkit_gateway/ros2/transports/ros2_parameter_transport.hpp
@@ -214,6 +214,18 @@ class Ros2ParameterTransport : public ParameterTransport {
   /// Created once in constructor - must be in DDS graph early for fast service discovery.
   std::shared_ptr<rclcpp::Node> param_node_;
 
+  /// Context of param_node_, plus the handle of the pre-shutdown callback registered on
+  /// it. param_node_ MUST be destroyed while the context is still valid: rclcpp's
+  /// ~NodeGraph calls GraphListener::remove_node(), which throws NodeNotFoundError once
+  /// the context has been shut down and the listener has dropped its node list. That
+  /// throw escapes an implicitly-noexcept destructor, so it is an immediate
+  /// std::terminate (SIGABRT) that no try/catch at the call site can intercept - the
+  /// only fix is to not be holding the node by then. The callback runs shutdown() ahead
+  /// of context invalidation; the destructor deregisters it and calls the (idempotent)
+  /// shutdown() for the ordinary case where the transport dies first.
+  std::shared_ptr<rclcpp::Context> context_;
+  rclcpp::PreShutdownCallbackHandle pre_shutdown_handle_;
+
   /// Cached AsyncParametersClient per target node, bounded by LRU eviction.
   mutable std::mutex clients_mutex_;
   BoundedLruCache<std::string, std::shared_ptr<rclcpp::AsyncParametersClient>> param_clients_;
--- a/src/ros2_medkit_gateway/src/ros2/transports/ros2_parameter_transport.cpp
+++ b/src/ros2_medkit_gateway/src/ros2/transports/ros2_parameter_transport.cpp
@@ -46,6 +46,17 @@ Ros2ParameterTransport::Ros2ParameterTransport(rclcpp::Node * node, double servi
   options.use_global_arguments(false);
   param_node_ = std::make_shared<rclcpp::Node>("_param_client_node", options);
 
+  // Tear the node down BEFORE the context is invalidated. Without this, an owner that
+  // outlives rclcpp::shutdown() (the gateway's own managers, or a plugin whose objects
+  // are destroyed from ~GatewayNode) drops param_node_ after the GraphListener has
+  // already cleared its node list, and ~NodeGraph's remove_node() throws
+  // NodeNotFoundError out of a noexcept destructor -> std::terminate, exit -6. It is a
+  // race, so it shows up as an intermittent abort on Ctrl+C rather than a reliable one.
+  context_ = param_node_->get_node_base_interface()->get_context();
+  pre_shutdown_handle_ = context_->add_pre_shutdown_callback([this] {
+    shutdown();
+  });
+
   // Store own node FQN for self-query detection.
   own_node_fqn_ = node_->get_fully_qualified_name();
 
@@ -54,7 +65,13 @@ Ros2ParameterTransport::Ros2ParameterTransport(rclcpp::Node * node, double servi
 }
 
 Ros2ParameterTransport::~Ros2ParameterTransport() {
-  shutdown();
+  // Deregister first: the callback captures `this`, and once we return the object is
+  // gone. remove_pre_shutdown_callback() is a no-op if the callback already ran during a
+  // normal rclcpp shutdown, and Context serializes it against a callback in flight.
+  if (context_) {
+    context_->remove_pre_shutdown_callback(pre_shutdown_handle_);
+  }
+  shutdown();  // idempotent: a no-op when the pre-shutdown callback already ran
 }
 
 void Ros2ParameterTransport::shutdown() {
--- a/src/ros2_medkit_gateway/src/ros2/transports/ros2_parameter_transport.cpp
+++ b/src/ros2_medkit_gateway/src/ros2/transports/ros2_parameter_transport.cpp
@@ -46,28 +46,46 @@ Ros2ParameterTransport::Ros2ParameterTransport(rclcpp::Node * node, double servi
   options.use_global_arguments(false);
   param_node_ = std::make_shared<rclcpp::Node>("_param_client_node", options);
 
-  // Tear the node down BEFORE the context is invalidated. Without this, an owner that
-  // outlives rclcpp::shutdown() (the gateway's own managers, or a plugin whose objects
-  // are destroyed from ~GatewayNode) drops param_node_ after the GraphListener has
-  // already cleared its node list, and ~NodeGraph's remove_node() throws
-  // NodeNotFoundError out of a noexcept destructor -> std::terminate, exit -6. It is a
-  // race, so it shows up as an intermittent abort on Ctrl+C rather than a reliable one.
-  context_ = param_node_->get_node_base_interface()->get_context();
-  pre_shutdown_handle_ = context_->add_pre_shutdown_callback([this] {
-    shutdown();
-  });
-
   // Store own node FQN for self-query detection.
   own_node_fqn_ = node_->get_fully_qualified_name();
 
   RCLCPP_INFO(node_->get_logger(), "Ros2ParameterTransport initialized (timeout=%.1fs, negative_cache=%.0fs)",
               service_timeout_sec_, negative_cache_ttl_sec_);
+
+  // Release param_node_ while the GraphListener is still running. The listener is stopped
+  // from an on_shutdown callback, which Context::shutdown() runs AFTER rcl_shutdown and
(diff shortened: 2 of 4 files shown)


src/ros2_medkit_gateway/test/test_ros2_parameter_transport_shutdown.cpp
```diff
@@ -0,0 +1,97 @@
...
+  // without the fix. One round trip (against a node that does not exist - it just has to
+  // reach wait_for_service, bounded by kServiceTimeoutSec) puts _param_client_node in the
+  // listener, which is the state every real gateway is in by shutdown time.
+  transport->list_parameters("/no_such_node_for_graph_registration");
+
+  // Invalidate the context while the transport still holds its internal node. The
+  // pre-shutdown callback registered in the constructor must release it here.
+  rclcpp::shutdown();
+
+  // Before the fix this dropped param_node_ with the GraphListener already down:
```
mfaferek93: This file argues with itself: the header (lines 17-24) says both tests "pass with and without the fix", while this test calls the round trip "REQUIRED to reproduce" (line 62) and says pre-fix "the process aborted instead of reaching any assertion below". Both cannot be true, and the header is the right half - I reverted the two ctor statements locally, rebuilt, and ran this binary 3 times: 2/2 tests pass every run, no abort.

rclcpp explains why: once list_parameters has put _param_client_node in the GraphListener, post-shutdown remove_node() just erases it from the still-intact list (jazzy graph_listener.cpp:305-307; __shutdown never clears node_graph_interfaces_, humble identical). NodeNotFoundError needs the node to be ABSENT, i.e. the half-registration race I describe in the hpp comment - which this single-threaded sequence cannot produce.

As written, lines 62-76 tell a future reader this is an abort-regression test: revert the fix, run it, see green, conclude the callback was dead code. Suggest rewording line 62 to "required for ~NodeGraph to call remove_node() at all" and rewriting or dropping the "Before the fix..." block to match the honest-scope header.

src/ros2_medkit_gateway/include/ros2_medkit_gateway/ros2/transports/ros2_parameter_transport.hpp
```diff
@@ -214,6 +214,18 @@ class Ros2ParameterTransport : public ParameterTransport {
   /// Created once in constructor - must be in DDS graph early for fast service discovery.
   std::shared_ptr<rclcpp::Node> param_node_;
 
+  /// Context of param_node_, plus the handle of the pre-shutdown callback registered on
+  /// it. param_node_ MUST be destroyed while the context is still valid: rclcpp's
+  /// ~NodeGraph calls GraphListener::remove_node(), which throws NodeNotFoundError once
+  /// the context has been shut down and the listener has dropped its node list. That
```
mfaferek93: "the listener has dropped its node list" is not what rclcpp does on any branch CI runs. GraphListener::__shutdown() only joins the listener thread and finalizes the wait set (jazzy graph_listener.cpp:331-343); node_graph_interfaces_ is only ever mutated by add_node's push_back and remove_node_'s erase. Post-shutdown remove_node() takes the direct branch (305-307) and cleanly erases a still-registered node; NodeNotFoundError (295) fires only for a node ABSENT from the list. A registered node destroyed after rclcpp::shutdown() tears down cleanly - a standalone repro (registration proven via count_graph_users()==1) exits clean 5/5.

The source-consistent route to the field abort is a half-registered node: a FIRST-ever graph use racing listener shutdown. get_graph_event() consumes should_add_to_graph_listener_, then add_node() throws GraphListenerShutdownError (node_graph.cpp:605-606, graph_listener.cpp:260-261), leaving the flag consumed but the node never listed; ~NodeGraph then calls remove_node() and NodeNotFoundError escapes the noexcept dtor. A repro of that aborts 5/5 with the exact field signature, and wait_for_service reaches get_graph_event() before its rclcpp::ok check (client.cpp:141 vs 154), so a request racing Ctrl+C gets there. It also matches the 6-in-33 intermittency - the mechanism as written here would abort every run.

The fix is still sound against the real mechanism: every wait_for_service site runs under spin_mutex_ (cpp 353/470/644/1097), the pre-shutdown callback blocks on it, and pre-shutdown callbacks run before the on_shutdown one that stops the listener - so no first graph use can straddle listener shutdown, and param_node_ dies while the listener is up. No code change needed - please reword this comment, the ctor comment ("has already cleared its node list"), and the test header, so the next person debugging this crash class is not sent after a teardown that provably works.

src/ros2_medkit_gateway/src/ros2/transports/ros2_parameter_transport.cpp
```diff
@@ -54,7 +65,13 @@ Ros2ParameterTransport::Ros2ParameterTransport(rclcpp::Node * node, double servi
 }
 
 Ros2ParameterTransport::~Ros2ParameterTransport() {
-  shutdown();
+  // Deregister first: the callback captures `this`, and once we return the object is
+  // gone. remove_pre_shutdown_callback() is a no-op if the callback already ran during a
```
mfaferek93: The no-op half of this sentence is wrong in a way that invites a future bug: fired pre-shutdown callbacks are NOT auto-removed. Context::shutdown() invokes them from a copy and never erases (jazzy context.cpp:357-371; humble iterates the live set), so after a normal shutdown this call finds the still-registered entry and erases it - real work, not a no-op. And the erase is load-bearing: the default context is a process-lifetime static that rclcpp::init() re-inits, and neither shutdown() nor clean_up() clears the callback lists, so an entry left behind fires on the NEXT init/shutdown cycle against a freed transport - the exact hazard the second test describes, and the new test binary runs two such cycles in one process. As written the comment invites deleting the "redundant" remove on the already-shut-down path. Suggest: "the callback stays registered after it fires; this erase also stops it firing again on a later init/shutdown cycle". The serialization half checks out - both branches fire callbacks under pre_shutdown_callbacks_mutex_, the same mutex remove takes (jazzy 359/488, humble 312/436).

src/ros2_medkit_gateway/src/ros2/transports/ros2_parameter_transport.cpp
```diff
@@ -46,6 +46,17 @@ Ros2ParameterTransport::Ros2ParameterTransport(rclcpp::Node * node, double servi
...
   param_node_ = std::make_shared<rclcpp::Node>("_param_client_node", options);
 
+  // Tear the node down BEFORE the context is invalidated. Without this, an owner that
+  // outlives rclcpp::shutdown() (the gateway's own managers, or a plugin whose objects
+  // are destroyed from ~GatewayNode) drops param_node_ after the GraphListener has
+  // already cleared its node list, and ~NodeGraph's remove_node() throws
+  // NodeNotFoundError out of a noexcept destructor -> std::terminate, exit -6. It is a
+  // race, so it shows up as an intermittent abort on Ctrl+C rather than a reliable one.
+  context_ = param_node_->get_node_base_interface()->get_context();
+  pre_shutdown_handle_ = context_->add_pre_shutdown_callback([this] {
```
mfaferek93: Registration is not the last statement of the ctor: own_node_fqn_ = node_->get_fully_qualified_name() below is a std::string assignment that can throw (bad_alloc only, so a future-edit trap rather than a live bug today). If it throws, ~Ros2ParameterTransport never runs - and destroying the pre_shutdown_handle_ member deregisters nothing, the handle is just a weak_ptr while the Context owns the shared_ptr to the std::function capturing this. The stale callback then fires on the next Context::shutdown() (at latest from ~Context, which calls shutdown() itself) and runs shutdown() on freed memory. Zero-cost hardening: move the FQN/log lines above this block so add_pre_shutdown_callback is the ctor's final statement.

src/ros2_medkit_gateway/src/ros2/transports/ros2_parameter_transport.cpp
```diff
@@ -46,6 +46,17 @@ Ros2ParameterTransport::Ros2ParameterTransport(rclcpp::Node * node, double servi
...
+    shutdown();
```
mfaferek93: Signal-path tradeoff worth making consciously: this moves shutdown()'s spin_mutex_ wait inside Context::shutdown, BEFORE rcl_shutdown, so the context is still valid while it blocks. An in-flight wait_for_service under that mutex then runs its full remaining timeout - its only early exit is rclcpp::ok, checked per <=100ms slice (jazzy client.cpp:154-165) - where pre-PR the teardown ran after rcl_shutdown and an in-flight wait bailed within ~100ms. Worst realistic case at the default parameter_service_timeout_sec=2.0 (gateway_node.cpp:164): a request against a dead node mid-wait adds ~2s to Ctrl+C (~4s if a second queued request grabs spin_mutex_ first), gating rcl_shutdown for every node in the process; negligible when targets are healthy. I checked it cannot deadlock: while pre-shutdown callbacks run, the context holds only init_mutex_ + pre_shutdown_callbacks_mutex_, and the op thread touches neither (ok() is lock-free; the spin executor registers via on_shutdown_callbacks_mutex_). If you want to cap the stall, check shutdown_requested_ right before each wait_for_service (including the one in cache_default_values) - spin_for already bails on it.

src/ros2_medkit_gateway/include/ros2_medkit_gateway/ros2/transports/ros2_parameter_transport.hpp
bburda: I checked this in the rclcpp source and it matches what you describe. `__shutdown()` only interrupts, joins the listener thread and calls `cleanup_wait_set()`. It never touches `node_graph_interfaces_`. `remove_node()` takes the `is_shutdown()` branch and erases directly, and `remove_node_` only throws when the node is not in the list. So a registered node is fine after shutdown, and my comment was wrong.

The comment now describes the half-registration path instead: `get_graph_event()` consumes `should_add_to_graph_listener_` before it calls `add_node()`, `add_node()` throws `GraphListenerShutdownError` once the listener is down, so the flag is spent while the node was never listed, and `~NodeGraph` then calls `remove_node()` on an absent node.

I also added a line saying the registered-node path provably works, so the next person does not try to fix it. Same wording corrected in the constructor comment and the test header.

src/ros2_medkit_gateway/test/test_ros2_parameter_transport_shutdown.cpp
bburda: The file did say both things. Both parts are rewritten.

Line 62 now says the round trip is needed for `~NodeGraph` to call `remove_node()` at all, not to reproduce an abort. The "Before the fix..." block is gone. It now says that dropping a fully registered node after shutdown is clean in rclcpp, so this test does not abort with or without the callback, and a green run here proves nothing about the abort.

The header keeps the scope note and adds why the two tests share one process: the first one fires the callback, so the second init/shutdown cycle is what would hit a stale callback if the destructor did not erase it.

src/ros2_medkit_gateway/src/ros2/transports/ros2_parameter_transport.cpp
bburda: Confirmed in `context.cpp`: `shutdown()` copies `pre_shutdown_callbacks_` and calls the copy, but never erases. So the entry is still registered after it fires and this remove is real work.

The comment says that now, plus why it matters past one cycle: the default context is a process-lifetime static that a later `rclcpp::init()` reuses without clearing the callback lists, so a leftover entry would fire on the next shutdown against a freed transport. I noted the same in the test file, since the second test depends on the first having run.

src/ros2_medkit_gateway/src/ros2/transports/ros2_parameter_transport.cpp
bburda: Moved. The `own_node_fqn_` assignment and the `RCLCPP_INFO` now run before the registration, so `add_pre_shutdown_callback` is the last statement of the constructor. The comment says it has to stay last, and why: the handle is a `weak_ptr`, so destroying it deregisters nothing.

src/ros2_medkit_gateway/src/ros2/transports/ros2_parameter_transport.cpp
bburda: I took the guard option. All four `wait_for_service` sites check `shutdown_requested_` first: `list_parameters`, `get_parameter`, `set_parameter` and `cache_default_values`. `shutdown()` sets that flag before it blocks on `spin_mutex_`, so the check sees it.

The three `ParameterResult` sites return `SHUT_DOWN`, `cache_default_values` returns `false`. None of them mark the negative cache, since this says nothing about whether the target node is reachable. The full reason is written at the `list_parameters` site and the other three point to it.

src/ros2_medkit_gateway/include/ros2_medkit_gateway/ros2/transports/ros2_parameter_transport.hpp
```diff
@@ -214,6 +214,18 @@ class Ros2ParameterTransport : public ParameterTransport {
...
+  /// throw escapes an implicitly-noexcept destructor, so it is an immediate
+  /// std::terminate (SIGABRT) that no try/catch at the call site can intercept - the
+  /// only fix is to not be holding the node by then. The callback runs shutdown() ahead
+  /// of context invalidation; the destructor deregisters it and calls the (idempotent)
+  /// shutdown() for the ordinary case where the transport dies first.
+  std::shared_ptr<rclcpp::Context> context_;
+  rclcpp::PreShutdownCallbackHandle pre_shutdown_handle_;
+
```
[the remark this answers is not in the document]
bburda: `add_pre_shutdown_callback`, `remove_pre_shutdown_callback` and `PreShutdownCallbackHandle` are available in Humble. In rclcpp on the humble branch, `rclcpp/include/rclcpp/context.hpp` declares `PreShutdownCallbackHandle` at line 64, `add_pre_shutdown_callback` at line 271 and `remove_pre_shutdown_callback` at line 281. No version gate is needed here. The Humble jobs on this PR build and test green.

src/ros2_medkit_gateway/src/ros2/transports/ros2_parameter_transport.cpp
```diff
@@ -46,6 +46,17 @@ Ros2ParameterTransport::Ros2ParameterTransport(rclcpp::Node * node, double servi
...
+  });
```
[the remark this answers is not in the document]
bburda: `add_pre_shutdown_callback`, `remove_pre_shutdown_callback` and `PreShutdownCallbackHandle` are available in Humble. In rclcpp on the humble branch, `rclcpp/include/rclcpp/context.hpp` declares `PreShutdownCallbackHandle` at line 64, `add_pre_shutdown_callback` at line 271 and `remove_pre_shutdown_callback` at line 281. No version gate is needed here. The Humble jobs on this PR build and test green.

mfaferek93 (approved):

Issue #570, opened 2026-07-29
Detect silent faults in the ROS 2 graph


bburda:
## Summary

Some failures leave the graph looking healthy. Every node is up, nothing logs an error, `/health` is green, and the robot still does not work.

Examples we keep hitting:

- A publisher offers BEST_EFFORT and a subscriber requests RELIABLE. DDS never matches the pair, so the topic exists, both nodes are alive, and no message is ever delivered.
- A remap typo. One node publishes `/scan_raw`, another subscribes `/scan_row`. Both sides are healthy and neither is talking to the other.
- A node dies. The thing that would have reported it is the thing that died.
- A TF broadcaster stops publishing while its process stays alive. Localization and planning break; the graph shows nothing wrong.
- A parameter drifts away from the value the robot was commissioned with.
- A node the operator expects to be active is stuck in `unconfigured` after a failed bringup.
- A pipeline stage goes silently late, so data still flows but too old to use.

None of these produce a fault today. An operator finds them by noticing the robot behaves badly and then reading topics by hand.

## Proposed solution

A gateway plugin that watches the ROS 2 graph and raises a fault for each of these classes, under frozen `GRAPH_*` fault codes so downstream consumers can match on them.

Two things it has to get right to be usable:

- **No bringup noise.** A joining node, a topic still being discovered, and a lifecycle node mid-transition all look like failure for the first few ticks. Raises need to be held until the graph settles.
- **No silent healing.** Once a fault is raised it must not disappear because the evidence went away. A dead node that stops being visible is still dead. Suppression should be opt-in and explicit, never a side effect.

Detectors that need an operator contract (expected parameter values, required-active nodes, per-topic timing budgets) should watch nothing until configured, so the plugin cannot false-positive on a graph nobody set up.

---

## Breakdown and order

One change per fault class, each with its own issue. They merge one at a time: a PR here that targets another PR's branch triggers no workflow at all, because `ci.yml`, `quality.yml` and `docs.yml` all filter on `branches: [main]`. So each one goes in before the next opens.

| # | Change | Issue | Depends on |
|---|--------|-------|------------|
| 1 | Plugin scaffolding + `qos_mismatch` | #572 | - |
| 2 | `orphan` | #577 | 1 |
| 3 | `param_drift` | #579 | 1 |
| 4 | `lifecycle_expectation` | #585 | 1 |
| 5 | Suppression framework + `node_death` | #624 | 1 |
| 6 | `tf_stale` | | 5 |
[...]

PR #571, opened 2026-07-29
feat(graph_watchdog): plugin scaffolding + QoS-mismatch detection

### LINKED ISSUE #572: QoS mismatch silently starves a subscriber
Part of #570.

## Summary

A publisher offers `BEST_EFFORT` and a subscriber requests `RELIABLE`. DDS applies the Request-vs-Offered rule, decides the pair is incompatible, and never matches it. Both nodes are alive, the topic exists, `ros2 topic list` shows it, and not one message is delivered.

Nothing reports this. The subscriber just never gets a callback. An operator finds it by noticing the robot behaves as if the data were missing, then running `ros2 topic info -v` by hand on a topic they already suspect.

The same applies to durability, and to deadline and liveliness lease, where the publisher offering a weaker guarantee than the subscriber requests kills the match.

A partial case is worse to spot: a topic with several publishers where only one is incompatible. That producer.s data is silently discarded while the topic still looks alive and carries traffic from the others.

## Proposed solution

Check every publisher x subscriber pair on every topic against the RxO rule and raise `GRAPH_QOS_MISMATCH`.

Report two levels, because they mean different things to an operator:

- a subscriber incompatible with **every** publisher receives nothing at all
- a subscriber incompatible with **some** publisher still gets data, but one producer is being dropped

The fault should name the affected subscriber FQNs, so the operator does not have to go find them. Endpoint discovery is not atomic, so a subscriber can be visible before a publisher.s QoS is: that needs a persistence threshold or it reports a mismatch every time the graph changes.

bburda:
## Summary

Adds `ros2_medkit_graph_watchdog`, a `GatewayPlugin` for failures where every node is up, nothing logs an error, and the robot is still broken.

This is the first of seven changes. It carries the plugin scaffolding and one detector, so the seam is proven end to end before the rest arrive. Each remaining silent-fault class lands separately against its own issue; their codes are already reserved in the frozen `GRAPH_*` namespace so the contract does not churn.

**Scaffolding**

- detector interface, self-registering registry, nested per-detector config
- unknown config keys and unknown detector ids are warned about at startup, so a typo cannot silently do nothing
- a central reliability gate holding every raise until an entity has been continuously present for `warmup_cycles` ticks, so bringup churn does not read as failure
- `AggregatedFault`: one graph-level fault per code, level-triggered, because fault_manager identifies a fault by `fault_code` alone
- one synthetic App published through `IntrospectionProvider`, so `GRAPH_*` faults resolve to a real entity and roll up to the host component
- a dedicated tick thread, never the gateway executor, since detectors do blocking reads

**`qos_mismatch` (`GRAPH_QOS_MISMATCH`)**

Applies the RxO rule to every publisher x subscriber pair on every topic. Reliability, durability, deadline and liveliness lease are checked; duration comparisons saturate instead of wrapping.

Two levels, because they mean different things to an operator:

- **starved** (ERROR): incompatible with every publisher, so it receives nothing
- **partial** (WARN): incompatible with some publisher, so it still gets data while one producer is silently discarded

The fault names the affected subscriber FQNs. A persistence threshold absorbs non-atomic endpoint discovery, where a subscriber can be visible before a publisher QoS is.

---

## Issue

- closes #572

Part of #570.

---

## Type

- [ ] Bug fix
- [x] New feature or tests
- [ ] Breaking change
- [ ] Documentation only

---

## Testing

18 tests, all green locally on Jazzy: 12 gtest targets (policy units plus integration tests driving real `rclcpp` nodes and real DDS), 1 launch test running a full e2e against a real gateway process, real fault_manager and a real ROS graph, and 5 linters.

Test domains 75-89 are carved out of the gateway range, which shrinks to 30-74 and keeps 12 free slots (it consumes 33 today). Each gtest that creates live ROS nodes gets its own domain; launch tests share one under a `RESOURCE_LOCK`. Sharing a domain between gtests let a publisher outlive its test and change the next one.s result.

    colcon build --packages-select ros2_medkit_graph_watchdog
    colcon test --packages-select ros2_medkit_graph_watchdog && colcon test-result --verbose

---

## Checklist

Two things worth flagging rather than leaving to be found:

- The plugin compiles three gateway sources by path (`lifecycle_status_helpers.cpp`, `ros2_lifecycle_state_reader.cpp`, `ros2_parameter_transport.cpp`) because `ros2_medkit_gateway` exports include dirs but installs no library target. Fine in a workspace build, but it would not survive a released source package, so it needs resolving before this is ever bloom-released.
- `flake8` and `pep257` are excluded, matching every other package here: nothing in this repo lints Python today, including the package holding all the other launch tests. `copyright`, `clang_format`, `cppcheck`, `lint_cmake` and `xmllint` are on.

### DIFF
--- a/src/ros2_medkit_gateway/src/core/managers/operation_manager.cpp
+++ b/src/ros2_medkit_gateway/src/core/managers/operation_manager.cpp
@@ -298,8 +298,12 @@ ActionSendGoalResult OperationManager::send_component_action_goal(const std::str
 }
 
 namespace {
-/// Wall-clock budget for one CancelGoal round-trip. See the call site for why it is not tighter.
-constexpr std::chrono::duration<double> kCancelGoalTimeout{15.0};
+/// Floor for the CancelGoal budget. Cancel is a service call to the action server plus that
+/// server's own handling, so it is bounded by the SERVER, not by us; a very small configured
+/// service timeout would otherwise report a live cancel as a failure. The configured
+/// service_call_timeout_sec still wins whenever it is larger, so lowering it keeps bounding
+/// cancel like every other call in this file.
+constexpr double kCancelGoalFloorSec = 15.0;
 }  // namespace
 
 ActionCancelResult OperationManager::cancel_action_goal(const std::string & action_path, const std::string & goal_id) {
@@ -327,7 +331,9 @@ ActionCancelResult OperationManager::cancel_action_goal(const std::string & acti
   // handling, so it is bounded by the SERVER, not by us. 5s was enough on an idle machine and
   // not on a loaded one: under a busy CI container the request timed out and the cancel was
   // reported as a vendor error while the goal had in fact been accepted for cancellation.
-  result = action_transport_->cancel_goal(action_path, goal_id, kCancelGoalTimeout);
+  const auto cancel_timeout =
+      std::chrono::duration<double>(std::max(static_cast<double>(service_call_timeout_sec_), kCancelGoalFloorSec));
+  result = action_transport_->cancel_goal(action_path, goal_id, cancel_timeout);
 
   if (result.success && result.return_code == 0) {
     update_goal_status(goal_id, ActionGoalStatus::CANCELING);
--- a/src/ros2_medkit_gateway/src/core/managers/operation_manager.cpp
+++ b/src/ros2_medkit_gateway/src/core/managers/operation_manager.cpp
@@ -15,8 +15,8 @@
 #include "ros2_medkit_gateway/core/managers/operation_manager.hpp"
 
 #include <algorithm>
-#include <chrono>
 #include <cctype>
+#include <chrono>
 #include <iomanip>
 #include <random>
 #include <regex>
--- a/src/ros2_medkit_gateway/src/core/managers/operation_manager.cpp
+++ b/src/ros2_medkit_gateway/src/core/managers/operation_manager.cpp
@@ -15,6 +15,7 @@
 #include "ros2_medkit_gateway/core/managers/operation_manager.hpp"
 
 #include <algorithm>
+#include <chrono>
 #include <cctype>
 #include <iomanip>
 #include <random>
@@ -296,6 +297,11 @@ ActionSendGoalResult OperationManager::send_component_action_goal(const std::str
   return send_action_goal(action_path, resolved_type, goal, entity_id);
 }
 
+namespace {
+/// Wall-clock budget for one CancelGoal round-trip. See the call site for why it is not tighter.
+constexpr std::chrono::duration<double> kCancelGoalTimeout{15.0};
+}  // namespace
+
 ActionCancelResult OperationManager::cancel_action_goal(const std::string & action_path, const std::string & goal_id) {
   ActionCancelResult result;
   result.success = false;
@@ -317,7 +323,11 @@ ActionCancelResult OperationManager::cancel_action_goal(const std::string & acti
     return result;
   }
 
-  result = action_transport_->cancel_goal(action_path, goal_id, std::chrono::duration<double>(5.0));
+  // A CancelGoal round-trip is a service call to the action server plus that server's own
+  // handling, so it is bounded by the SERVER, not by us. 5s was enough on an idle machine and
+  // not on a loaded one: under a busy CI container the request timed out and the cancel was
+  // reported as a vendor error while the goal had in fact been accepted for cancellation.
+  result = action_transport_->cancel_goal(action_path, goal_id, kCancelGoalTimeout);
 
   if (result.success && result.return_code == 0) {
     update_goal_status(goal_id, ActionGoalStatus::CANCELING);
--- a/src/ros2_medkit_plugins/ros2_medkit_graph_watchdog/CMakeLists.txt
+++ b/src/ros2_medkit_plugins/ros2_medkit_graph_watchdog/CMakeLists.txt
@@ -87,6 +87,14 @@ target_link_libraries(ros2_medkit_graph_watchdog nlohmann_json::nlohmann_json)
 install(TARGETS ros2_medkit_graph_watchdog LIBRARY DESTINATION lib/${PROJECT_NAME})
 
 if(BUILD_TESTING)
+  include(ROS2MedkitTestDomain)
+  # This package owns 75-89. The gtests below take one domain each through the shared macro,
+  # so the allocation table cannot drift from the literals and exhaustion is a configure
+  # error rather than a silent collision. The launch tests cannot use it: it writes
+  # ENVIRONMENT wholesale and they also need GATEWAY_TEST_PORT and the plugin path, so they
+  # keep an explicit assignment and share one domain under a RESOURCE_LOCK.
+  medkit_init_test_domains(START 75 END 89)
+
   find_package(ament_lint_auto REQUIRED)
   # uncrustify/cpplint conflict with project-wide clang-format (120 cols vs 100).
   # flake8/pep257 are excluded to match the rest of the workspace: no package here
@@ -145,7 +153,7 @@ if(BUILD_TESTING)
   ament_add_gtest(test_watchdog_clock test/test_watchdog_clock.cpp)
   target_include_directories(test_watchdog_clock PRIVATE include ${GATEWAY_SRC_INCLUDE_DIR})
   medkit_target_dependencies(test_watchdog_clock rclcpp rosgraph_msgs)
-  set_tests_properties(test_watchdog_clock PROPERTIES ENVIRONMENT "ROS_DOMAIN_ID=75")
+  medkit_set_test_domain(test_watchdog_clock)
 
   ament_add_gtest(test_graph_watchdog_plugin
     test/test_graph_watchdog_plugin.cpp
@@ -159,7 +167,7 @@ if(BUILD_TESTING)
   medkit_target_dependencies(test_graph_watchdog_plugin
     ros2_medkit_gateway rclcpp ros2_medkit_msgs lifecycle_msgs)
   target_link_libraries(test_graph_watchdog_plugin nlohmann_json::nlohmann_json)
-  set_tests_properties(test_graph_watchdog_plugin PROPERTIES ENVIRONMENT "ROS_DOMAIN_ID=76")
+  medkit_set_test_domain(test_graph_watchdog_plugin)
 
   ament_add_gtest(test_lifecycle_watcher
     test/test_lifecycle_watcher.cpp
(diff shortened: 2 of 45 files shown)


src/ros2_medkit_gateway/src/core/managers/operation_manager.cpp
```diff
@@ -317,7 +323,11 @@ ActionCancelResult OperationManager::cancel_action_goal(const std::string & acti
     return result;
   }
 
-  result = action_transport_->cancel_goal(action_path, goal_id, std::chrono::duration<double>(5.0));
+  // A CancelGoal round-trip is a service call to the action server plus that server's own
+  // handling, so it is bounded by the SERVER, not by us. 5s was enough on an idle machine and
+  // not on a loaded one: under a busy CI container the request timed out and the cancel was
+  // reported as a vendor error while the goal had in fact been accepted for cancellation.
+  result = action_transport_->cancel_goal(action_path, goal_id, kCancelGoalTimeout);
```
mfaferek93: `kCancelGoalTimeout` is a file-local constant, so cancel is now the only call in this file that ignores `service_call_timeout_sec_`. That parameter is declared in `gateway_node.cpp:596` (default 10) and its header doc says it is "applied to every service / action call"; the three sibling call sites (lines 120, 243, 356) still use it.

Two consequences: an operator who lowers the parameter no longer bounds cancel at all (`test/test_operation_handlers.cpp:256` launches the gateway with `service_call_timeout_sec:=1` for exactly that reason), and the HTTP worker thread serving a cancel can now sit on an unresponsive action server for 15s plus the 2s `wait_for_service` in `ros2_action_transport.cpp:279`, up from 7s.

Raising the budget is right; hardcoding it is what breaks the contract. Derive it from `service_call_timeout_sec_` (e.g. `std::max(service_call_timeout_sec_, floor)`) or declare a real parameter for cancel.

src/ros2_medkit_plugins/ros2_medkit_graph_watchdog/include/ros2_medkit_graph_watchdog/detector_config_keys.hpp
```diff
@@ -0,0 +1,81 @@
...
+
+// Deliberately ROS-free (no detector.hpp, no rclcpp): the pure policy headers that need
+// this - param_drift_policy.hpp and friends - are unit-tested without a ROS build.
+namespace ros2_medkit_graph_watchdog {
+
+/// Keys the PLUGIN injects into every detector's config before calling configure()
+/// (GraphWatchdogPlugin::set_context). They are plumbing, not a detector's own keys, so a
+/// detector that ignores one must never report it as an operator typo.
+inline const std::set<std::string> & plugin_injected_detector_keys() {
+  static const std::set<std::string> keys{"tick_interval_ms"};
```
mfaferek93: `plugin_injected_detector_keys()` lists only `tick_interval_ms`, but `set_context()` also injects `prune_grace` into every detector's config (`src/graph_watchdog_plugin.cpp:206`). `collect_unknown_detector_keys()` then reports the plugin's own injected key as an operator typo.

Confirmed on the e2e run in this branch, with no user config at all:

```
[gateway_node-1] [WARN] [ros2_medkit_gateway]: graph_watchdog qos_mismatch: unknown config key 'prune_grace'; it has no effect (known keys: allowlist, grace)
```

So the one mechanism that exists to catch config typos cries wolf on every boot, and every future detector that does not happen to read `prune_grace` inherits it. An operator acting on the message deletes a key they may legitimately have set elsewhere.

Fix: add `"prune_grace"` to this set.

src/ros2_medkit_plugins/ros2_medkit_graph_watchdog/src/detectors/qos_mismatch_detector.cpp
```diff
@@ -0,0 +1,245 @@
...
+    // a graph that churns topic names.
+    for (auto it = streak_.begin(); it != streak_.end();) {
+      it = seen_topics.count(it->first) == 0 ? streak_.erase(it) : std::next(it);
+    }
+
+    // One fault code carries one severity, so it reflects the worst finding in the sweep:
+    // ERROR as soon as any subscriber anywhere receives nothing at all, WARN when every
+    // finding is a degraded-but-alive topic.
+    AggregatedFault aggregated{graph_fault_codes::kQosMismatch, any_starved ? kStarvedSeverity : kPartialSeverity};
+    aggregated.emit(ctx, affected);
```
mfaferek93: `emit()` lists affected topics lexicographically and `describe_ordered()` truncates at 480 chars (`aggregated_fault.hpp:69`). One entry from the e2e run is `/probe: reliability: publisher BEST_EFFORT vs subscriber RELIABLE (receives nothing: /qos_e2e_sub)`, about 95 characters, so the cap is hit at roughly five affected topics.

Past that point a newly starved subscriber on a late-sorting topic produces no operator-visible change at all: the fault code is already raised, the description is already full, and severity only moves if the new finding happens to be the first `Starved` one. That is the failure this plugin is supposed to remove, not create.

`aggregated_fault.hpp:83-90` documents this exact hazard and adds `emit_ordered()` for it, but this detector uses the plain `emit()`. Track the topics whose streak crossed `grace_` on this tick and pass them first via `emit_ordered()`.

src/ros2_medkit_gateway/src/core/managers/operation_manager.cpp
bburda: You are right, hardcoding it breaks the contract. The budget now comes from the parameter:

```cpp
const auto cancel_timeout =
    std::chrono::duration<double>(std::max(static_cast<double>(service_call_timeout_sec_), kCancelGoalFloorSec));
```

So lowering `service_call_timeout_sec` bounds cancel again, and the floor only applies when the configured value is smaller than it. The test that launches with `service_call_timeout_sec:=1` keeps working.

On why there is a floor at all: cancel is a service call to the action server plus that server's own handling, so a very small value can report a live cancel as a failure.

I should also say the 15s was a stopgap, not a fix. I measured the cancel round trip at about 1 ms in every setup I tried, including two pinned cores under load and CycloneDDS, so the 5s failure in CI is a lost response, not a tight budget. I opened separate issues for the real cause and for the wrong error code it produces.

src/ros2_medkit_plugins/ros2_medkit_graph_watchdog/include/ros2_medkit_graph_watchdog/detector_config_keys.hpp
bburda: Fixed, `prune_grace` is in the set now.

This was the worst kind of bug in that mechanism: it fired on every boot with no user config, so the warning that exists to catch typos was training people to ignore it.

src/ros2_medkit_plugins/ros2_medkit_graph_watchdog/src/detectors/qos_mismatch_detector.cpp
bburda: Fixed. The detector now collects the topics whose streak crossed `grace_` on this tick and passes them to `emit_ordered()`:

```cpp
if (streak_[topic] == grace_ + 1) {
  newly_affected.push_back(topic);
}
...
aggregated.emit_ordered(ctx, affected, newly_affected);
```

So a newly starved subscriber is named first and cannot be cut by the 480 char cap, whatever its topic name sorts as.

src/ros2_medkit_plugins/ros2_medkit_graph_watchdog/CMakeLists.txt
```diff
@@ -0,0 +1,623 @@
...
+  ament_add_gtest(test_detector_config test/test_detector_config.cpp)
+  target_include_directories(test_detector_config PRIVATE include ${GATEWAY_SRC_INCLUDE_DIR})
+  medkit_target_dependencies(test_detector_config rclcpp ros2_medkit_msgs)
+  target_link_libraries(test_detector_config nlohmann_json::nlohmann_json)
+
+  ament_add_gtest(test_watchdog_clock test/test_watchdog_clock.cpp)
+  target_include_directories(test_watchdog_clock PRIVATE include ${GATEWAY_SRC_INCLUDE_DIR})
+  medkit_target_dependencies(test_watchdog_clock rclcpp rosgraph_msgs)
+  set_tests_properties(test_watchdog_clock PROPERTIES ENVIRONMENT "ROS_DOMAIN_ID=75")
+
```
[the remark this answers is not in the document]
bburda: Fixed. The package now calls `medkit_init_test_domains(START 75 END 89)` and every gtest that creates live ROS nodes takes its domain through `medkit_set_test_domain`, so exhaustion is a configure error instead of a silent reuse.

The launch tests are the exception and keep an explicit assignment. `medkit_set_test_domain` writes the whole `ENVIRONMENT` property, and those tests also need `GATEWAY_TEST_PORT` and the plugin path, so the macro would drop them. They share one domain under a `RESOURCE_LOCK`. The reason is written at the call site.

src/ros2_medkit_plugins/ros2_medkit_graph_watchdog/src/graph_watchdog_plugin.cpp
```diff
@@ -0,0 +1,391 @@
...
+    if (!gate_) {
+      res.send_error(503, ros2_medkit_gateway::ERR_SERVICE_UNAVAILABLE,
+                     "graph_watchdog reliability gate not initialized");
+      return;
+    }
+    res.send_json(gate_->status_json());
+  };
+  std::vector<PluginRoute> routes;
+  routes.push_back({"GET", R"(x-medkit-watchdog)", std::move(handler)});
+  return routes;
```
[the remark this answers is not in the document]
bburda: Added to `docs/api/rest.rst`, next to the other plugin route. It documents what the payload carries and what an unarmed entity means, since "present but not armed" during bringup is the thing people will ask about first.

src/ros2_medkit_plugins/ros2_medkit_graph_watchdog/test/e2e/test_config_plumbing_e2e.test.py
```diff
@@ -0,0 +1,192 @@
...
+        demo_nodes=['temp_sensor'],
+        port=PORT,
+    )
+
+
+class TestConfigPlumbingPositive(unittest.TestCase):
+    """Positive control: the nested `expect` config reaches param_drift."""
+
+    @unittest.skipUnless(SCENARIO == 'positive', 'runs only under the "positive" scenario launch')
+    def test_expect_rule_raises_from_nested_config(self):
```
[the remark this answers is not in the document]
bburda: This file is no longer in the pull request. The change was cut down to the plugin scaffolding plus one detector, so the config plumbing e2e moves to the change that adds `param_drift`, which is what its scenarios exercise.

The point still stands and I will apply it there.

src/ros2_medkit_plugins/ros2_medkit_graph_watchdog/test/e2e/test_tf_publisher_liveness_e2e.test.py
```diff
@@ -0,0 +1,225 @@
...
+    thread.start()
+    return thread
+
+
+class TestTfPublisherLivenessDeadPublisher(unittest.TestCase):
+    """A total /tf blackout (the ONLY publisher destroyed entirely) never raises
+    GRAPH_TF_STALE once require_publisher_alive is on."""
+
+    @unittest.skipUnless(SCENARIO == 'dead_publisher', 'runs only under the "dead_publisher" scenario launch')
+    def test_dead_publisher_suppresses_tf_stale(self):
```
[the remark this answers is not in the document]
bburda: This file is no longer in the pull request either. It comes back with the `tf_stale` detector, and I will split the scenarios then.

mfaferek93 (approved):

PR #573, opened 2026-07-29
Stop PASSED events re-dating a fault, and count real SSE loss

mfaferek93:
Two defects found on a live diagnostic box, plus the fix for each.

## A stale fault reads as freshly active

`PLC_COMMS_LOST` on the Beckhoff ADS instance sat `CONFIRMED` for hours while the link was
demonstrably alive, and its `last_occurred` matched the plugin's clear log to the millisecond.

The `CONFIRMED` latch itself is **not** a bug: `compute_debounce_status` only returns `HEALED` when
healing is enabled, the box runs `healing=disabled`, and that is the documented default. What is a
bug is that PASSED events advanced `last_occurred`, so a fault cleared three hours ago presented as
current. The clear instant already has its own field, `last_passed_ns`; `last_occurred` is now
written only on FAILED.

Found next to it and fixed here too: no event was published on the `CONFIRMED -> HEALED`
transition. The shipped `bringup_params.yaml` sets `healing_enabled: true`, so on the default stack
a healed fault stayed on every SSE consumer's screen indefinitely. It now publishes `EVENT_CLEARED`
with `status: HEALED` - deliberately the existing event name, since a new one would be silently
dropped by any UI not listening for it.

## The SSE drop counter measured the wrong thing

The handler trimmed a shared 100-entry replay buffer FIFO and counted every eviction as a
client-visible drop, with no notion of clients at all. On the box the counter passed 20,000 while
`SSE fault client connected` appeared zero times in the log - nothing had ever been lost, the ring
was simply rotating, and the "slow or disconnected clients" text was fiction.

Eviction is now cursor-aware. Entries every live client has already received go first and cost
nothing. Entries superseded by a newer event for the same fault code go next, so a lagging client
still converges on the current state of every fault; entries carrying `auto_cleared_codes` are
exempt, because those symptom codes appear nowhere else. Only what neither rule covers is counted
and logged as an actual loss, with the slow-client, coalesced and reaped counts in the message.
A client with no completed write for three keepalive intervals is reaped and excluded from the
accounting rather than inflating backpressure forever.

Because coalescing erases mid-deque, the stream loop now formats under the lock and writes outside
it; the previous code held a `deque` iterator across `sink.write`.

## Verification

Built and tested in the CI image (Jazzy, CycloneDDS). Gateway 3959 tests, fault_manager 527 tests,
zero failures. `clang-format-18` run deliberately rather than left to be skipped: 438 files in the
gateway, 36 in fault_manager, clean.

Each fix has a test that fails when it is reverted:

- `FaultStorageTest.PassedEventDoesNotAdvanceLastOccurred` and the sqlite equivalent
- `HealingFaultEventPublishingTest` - "no event published for the CONFIRMED -> HEALED transition"
- `BufferRotationWithNoClientsIsNotADrop`, `LaggingClientKeepsRareFaultWhenFlapFillsBuffer`,
  `StalledClientIsReapedInsteadOfCountedAsBackpressure`

`test_graph_provider_{sse,greenwave,stale}` in `ros2_medkit_integration_tests` fail in this
container, and also fail on an unmodified `origin/main` build with the same signature
(`rclcpp::graph_listener::NodeNotFoundError` aborting the gateway). Pre-existing, not from this
change.

### DIFF
--- a/src/ros2_medkit_fault_manager/src/sqlite_fault_storage.cpp
+++ b/src/ros2_medkit_fault_manager/src/sqlite_fault_storage.cpp
@@ -191,6 +191,19 @@ void SqliteFaultStorage::initialize_schema() {
     }
   }
 
+  // Migration: releases that advanced last_occurred_ns on PASSED events left
+  // inflated rows behind, and a latched CONFIRMED fault that only ever heals
+  // would keep the wrong timestamp forever. last_failed_ns holds the true
+  // last occurrence; last_occurred_ns can only exceed it via that old bug.
+  if (sqlite3_exec(db_,
+                   "UPDATE faults SET last_occurred_ns = last_failed_ns "
+                   "WHERE last_failed_ns > 0 AND last_occurred_ns > last_failed_ns",
+                   nullptr, nullptr, &err_msg) != SQLITE_OK) {
+    std::string error = err_msg ? err_msg : "Unknown error";
+    sqlite3_free(err_msg);
+    throw std::runtime_error("Failed to repair last_occurred_ns rows: " + error);
+  }
+
   // Create snapshots table for storing topic data captured when faults are confirmed
   const char * create_snapshots_table_sql = R"(
     CREATE TABLE IF NOT EXISTS snapshots (
--- a/src/ros2_medkit_fault_manager/src/sqlite_fault_storage.cpp
+++ b/src/ros2_medkit_fault_manager/src/sqlite_fault_storage.cpp
@@ -608,7 +608,7 @@ SqliteFaultStorage::list_faults(bool filter_by_severity, uint8_t severity,
   // Build query
   std::string sql =
       "SELECT fault_code, severity, description, first_occurred_ns, last_occurred_ns, "
-      "occurrence_count, status, reporting_sources FROM faults WHERE status IN (";
+      "occurrence_count, status, reporting_sources, last_passed_ns FROM faults WHERE status IN (";
   for (size_t i = 0; i < status_filter.size(); ++i) {
     if (i > 0) {
       sql += ", ";
@@ -646,6 +646,7 @@ SqliteFaultStorage::list_faults(bool filter_by_severity, uint8_t severity,
     fault.occurrence_count = static_cast<uint32_t>(stmt.column_int64(5));
     fault.status = stmt.column_text(6);
     fault.reporting_sources = parse_json_array(stmt.column_text(7));
+    fault.last_passed = rclcpp::Time(stmt.column_int64(8), RCL_SYSTEM_TIME);
 
     result.push_back(fault);
   }
@@ -658,7 +659,7 @@ std::optional<ros2_medkit_msgs::msg::Fault> SqliteFaultStorage::get_fault(const 
 
   SqliteStatement stmt(db_,
                        "SELECT fault_code, severity, description, first_occurred_ns, last_occurred_ns, "
-                       "occurrence_count, status, reporting_sources FROM faults WHERE fault_code = ?");
+                       "occurrence_count, status, reporting_sources, last_passed_ns FROM faults WHERE fault_code = ?");
   stmt.bind_text(1, fault_code);
 
   if (stmt.step() != SQLITE_ROW) {
@@ -678,6 +679,7 @@ std::optional<ros2_medkit_msgs::msg::Fault> SqliteFaultStorage::get_fault(const 
   fault.occurrence_count = static_cast<uint32_t>(stmt.column_int64(5));
   fault.status = stmt.column_text(6);
   fault.reporting_sources = parse_json_array(stmt.column_text(7));
+  fault.last_passed = rclcpp::Time(stmt.column_int64(8), RCL_SYSTEM_TIME);
 
   return fault;
 }
@@ -1178,7 +1180,7 @@ std::vector<ros2_medkit_msgs::msg::Fault> SqliteFaultStorage::get_all_faults() c
 
   SqliteStatement stmt(db_,
                        "SELECT fault_code, severity, description, first_occurred_ns, last_occurred_ns, "
-                       "occurrence_count, status, reporting_sources FROM faults");
+                       "occurrence_count, status, reporting_sources, last_passed_ns FROM faults");
 
   std::vector<ros2_medkit_msgs::msg::Fault> result;
   while (stmt.step() == SQLITE_ROW) {
@@ -1195,6 +1197,7 @@ std::vector<ros2_medkit_msgs::msg::Fault> SqliteFaultStorage::get_all_faults() c
     fault.occurrence_count = static_cast<uint32_t>(stmt.column_int64(5));
     fault.status = stmt.column_text(6);
     fault.reporting_sources = parse_json_array(stmt.column_text(7));
+    fault.last_passed = rclcpp::Time(stmt.column_int64(8), RCL_SYSTEM_TIME);
 
     result.push_back(fault);
   }
--- a/src/ros2_medkit_fault_manager/src/sqlite_fault_storage.cpp
+++ b/src/ros2_medkit_fault_manager/src/sqlite_fault_storage.cpp
@@ -510,16 +510,15 @@ bool SqliteFaultStorage::report_fault_event(const std::string & fault_code, uint
 
       std::string new_status = compute_debounce_status(debounce_counter, current_status, config);
 
+      // last_occurred_ns is deliberately NOT touched: a PASSED event is the fault
+      // ENDING, not occurring. Bumping it made a long-stale CONFIRMED fault look
+      // freshly active. The PASSED instant is recorded in last_passed_ns.
       SqliteStatement update_stmt(
-          db_,
-          "UPDATE faults SET last_occurred_ns = ?, last_passed_ns = ?, status = ?, debounce_counter "
-          "= ? WHERE "
-          "fault_code = ?");
+          db_, "UPDATE faults SET last_passed_ns = ?, status = ?, debounce_counter = ? WHERE fault_code = ?");
       update_stmt.bind_int64(1, timestamp_ns);
-      update_stmt.bind_int64(2, timestamp_ns);
-      update_stmt.bind_text(3, new_status);
-      update_stmt.bind_int(4, debounce_counter);
-      update_stmt.bind_text(5, fault_code);
+      update_stmt.bind_text(2, new_status);
+      update_stmt.bind_int(3, debounce_counter);
+      update_stmt.bind_text(4, fault_code);
 
       if (update_stmt.step() != SQLITE_DONE) {
         throw std::runtime_error(std::string("Failed to update fault: ") + sqlite3_errmsg(db_));
--- a/src/ros2_medkit_gateway/README.md
+++ b/src/ros2_medkit_gateway/README.md
@@ -793,8 +793,9 @@ Real-time fault event stream using Server-Sent Events (SSE). Clients receive ins
 - **Real-time notifications**: Events pushed instantly when fault state changes
 - **Automatic reconnection**: Supports `Last-Event-ID` header for seamless reconnection
 - **Keepalive**: Sends `:keepalive` comment every 30 seconds to prevent timeouts
-- **Event buffer**: Buffers up to 100 recent events for reconnecting clients. Under overflow the buffer drops entries every live client has already received, then entries superseded by a newer event for the same fault code, so a lagging client still converges on the current state of every fault
+- **Event buffer**: Buffers up to 100 recent events for reconnecting clients. Under overflow the buffer evicts entries every live client has already received, then `fault_updated` entries superseded by a newer event for the same fault code - a lagging client still converges on the current state of every fault and loses no status transition. Anything beyond that (a superseded transition, or an event with no newer sibling) is genuinely lost for lagging clients: those losses are counted and logged as drops, and an affected client should refetch `GET /api/v1/faults` to resynchronize
 - **Entity context (SOVD payload extension)**: When the gateway can resolve the fault's first reporting source back to an entity, the payload carries an `x-medkit` object with `entity_type` and `entity_id` fields so consumers can hit `/{entity_type}/{entity_id}/bulk-data/rosbags/{fault_code}` directly without enumerating entities
+- **Correlation payload**: when a root-cause event auto-clears correlated symptom faults, the payload carries their codes in `auto_cleared_codes` (omitted when empty); those symptoms get no event of their own
 
 **Event Types:**
 - `fault_confirmed` - Fault transitioned to CONFIRMED status
@@ -833,7 +834,7 @@ data: {"event_type":"fault_cleared","fault":{"fault_code":"MOTOR_OVERHEAT",...},
 curl -N -H "Last-Event-ID: 5" http://localhost:8080/api/v1/faults/stream
 ```
 
-This replays any buffered events with ID > 5, then continues streaming new events.
+This replays any buffered events with ID > 5, then continues streaming new events. Replayed `id` sequences may contain holes (e.g. 41, 43, 44) where buffer eviction removed superseded events; ids are monotonically increasing but never guaranteed contiguous. A `Last-Event-ID` beyond the newest issued id is treated as "caught up": the client receives only events published after it connected.
 
 #### GET /api/v1/components/{component_id}/faults
 
--- a/src/ros2_medkit_gateway/README.md
+++ b/src/ros2_medkit_gateway/README.md
@@ -793,13 +793,13 @@ Real-time fault event stream using Server-Sent Events (SSE). Clients receive ins
 - **Real-time notifications**: Events pushed instantly when fault state changes
 - **Automatic reconnection**: Supports `Last-Event-ID` header for seamless reconnection
 - **Keepalive**: Sends `:keepalive` comment every 30 seconds to prevent timeouts
-- **Event buffer**: Buffers up to 100 recent events for reconnecting clients
+- **Event buffer**: Buffers up to 100 recent events for reconnecting clients. Under overflow the buffer drops entries every live client has already received, then entries superseded by a newer event for the same fault code, so a lagging client still converges on the current state of every fault
 - **Entity context (SOVD payload extension)**: When the gateway can resolve the fault's first reporting source back to an entity, the payload carries an `x-medkit` object with `entity_type` and `entity_id` fields so consumers can hit `/{entity_type}/{entity_id}/bulk-data/rosbags/{fault_code}` directly without enumerating entities
 
 **Event Types:**
 - `fault_confirmed` - Fault transitioned to CONFIRMED status
 - `fault_updated` - Fault data changed (occurrence_count, sources, etc.)
-- `fault_cleared` - Fault was cleared via ClearFault service
+- `fault_cleared` - Fault ended: cleared via ClearFault service, or auto-healed by PASSED events (check `fault.status` for `CLEARED` vs `HEALED`)
 
 **Example:**
 ```bash
--- a/src/ros2_medkit_gateway/include/ros2_medkit_gateway/http/handlers/sse_fault_handler.hpp
+++ b/src/ros2_medkit_gateway/include/ros2_medkit_gateway/http/handlers/sse_fault_handler.hpp
@@ -82,7 +82,11 @@ class SSEFaultHandler {
   SSEFaultHandler(HandlerContext & ctx, std::shared_ptr<SSEClientTracker> client_tracker,
                   std::chrono::milliseconds keepalive_interval);
 
-  /// Destructor - cleanup subscription
+  /// Destructor. Signals shutdown, then blocks until every stream closure has
+  /// been destroyed: the per-stream unregister deleter locks queue_mutex_ and
+  /// touches clients_, so no closure may outlive the handler. The HTTP server
+  /// must already be stopping (request_shutdown() makes every loop return
+  /// false on its next wakeup) or destruction will wait for it.
   ~SSEFaultHandler();
 
   // Disable copy/move
--- a/src/ros2_medkit_gateway/include/ros2_medkit_gateway/http/handlers/sse_fault_handler.hpp
+++ b/src/ros2_medkit_gateway/include/ros2_medkit_gateway/http/handlers/sse_fault_handler.hpp
@@ -20,7 +20,6 @@
 #include <cstdint>
 #include <deque>
 #include <functional>
-#include <limits>
 #include <memory>
 #include <mutex>
 #include <optional>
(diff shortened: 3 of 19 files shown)


src/ros2_medkit_gateway/src/http/handlers/sse_fault_handler.cpp
```diff
@@ -148,69 +234,108 @@ uint64_t parse_last_event_id(std::string_view value) {
...
+  cursor->last_progress = std::chrono::steady_clock::now();
+}
+
 std::function<bool(httplib::DataSink &)> SSEFaultHandler::make_stream_loop(uint64_t initial_last_event_id) {
-  return [this, last_event_id = initial_last_event_id](httplib::DataSink & sink) mutable -> bool {
-    // First, send any buffered events the client missed (for reconnection)
-    {
-      std::lock_guard<std::mutex> lock(queue_mutex_);
+  auto cursor = std::make_shared<ClientCursor>();
+  cursor->last_delivered = initial_last_event_id;
```
bburda, line 254: Could you clamp the value that seeds the cursor? `parse_last_event_id` only maps a throw to 0, and `std::stoull` accepts a negative string: `Last-Event-ID: -1` returns 18446744073709551615, which is exactly `kNothingOwed`. I compiled this to be sure.

One header then breaks three things at once, all without an error or a log line:

- The stream goes blind. `collect_pending` tests `queued.id > last_event_id`, which can never be true, so the client gets nothing for as long as it stays connected.
- The watermark aliases. If this is the only client, `delivered_watermark_locked()` returns `kNothingOwed`, so the whole buffer is evicted as "every live client already has it" and no loss is counted. A connected client looks the same as no client.
- Keepalives stop. `keepalive_due` needs `status == timeout`, but every `notify_all` returns `no_timeout`, so with steady fault traffic this client never writes and never records progress.

Capping the parsed value at `next_event_id_ - 1` would cover it. It would also help to use a sentinel a client cannot reach, for example `std::optional<uint64_t>` for the watermark.

src/ros2_medkit_gateway/src/http/handlers/sse_fault_handler.cpp
```diff
@@ -89,6 +92,82 @@ void SSEFaultHandler::request_shutdown() {
...
+  for (const auto & cursor : clients_) {
+    if (cursor->stalled) {
+      continue;  // dead client: its backlog is not backpressure
+    }
+    watermark = std::min(watermark, cursor->last_delivered);
+  }
+  return watermark;
+}
+
+std::deque<SSEFaultHandler::QueuedEvent>::iterator SSEFaultHandler::find_superseded_locked() {
```
bburda: This keys only on `fault_code` and ignores `event_type`, so a status transition can be erased and reported as coalesced. Take three buffered events for the same code: id 10 `fault_confirmed`, id 11 `fault_cleared` (the heal event this PR adds), id 12 `fault_confirmed`. Under pressure ids 10 and 11 are both erased and a lagging client sees only id 12.

Two results. A client using `addEventListener('fault_confirmed', ...)`, which the README recommends, never sees the confirm if the survivor is a `fault_updated`. And a fault that ended and came back reads as one unbroken fault, so downtime counting is wrong. Neither is counted as a drop, and the header documents `coalesced_events` as explicitly not a loss.

Could you exempt transitions the same way `auto_cleared_codes` is exempt, so a `fault_updated` may only supersede an earlier `fault_updated` for the same code?

src/ros2_medkit_gateway/test/test_sse_fault_handler.cpp
```diff
@@ -399,6 +399,101 @@ TEST_F(SSEFaultHandlerTest, BufferEvictsOldestEventWhenFull) {
...
+  // Observed on a live diagnostic box: 19,000+ "events dropped ... slow or
+  // disconnected clients" with zero SSE clients ever connected. Rotating a
+  // replay buffer nobody is reading loses nothing and must not be reported as
+  // backpressure.
+  for (int i = 1; i <= 150; ++i) {
+    enqueue_event(make_fault_event(FaultEvent::EVENT_CONFIRMED, "FAULT_" + std::to_string(i), i));
+  }
+
+  EXPECT_EQ(handler_->connected_clients(), 0u);
+  EXPECT_EQ(handler_->dropped_events(), 0u);
```
bburda, line 434: Can you add a test where a real loss happens? Every assertion on this counter in the file is `EXPECT_EQ(..., 0u)` (this line, line 430, and line 491), so `stats.dropped > 0` is never reached. That leaves the genuine-loss branch, the `slow_clients` tally, and the whole new WARN untested, including the new first-loss condition.

That is the behaviour the PR title promises, and today you could hardcode the counter to never increment and the suite would stay green.

It is a small variant of a test you already have: register a cursor the way `LaggingClientKeepsRareFaultWhenFlapFillsBuffer` does, then enqueue 150 distinct fault codes at the default keepalive so nothing is reaped and nothing can be coalesced. That should give 50 drops and one slow client.

src/ros2_medkit_gateway/test/test_sse_fault_handler.cpp
```diff
@@ -399,6 +399,101 @@ TEST_F(SSEFaultHandlerTest, BufferEvictsOldestEventWhenFull) {
...
+  EXPECT_GT(handler_->coalesced_events(), 0u);
+
+  auto output = read_stream_once(res, 1);
+  EXPECT_EQ(parse_sse_payload(output)["fault"]["fault_code"], "PLC_COMMS_LOST");
+
+  release_stream(res);
+}
+
+TEST_F(SSEFaultHandlerTest, CoalescedReplayEndsOnTheCurrentStateOfEachFault) {
```
bburda: This test does not exercise coalescing. The 151 events are enqueued before `handle_stream` runs, so no cursor exists while the buffer is being trimmed. `delivered_watermark_locked()` returns `kNothingOwed`, every eviction takes the plain `pop_front` branch, and nothing is ever superseded.

The assertion holds under the old FIFO code too, so this passes on unmodified main. The name and the "Convergence contract" comment promise coverage the test does not give.

Could you open the stream before the burst, as the lagging-client test does?

src/ros2_medkit_fault_manager/src/sqlite_fault_storage.cpp
```diff
@@ -510,16 +510,15 @@ bool SqliteFaultStorage::report_fault_event(const std::string & fault_code, uint
...
 
+      // last_occurred_ns is deliberately NOT touched: a PASSED event is the fault
+      // ENDING, not occurring. Bumping it made a long-stale CONFIRMED fault look
+      // freshly active. The PASSED instant is recorded in last_passed_ns.
       SqliteStatement update_stmt(
-          db_,
-          "UPDATE faults SET last_occurred_ns = ?, last_passed_ns = ?, status = ?, debounce_counter "
-          "= ? WHERE "
-          "fault_code = ?");
+          db_, "UPDATE faults SET last_passed_ns = ?, status = ?, debounce_counter = ? WHERE fault_code = ?");
```
bburda, line 530: Could you backfill existing rows here? Databases written by the current release already hold a `last_occurred_ns` that a PASSED event inflated, and nothing corrects them.

This hits the default deployment: `storage_type` defaults to sqlite with a persistent database path. It also hits exactly the case this PR is about. A latched CONFIRMED fault whose source only sends PASSED after it recovers may never get another FAILED, so the wrong timestamp stays forever and the box does not repair itself on upgrade.

The right value is already in the same row, and this file already has the migration idiom used for `confirmed_at_ns`:

```sql
UPDATE faults SET last_occurred_ns = last_failed_ns
WHERE last_failed_ns > 0 AND last_occurred_ns > last_failed_ns;
```

src/ros2_medkit_gateway/src/http/handlers/sse_fault_handler.cpp
```diff
@@ -148,69 +234,108 @@ uint64_t parse_last_event_id(std::string_view value) {
...
+      bool keepalive_due = false;
+      {
+        std::unique_lock<std::mutex> lock(queue_mutex_);
+        if (shutdown_flag_.load()) {
+          return false;  // Handler is shutting down
+        }
 
-      // Wait for new event or timeout
-      auto status = queue_cv_.wait_for(lock, timeout);
+        const auto status = queue_cv_.wait_for(lock, timeout);
```
bburda: This waits before it looks at the queue, and the loop holds no lock between iterations. So an event pushed after `flush` returns and before this line is reached loses its `notify_all`, and it then waits for the full keepalive interval, which is 30 seconds in production.

Nothing is lost, but the README says events are pushed instantly when fault state changes, and a single confirmed fault followed by a quiet period can be up to 30 seconds late. The old code held the lock across iterations, so the window is wider now.

The predicate form of `wait_for` checks before it waits and would close it:

```cpp
queue_cv_.wait_for(lock, timeout, [&]{
  return shutdown_flag_.load() || cursor->stalled || has_pending_locked(last_event_id);
});
```

Then take `keepalive_due` from its `false` return.

src/ros2_medkit_gateway/README.md
```diff
@@ -793,13 +793,13 @@ Real-time fault event stream using Server-Sent Events (SSE). Clients receive ins
 - **Real-time notifications**: Events pushed instantly when fault state changes
 - **Automatic reconnection**: Supports `Last-Event-ID` header for seamless reconnection
 - **Keepalive**: Sends `:keepalive` comment every 30 seconds to prevent timeouts
-- **Event buffer**: Buffers up to 100 recent events for reconnecting clients
+- **Event buffer**: Buffers up to 100 recent events for reconnecting clients. Under overflow the buffer drops entries every live client has already received, then entries superseded by a newer event for the same fault code, so a lagging client still converges on the current state of every fault
```
bburda: This describes two of the three eviction branches. `evict_to_capacity_locked` also has a third one, where the entry is owed to a live client, no newer event exists for that fault code, and it is dropped for real.

So the closing phrase "still converges on the current state of every fault" is only true while that branch never runs. The header comment states the loss case correctly, so the user-facing text is weaker than the internal one. Could you add the loss case here?

While you are in this section: after coalescing, a replayed `id` sequence can have holes for the first time (for example 41, 43, 44). Nothing promised contiguity, but the old text implied FIFO, so one sentence near the `Last-Event-ID` example would help a client author.

src/ros2_medkit_gateway/src/http/handlers/sse_fault_handler.cpp
```diff
@@ -148,69 +234,108 @@ uint64_t parse_last_event_id(std::string_view value) {
...
+  cursor->last_progress = std::chrono::steady_clock::now();
+  {
+    std::lock_guard<std::mutex> lock(queue_mutex_);
+    clients_.push_back(cursor);
+  }
+
+  // Deregisters when the content provider (and with it this closure) is
+  // destroyed, which is the only point at which the connection is certainly
+  // gone for both the typed and the legacy entry.
+  auto unregister = std::shared_ptr<void>(nullptr, [this, cursor](void *) {
```
bburda, line 263: This deleter dereferences `this` and locks `queue_mutex_` when the content provider is destroyed, which can be on any thread and at any time. Nothing in the handler waits for in-flight stream loops: `request_shutdown()` sets a flag and notifies, and the destructor does not join.

Today production is safe, but only because a different class joins the server thread before the handler is destroyed, and in this class the handler member is destroyed before the HTTP server. I could not build a failing sequence in tree, so this is a question rather than a defect. Since this PR adds the first code path that touches handler state at provider destruction time, would you consider making the invariant explicit, for example an active-stream count that `request_shutdown()` waits on?

src/ros2_medkit_gateway/include/ros2_medkit_gateway/http/handlers/sse_fault_handler.hpp
```diff
@@ -215,6 +289,13 @@ class SSEFaultHandler {
 
   /// Keepalive interval in seconds
   static constexpr int kKeepaliveIntervalSec = 30;
+
+  /// A client that has not completed a write (event or keepalive) for this many
+  /// keepalive intervals is dead, not slow.
+  static constexpr int kStallIntervals = 3;
```
bburda: Can you check whether this can ever be reached? The budget is `keepalive_interval_ * kStallIntervals`, which is 90 seconds in production, but the HTTP server sets no write timeout, so cpp-httplib's 5 second default applies. A dead socket fails `sink.write` and the loop exits about 85 seconds before the cursor could be marked stalled.

`reap_stalled_clients_locked` is also only called from `evict_to_capacity_locked`, so it only runs when the buffer is already over capacity. As far as I can tell the only client that reaches the budget is one whose `pending` is always empty while notifications keep arriving, which is the `Last-Event-ID` case I raised separately.

If that is right, the reaping and the reaped count in the WARN are dead weight, and the watermark alone already fixes the reported bug.

src/ros2_medkit_gateway/src/http/handlers/sse_fault_handler.cpp
```diff
@@ -89,6 +92,82 @@ void SSEFaultHandler::request_shutdown() {
...
+      ++stats.coalesced;
+      continue;
+    }
+    event_queue_.pop_front();  // distinct fault codes only: this one is lost
+    ++stats.dropped;
+  }
+
+  if (stats.dropped > 0) {
+    for (const auto & cursor : clients_) {
+      if (!cursor->stalled && cursor->last_delivered < event_queue_.back().id) {
```
bburda: This compares against `event_queue_.back().id`, which is the event pushed a few lines earlier in the same locked section. No client can have received it yet, so every non-stalled cursor matches and the count is just the number of live clients.

With four clients and one genuinely behind, the WARN says four slow clients. Comparing against `event_queue_.front().id` would report the clients that are actually lagging. It does correctly skip stalled cursors.

src/ros2_medkit_gateway/src/http/handlers/sse_fault_handler.cpp
```diff
@@ -89,6 +92,82 @@ void SSEFaultHandler::request_shutdown() {
...
+  return watermark;
+}
+
+std::deque<SSEFaultHandler::QueuedEvent>::iterator SSEFaultHandler::find_superseded_locked() {
+  std::unordered_map<std::string, std::size_t> newest_index;
+  for (std::size_t i = 0; i < event_queue_.size(); ++i) {
+    newest_index[event_queue_[i].event.fault.fault_code] = i;
+  }
+  for (std::size_t i = 0; i < event_queue_.size(); ++i) {
+    // An auto-clear cascade lists its symptoms only here; those codes get no
```
bburda, line 119: The reasoning here does not hold for SSE clients: `format_sse_event` serialises only `event_type`, `fault`, `timestamp` and the `x-medkit` object, so `auto_cleared_codes` never reaches the wire in either case.

The exemption itself is fine and conservative, so this is only about the comment. Could you either reword it, or serialise the field so the justification becomes true?

bburda (commented):
A few things outside this diff that I think should move with it.

Five places still describe the old contract:

- `src/ros2_medkit_msgs/msg/Fault.msg` line 38 says `last_occurred` is the time the fault was "last reported (FAILED or PASSED event)". That is the wire contract for the field you changed, in the same package whose `FaultEvent.msg` this PR does edit.
- `src/ros2_medkit_msgs/README.md` line 25 repeats it, and line 71 still reads "EVENT_CLEARED | Fault transitions to CLEARED".
- `docs/api/messages.rst` line 34 repeats it too, and line 131 still says `fault_cleared` means cleared via the ClearFault service. This one is published to the docs site.
- The Postman collection has the same sentence for `fault_cleared` ("Fault was manually cleared").

One consumer in an untouched file needs the same status check: `src/ros2_medkit_gateway/src/trigger_fault_subscriber.cpp` lines 69 to 73 map `fault_cleared` to `ChangeType::DELETED` without reading `fault.status`, so an auto-heal now raises a fault trigger with delete semantics. On the shipped bringup config healing is on, so this is the common path rather than an edge case.

On coverage: neither fix is proven end to end. `test_sse.test.py` launches with `fault_manager=False`, so no integration test has ever seen a fault frame on `/faults/stream`. Worth noting that the graph provider failures you mention are a different handler (cyclic subscriptions, no replay buffer and no cursors), so they do not explain this gap. The heal fix has the better evidence, since its test drives a real node over a real service and topic, but nothing carries it through the gateway to an HTTP client.

Last one is a design call rather than a defect. `last_passed_ns` is storage only. It is not in `Fault.msg`, not in `fault_to_json` and not in any DTO, so a consumer that used `last_occurred` as "when did this fault last report anything" has lost that with no replacement on the wire. Do you want to expose `last_passed`, or record the loss in the release notes?

src/ros2_medkit_gateway/src/http/handlers/sse_fault_handler.cpp
mfaferek93: Fixed in de4ba176. parse_last_event_id is digits-only now ("-1" falls back to full replay) and make_stream_loop clamps the cursor seed to the newest issued id, so an overshooting header means caught-up, not blind. The watermark is a std::optional, no aliasable sentinel left. Covered by HugeLastEventIdIsClampedToNewestIssuedId and NegativeLastEventIdFallsBackToFullReplay.

src/ros2_medkit_gateway/src/http/handlers/sse_fault_handler.cpp
mfaferek93: Fixed in de4ba176, with one deliberate difference from the literal rule: only fault_updated entries coalesce silently, but when everything owed is a transition, a superseded transition is still evicted before a code's only buffered event - now counted and logged as a real drop instead of "coalesced". Strictly never evicting transitions would push the rare standing fault out again (the flap case this PR started from); this keeps that fault alive while making the history loss honest. The flap test now asserts dropped==51 / coalesced==0, and the reworked convergence test pins that an update storm cannot erase the confirm.

src/ros2_medkit_gateway/test/test_sse_fault_handler.cpp
mfaferek93: Added in de4ba176: OwedDistinctEventsLostUnderPressureAreCountedAndLogged registers a cursor, floods 150 distinct codes and asserts dropped==50 plus the first-loss WARN attributing 1 slow client. The end-to-end gap from your review body is closed in 1762beb2: test_sse_fault_stream.test.py runs the real fault manager, drives FAILED then two PASSED over ReportFault and asserts fault_confirmed and then fault_cleared with status HEALED arrive at an HTTP client on /faults/stream. Verified green in the CI image.

src/ros2_medkit_gateway/test/test_sse_fault_handler.cpp
mfaferek93: Fixed in de4ba176: the stream opens before the burst, and the burst is an update storm so coalescing actually runs (coalesced>0 and dropped==0 asserted). It also pins that the initial fault_confirmed survives the storm and the replay ends on the final cleared state. Renamed to CoalescedReplayKeepsTransitionsAndEndsOnCurrentState.

src/ros2_medkit_fault_manager/src/sqlite_fault_storage.cpp
mfaferek93: Done in 322f36eb with exactly that statement, plus ReopenRepairsLastOccurredInflatedByOldPassedBug: writes a row, inflates last_occurred_ns via raw SQL, reopens the store and asserts the repair. On the review-body design call: last_passed is on the wire as of ad61c811 (Fault.msg, both backends, fault JSON - absent when the fault never passed), so "when did this fault last report anything" stays answerable as max(last_occurred, last_passed).

src/ros2_medkit_gateway/src/http/handlers/sse_fault_handler.cpp
mfaferek93: Fixed in f4c20828 with the predicate form; the check is O(1) (back().id > last_event_id, ids are monotonic) and keepalive_due now derives from the false return.

src/ros2_medkit_gateway/README.md
mfaferek93: Both in adf93b19: the loss branch is documented with a refetch hint, and the Last-Event-ID example now says ids stay monotonic but not contiguous after coalescing, and that an above-newest id means caught-up. The stale contract text from your review body is in the same commit (msgs README, messages.rst, Postman; Fault.msg itself in ad61c811). On trigger_fault_subscriber: kept DELETED deliberately - ChangeType has no reader today (TriggerManager is the only notifier subscriber and evaluates change.value only), a heal leaves the default faults view exactly like a manual clear, and the payload status carries the CLEARED/HEALED distinction; adf93b19 pins that intent in a comment. Triggers firing on heal is the point of the fix: the fault's end is now observable.

src/ros2_medkit_gateway/src/http/handlers/sse_fault_handler.cpp
mfaferek93: Made explicit in 4143c969: the destructor now blocks until clients_ is empty (the unregister deleter notifies after erasing), so a closure outliving the handler deadlocks loudly instead of dereferencing a dead object. It sits in the destructor rather than request_shutdown because request_shutdown runs before the server stops, while providers are still legitimately alive; in the production order (RESTServer::stop joins the server before members are destroyed) the wait returns immediately.

src/ros2_medkit_gateway/include/ros2_medkit_gateway/http/handlers/sse_fault_handler.hpp
mfaferek93: Confirmed and removed in de4ba176. Nothing overrides the server write timeout, so cpp-httplib's 5s default fails a dead socket out of the loop long before the 90s budget, and with Last-Event-ID clamped the empty-pending-forever case is gone too. The watermark alone carries the fix; the reap machinery, its share of the WARN and its test are deleted.

src/ros2_medkit_gateway/src/http/handlers/sse_fault_handler.cpp
mfaferek93: Fixed in de4ba176. The count now keys on the highest id actually lost in the pass, which is exactly "clients owed at least one lost event"; front().id would still overcount a client that had everything up to the drop point but missed nothing.

src/ros2_medkit_gateway/src/http/handlers/sse_fault_handler.cpp
mfaferek93: Serialised in 9f4188dc (omitted when empty), so the exemption now protects data that reaches the wire. The gateway README documents the field and StreamCarriesAutoClearedCodes covers presence and absence.

src/ros2_medkit_msgs/msg/FaultEvent.msg
```diff
@@ -30,7 +30,9 @@ builtin_interfaces/Time timestamp
 # Emitted when a fault transitions from PREFAILED to CONFIRMED status
 # (e.g., after debounce counter reaches confirmation threshold).
 string EVENT_CONFIRMED = "fault_confirmed"
-# Emitted when a fault transitions to CLEARED status via ClearFault service.
+# Emitted when a fault ends: CLEARED via the ClearFault service, or HEALED when
+# PASSED events drive it past the healing threshold. Check fault.status to tell
+# the two apart.
```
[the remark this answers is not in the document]
mfaferek93: Done in adf93b19: the messages.rst FaultEvent table now describes both endings with the fault.status distinction; msgs README and the Postman collection updated to match.

bburda (approved):

PR #578, opened 2026-07-29
feat(graph_watchdog): orphan detector (GRAPH_ORPHAN)

### LINKED ISSUE #577: A topic-name typo leaves both sides healthy and silently disconnected
Part of #570.

## Summary

One node publishes `/scan_raw`, another subscribes `/scan_row`. Both nodes are up, both report healthy, and neither is talking to the other. Nothing in ROS 2 notices, because from the graph point of view these are simply two different topics that each happen to have one side.

The same shape appears from a wrong remap in a launch file, a renamed topic where only one side was updated, or a namespace that was applied to one node and not the other.

This is hard to see by hand. `ros2 topic list` shows both names and they look plausible. The operator usually finds it by noticing the robot behaves as if the data never arrives, then comparing topic names character by character.

## Proposed solution

Look for a topic that has publishers but no subscribers, or subscribers but no publishers, and check whether a topic with the complementary shape exists nearby: same namespace, same message type, and a name within a very small edit distance.

Two things it has to be careful about.

The fault should name both sides and say plainly that it is a candidate. The tool cannot tell a typo from a counterpart that simply has not started yet or has just exited, so it should say both possibilities rather than assert one and send someone renaming things.

It also needs a persistence requirement. During a staged bringup one side legitimately exists without the other for a while, and that looks exactly like a typo until the second node appears. Reporting on a single sweep would make this fire on every launch.

bburda:
## Summary

Second of the seven silent-fault detectors. Raises `GRAPH_ORPHAN` for a topic with publishers but no subscribers, or the reverse, when a topic with the complementary shape exists nearby: same namespace, same message type, name within a very small edit distance.

That is what a remap typo looks like from the graph. One node publishes `/scan_raw`, another subscribes `/scan_row`, both are healthy, and neither is talking to the other.

Two deliberate choices:

**The fault names both sides and calls it a candidate.** Nothing in the graph can tell a typo from a counterpart that has just exited or has not started yet, so the description says both and asks the operator to check before renaming.

**A pair must hold for more than `grace` sweeps, default 10.** During a staged bringup one side legitimately exists without the other, which is indistinguishable from a typo until the second node appears. Reporting on a single sweep would fire on every launch.

---

## Issue

- closes #577

Part of #570.

---

## Type

- [ ] Bug fix
- [x] New feature or tests
- [ ] Breaking change
- [ ] Documentation only

---

## Testing

21 tests green locally on Jazzy, up from 18. New: a policy unit test for the near-miss matching and the edit distance, an integration test driving real publishers and subscribers over real DDS, and an e2e against a real gateway asserting over `/faults`.

    colcon build --packages-select ros2_medkit_graph_watchdog
    colcon test --packages-select ros2_medkit_graph_watchdog && colcon test-result --verbose

---

## Checklist

### DIFF
--- a/src/ros2_medkit_plugins/ros2_medkit_graph_watchdog/CMakeLists.txt
+++ b/src/ros2_medkit_plugins/ros2_medkit_graph_watchdog/CMakeLists.txt
@@ -190,6 +190,9 @@ if(BUILD_TESTING)
   # LifecycleShutdownSuppressor: needs a REAL ReliabilityGate (departed_lifecycle_state_of()
   # reads live from ctx.gate) - same real-Node/ROS_DOMAIN_ID rationale as test_reliability_gate
   # above, unlike test_allowlist_suppressor (pure interface, no gate needed).
+  ament_add_gtest(test_orphan_policy test/test_orphan_policy.cpp)
+  target_include_directories(test_orphan_policy PRIVATE include)
+
   ament_add_gtest(test_qos_policy test/test_qos_policy.cpp)
   target_include_directories(test_qos_policy PRIVATE include)
   medkit_target_dependencies(test_qos_policy rmw)   # rmw is transitively found via rclcpp; no find_package needed
@@ -214,7 +217,20 @@ if(BUILD_TESTING)
     ros2_medkit_gateway rclcpp ros2_medkit_msgs lifecycle_msgs std_msgs)
   target_link_libraries(test_qos_mismatch_integration nlohmann_json::nlohmann_json)
   medkit_set_test_domain(test_qos_mismatch_integration)
-  set_tests_properties(test_qos_mismatch_integration PROPERTIES TIMEOUT 120 RESOURCE_LOCK graph_watchdog_integration_domain)
+
+  ament_add_gtest(test_orphan_integration
+    test/test_orphan_integration.cpp
+    src/detectors/orphan_detector.cpp
+    src/detector_registry.cpp
+    src/reliability_gate.cpp
+    src/lifecycle_watcher.cpp
+    ${LIFECYCLE_STATE_READER_SOURCES})
+  target_include_directories(test_orphan_integration PRIVATE include)
+  medkit_target_dependencies(test_orphan_integration
+    ros2_medkit_gateway rclcpp ros2_medkit_msgs lifecycle_msgs std_msgs sensor_msgs)
+  target_link_libraries(test_orphan_integration nlohmann_json::nlohmann_json)
+  medkit_set_test_domain(test_orphan_integration)
+  set_tests_properties(test_orphan_integration PROPERTIES TIMEOUT 120 RESOURCE_LOCK graph_watchdog_integration_domain)
 
   # === QoS-mismatch e2e (launch_testing) ===
   #
@@ -232,6 +248,13 @@ if(BUILD_TESTING)
     LABELS "integration;e2e"
     # Shares the e2e domain and its lock rather than consuming another slot.
     ENVIRONMENT "GATEWAY_TEST_PORT=19120;ROS_DOMAIN_ID=89;GRAPH_WATCHDOG_PLUGIN_PATH=${_WATCHDOG_E2E_SO}"
+    RESOURCE_LOCK graph_watchdog_e2e_domain)
+
+  add_launch_test(test/e2e/test_orphan_e2e.test.py TIMEOUT 180)
+  set_tests_properties(test_e2e_test_orphan_e2e.test.py PROPERTIES
+    LABELS "integration;e2e"
+    # Shares the e2e domain and its lock rather than consuming another slot.
+    ENVIRONMENT "GATEWAY_TEST_PORT=19130;ROS_DOMAIN_ID=89;GRAPH_WATCHDOG_PLUGIN_PATH=${_WATCHDOG_E2E_SO}"
     RESOURCE_LOCK graph_watchdog_e2e_domain)
 
   ros2_medkit_relax_vendor_warnings()
--- a/src/ros2_medkit_plugins/ros2_medkit_graph_watchdog/README.md
+++ b/src/ros2_medkit_plugins/ros2_medkit_graph_watchdog/README.md
@@ -7,7 +7,7 @@ Detectors read the graph and raise faults through a `ReportFault` service client
 gateway node; the faults surface via FaultManager on the gateway `/faults` API.
 
 This package carries the plugin skeleton, the central reliability gate that holds raises
-until the graph has quiesced, and the first detector, `qos_mismatch`. The remaining
+until the graph has quiesced, and two detectors, `qos_mismatch` and `orphan`. The remaining
 silent-fault classes land in follow-up changes, each against its own issue; their fault
 codes are already reserved in the frozen `GRAPH_*` namespace (see "Fault codes").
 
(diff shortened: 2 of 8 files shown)


src/ros2_medkit_plugins/ros2_medkit_graph_watchdog/include/ros2_medkit_graph_watchdog/orphan_policy.hpp
```diff
@@ -0,0 +1,137 @@
...
+    int best = max_edit_distance + 1;
+    const TopicEndpointCounts * match = nullptr;
+    for (const auto & u : topics) {
+      if (&u == &t || is_system_topic(u.name) || u.type != t.type) {
+        continue;
+      }
+      // Only pair near-misses in the SAME namespace, so a leaf typo (/scann vs /scan)
+      // matches but distinct per-robot topics (/robot1/scan vs /robot2/scan, also
+      // distance 1) do not - that is a fleet layout, not a typo.
+      if (parent_namespace(u.name) != parent_namespace(t.name)) {
```
mfaferek93: The same-namespace guard only helps when the index sits in the namespace segment. When it sits in the leaf, both topics share a parent namespace and the guard is a no-op. I compiled this header standalone and ran `find_orphans` at `max_edit_distance=1`:

- `/camera1/image` (pub-only) + `/camera2/image` (sub-only) -> 0 findings. Guard works.
- `/image1` (pub-only) + `/image2` (sub-only) -> 1 finding.
- `/cameras/image1` + `/cameras/image2` -> 1 finding.
- `/lidar_1` (pub-only) + `/lidar_2` (sub-only) -> 1 finding.

Two independently one-sided siblings on a multi-sensor rig (nothing recording `/lidar_1`, `/lidar_2`'s driver down) is an ordinary state, and it raises `GRAPH_ORPHAN` at `SEVERITY_ERROR` telling the operator the two names are a possible mismatch. `OneSidedNextToHealthyConnectedSiblingIsNotOrphan` covers only the case where the sibling is fully connected, so this shape is untested. The detector defaults to `mode: raise`, so it ships on.

Suggest skipping a pair whose names are equal once digit runs are collapsed (`/lidar_1` and `/lidar_2` both become `/lidar_#`): a difference confined to a numeric field is a sibling index, not a typo. `/scam` vs `/scan` and `/scan` vs `/scan_1` are unaffected.

src/ros2_medkit_plugins/ros2_medkit_graph_watchdog/src/detectors/orphan_detector.cpp
```diff
@@ -0,0 +1,170 @@
...
+      }
+      affected[key] = finding.reason;
+    }
+    // Drop streaks for pairs that stopped matching, so a pair has to hold CONSECUTIVE ticks
+    // and the map cannot grow without bound on a graph that churns topic names.
+    for (auto it = streak_.begin(); it != streak_.end();) {
+      it = seen_pairs.count(it->first) == 0 ? streak_.erase(it) : std::next(it);
+    }
+
+    aggregated_.emit(ctx, affected);
```
mfaferek93: `emit()` lists pairs lexicographically and `AggregatedFault::kMaxDescriptionChars` is 480. Each orphan reason is about 215 chars, so two entries fill the cap. I ran the real `describe_ordered` (extracted verbatim from `aggregated_fault.hpp`) over four findings from `find_orphans`: description length 480, `/aaa*` and `/bbb*` named, `/ccc*` cut mid-sentence, `/zzz*` absent entirely.

This is the defect `de21e84f` fixed one commit earlier in `qos_mismatch_detector.cpp` with `newly_affected` + `emit_ordered`, and it bites about 2.5x sooner here because each entry is roughly four times longer than a QoS entry. A newly detected pair sorting after the first two changes nothing an operator can see: the code is already raised and the text is already full.

Same shape of fix: collect keys where `streak_[key] == grace_ + 1` on this tick and call `emit_ordered` with them. Shortening the per-pair detail would buy headroom too.

src/ros2_medkit_plugins/ros2_medkit_graph_watchdog/README.md
```diff
@@ -43,6 +43,14 @@ read produces one startup warning naming the key and listing the ones that exist
 typo like `allow_list:` for `allowlist:` cannot silently do nothing. A typo in the detector
 id itself is warned about the same way, with the registered ids listed.
 
+### `orphan` keys
+
+| Key | Type | Default | Meaning |
+|-----|------|---------|---------|
+| `mode` | string | `raise` | As above. |
+| `grace` | int | `10` | Consecutive sweeps a candidate pair must hold before it is reported. A staged bringup or a restart makes one side missing for a short time, which looks exactly like a typo until the other side appears. |
+| `allowlist` | string[] | `[]` | Topic names never reported. Exact match only. |
```
mfaferek93: `allowlist` is not implemented for `orphan`. `OrphanDetector::known_keys()` is `{"max_edit_distance", "grace"}` and `configure()` never reads it, so `detectors.orphan.allowlist` suppresses nothing and instead produces `unknown config key 'allowlist'` on every boot, from the mechanism line 43 of this same file advertises as the thing that stops a config typo silently doing nothing. The second orphan table at line 118 lists a different key set with no allowlist, so the two tables contradict each other.

This matters more given the sibling-topic false positive on `/lidar_1` + `/lidar_2`: an allowlist is the only documented way to silence a finding short of turning the detector off, and it does not exist. Either implement it (skip listed names in the sweep, add the key to `known_keys()`) or drop the row.

src/ros2_medkit_plugins/ros2_medkit_graph_watchdog/CMakeLists.txt
```diff
@@ -190,6 +190,10 @@ if(BUILD_TESTING)
   # LifecycleShutdownSuppressor: needs a REAL ReliabilityGate (departed_lifecycle_state_of()
   # reads live from ctx.gate) - same real-Node/ROS_DOMAIN_ID rationale as test_reliability_gate
   # above, unlike test_allowlist_suppressor (pure interface, no gate needed).
+  ament_add_gtest(test_orphan_policy test/test_orphan_policy.cpp)
+  target_include_directories(test_orphan_policy PRIVATE include ${GATEWAY_SRC_INCLUDE_DIR})
+  target_link_libraries(test_orphan_policy nlohmann_json::nlohmann_json)
```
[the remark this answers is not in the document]
bburda: Removed. `test_orphan_policy` now only gets `target_include_directories(... PRIVATE include)`. The gateway include dir and the `nlohmann_json` link came from copying the qos test block, and the policy header needs neither.

src/ros2_medkit_plugins/ros2_medkit_graph_watchdog/CMakeLists.txt
```diff
@@ -214,6 +218,19 @@ if(BUILD_TESTING)
...
+    src/detector_registry.cpp
+    src/reliability_gate.cpp
+    src/lifecycle_watcher.cpp
+    ${LIFECYCLE_STATE_READER_SOURCES})
+  target_include_directories(test_orphan_integration PRIVATE include)
+  medkit_target_dependencies(test_orphan_integration
+    ros2_medkit_gateway rclcpp ros2_medkit_msgs lifecycle_msgs std_msgs sensor_msgs)
+  target_link_libraries(test_orphan_integration nlohmann_json::nlohmann_json)
+  medkit_set_test_domain(test_orphan_integration)
   set_tests_properties(test_qos_mismatch_integration PROPERTIES TIMEOUT 120 RESOURCE_LOCK graph_watchdog_integration_domain)
```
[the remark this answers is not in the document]
bburda: Fixed. `test_orphan_integration` now carries `TIMEOUT 120` and `RESOURCE_LOCK graph_watchdog_integration_domain`, same as the qos integration test. The block also still had a `set_tests_properties` call naming `test_qos_mismatch_integration`, so the qos test received its properties twice and the orphan test received none.

src/ros2_medkit_plugins/ros2_medkit_graph_watchdog/README.md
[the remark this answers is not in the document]
bburda: The table did not match the code. I implemented the key instead of dropping the row: `configure()` reads `allowlist`, it is listed in `known_keys()` so it no longer trips the unknown-key warning, and the sweep skips a pair when either side is listed and clears its streak. Matching is exact, never a prefix. Both orphan tables now carry the same keys. An integration test covers the skip, and the e2e passes the array through real gateway parameters.

src/ros2_medkit_plugins/ros2_medkit_graph_watchdog/include/ros2_medkit_graph_watchdog/orphan_policy.hpp
bburda: Reproduced all four cases. Implemented the collapse you suggested: `collapse_digit_runs()` replaces every run of digits with a single `#`, and a pair is skipped when the collapsed names are equal. `/lidar_1` and `/lidar_2` both become `/lidar_#`, so they never pair.

Collapsing runs rather than comparing character positions also covers `/lidar_9` against `/lidar_10`, where the index changes length. I wrote the per-character check first and it missed that case. `/scam` vs `/scan` and `/scan` vs `/scan_1` are unaffected, and `/scan1` vs `/scan` is still reported, because `/scan#` is not `/scan`.

The cost is that a typo which is itself a digit is no longer reported. That is written at the function and in the README, next to the allowlist as the way out.

Your point about the guard being a no-op in the leaf also made its only test vacuous: `/robot1/scan` vs `/robot2/scan` is now rejected by the new rule, so the same-namespace guard could be deleted with every test still green. Added a letter-difference case (`/left/scan` vs `/lift/scan`) that fails without the guard.

Separately, the namespace now has its own edit budget, `namespace_edit_distance`, default `0`. That keeps the exact-match behaviour, and raising it catches a misspelled namespace such as `/robott/scan` against `/robot/scan`, which was invisible before. It stays opt-in because a fleet named with letters would pair every robot with its neighbour.

Tests: your four cases, the length-changing index at `max_edit_distance` 2 where only this rule can spare it, the digit typo we knowingly drop, and a letter difference of the same shape that must still be reported.

src/ros2_medkit_plugins/ros2_medkit_graph_watchdog/src/detectors/orphan_detector.cpp
bburda: Fixed the same way. The sweep collects the keys that reach `streak_[key] == grace_ + 1` on this tick and passes them to `emit_ordered`, so a newly detected pair is named first instead of being cut off after the first two entries. I left the per-pair text as it is, since ordering is what decides whether a new pair is visible at all.

src/ros2_medkit_plugins/ros2_medkit_graph_watchdog/README.md
bburda: Implemented rather than dropped, for the reason you give. The sibling case is fixed at the source now, but a heuristic over names still needs a way out that is not turning the detector off. `configure()` reads `allowlist`, it is in `known_keys()`, and the sweep skips a listed topic and clears its streak so a silenced topic does not accumulate one. Both tables now list the same keys, and the second one also gained `namespace_edit_distance`. The e2e runs an allowlisted pair next to a reported one, which is also the only place the array is proven to survive the parameter path into `configure()`.

mfaferek93 (approved):

PR #580, opened 2026-07-29
feat(graph_watchdog): param_drift detector (GRAPH_PARAM_DRIFT)

### LINKED ISSUE #579: A parameter changes at runtime and nothing reports it
Part of #570.

## Summary

Someone sets a gain, a frame name or `use_sim_time` from the command line during a test and never reverts it. A launch file overrides a value that a later config file was supposed to own. The node reads the parameter once, keeps running, and logs nothing. The robot behaves slightly differently and everything reports healthy.

`ros2 param get` shows the current value, but nothing shows that it is not the value the node started with, or not the value it was supposed to have. To find it by hand you already need to suspect the right parameter.

## Proposed solution

Read node parameters and report a value that no longer matches its reference, with two ways to define that reference:

- self-capture: remember the value the parameter had when the node armed, and report later changes. This needs no configuration and covers the "someone changed it during the run" case.
- a pinned expected value in config. This covers the case self-capture cannot see by construction, a value that was already wrong on the first tick.

Three things it has to be careful about.

Reading parameters is a service call into another node. Every detector so far reads the local graph cache and is free to be exhaustive every tick. This one is not, so it needs a budget on calls per tick, and it must not run on the gateway executor, or one node that stops answering stalls the gateway itself.

A changed value is not automatically a broken robot. The fault should say what moved and leave the judgement to whoever reads it, so WARN rather than ERROR.

A node that is missing from one sweep must not immediately lose its captured baseline. Otherwise a discovery gap turns into a fresh capture and a real drift disappears behind it.

## Also in scope

Nested per-detector configuration has no end-to-end proof in this package today. Every test calls `configure()` with hand-built JSON, so nothing covers the path from YAML through the gateway into a detector. This detector's pinned-value rule is the natural place to prove it.

bburda:
## Summary

Adds the `param_drift` detector. It raises `GRAPH_PARAM_DRIFT` when a node parameter no longer matches its reference, either the value self-captured when the node armed or a value pinned with `expect`.

Unlike the two detectors already here, this one makes service calls into other nodes instead of reading the local graph cache. The sweep is therefore bounded by `max_reads_per_tick` and walks the graph round-robin, which bounds the load on watched nodes and pays with coverage latency in both directions: a drift is invisible until its node's turn comes, and so is a repair. The reads run on a thread the plugin owns, never on the gateway executor.

It also brings the config-plumbing e2e, which is the first proof in this package that nested per-detector config reaches a detector at all.

The README referred twice to a "Closing the loop" section on healing configuration that was never carried over. It belongs with the first detector whose clear story depends on it, so it is restored here.

---

## Issue

- closes #579

---

## Type

- [ ] Bug fix
- [x] New feature or tests
- [ ] Breaking change
- [ ] Documentation only

---

## Testing

26/26 green locally on Jazzy.

- `test_param_drift_policy`: the pure comparison and config rules, glob matching, the `baseline` and `expect` combinations.
- `test_param_drift_integration`: real nodes with real parameters over the real parameter service against a fake `ReportFault` service. Covers the round-robin budget, the grace-based prune when a node vanishes, the clear path, and a node that never answers, which must not hang the tick.
- `test_config_plumbing_e2e`: three real gateway launches. One proves the nested `expect` reader, one `mode: "off"` as a string, one the bare `mode: off` that the ROS parser types as a YAML 1.1 boolean, which is the form operators actually write.

---

## Checklist

### DIFF
--- a/src/ros2_medkit_plugins/ros2_medkit_graph_watchdog/CMakeLists.txt
+++ b/src/ros2_medkit_plugins/ros2_medkit_graph_watchdog/CMakeLists.txt
@@ -28,6 +28,9 @@ find_package(ament_cmake REQUIRED)
 find_package(ros2_medkit_gateway REQUIRED)
 find_package(ros2_medkit_msgs REQUIRED)
 find_package(rclcpp REQUIRED)
+# The reused gateway parameter transport includes rcl_interfaces message headers directly;
+# it built only through rclcpp's transitive exports.
+find_package(rcl_interfaces REQUIRED)
 find_package(nlohmann_json REQUIRED)
 find_package(lifecycle_msgs REQUIRED)
 find_package(tf2_msgs REQUIRED)
@@ -82,7 +85,7 @@ add_library(ros2_medkit_graph_watchdog MODULE
   ${DETECTOR_SRCS})
 target_include_directories(ros2_medkit_graph_watchdog PRIVATE include ${GATEWAY_SRC_INCLUDE_DIR})
 medkit_target_dependencies(ros2_medkit_graph_watchdog
-  ros2_medkit_gateway rclcpp ros2_medkit_msgs lifecycle_msgs tf2_msgs geometry_msgs)
+  ros2_medkit_gateway rclcpp rcl_interfaces ros2_medkit_msgs lifecycle_msgs tf2_msgs geometry_msgs)
 target_link_libraries(ros2_medkit_graph_watchdog nlohmann_json::nlohmann_json)
 install(TARGETS ros2_medkit_graph_watchdog LIBRARY DESTINATION lib/${PROJECT_NAME})
 
@@ -121,6 +124,8 @@ if(BUILD_TESTING)
 
   find_package(ament_cmake_gtest REQUIRED)
   find_package(rosgraph_msgs REQUIRED)
+  # test_param_drift_integration.cpp includes rcutils/logging.h directly.
+  find_package(rcutils REQUIRED)
 
   ament_add_gtest(test_warmup_tracker test/test_warmup_tracker.cpp)
   target_include_directories(test_warmup_tracker PRIVATE include ${GATEWAY_SRC_INCLUDE_DIR})
@@ -187,11 +192,13 @@ if(BUILD_TESTING)
   target_link_libraries(test_reliability_gate nlohmann_json::nlohmann_json)
   medkit_set_test_domain(test_reliability_gate)
 
-  # LifecycleShutdownSuppressor: needs a REAL ReliabilityGate (departed_lifecycle_state_of()
-  # reads live from ctx.gate) - same real-Node/ROS_DOMAIN_ID rationale as test_reliability_gate
-  # above, unlike test_allowlist_suppressor (pure interface, no gate needed).
+  # Pure matching logic over hand-built inputs: no rclcpp::Node, so no domain is needed.
   ament_add_gtest(test_orphan_policy test/test_orphan_policy.cpp)
   target_include_directories(test_orphan_policy PRIVATE include)
+
+  ament_add_gtest(test_param_drift_policy test/test_param_drift_policy.cpp)
+  target_include_directories(test_param_drift_policy PRIVATE include)
+  target_link_libraries(test_param_drift_policy nlohmann_json::nlohmann_json)
 
   ament_add_gtest(test_qos_policy test/test_qos_policy.cpp)
   target_include_directories(test_qos_policy PRIVATE include)
@@ -202,9 +209,10 @@ if(BUILD_TESTING)
   medkit_target_dependencies(test_aggregated_fault ros2_medkit_gateway rclcpp ros2_medkit_msgs)
   target_link_libraries(test_aggregated_fault nlohmann_json::nlohmann_json)
 
-  # Pure suppressor-chain interface: no raise_fault/clear_fault calls (no ReliabilityGate
-  # sources needed), no live rclcpp::Node, so no ROS_DOMAIN_ID is required - same as
-  # test_qos_policy above.
+  # Real DDS endpoints read through the live graph API, so each of these takes a domain of its
+  # own via medkit_set_test_domain. The two that ALSO register a fake /fault_manager/report_fault
+  # service additionally share a RESOURCE_LOCK, because they spin real nodes for many seconds and
+  # running them concurrently on this box starves the discovery each one waits on.
   ament_add_gtest(test_qos_mismatch_integration
     test/test_qos_mismatch_integration.cpp
     src/detectors/qos_mismatch_detector.cpp
@@ -232,6 +240,27 @@ if(BUILD_TESTING)
   medkit_set_test_domain(test_orphan_integration)
   set_tests_properties(test_orphan_integration PROPERTIES TIMEOUT 120 RESOURCE_LOCK graph_watchdog_integration_domain)
 
+  ament_add_gtest(test_param_drift_integration
+    test/test_param_drift_integration.cpp
+    src/detectors/param_drift_detector.cpp
+    src/detector_registry.cpp
+    src/reliability_gate.cpp
+    src/lifecycle_watcher.cpp
+    ${LIFECYCLE_STATE_READER_SOURCES}
+    ${PARAMETER_TRANSPORT_SOURCES})
+  target_include_directories(test_param_drift_integration PRIVATE include ${GATEWAY_SRC_INCLUDE_DIR})
+  medkit_target_dependencies(test_param_drift_integration
+    ros2_medkit_gateway rclcpp rcl_interfaces rcutils ros2_medkit_msgs lifecycle_msgs)
+  target_link_libraries(test_param_drift_integration nlohmann_json::nlohmann_json)
+  medkit_set_test_domain(test_param_drift_integration)
+  # 300 s, not 120: several cases here are paced off the detector's own attempt and tick horizons
+  # (sixty failed reads, sixty unarmed ticks, a read that answers past the per-read bound) and each
+  # of those costs real seconds that no budget knob can compress away. A timeout the suite can
+  # outgrow kills the process before gtest prints which case failed, which is the one thing a
+  # timeout must never do to a diagnostic suite.
+  set_tests_properties(test_param_drift_integration PROPERTIES
+    TIMEOUT 300 RESOURCE_LOCK graph_watchdog_integration_domain)
+
   # === QoS-mismatch e2e (launch_testing) ===
   #
   # Acceptance gate: proves GRAPH_QOS_MISMATCH raises and clears through the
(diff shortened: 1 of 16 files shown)


src/ros2_medkit_plugins/ros2_medkit_graph_watchdog/src/detectors/param_drift_detector.cpp
```diff
@@ -0,0 +1,413 @@
...
+    for (const auto & id : present) {
+      misses_.emplace(id, 0);
+    }
+  }
+
+  /// Level-triggered aggregation (D2/D3): one graph-level fault carrying every app's drift.
+  void emit_aggregated(DetectorContext & ctx) {
+    std::string source = kGraphSourceId;
+    if (ctx.snapshot && !ctx.snapshot->components.empty() && !ctx.snapshot->components.front().id.empty()) {
+      source = ctx.snapshot->components.front().id;
```
mfaferek93: In a real gateway the snapshot always carries the host Component, so this raises GRAPH_PARAM_DRIFT under the host component id, not `graph_watchdog`. HostInfoProvider never sets `external`, and a non-external component never claims its bare id in a scope set (fault_scope.cpp, the #516 guardrail), so the fault matches no entity scope: it shows in flat `/faults` but `/apps/graph_watchdog/faults` and `/components/<host>/faults` both drop it - the exact reachable-from-no-endpoint scoping the aggregated_fault.hpp comment documents. It also contradicts this PR's README ("Every GRAPH_* fault is raised with `source_id: graph_watchdog`") and diverges from qos_mismatch/orphan, which go through `graph_source_id()`. On a box with PLC plugins `components.front()` can even be an external PLC component, and the drift lands on the PLC's fault page. No test sees this: every C++ test runs with an empty components list (the fallback), and the e2e polls flat `/faults`. Use `graph_source_id(ctx)` like the other two detectors.

src/ros2_medkit_plugins/ros2_medkit_graph_watchdog/src/detectors/param_drift_detector.cpp
```diff
@@ -0,0 +1,413 @@
...
+
+namespace {
+// Fault severity scale: SEVERITY_WARN=1, SEVERITY_ERROR=2 (ros2_medkit_msgs/msg/Fault.msg).
+// The detector reports the drift and does not grade its impact: whether a changed parameter
+// matters is a question about the robot, not about the graph, so triage belongs to whoever
+// reads the fault. WARN keeps it visible without claiming the run is broken.
+constexpr std::uint8_t kParamDriftSeverity = ros2_medkit_msgs::msg::Fault::SEVERITY_WARN;
+constexpr double kParamServiceTimeoutSec = 0.5;  // per-read timeout, same as the lifecycle GetState read
+constexpr double kNegativeCacheTtlSec = 10.0;    // unreachable-node backoff
+constexpr int kForgetGrace = 2;                  // missed-sweep grace before forgetting, mirrors WarmupTracker
```
mfaferek93, line 55: A restart faster than this grace keeps the stale baseline forever. A node absent for <= kForgetGrace sweeps (~2 s at the default 1 s tick, or never observed absent at all) is not forgotten, and WarmupTracker's forget grace is the same 2, so it comes back still armed and gets evaluated against the pre-restart baseline. The standard flow - edit YAML, respawn the node (launch respawn defaults to 0 s delay) - then raises GRAPH_PARAM_DRIFT permanently on a healthy robot, and the README's recovery path ("restart the node, and the detector captures the new value as a fresh baseline") silently fails; the only exits are a deliberately slow restart or a gateway restart. Consider dropping the baseline on any observed absence - a churn-triggered re-capture loses at most one drift window, while a stale baseline latches a false positive with no way out - or document that only an outage longer than the prune grace re-baselines.

src/ros2_medkit_plugins/ros2_medkit_graph_watchdog/src/detectors/param_drift_detector.cpp
```diff
@@ -0,0 +1,413 @@
...
+    return true;
+  }
+  for (const auto & name : io.expect_names) {
+    {
+      std::lock_guard<std::mutex> lk(io.mutex);
+      if (io.stop) {
+        return false;  // shutting down mid-read: incomplete view -> skip (no spurious clear)
+      }
+    }
+    const auto result = io.transport->get_parameter(fqn, name);
```
mfaferek93, line 314: `max_reads_per_tick` paces node visits, but the docs promise service calls. Each pinned name here costs the watched node three service round-trips (list + get + describe inside `get_parameter`), issued back to back inside one budget slot, so E pins fire ~3E calls at a single node per visit; baseline mode is 2 RPCs per visit and 4-5 on first contact (`cache_default_values` runs its own list + get before `list_parameters` repeats them). The README table says "Parameter-service calls this detector may spend per tick" and reader_loop's comment says "max_reads_per_tick transport calls per tick period" - an operator lowering the knob to protect a loaded node still gets a multiple of it. Keep the per-node read atomic (ExpectBudgetTruncationDoesNotFlap pins that), but charge the slot for the calls actually made, e.g. scale the pacing wait by the call count, or add a batch get-by-names to the transport so an expect visit is one round trip.

src/ros2_medkit_plugins/ros2_medkit_graph_watchdog/src/detectors/param_drift_detector.cpp
bburda: Fixed, it uses `graph_source_id(ctx)` now like the other two.

The empty components list in every test is what hid it. `SourceIdIgnoresTheHostComponent` now seeds a host Component and asserts the fault still comes under `graph_watchdog`. The design doc described the old fallback too, that is corrected.

src/ros2_medkit_plugins/ros2_medkit_graph_watchdog/src/detectors/param_drift_detector.cpp
bburda: I tried dropping the baseline on any absence and reverted it. On the first missed sweep, normal discovery churn re-captures a node on an already drifted value, so a confirmed fault heals while the parameter is still wrong and can never fire again.

It drops on the third missed sweep now, and the README says what that means in practice: a respawn with no delay is often never seen as an absence, so it does not re-baseline. Use `respawn_delay`, or stop and start. `ChurnMustNotLaunderARaisedDrift` and `ARestartLengthAbsenceReBaselinesInsteadOfReportingDrift` pin both sides.

src/ros2_medkit_plugins/ros2_medkit_graph_watchdog/src/detectors/param_drift_detector.cpp
bburda: Agreed, and your numbers were right. It is charged in round trips now: 2 for a baseline read, 3 for a declared pin, 2 for an undeclared one.

A new e2e measures it against a node that counts inbound requests: baseline 4 cold and 2 warm, declared pin 3, undeclared pin 1. Your "4-5 on first contact" is 4.

The undeclared pin is charged 2 while it costs 1, because a second path returns NOT_FOUND after list + get and the error code does not separate them. The README table splits cost from charge.

Still no pacing inside a visit: 4 pins at budget 8 means 12 round trips in one tick period, then idle. It is an average rate, not a per-tick ceiling, and the README says so now. I did not add the batch get-by-names, that is gateway transport.

src/ros2_medkit_plugins/ros2_medkit_graph_watchdog/test/e2e/test_config_plumbing_e2e.test.py
```diff
@@ -0,0 +1,192 @@
...
+        demo_nodes=['temp_sensor'],
+        port=PORT,
+    )
+
+
+class TestConfigPlumbingPositive(unittest.TestCase):
+    """Positive control: the nested `expect` config reaches param_drift."""
+
+    @unittest.skipUnless(SCENARIO == 'positive', 'runs only under the "positive" scenario launch')
+    def test_expect_rule_raises_from_nested_config(self):
```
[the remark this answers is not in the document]
bburda: Fixed.

src/ros2_medkit_plugins/ros2_medkit_graph_watchdog/CMakeLists.txt
```diff
@@ -232,6 +240,27 @@ if(BUILD_TESTING)
...
+    ros2_medkit_gateway rclcpp rcl_interfaces rcutils ros2_medkit_msgs lifecycle_msgs)
+  target_link_libraries(test_param_drift_integration nlohmann_json::nlohmann_json)
+  medkit_set_test_domain(test_param_drift_integration)
+  # 300 s, not 120: several cases here are paced off the detector's own attempt and tick horizons
+  # (sixty failed reads, sixty unarmed ticks, a read that answers past the per-read bound) and each
+  # of those costs real seconds that no budget knob can compress away. A timeout the suite can
+  # outgrow kills the process before gtest prints which case failed, which is the one thing a
+  # timeout must never do to a diagnostic suite.
+  set_tests_properties(test_param_drift_integration PROPERTIES
+    TIMEOUT 300 RESOURCE_LOCK graph_watchdog_integration_domain)
```
mfaferek93, line 262: The sanitizer jobs never see this 300. quality.yml extends ctest budgets by rewriting only the literals `TIMEOUT "60"` -> 180 and `TIMEOUT "120"` -> 360, then runs `ctest -LE linter` over every package - so under ASan/TSan this suite, the largest gtest in the plugin tree with pacing that already costs well over 100 s natively, keeps 300 s while the 14x smaller orphan suite gets 360. And a ctest kill eats gtest's output, which is exactly what your comment above says this value exists to prevent.

Same gap for the new e2e budgets (240/300) - neither matches the sed. Either register these at sed-matched values or teach quality.yml the new literals in this PR, so the first TSan flake here doesn't come back as a corpse with no test name.

src/ros2_medkit_plugins/ros2_medkit_graph_watchdog/test/e2e/test_param_drift_e2e.test.py
```diff
@@ -0,0 +1,148 @@
...
+        cls._stop.set()
+        cls._spin.join(timeout=5.0)
+        cls._node.destroy_node()
+        rclpy.shutdown()
+
+    def test_01_runtime_change_raises_and_names_the_parameter(self):
+        # Give the detector time to capture the baseline BEFORE changing anything. Capturing must be
+        # silent, so a fault appearing here at all would mean the detector reports its own first
+        # read as drift.
+        self.assertIsNone(
```
mfaferek93, line 154: This absence window is ungated - and harness.py's own `wait_until_watchdog_armed` docstring says any absence assertion must gate on it first (only test_config_plumbing imports it). Two ways this bites on a slow runner:

1. `poll_faults` swallows connection errors, so if the stack is still coming up, 15 s of nothing "proves" silent capture vacuously.
2. If bringup crosses 15 s (sanitizer launch runs, loaded shared runners), the `set_parameters` below lands before the detector's first read - it silently baselines 0.2 as truth, no fault can ever raise, and the test dies at the 60 s poll with a message pointing at the wrong cause.

The C++ suite gates every drift scenario on `wait_for_baseline_evidence`; this file needs the same gate before the absence check and the flip.

src/ros2_medkit_plugins/ros2_medkit_graph_watchdog/src/detectors/param_drift_detector.cpp
```diff
@@ -0,0 +1,413 @@
...
+
+// Reading a node's parameters is a BLOCKING service round-trip (up to kParamServiceTimeoutSec
+// each, and observed to hang PAST that on a real Nav2 node stuck mid-transition, since the
+// underlying DDS recv does not always honour the spin timeout). Doing it inline on the tick
+// thread stalls the WHOLE detection loop - node_death, tf_stale, everything - the moment ONE
+// node's parameter service is slow. So the reads run on a dedicated background thread and the
+// tick thread only ever reads the cache. State shared between the two threads lives here,
+// behind a shared_ptr so a reader stuck in a hung read keeps it alive (the detector detaches
+// rather than joins on teardown - a hung blocking read cannot be interrupted, so joining it
+// would hang the gateway shutdown).
```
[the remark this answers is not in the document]
bburda: You were right, the comment was wrong. It now says the destructor joins the reader thread, and that the transport is shut down first so a blocking read returns instead of holding the join.

src/ros2_medkit_plugins/ros2_medkit_graph_watchdog/src/detectors/param_drift_detector.cpp
```diff
@@ -0,0 +1,413 @@
...
+        return io->stop || !io->targets.empty();
+      });
+      if (io->stop) {
+        break;
+      }
+      target = io->targets[idx % io->targets.size()];
+      ++idx;
+      const int reads = io->max_reads_per_tick > 0 ? io->max_reads_per_tick : 1;
+      min_gap = std::chrono::milliseconds(io->tick_interval_ms / reads);
+    }
```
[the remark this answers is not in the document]
bburda: Fixed. `read_gap` works in microseconds now, rounds up and floors at one, so the gap cannot reach zero. `ReadGapNeverReachesZeroEvenForAnAbsurdBudget` and `ReadGapNeverLetsTheAchievableRateExceedTheBudget` cover it, including the sub-millisecond slot.

src/ros2_medkit_plugins/ros2_medkit_graph_watchdog/CMakeLists.txt
bburda: You are right, and it is wider than the three values here. On the current tree the literal pair misses 62 of 361 `TIMEOUT` declarations. This suite fell out of its reach when I raised it from 120 to 300, so I made the problem.

The rewrite is generic now: `perl` scales any `TIMEOUT "<n>"` by three, in both the ASan and the TSan step. 60 and 120 still become 180 and 360, so no other package changes. Checked against a real generated `CTestTestfile.cmake` rather than by reading the regex: 30/60/90/120/180/200/240/300/420 all scale, none missed.

src/ros2_medkit_plugins/ros2_medkit_graph_watchdog/test/e2e/test_param_drift_e2e.test.py
bburda: Gated on `wait_until_watchdog_armed` now. It took an `app_id` argument so it waits for that entity to be armed, not just for the plugin to answer.

I checked it the way you implied and it was worse than a slow runner. With `GRAPH_WATCHDOG_PLUGIN_PATH` pointed at a nonexistent `.so` the whole file passed: the 15 s absence, the heal test, and the exit codes, against a gateway logging `Loaded 0 plugin(s)`. Now it fails naming the cause.

`test_02` had the same hole, since `poll_cleared` is also an absence, so it starts with a presence check. Its budget went 240 to 300 because the internal deadlines now sum to 225 s.

mfaferek93 (approved):

PR #587, opened 2026-08-04
Detect a node that must be active but is stuck in another lifecycle state

### LINKED ISSUE #585: A node that must be active is stuck in another lifecycle state
Part of #570.

## Summary

A lifecycle node comes up and never reaches `active`. The bringup order changed, a
configure callback returned FAILURE, or nobody activated the node after a manual
transition. The process runs, the node is in the graph, and nothing reports a problem.

Other detectors cannot see this. The node publishes nothing, but a node that is not active
is not supposed to publish, so silence is correct here and proves nothing. To find it by
hand the operator has to suspect the right node and run `ros2 lifecycle get` on it.

## Proposed solution

Let the operator name the nodes that must be active, and raise `GRAPH_NODE_INACTIVE` when
one of them is present but stays in another state for longer than a grace period.

The detector watches nothing until it is configured. A node in `unconfigured` is only a
fault if somebody expected it to be active, and the plugin cannot know that by itself.

Three things it has to be careful about.

A node that is absent is not this detector's job. That is the presence class
(`GRAPH_NODE_DISAPPEARED`), and reporting it here too would report one failure twice under
two codes.

Clearing has to be as strict as raising. A clear says every required node is healthy. If
the detector has not read a state yet, it knows nothing and must say nothing. This matters
most right after a gateway restart, when a fault is already stored and no state has been
read yet: a clear at that moment would heal a fault that is still real, and the fault
manager would count it toward healing.

Node identity can move. An app id can stay the same while it points at a different node.
Anything cached per node has to notice that, or one node's lifecycle state gets enforced
against another node.

## Also in scope

The shared lifecycle state cache keys entries by app id and never re-checks whether the id
still points at the same node. A node that is replaced under the same id keeps the old
node's state and the old event subscription. This detector is the first one that acts on
that state directly, so the cache needs to treat a moved binding as a new node: drop the
old label, subscribe again, and read the new one.

bburda:
## Summary

Adds the `lifecycle_expectation` detector to the graph watchdog plugin. The operator lists the
nodes that must be active, and the detector reports the ones that are not. With no `require_active`
entries it does nothing at all, so it cannot false-positive on a graph nobody configured.

Config keys under `detectors.lifecycle_expectation`: `require_active` (list, empty by default),
`grace` (default 5, accepted 0..300), `tracked_node_cap` (default 512, accepted 1..16384), plus the
plugin-wide `prune_grace` and `mode`. An entry matches by app id, by full FQN, or by bare node name,
so one entry can cover several namespaces. Unknown keys and out-of-range values produce a startup
warning instead of being dropped in silence.

`grace` is capped at 300 rather than at the int range because it also decides how long a node that
left the graph while not-active sits unsettled, and `GRAPH_NODE_INACTIVE`'s clear is withheld for
every node while it does. At an unbounded maximum the code could neither raise nor heal for weeks.

Three fault codes, because "this node is not active" and "I cannot tell whether this node is
active" are different operator problems and the fault store keys a record by fault code alone:

| Code | Severity | Meaning |
|---|---|---|
| `GRAPH_NODE_INACTIVE` | ERROR | measured, and the state is not `active` |
| `GRAPH_NODE_UNREADABLE` | WARN | managed, but the lifecycle state cannot be read at all |
| `GRAPH_NODE_NOT_MANAGED` | WARN | the node named has no lifecycle to read |

A node is content of at most one of the three at a time. Sharing one record would mean a single
unreadable node keeps it raised forever while a healed node never receives a debounce step toward
healing.

---

## One state machine, because counters cancel each other

Per node, per tick, the tracker classifies exactly one observed state - `ACTIVE`, `INACTIVE`,
`UNREADABLE`, `NOT_MANAGED`, or `ABSENT` - and keeps exactly two clocks:

- **violation streak** - advances on `INACTIVE`, resets ONLY on `ACTIVE`, reported past `grace`
- **unmeasured clock** - advances on `UNREADABLE` or `NOT_MANAGED`, resets ONLY on a real
  measurement. It is deliberately blind to which of the two causes it saw

That blindness is the point. With one counter per cause, a node flapping between two causes resets
each counter with the other cause and stays invisible to both. A cause-blind clock keeps advancing
instead, which closes every such pair at once, including pairs nobody has enumerated.

When the unmeasured clock matures, its code takes the node and the violation streak is released, so
`GRAPH_NODE_INACTIVE` can heal for it. The node returning to `INACTIVE` re-earns grace.

---

## Absence continues a violation; it does not erase one

Any rule that erases accumulated evidence is a hole for a node that periodically triggers that rule.
A node in a restart loop - start, crash, respawn delay, start - triggers absence by construction,
and it is the case this detector most exists to catch. So:

| Last real observation | What sustained absence does |
|---|---|
| `ACTIVE` | nothing. A healthy departure starts no violation |
| `INACTIVE` | keeps advancing the violation streak |
| `UNREADABLE` / `NOT_MANAGED` | keeps advancing the unmeasured clock |

Up to `absence_grace` (a fixed 3 ticks) everything is held unchanged, so an ordinary blink costs
nothing. Past it the clocks resume rather than restart.

Two consequences worth stating plainly for anyone consuming these faults. **A departure never heals
a fault, within a gateway lifetime**: a node that leaves while violating stays reported until it
returns and reads `active`, or until the operator changes the configuration. A gateway restart
re-baselines, because the restarted detector has no measurements and an entry that matches nothing
cannot be told apart from a misspelt one; that boundary is pinned by a scenario rather than left to
be discovered. **A departure never starts one**: a node measured `active` that shuts down raises
nothing. Reporting a healthy departure is what `GRAPH_NODE_DISAPPEARED` owns; it has a fault code
but no detector yet, and #570 sequences that work behind the suppression framework it needs.

Because "healthy departure" decides whether a fault is raised at all, the classification does not
turn on a single tick. A present, healthy, managed node reads as unmanaged for one sweep whenever
its `get_state` path is missing from that sweep, and without settling, one such tick before a clean
shutdown would leave a permanent fault about a node that was fine.

Bookkeeping is bounded by evidence rather than by age. `prune_grace` reclaims only IDLE entries -
both clocks at zero and no matured ownership, so nothing to lose - and an entry carrying evidence is
never reclaimed by age however long the node stays gone. `tracked_node_cap` bounds the map instead,
and the order at the cap matters: idle entries are reclaimed first, then entries for departed nodes
are collapsed into a count that still keeps the code raised, and only if every tracked node is
present and carrying evidence is a newcomer refused. **A present node therefore always wins a slot.**
Without that order, a fleet whose robots come and go under their own namespaces fills the map with
the dead, and a genuinely broken present node is refused, never checked, and silently reported
healthy - the exact failure this detector exists to prevent.

Refusing to track a required node also withholds the `GRAPH_NODE_INACTIVE` clear: a detector that
declined to check a node cannot assert that every node is healthy. Saturation is reported on
`GET /x-medkit-watchdog` under `detectors.lifecycle_expectation`, not only in the log, and it is
re-armed when it ends so a later real saturation is not silenced by an earlier one.

---

## A clear has to prove as much as a raise

The aggregated fault is level triggered, so an empty result would normally send a clear every tick.
That is wrong when the detector has not measured anything. `GRAPH_NODE_INACTIVE` now sends nothing
while an entry has not matched a node yet, or a matched node's state has never been read, or a
matched node reads not-active but has not passed `grace`. A raise is never withheld, and each hold
is bounded, so a typo in `require_active` cannot block healing forever.

Without this a gateway restart heals a fault that is still real: the stored fault is `CONFIRMED`,
the restarted detector has empty state, and the clears it sends before the first read are enough to
reach `HEALED`. There is an e2e for exactly this - it kills the gateway by PID, waits for the port
to go down, waits for the plugin to arm again, and fails if a `PASSED` arrives while the node is
still inactive.

---

## A node can be replaced under the same app id

The shared lifecycle cache keys its tracked map by `App::id`. An id can stay the same across a graph
sweep and point at a different node, and the cache then kept the old node's state and its old
transition subscription. An entry is now identified by its binding instead: the pair (fqn,
`GetState` path), both recorded when the node is first seen. `update()` compares that pair every
tick. When it changes, the old binding is written down as a departure under its own fqn, the entry
is dropped, and the id is seeded again like any new node. For this detector that means the old
node's label is never applied to the node the id points at now.

The `orphan` and `qos_mismatch` detectors are untouched apart from one comment fix each.

A separate commit clears 20 clang-tidy findings in this package that no gate was reporting. They
have nothing to do with the detector. Why the package has no clang-tidy target, and what it would
take to add one, is written up in #605.

---

## Issue

- closes #585

---

## Type

- [ ] Bug fix
- [x] New feature or tests
- [ ] Breaking change
- [ ] Documentation only

---

## Testing

Full package suite green on Jazzy locally: **611 tests, 0 failures, 44 skipped** (the skips are
cppcheck). 66 tracker unit cases, 91 integration cases, 15 in the lifecycle cache suite, and 12 e2e
scenarios for this detector across 84 cases.

Everything above is driven end to end rather than argued. Two fixture nodes make the awkward states
reachable on a real stack: one advertises the lifecycle services and holds its `GetState` responses
open without ever answering, the other drops its own lifecycle services on command and then exits.
Between them the e2e covers a sustained read failure and its clear by read; a node that leaves and
whose fault survives the departure; a crash loop of twelve real SIGTERM/respawn cycles, each proven
absent past the grace via `GET /apps`; a cap saturated by departed nodes, where a present broken
node must still be reported and no clear may be emitted while one is refused; a healthy node whose
services blink for one sweep before a clean shutdown, which must raise nothing while a genuinely
unmanaged departure still does; a configuration at the old unbounded `grace`, which must be refused;
and a gateway restart with a departed node's fault outstanding, which pins the re-baseline boundary.

The tests sweep the config space at both ends of every documented range (`grace` at 0 and 300 plus
one past each, `tracked_node_cap` at 1 and 16384 plus one past each, `prune_grace` at 0, 1, 3600 and
3601, an empty and a non-array `require_active`, unknown and misspelt keys), scale past the
description cap and past the configured tracked-node cap, and change during a run: nodes appearing
late, vanishing after being reported, returning under a new binding, returning after their entry was
collapsed, and alternating between every pair of unmeasured causes.

Every scenario that asserts a fault is ABSENT proves the channel it watched was alive for the whole
window, not just at the start, and distinguishes "asked, and there is no such fault" from "could not
ask". The harness carries a test of that test: it points a scenario at a fault surface that dies
mid-window and fails if the silence assertion passes.

Reviewers can check the restart behaviour with `ctest -R test_lifecycle_expectation_e2e_main`, the
crash loop with `ctest -R test_lifecycle_expectation_e2e_restart_loop`, and the config validation
with `ctest -R test_lifecycle_expectation_integration`.

---

## Checklist

### DIFF
--- a/src/ros2_medkit_plugins/ros2_medkit_graph_watchdog/include/ros2_medkit_graph_watchdog/lifecycle_expectation_tracker.hpp
+++ b/src/ros2_medkit_plugins/ros2_medkit_graph_watchdog/include/ros2_medkit_graph_watchdog/lifecycle_expectation_tracker.hpp
@@ -1,0 +1,929 @@
+// Copyright 2026 bburda
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+#pragma once
+
+#include <algorithm>
+#include <cstddef>
+#include <iterator>
+#include <limits>
+#include <map>
+#include <optional>
+#include <set>
+#include <string>
+#include <utility>
+#include <vector>
+
+namespace ros2_medkit_graph_watchdog {
+
+/// Consecutive ticks a require_active entry may match NO node at all before it is worth
+/// warning about. A required node that never came up is invisible to the presence class
+/// (GRAPH_NODE_DISAPPEARED needs the node present and online first), so nothing else
+/// reports it.
+inline constexpr int kDefaultNoMatchWarnTicks = 10;
+/// Consecutive ticks a matched node may be genuinely ABSENT from the snapshot before its
+/// clocks start moving again. Inside this budget a node's whole state is simply HELD:
+/// zeroing it on the first blink would mean a node dropping out of one snapshot in every
+/// few never accumulates enough consecutive present ticks on either clock below to be
+/// reported at all. PAST this budget absence CONTINUES whatever the node's last real
+/// observation had started - it never restarts a clock and never discards one. See
+/// "Absence continues, it never erases" in LifecycleExpectationTracker's class doc.
+inline constexpr int kDefaultAbsenceGrace = 3;
+
+/// Default bound on how many nodes this tracker keeps state for at once - the operator's
+/// `tracked_node_cap`. Evidence is never reclaimed by AGE (an age horizon is itself an
+/// evasion for a node that touches it periodically), so this is what bounds the map: at the
+/// cap, IDLE entries - both clocks at zero and no matured ownership, so nothing to lose -
+/// are reclaimed first, then DEPARTED entries are collapsed into a count (see
+/// kMaxNamedDepartedEntries), and only if every tracked node is PRESENT and carrying
+/// evidence is a newly-seen node refused and the saturation surfaced to the operator. 512
+/// against a realistic graph: this map only ever holds nodes a require_active entry MATCHED,
+/// and a large single-robot ROS 2 graph (a full Nav2 stack plus perception) runs to roughly
+/// a hundred nodes while a ten-robot fleet sharing one domain runs to a few hundred - so
+/// even an operator whose entries match every node in the graph stays well under the cap, at
+/// a cost of a few hundred KB to hold it. Growth past it can only come from identity CHURN
+/// (respawns under ever-new fqns), which is exactly what the cap exists to bound. Same
+/// value, for the same reason, as ros2_medkit_log_bridge's own max_tracked_nodes default.
+inline constexpr int kDefaultTrackedNodeCap = 512;
+
+/// The most DEPARTED entries kept individually NAMED when the cap has to make room for a
+/// present node; the rest are collapsed into a per-code count (see "A present node always
+/// wins a slot" in the class doc). Three, and the number falls straight out of the
+/// description budget: a detail is capped at kMaxLifecycleDetailChars and
+/// AggregatedFault::kMaxDescriptionChars holds at most three of them
+/// (3 * 150 + 2 * 2 = 454 <= 480, while a fourth would not fit - the arithmetic
+/// lifecycle_expectation_detector.cpp already static_asserts). So a fourth named departed
+/// entry could never appear in any description however the ordering fell out; past three
+/// they are pure cost, and the cost they impose is a slot a PRESENT node needs.
+inline constexpr int kMaxNamedDepartedEntries = 3;
+
+/// Consecutive matched ticks an UNMEASURED observation (kUnreadable or kNotManaged) must
+/// hold before ABSENCE is allowed to continue it - see "Absence continues the last
+/// CORROBORATED observation" in the class doc. A present, healthy, managed node reads
+/// kNotManaged for a tick whenever its get_state path is missing from one sweep
+/// (LifecycleWatcher::update() drops a tracked id whose path is absent from the current
+/// sweep, and discover_apps() can yield an app with no services when a sweep races service
+/// enumeration), so a single such reading must never be what decides that a node departed
+/// unmeasurable. Six: no single racing sweep produces six consecutive readings, it is the
+/// same bar the detector's own typo warning already sets (kUnmanagedWarnTicks fires on the
+/// sixth consecutive unmanaged tick, guarding against this exact transient), and it is a
+/// tenth of kDefaultUnmeasuredHoldTicks - so a node that is genuinely unmeasurable when it
+/// leaves has corroborated that long before the hold which would report it.
+inline constexpr int kDefaultObservationSettleTicks = 6;
+
+/// Consecutive ticks a node's UNMEASURED clock (see LifecycleObservedState below) may
+/// climb before the cause it is currently climbing under takes ownership of the node and
+/// gets its own fault code. "Frozen one past itself" like every other bounded clock here:
+/// past the bound the stored counter stops incrementing, so it only needs to distinguish
+/// "still climbing" from "matured", never how far past. Not configurable via detector
+/// config (mirrors the previous kUnmeasuredHoldTicks, which was a detector-file-local
+/// constant) - exposed here as a constructor default so a test can shrink it without
+/// waiting out 60 real ticks.
+inline constexpr int kDefaultUnmeasuredHoldTicks = 60;
+
+/// Longest label a REAL lifecycle implementation can produce on `~/transition_event`: the
+/// four primary states this detector enforces against (`unconfigured`, `inactive`,
+/// `active`, `finalized`) plus the six transition states lifecycle_msgs/msg/State.msg
+/// defines (`configuring`, `cleaningup`, `shuttingdown`, `activating`, `deactivating`,
+/// `errorprocessing`). The longest of the ten is `errorprocessing` at 15 characters.
+/// kMaxLifecycleLabelChars is more than double that, so no conforming label is ever
+/// affected by the trim below - only a value from a non-conforming remote publisher (the
+/// label arrives verbatim off a remote `~/transition_event`, see lifecycle_watcher.cpp)
+/// spends the budget. This is the PRIMARY guard (R13): the label is the untrusted,
+/// unbounded part of the detail, so it is trimmed on its own rather than as part of a
+/// whole-detail head-and-tail that would also mangle the node name.
+inline constexpr std::size_t kMaxLifecycleLabelChars = 32;
+
+/// Backstop on the WHOLE per-node detail string (after the label above is already
+/// trimmed), for a pathological fqn or a long `require_active` "required by" list rather
+/// than the label - R13's secondary guard. Sized so at least 3 maximally-long details
+/// still fit inside one description under AggregatedFault::kMaxDescriptionChars (480,
+/// joined by "; "): 3 * 150 + 2 * 2 = 454 <= 480, while 4 maximally-long details would not:
+/// 4 * 150 + 3 * 2 = 606 > 480. lifecycle_expectation_detector.cpp static_asserts this
+/// arithmetic against the real AggregatedFault::kMaxDescriptionChars constant, since this
+/// header does not include aggregated_fault.hpp (see its own "Pure." class doc).
+inline constexpr std::size_t kMaxLifecycleDetailChars = 150;
+
+/// Truncates `text` to at most `max_chars`, replacing the tail with "..." when it does not
+/// fit whole - the same suffix-ellipsis convention AggregatedFault::describe_ordered uses
+/// for the overall description, so a trimmed label or detail reads the same way a trimmed
+/// description does. Shared by every detail builder below.
+inline std::string trim_to(const std::string & text, std::size_t max_chars) {
+  if (text.size() <= max_chars) {
+    return text;
+  }
+  constexpr std::size_t kMarkerChars = 3;  // "..."
+  const std::size_t keep = max_chars > kMarkerChars ? max_chars - kMarkerChars : 0;
+  return text.substr(0, keep) + "...";
+}
+
+/// One live node matched by a require_active entry, with its lifecycle state.
+struct LifecycleMatch {
+  std::string entry;  ///< the require_active entry that matched
+  std::string fqn;    ///< the node's stable App::effective_fqn()
+  std::optional<std::string> state;
+};
+
+/// The one observed fact this whole state machine runs on: what a MATCHED node's read
+/// tells us this tick. A node that did not match anything this tick never reaches
+/// classify_observed_state() at all - it is ABSENT, and absence is handled by a wholly
+/// separate path (the per-node absence budget in update()), not by this enum, because
+/// absence carries no label to classify.
+enum class LifecycleObservedState {
+  kActive,      ///< label == "active"
+  kInactive,    ///< label non-empty and != "active" - the one state that ever counts
+                ///< toward the violation clock
+  kUnreadable,  ///< optional("") - a MANAGED node whose label has never been read
+  kNotManaged,  ///< nullopt - no tracked lifecycle at all this tick
+};
+
+/// Pure function of the label alone.
+inline LifecycleObservedState classify_observed_state(const std::optional<std::string> & state) {
+  if (!state.has_value()) {
+    return LifecycleObservedState::kNotManaged;
+  }
+  if (state->empty()) {
+    return LifecycleObservedState::kUnreadable;
+  }
+  return *state == "active" ? LifecycleObservedState::kActive : LifecycleObservedState::kInactive;
+}
+
+/// Which of the two "I cannot measure this" causes a node's unmeasured clock most
+/// recently climbed under. Meaningful in two different ways depending on
+/// `NodeState::unmeasured_matured` (private below): while the clock is still climbing it
+/// is simply "what did the last unmeasured tick look like" (refreshed every such tick, for
+/// the withheld-clear log's own bookkeeping); once matured it is frozen - see "Sticky vs.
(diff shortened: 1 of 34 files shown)


src/ros2_medkit_plugins/ros2_medkit_graph_watchdog/include/ros2_medkit_graph_watchdog/lifecycle_expectation_tracker.hpp
```diff
@@ -0,0 +1,278 @@
...
+        ++it;
+      }
+    }
+
+    // Pending: streaks that have started but not passed grace. Bounded by the SAME absence
+    // budget the streak itself is, so a node the tracker has already handed to the presence
+    // class holds nothing back - and a streak whose absence bookkeeping is gone entirely
+    // cannot resurrect as a hold either.
+    for (const auto & miss : misses_) {
+      if (miss.second <= 0 || miss.second > grace_) {
```
mfaferek93: The comment is doing the damage here: past-grace implies "already in `affected`" only when the node was seen this tick. `report.affected` is written at :184 inside the loop over `nodes`, which is built from `matches`.

On an absent tick the node is in neither set: not in `seen_fqns` so not in `affected`, `misses_` preserved by the absence loop above, skipped here so not in `pending`. In the detector `guard.read` is sticky (`lifecycle_expectation_detector.cpp:351` is the only true-assignment) so the unread leg at :371 excludes it, and `ever_matched_` keeps `unmatched_blocking` false. All three legs of the guard at :379 stay quiet and `emit(ctx, {})` sends EVENT_PASSED on the very first blink, while the absence loop right above is deliberately preserving the streak for 3 ticks.

I compiled the header standalone (grace=1, absence_grace=3) to be sure:

```
present tick 1  affected=1 pending=0     <- reported
ABSENT  tick 1  affected=0 pending=0 tracked=1
ABSENT  tick 2  affected=0 pending=0 tracked=1
ABSENT  tick 3  affected=0 pending=0 tracked=1
ABSENT  tick 4  affected=0 pending=0     (streak erased - by design)
```

So a node measured stuck *and already reported* gets strictly less protection than one below grace, which `PendingStopsWhenAbsenceOutlivesTheAbsenceGrace` pins as holding through absent ticks 1-2. Under bringup_params.yaml the hysteresis latch absorbs a single blink, but it still stamps `last_passed` on a still-violating fault and burns the healing debounce; at `healing_threshold: 0` or `1` (the value the package README's own config block recommends) it flips to HEALED.

Gating on presence keeps both behaviours:

```cpp
if (miss.second <= 0 || (miss.second > grace_ && seen_fqns.count(miss.first) != 0)) {
```

Patched a copy and re-ran: absent ticks 1-3 hold the clear, tick 4 hands over as designed. `PendingStopsWhenAbsenceOutlivesTheAbsenceGrace` and `AbsenceAfterRaiseClearsNotNodeDeathsDomain` both still pass. Dropping the `> grace_` term outright would break the first.

src/ros2_medkit_plugins/ros2_medkit_graph_watchdog/include/ros2_medkit_graph_watchdog/lifecycle_expectation_tracker.hpp
```diff
@@ -0,0 +1,278 @@
...
+      const int m = ++misses_[fqn];
+      if (m > grace_) {
+        report.affected[fqn] = "node " + fqn + " expected active but is " + *node.state + " (required by " +
+                               join_quoted(node.entries) + ")";
+      }
+    }
+
+    // A matched node that left the snapshot (or came back unread) keeps its streak for
+    // absence_grace_ ticks; past that the streak is discarded, since absence itself
+    // belongs to the presence class (GRAPH_NODE_DISAPPEARED).
```
mfaferek93: This handoff is load-bearing in several places (also `lifecycle_expectation_detector.cpp:243-251` and :296-300, plus `AbsenceAfterRaiseClearsNotNodeDeathsDomain`), but `GRAPH_NODE_DISAPPEARED` exists only in this PR's test comments. No code definition, no registered detector (`REGISTER_DETECTOR` yields exactly qos_mismatch, orphan, param_drift, lifecycle_expectation), nothing under any branch. `graph_watchdog_plugin.cpp:44` already reads `detectors.node_death.miss_grace` and cites a `node_death_detector.cpp` that is not in the tree.

So the terminal case today is: required node stuck inactive raises correctly, its process then dies outright, everything clears, nobody reports it. Fine if node_death is landing right behind this - just worth saying so on the PR, or landing them together.

src/ros2_medkit_plugins/ros2_medkit_graph_watchdog/src/detectors/lifecycle_expectation_detector.cpp
```diff
@@ -0,0 +1,470 @@
...
+        }
+        const std::optional<std::string> state = ctx.gate ? ctx.gate->lifecycle_state_of(app.id) : std::nullopt;
+        if (state.has_value()) {
+          entry_has_managed_match.insert(id);
+        }
+        // One match per (entry, node): the tracker keys violations by NODE, so two
+        // namesakes are both reported instead of one silently replacing the other.
+        matches.push_back(LifecycleMatch{id, fqn, state});
+        bool & read = matched_read[fqn];
+        read = read || (state.has_value() && !state->empty());
```
mfaferek93: `read` collapses two cases the tracker deliberately keeps apart: `nullopt` (no lifecycle to be wrong about, `lifecycle_expectation_tracker.hpp:251`) and `""` (a seed that has not answered, aged like absence at :173). Both land as `!read` and both get the same `kUnmeasuredHoldTicks` bound.

That bound is right for `nullopt` - a typo must not block healing forever, which is what the comment at :337-341 argues. It is wrong for `""`. A managed node hung inside `on_configure` on a single-threaded executor cannot serve `get_state`, and its `configuring` transition event went out before the volatile subscription matched. LifecycleWatcher seeds once plus `kReseedAttempts = 2` (`lifecycle_watcher.cpp:48`, budget set at first sighting and never replenished), all timing out, so the label stays `""` for the life of the process. Then: the tracker `continue`s at :173 so no streak ever starts and `pending` is empty, `guard.unmeasured` freezes past 60 and drops out of the scan at :371, `unmatched_blocking` is false because the entry did match - and from tick 61 on the detector emits a clear every tick.

That is a positive "this node is fine" about a node it has never measured, which is the failure mode the detector exists to catch. Suggest tracking the two apart: keep the bounded release for `nullopt`, and for `""` either keep withholding or re-warn instead of going quiet.

Worth noting the integration fixture cannot catch this - its apps carry no services (test:30-34), so a node without an injected label is simply absent from `LifecycleWatcher::tracked` and `Some("")` never reaches the guard.

src/ros2_medkit_plugins/ros2_medkit_graph_watchdog/src/detectors/lifecycle_expectation_detector.cpp
```diff
@@ -0,0 +1,470 @@
...
+        ++unread;
+      }
+    }
+    if (report.affected.empty() && (unread > 0 || !report.pending.empty() || unmatched_blocking)) {
+      report_withheld_clear(ctx, unread, unread_example, report.pending, unmatched_blocking);
+      return;  // nothing measured - and a young streak - is not the same as healthy
+    }
+    withheld_ticks_ = 0;
+    withheld_reported_ = false;
+    aggregated_.emit(ctx, report.affected);
```
mfaferek93: This is the only aggregating detector calling plain `emit()`. `orphan_detector.cpp:187`, `qos_mismatch_detector.cpp:203` and `param_drift_detector.cpp:1068` all pass a new-first order to `emit_ordered` so the 480-char cap cannot hide a fresh entity.

The cap applies here too, `source_id` is the constant `graph_watchdog`, and the record is keyed by code alone - so the description is the only thing carrying which node broke. A detail runs ~97 chars, so 4-5 fit. With `require_active: ["controller_server"]` over /robot01../robot12: once robot01-05 are stuck and CONFIRMED, robot12 going inactive produces a byte-identical description and is never named. `TwentyFiveStuckNodesRaiseOneCappedFault` pins the lexicographic order, so I assume this was deliberate - but the consequence looks unintended.

Same line, secondary: `*node.state` is `msg->goal_state.label` taken verbatim off a remote `~/transition_event` (`lifecycle_watcher.cpp:272`). The cap keeps it out of the store, but with no per-node trim (param_drift has one at :1064) one long label eats the whole budget and masks every other affected node.

src/ros2_medkit_plugins/ros2_medkit_graph_watchdog/include/ros2_medkit_graph_watchdog/lifecycle_expectation_tracker.hpp
bburda: You were right. The first fix was the presence gating you wrote.

After that we changed the model, so the fix now sits lower down. Absence does not erase anything. While a node is absent inside the grace, it keeps its whole state. After the grace, absence keeps counting on the clock the node was already on. So a node that was reported stays reported while it blinks. A node that leaves while it is still violating stays reported until it comes back and reads `active`.

We made the bigger change because of a pattern. Any rule that throws away what we already measured is a way out of the detector. A node only has to hit that rule again and again. A node that keeps crashing and restarting does exactly that, and it is the node we most want to catch.

There is an end to end test for your case. It runs the fault manager with `healing_threshold 1`. It blinks the node twice and fails if a `PASSED` arrives. Each blink also checks that the node was away for longer than one tick. If the gap is shorter, the tracker can miss it, and then the blink tests nothing.

src/ros2_medkit_plugins/ros2_medkit_graph_watchdog/include/ros2_medkit_graph_watchdog/lifecycle_expectation_tracker.hpp
bburda: The PR description says it now: the code exists, there is no detector for it, and that work has its own issue.

Your case is also fixed, not only written down. A departure does not heal a fault any more. If a required node is stuck inactive and then dies, `GRAPH_NODE_INACTIVE` stays raised. Absence keeps the violation going instead of clearing it. The other case is still open. A healthy node that disappears raises nothing. That one belongs to the disappeared code.

One limit, because it is easy to miss. This only holds inside one gateway process. After a gateway restart the detector starts from nothing. It has no measurements, and an entry that matches no node looks the same as a typo. So the fault can heal there. We added a test that pins this behaviour, so it is a known limit and not a surprise later.

src/ros2_medkit_plugins/ros2_medkit_graph_watchdog/src/detectors/lifecycle_expectation_detector.cpp
bburda: They are separate now. Each tick the tracker puts a node in one state. `nullopt` and an empty label are two different states. Each one has its own fault code. `GRAPH_NODE_UNREADABLE` is for a managed node whose label we cannot read. `GRAPH_NODE_NOT_MANAGED` is for a node with no lifecycle at all. Neither of them stops reporting after a timeout.

You were right that the integration fixture could not reach this. We added a demo node that advertises the lifecycle services and never answers `get_state`. The end to end test now drives a real read failure through the gateway and the fault manager. A second demo node drops its own lifecycle services when asked, which covers the `nullopt` side.

Splitting the two needed one extra guard. A healthy managed node reads as not managed for one sweep if its `get_state` path is missing from that sweep. If that is the last sweep before a clean shutdown, a healthy node would get a permanent fault. So a node now has to read the same way for several ticks in a row before we use that reading to describe a departure.

src/ros2_medkit_plugins/ros2_medkit_graph_watchdog/src/detectors/lifecycle_expectation_detector.cpp
bburda: Both are changed.

All three aggregates use `emit_ordered` now and pass the new nodes first. A node that crossed on this tick is named before older ones, so the description cap cannot hide it.

The label is also trimmed per node, at 32 characters. One whole detail is capped at 150. A static assert ties that to the real description cap, so three worst case details still fit. A long label from a remote transition event cannot take the whole budget any more.

mfaferek93 (approved):

PR #588, opened 2026-08-04
perf(cmake): analyse a package's translation units in parallel under clang-tidy

### LINKED ISSUE #586: Full clang-tidy gate runs one file at a time
## Summary

The full clang-tidy gate (`./scripts/test.sh tidy`, i.e. the `clang_tidy` CTest test) analyses one translation unit at a time on any number of cores.

`ament_clang_tidy` defaults to `--jobs 1`, and our wrapper `ros2_medkit_clang_tidy()` in `ros2_medkit_cmake/cmake/ROS2MedkitLinting.cmake` passes neither a `JOBS` argument nor the `ament_cmake_clang_tidy_JOBS` variable that the upstream macro also honours. CTest-level parallelism does not compensate: each package registers a single `clang_tidy` test, so `--ctest-args -j $(nproc)` only overlaps packages, and the largest package still runs serially inside its own test.

Observed on a 16-core machine: the gateway package has 244 entries in its compile database, and a full run sits at 45+ minutes with a single `clang-tidy` process at 100% CPU while the other 15 cores idle.

Two paths are not affected and stay as they are:

- CI does not use this code path at all. `quality.yml` drives `run-clang-tidy` (parallel by default) with `clang-tidy-cache`, and analyses only changed files on pull requests.
- The pre-push incremental script `scripts/clang-tidy-diff.sh` loops serially, but pre-commit already invokes it with several file batches in parallel, so it is not the bottleneck.

So this is a local-developer cost only, paid every time someone runs the full gate before pushing.

## Proposed solution

Pass the job count through the wrapper:

- add a `JOBS` argument to `ros2_medkit_clang_tidy()` and a `ROS2_MEDKIT_CLANG_TIDY_JOBS` cache variable defaulting to the host core count;
- run the `tidy` preset in `scripts/test.sh` with `--ctest-args -j 1`, because the parallelism now lives inside each package test - keeping both would multiply peak memory by the number of packages (7 packages register the test, and each `clang-tidy` process holds roughly half a gigabyte).

## Additional context

The default stays conservative in the sense that it matches the machine it runs on, and a developer who wants a quieter box can set `-DROS2_MEDKIT_CLANG_TIDY_JOBS=<n>` at configure time.

bburda:
# Pull Request

## Summary

The full clang-tidy gate analysed one translation unit at a time regardless of how many cores the machine had.

`ament_clang_tidy` defaults to `--jobs 1`, and `ros2_medkit_clang_tidy()` passed neither a `JOBS` argument nor the `ament_cmake_clang_tidy_JOBS` variable the upstream macro also honours. CTest-level parallelism did not compensate, because each package registers a single `clang_tidy` test, so `-j` only overlaps packages while the largest one sets the wall clock on its own.

This passes a job count through the wrapper, defaulting to the host core count, and runs the `tidy` preset with one CTest job. The parallelism now lives inside each package test; keeping both levels would multiply peak memory by the number of packages, and a `clang-tidy` process on this codebase holds roughly half a gigabyte.

Two paths are deliberately unchanged:

- CI does not use this code path. `quality.yml` drives `run-clang-tidy` with `clang-tidy-cache`, and analyses only changed files on pull requests.
- `scripts/clang-tidy-diff.sh` loops serially, but pre-commit already invokes it with several file batches at once, so it is not the bottleneck.

---

## Issue

- closes #586

---

## Type

- [ ] Bug fix
- [ ] New feature or tests
- [ ] Breaking change
- [ ] Documentation only

Build tooling / developer ergonomics.

---

## Testing

Measured on the `ros2_medkit_gateway` package (244 entries in its compile database), 16-core machine, same worktree and same build for both runs, `colcon test --ctest-args -j 1 -R '^clang_tidy$'`:

| `ROS2_MEDKIT_CLANG_TIDY_JOBS` | Wall clock |
|---|---|
| 1 (previous behaviour) | 2125 s |
| 16 (new default on this host) | 261 s |

Both runs report the same single pre-existing finding, so the change alters scheduling only, not what gets analysed.

To verify: configure with `-DENABLE_CLANG_TIDY=ON` and check the registered command in `build/<pkg>/CTestTestfile.cmake` - it now ends in `--jobs <n>`. `-DROS2_MEDKIT_CLANG_TIDY_JOBS=1` restores the old behaviour, and `ros2_medkit_clang_tidy(JOBS <n>)` overrides it for one package.

The serialisation is already being worked around in the tree. `ament_clang_tidy` defaults to a 300 s test timeout, and `ros2_medkit_gateway` registers its check with `TIMEOUT 3000` - a tenfold raise. `ros2_medkit_fault_manager` takes the default and, measured on the same 16-core machine, its serial run exceeds it and the test fails by timeout; with the job count passed through it finishes in roughly a minute. So this is not only a wall-clock cost: on a machine with cores to spare, a package that never raised its timeout has a red gate for no reason other than scheduling.

The `TIMEOUT 3000` on the test is left as is - it was close to being reached by the serial run and is now comfortably clear.

---

## Checklist

No behaviour change for callers; the `ros2_medkit_cmake` README gained a section describing the module, the new knob, and why the two levels of parallelism must not both be enabled.

### DIFF
--- a/scripts/test.sh
+++ b/scripts/test.sh
@@ -60,9 +60,13 @@ case "$PRESET" in
       "$@"
     ;;
   tidy)
-    echo "==> Running clang-tidy (this will take a while)"
+    echo "==> Running clang-tidy"
+    # One package at a time: each clang_tidy test already analyses its own
+    # translation units in parallel (ROS2_MEDKIT_CLANG_TIDY_JOBS, default one
+    # process per core). Running the package tests concurrently on top of that
+    # would multiply peak memory by the number of packages.
     colcon test "${COMMON_ARGS[@]}" \
-      --ctest-args -j "$(nproc)" -R "clang_tidy" \
+      --ctest-args -j 1 -R "clang_tidy" \
       "$@"
     ;;
   all)
--- a/scripts/test.sh
+++ b/scripts/test.sh
@@ -22,7 +22,7 @@ set -euo pipefail
 #   unit        - Unit tests only, no linters, no integration (DEFAULT)
 #   integ       - Integration tests only
 #   lint        - Linters only (clang-format, copyright, cmake-lint; NO clang-tidy)
-#   tidy        - clang-tidy only (slow, ~8-10 min)
+#   tidy        - clang-tidy only (slow; accepts --jobs N, see below)
 #   all         - Everything (equivalent to bare colcon test)
 #   <test_name> - Run a single test by CTest name regex
 #
@@ -32,6 +32,12 @@ set -euo pipefail
 #   ./scripts/test.sh lint                   # fast linters only
 #   ./scripts/test.sh test_health_handler    # single test
 #   ./scripts/test.sh unit --packages-select ros2_medkit_gateway
+#   ./scripts/test.sh tidy --jobs 8          # trade memory for speed
+#
+# The tidy preset defaults to a small number of clang-tidy processes per package
+# so one package fits an 8 GB machine. Each process peaks around 1.4 GiB, so
+# --jobs N costs roughly N x 1.4 GiB. The value sticks in the CMake cache and is
+# printed on every run.
 
 PRESET="${1:-unit}"
 shift 2>/dev/null || true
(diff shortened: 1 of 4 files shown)


scripts/test.sh
```diff
@@ -60,9 +60,13 @@ case "$PRESET" in
...
   tidy)
-    echo "==> Running clang-tidy (this will take a while)"
+    echo "==> Running clang-tidy"
+    # One package at a time: each clang_tidy test already analyses its own
+    # translation units in parallel (ROS2_MEDKIT_CLANG_TIDY_JOBS, default one
+    # process per core). Running the package tests concurrently on top of that
+    # would multiply peak memory by the number of packages.
     colcon test "${COMMON_ARGS[@]}" \
-      --ctest-args -j "$(nproc)" -R "clang_tidy" \
+      --ctest-args -j 1 -R "clang_tidy" \
```
mfaferek93: ctest -j is a no-op here - every package has exactly one clang_tidy test. The overlap comes from colcon: COMMON_ARGS still passes --parallel-workers $(nproc), so the three independent bridges each spawn nproc tidy jobs at once - the exact memory blowup the comment above says is prevented. Appending --parallel-workers 1 after "${COMMON_ARGS[@]}" fixes it (later flag wins). Also worth knowing: ament_clang_tidy exits 0 even when a tidy process gets OOM-killed, so this fails green.

src/ros2_medkit_cmake/README.md
```diff
@@ -24,6 +24,24 @@ Resolves dependency differences across ROS 2 distributions:
 - `medkit_target_dependencies()` - Drop-in replacement for `ament_target_dependencies` (removed on Lyrical)
 - `medkit_detect_compat_defs()` / `medkit_apply_compat_defs()` - Compile definitions for version-specific APIs
 
+### ROS2MedkitLinting
+
+Registers the package's `clang_tidy` CTest test behind `-DENABLE_CLANG_TIDY=ON`
+(off by default locally, on in CI).
```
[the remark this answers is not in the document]
bburda: You are right. The CTest target is not used in CI. quality.yml configures with `-DENABLE_CLANG_TIDY=OFF` and runs `run-clang-tidy` over the compilation database. It is a local gate only.

Fixed in three places that all said the same wrong thing: this section, the module table at the top of the README, and the linting entry in `design/index.rst`.

The 0.4.0 changelog entry still says "mandatory in CI". I left it on purpose. CI moved to `run-clang-tidy` in April, after that release, so the entry was true when it was written.

scripts/test.sh
bburda: All three are right.

On `-j`: there is one `clang_tidy` test per package, and colcon runs a separate ctest for each package, so `-j` had nothing to overlap. I removed it.

On `--parallel-workers`: I first did what you suggested and put `--parallel-workers 1` after `${COMMON_ARGS[@]}`. That was not enough. colcon still reads its own options after `--ctest-args`, so `./scripts/test.sh tidy --parallel-workers 8` ends up with three of them and the last one wins. I tried it, and `diagnostic_bridge` and `log_bridge` both started before either one finished. The pin is now after the caller's arguments, and the preset prints a note when it ignores an override like that. With that order the `Starting` and `Finished` lines stay in sequence, with or without a trailing flag.

Your OOM point made me change the default too. I measured the gateway package:

```
jobs=2   1078 s    2.6 GiB
jobs=4    595 s    4.7 GiB
jobs=8    359 s    9.4 GiB
jobs=16   272 s   17.4 GiB
```

Memory grows by about 1.2 GiB per job, but the time does not go down in the same way. One job per core meant 17.4 GiB for a single package. The default is now `min(host cores, 2)`, so one package fits an 8 GB machine. To change it, use `./scripts/test.sh tidy --jobs <n>`. It reconfigures the packages first, about 3 s each, and does not rebuild.

You were right about it failing green as well. `ament_clang_tidy` only counts the warnings it parsed, so a process that gets killed adds nothing to the result. `run_test.py` writes the child stderr into the per-test log, so the preset now looks for `failed with error code` in those logs, fails the run, and prints which packages were not fully analysed.

mfaferek93 (approved):

PR #590, opened 2026-08-04
Cut PR feedback time from 56 to about 35 minutes

### LINKED ISSUE #589: CI feedback takes 56 minutes and most of it is one build step
## Summary

A pull request waits about 56 minutes for an answer (p50, 42 runs over 13 days; p90 is
63.5 min). In the fastest quarter of iterations the next commit is pushed 21 minutes after
the previous one, so the answer arrives long after the author moved on.

The time is not spread out. It sits in two jobs.

- `sanitizer-asan` ends last in 98% of Quality runs on pull requests. Its build step alone
  is 35.2 minutes, which is 64% of the whole critical path.
- `coverage` ends last in 79% of CI runs. Inside its 12.7-minute report step, more than 99%
  is a single-threaded `lcov --capture` (measured 13.43 of 13.50 minutes in one job, 17.87
  of 17.95 in another). It produces no pass or fail signal.

## Why the build is slow

ccache is not caching. `CCACHE_MAXSIZE` is 500M everywhere, and every job that reaches that
ceiling shows hundreds of evictions during its own build: the cache is throwing away the
objects the same build is still producing. Measured hit rates at the ceiling are 0% to 49%.

The clearest measurement is one job compared with itself. The same clang-tidy job, same
workspace, same compiler, built in 5m19s at 81% cache hits and in 21m21s at 11% hits.

There is a second reason the caches keep getting evicted. The repository cache holds about
10.5 GB against a 10 GB limit, so GitHub is dropping entries all the time. About 3.1 GB of
that is intermediate Docker layers kept by `cache-to: type=gha,mode=max` in the image
publishing workflow, which only runs on pushes to `main`.

## Proposed solution

Free the cache budget first, then let ccache work, then remove the work that does not
gate anything:

- Publish images with `mode=min` so the layer cache stops taking 30% of the budget.
- Add `-g1` to sanitizer builds. The sanitizer module already overrides the optimisation
  level but never the debug level, so the instrumented tree carries full debug info and
  reaches about 27 GB. Reports still name file and line, because that needs the line table
  and the variable name in a stack overflow report comes from the sanitizer itself.
- Raise `CCACHE_MAXSIZE` for the sanitizer jobs now that there is room.
- Run `lcov --capture` in parallel.
- Merge `jazzy-build` into `jazzy-test`. The build artifact has exactly one consumer, so
  the split pays about 4.7 minutes of repeated container setup for no parallel work.
- Run `coverage` and the Pixi build on pushes to `main` instead of on every pull request.
  Pixi is `continue-on-error`, so it never gated anything, and the static half of the
  coverage scope check already runs on every pull request in the Quality workflow.

Expected result is about 34 to 37 minutes, with no test and no distribution removed.

## Not proposed

Dropping Humble or Lyrical saves nothing, because neither is on a critical path, and both
have caught real bugs before. Running sanitizers only at night would cut the number the
most, but `main` has no required status checks, so a red commit could land unnoticed.
Passing `-j` to ctest is tempting, but these integration and end to end tests are timing
sensitive and a flaky test is worse than a slow one.

bburda:
## Summary

Targets the 56-minute pull request feedback time measured in the issue, with an expected
result of 34 to 37 minutes. No test and no distribution is dropped from pull requests. Two
things do change and are called out below: the coverage job moves to `main`, and the image
publish job gives up its builder layer cache.

Measured on this branch. Each run changed something the caches depended on, so read them in
order:

| run | critical path | why |
|---|---|---|
| 1 | 59 min, `sanitizer-asan` | change 7 alters the sanitizer compile command line, so both sanitizer caches were invalid and built cold |
| 2 | 41 min, `jazzy-test` | sanitizers warm and at target: 99% hits, 36m55s and 34m47s. `jazzy-test` was restoring the ASan cache, see change 10 |
| 3 | 38 min, `jazzy-test` | its key prefix now matches nothing else, so it had nothing to restore and was cold by construction |

The sanitizer jobs reach the expected figure from run 2 onwards, which is what the issue
identified as 64% of the critical path. `jazzy-test` has not yet had a run with a warm cache
of its own; the next push is its first.

Ten changes, in the order they depend on each other:

1. `docker-publish.yml`: `cache-to` goes from `mode=max` to `mode=min`. It was holding about
   3.1 GB of intermediate layers in a repository cache that is already over its 10 GB limit,
   for a job that only runs on pushes to `main`. Nothing else can be fixed until this frees
   space. This costs build time in that job. The `Dockerfile` is two-stage and `mode=min`
   exports only the final image, so builder-stage layers are not restored at all. In
   practice the loss is smaller than it sounds: the `COPY src/` layers sit above the
   `rosdep` + `colcon build` layer, so any push touching `src/` already invalidated that
   layer under `mode=max` too. What such a push genuinely loses is the build-deps `apt-get`
   layer. The job is not on the pull request path, and the quota it was holding is what
   every pull request needs.
2. `ROS2MedkitSanitizers.cmake`: adds `-g1` to sanitizer builds. The module already
   overrode the build type's optimisation level but never its debug level, so
   `RelWithDebInfo` kept full debug info and the instrumented tree grew to about 27 GB,
   mostly DWARF. On one translation unit the object shrinks by 45% and its debug info by
   72%.
3. `quality.yml`: `CCACHE_MAXSIZE` goes to 2G for ASan and 1.5G for TSan. At 500M both jobs
   evicted hundreds of objects during their own build and never got above 18% hits.
4. `ci.yml`: `lcov --capture` gets `--parallel`. It was more than 99% of the 12.7-minute
   coverage report step and ran on one core.
5. `ci.yml`: `jazzy-build` is merged into `jazzy-test`. The build artifact had exactly one
   consumer, so the split paid about 4.7 minutes of repeated container and ROS setup for no
   parallel work. The job id stays `jazzy-test`, so `notify-demos` needs no change.
6. `ci.yml`: the `coverage` job runs on pushes to `main` only, and gains
   `--return-code-on-test-failure`. It was missing that flag while the two other test jobs
   have it, so a failing test there could not fail CI. The job read as a gate without being
   one.
7. `ROS2MedkitSanitizers.cmake`: adds `-UNDEBUG`. This is the compensation for change 6.
   The coverage job was the last pull request build compiling without `-DNDEBUG`, so on its
   own that change would have taken every `assert()` out of pull request builds, since the
   remaining jobs are `Release` or `RelWithDebInfo`. The two sanitizer jobs already run on
   every pull request, over a superset of the test selection the coverage job used, so
   putting the asserts there costs no CI time. Note the reach: the flag is directory-scoped,
   so it also enables assertions in headers compiled into a participating package.
   `nlohmann/json` routes `JSON_ASSERT` to `assert`, and the vendored `cpp-httplib` and
   `dynmsg` carry their own. That is the intent, but it does mean a latent bug in a header
   now shows up as an abort in the sanitizer jobs and nowhere else.
8. `pixi.yml`: runs on pushes to `main` plus a nightly cron instead of on every pull
   request. It is `continue-on-error`, so it never gated a pull request, and its cache is
   only written on `main`, so every pull request paid a 30.8-minute cold build.
9. `scripts/ccache_report.sh`: reports what ccache actually did. Changes 1 and 3 rest on the
   claim that ccache stops evicting during its own build, and nothing read the numbers that
   would show whether it worked. Every ccache-backed build step now runs `ccache -z` before
   compiling and writes hit rate, cache size, fill level and cleanup count to the job
   summary afterwards. The zeroing matters: ccache keeps its counters inside the cache
   directory and `actions/cache` restores that directory whole, so read as they come the
   numbers are the running total over every job that fed the entry rather than what this
   build did.

   It raises a workflow warning in two states. A cache at 90% of its ceiling or more that is
   still running cleanups is discarding objects the same build produced, so the ceiling is
   too small or the Actions quota is evicting the entry. A hit rate under 50% without that
   ceiling pressure means the cache was not restored, or the branch changed a compiler flag
   and invalidated the objects. Cleanups on their own prove nothing, because ccache trims a
   subdirectory when it passes `max_size/16` and a nearly empty cache reports a few. The
   script never fails the step, because a cold cache is a legitimate state.

10. `ci.yml`: `jazzy-test` gets its own ccache key prefix. It used `ccache-jazzy-`, which
    also matches `ccache-jazzy-asan-`, `-tsan-`, `-lint-` and `-tidy-`, and `restore-keys`
    takes the most recently created match. Run 2 shows it loading
    `ccache-jazzy-asan-818af7de`: 2 GB of instrumented objects it cannot use, 42 hits out of
    411 calls, then 219 cleanups as its own `Release` objects fought for the 500M ceiling on
    top of them. It was the critical path at 41m7s while the ASan job whose cache it had
    taken finished in 36m55s at 99% hits. The prefix is now `ccache-jazzy-test-`. This is
    older than this branch; it only became visible once change 9 made the numbers per-build.

Pull requests keep the static coverage scope check, which is in the Quality workflow and
fails if a package compiles production C++ without coverage instrumentation.

`QUALITY_DECLARATION.md` and `CONTRIBUTING.md` are corrected against what the workflows
actually do, including checks that were already inaccurate before this branch: the build is
not "full" (all three distro jobs skip `ros2_medkit_opcua`), clang-tidy is incremental on a
pull request, the Sphinx and OPC-UA workflows and two source gates also gate pull requests,
sanitizer instrumentation does not reach packages that omit the module, and no required
status check enforces any of it.

---

## Issue

- closes #589

---

## Type

- [x] Bug fix
- [ ] New feature or tests
- [ ] Breaking change
- [ ] Documentation only

---

## Testing

`lcov --parallel` produces the same file as the serial run. Same md5, all 163 source file
records equal, same line and function totals. The whole downstream chain of filters and
`genhtml` runs on the parallel output.

`-g1` keeps what a sanitizer report needs. An ASan stack buffer overflow built with the new
flags still names file and line in every frame and still names the overflowed variable,
because that name comes from the frame descriptor the sanitizer embeds, not from DWARF.
What is lost is inspecting local variables in a debugger on a core file.

`-UNDEBUG` was verified by running the suite with the asserts live. The workspace was built
`RelWithDebInfo` with the config flags overridden to `-O1 -g1`, which is what the sanitizer
jobs compile with minus the instrumentation, and with no `-DNDEBUG` in any compile command.
The coverage job's test selection then ran over 18 packages: 4571 tests, and no assertion
fired anywhere. No `Assertion ... failed`, no `SIGABRT`, no core dump. One integration test
failed, `test_external_app_fault_rollup`, and it is not related: it fails at the same rate
with the asserts compiled out (2 of 10 passing either way, run alternately to keep machine
load equal), its log carries the `Buffer is empty, cannot create bag file` warning, and a
branch that fixes that empty-buffer path passes it 10 times out of 10 on the same machine.

`ccache_report.sh` has its thresholds pinned by `scripts/test_ccache_report.sh`, which runs
in `format-lint`. It stubs ccache on `PATH` so the script runs unmodified, and its rows are
the numbers six jobs reported on the first run of this branch plus three constructed cases:
a full cache that is not evicting, a step that compiled nothing, and a cache that is both
pinned and cold. Against the earlier threshold, which warned on any cleanup at all, it
fails on two of those rows, which is what makes it a test rather than a constant.

Both the threshold and the zeroing came out of that first run rather than from reading the
code. It
reported cleanups in three jobs, one of them a cache at 15% of its ceiling, which is ccache
trimming a subdirectory and not a ceiling worth raising. Chasing that down showed the
counters were cumulative, which is what change 9 now handles.

The speedup itself cannot be measured locally and is not claimed here. The `ccache` rows in
the job summaries are the thing to read, and from the second run onwards they describe a
single build: cleanups should be zero or the cache well below its ceiling, and the hit rate
well above the 50% floor. For scale, a cold full `Release` build of the workspace stores
about 101 MiB across 376 entries locally, so the 500M ceilings hold several builds' worth
of objects.

Every workflow file still parses, and every `needs:` edge and `needs.<job>.*` expression in
all seven workflows was resolved against the declared job list, with no broken references.

---

## Checklist

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -118,10 +118,22 @@ jobs:
             log/
             build/*/test_results/
 
-  jazzy-build:
+  # Builds AND tests Jazzy. These were two jobs (jazzy-build -> jazzy-test)
+  # passing a tarred build/ + install/ tree between them. That split dates from
+  # when lint ran off the same artifact in parallel; lint has since moved to
+  # quality.yml, where format-lint and clang-tidy each do their own build, so
+  # the artifact was left with exactly one consumer and the split bought no
+  # parallelism at all. It cost 4.24 min of duplicated container bring-up and
+  # ROS setup in the second job plus 0.50 min of tar/upload in the first.
+  #
+  # The job id stays jazzy-test because notify-demos gates on it by name.
+  jazzy-test:
     runs-on: ubuntu-latest
     container:
       image: ubuntu:noble
+    # The more generous of the two timeouts this job replaces (build had 60,
+    # test had 45). One job now covers both phases, and a cold ccache makes the
+    # build phase the variable one, so keep the headroom rather than the sum.
     timeout-minutes: 60
     defaults:
       run:
@@ -158,7 +170,9 @@ jobs:
       - name: Install dependencies
         run: |
           apt-get update
-          apt-get install -y ros-jazzy-test-msgs
+          apt-get install -y \
+            ros-jazzy-test-msgs \
+            ros-jazzy-rmw-cyclonedds-cpp
           source /opt/ros/jazzy/setup.bash
           rosdep update
           rosdep install --from-paths src --ignore-src -y
@@ -175,61 +189,6 @@ jobs:
             --cmake-args -DCMAKE_BUILD_TYPE=Release \
             --event-handlers console_direct+
           ccache -s
-
-      - name: Package build artifacts
-        run: tar cf /tmp/jazzy-build.tar build/ install/
-
-      - name: Upload build artifacts
-        uses: actions/upload-artifact@v4
-        with:
-          name: jazzy-build
-          path: /tmp/jazzy-build.tar
-          retention-days: 1
-
-  jazzy-test:
-    needs: jazzy-build
-    runs-on: ubuntu-latest
-    container:
-      image: ubuntu:noble
-    timeout-minutes: 45
-    defaults:
-      run:
-        shell: bash
-
-    steps:
-      - name: Install Git
-        run: |
-          apt-get update
-          apt-get install -y git
-
-      - name: Checkout repository
-        uses: actions/checkout@v4
-
-      - name: Pre-install ROS 2 apt source
-        uses: ./.github/actions/ros-apt-source
-
-      - name: Set up ROS 2 Jazzy
-        uses: ros-tooling/setup-ros@v0.7
-        with:
-          required-ros-distributions: jazzy
-
-      - name: Install dependencies
-        run: |
-          apt-get update
-          apt-get install -y \
-            ros-jazzy-test-msgs \
-            ros-jazzy-rmw-cyclonedds-cpp
-          source /opt/ros/jazzy/setup.bash
-          rosdep update
-          rosdep install --from-paths src --ignore-src -y
-
-      - name: Download build artifacts
-        uses: actions/download-artifact@v4
-        with:
-          name: jazzy-build
-
-      - name: Extract build artifacts
-        run: tar xf jazzy-build.tar && rm jazzy-build.tar
 
       - name: Run unit and integration tests
         env:
@@ -258,6 +217,15 @@ jobs:
             build/*/test_results/
 
   coverage:
+    # Push to main only. This job owned CI's critical path in 79% of PR runs at
+    # ~40 min, and what a PR loses by not running it is narrow: its Codecov
+    # upload was already gated to push (see the last step), and the source-tree
+    # half of the coverage-scope gate - the half that catches a new package
+    # compiling C++ without opting into instrumentation - runs on every PR in
+    # quality.yml's format-lint job via check_coverage_packages.sh --static-only.
+    # What moves to main-only is the runtime half, which needs .gcda files and
+    # therefore needs this job's build and test run.
+    if: github.event_name == 'push' && github.ref == 'refs/heads/main'
     runs-on: ubuntu-latest
     container:
       image: ubuntu:noble
@@ -319,14 +287,26 @@ jobs:
       - name: Run unit and integration tests for coverage
         run: |
           source /opt/ros/jazzy/setup.bash
-          colcon test \
+          # --return-code-on-test-failure, as in the two other colcon test
+          # invocations in this file. Without it colcon exits 0 whatever the
+          # tests did, so this job ran the full Jazzy suite under a coverage
+          # build and could not fail CI on a failure - it read as a gate
+          # without being one.
+          colcon test --return-code-on-test-failure \
             --packages-skip ros2_medkit_opcua \
             --ctest-args -LE linter \
             --event-handlers console_direct+
 
       - name: Generate coverage report
         run: |
+          # --parallel: this single lcov --capture was measured at >99% of the
+          # whole step (13.43 min of 13.50, 17.87 of 17.95) and is otherwise
+          # single-threaded, while --extract, --remove, --list and genhtml
+          # together finish in under six seconds. It forks per .gcda directory,
+          # so the output is the same records from the same gcov runs, only
+          # gathered concurrently.
           lcov --capture --directory build --output-file coverage.raw.info \
+            --parallel "$(nproc)" \
             --ignore-errors mismatch,negative,empty,gcov
 
           if [ ! -s coverage.raw.info ] || ! grep -q 'SF:' coverage.raw.info; then
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -165,9 +165,16 @@ jobs:
         uses: actions/cache@v4
         with:
           path: /root/.cache/ccache
-          key: ccache-jazzy-${{ github.sha }}
+          # ccache-jazzy-test-, not ccache-jazzy-. The shorter prefix also
+          # matches ccache-jazzy-asan-, -tsan-, -lint- and -tidy-, and
+          # restore-keys takes the most recently created match, so this job kept
+          # restoring whichever other Jazzy job finished last. It was measured
+          # loading the ASan cache: 2 GB of instrumented objects it cannot use,
+          # 10% hits, and then 219 cleanups as its own Release objects fought
+          # for the 500M ceiling on top of them.
+          key: ccache-jazzy-test-${{ github.sha }}
           restore-keys: |
-            ccache-jazzy-
+            ccache-jazzy-test-
 
       - name: Install dependencies
         run: |
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -82,6 +82,7 @@ jobs:
             --cmake-args -DCMAKE_BUILD_TYPE=Release \
             --event-handlers console_direct+
           ccache -s
+          ./scripts/ccache_report.sh "${{ matrix.ros_distro }}"
 
       - name: Run unit and integration tests
         # The suite grew past the old 15-minute budget: on lyrical it was already taking
@@ -189,6 +190,7 @@ jobs:
             --cmake-args -DCMAKE_BUILD_TYPE=Release \
             --event-handlers console_direct+
           ccache -s
+          ./scripts/ccache_report.sh jazzy-test
 
       - name: Run unit and integration tests
         env:
@@ -225,6 +227,14 @@ jobs:
     # quality.yml's format-lint job via check_coverage_packages.sh --static-only.
     # What moves to main-only is the runtime half, which needs .gcda files and
     # therefore needs this job's build and test run.
+    #
+    # This was also the only PR job building without -DNDEBUG, so on its own it
+    # would have taken every assert() out of pull request builds - the other
+    # jobs are Release or RelWithDebInfo. ROS2MedkitSanitizers.cmake now passes
+    # -UNDEBUG, which puts the asserts back in the two sanitizer jobs. Those run
+    # ctest -LE linter per package, a superset of the selection here - they do
+    # not skip ros2_medkit_opcua - so the asserts are still exercised against
+    # unit and integration tests on every PR.
     if: github.event_name == 'push' && github.ref == 'refs/heads/main'
     runs-on: ubuntu-latest
     container:
@@ -283,6 +293,7 @@ jobs:
             --cmake-args -DCMAKE_BUILD_TYPE=Debug -DENABLE_COVERAGE=ON \
             --event-handlers console_direct+
           ccache -s
+          ./scripts/ccache_report.sh coverage
 
       - name: Run unit and integration tests for coverage
         run: |
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -77,6 +77,7 @@ jobs:
           CCACHE_SLOPPINESS: pch_defines,time_macros
         run: |
           source /opt/ros/${{ matrix.ros_distro }}/setup.bash
+          ccache -z
           colcon build --symlink-install \
             --packages-skip ros2_medkit_opcua \
             --cmake-args -DCMAKE_BUILD_TYPE=Release \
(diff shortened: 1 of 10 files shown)


.github/workflows/docker-publish.yml
```diff
@@ -42,5 +42,13 @@ jobs:
...
-                  cache-to: type=gha,mode=max,scope=${{ matrix.ros_distro }}
+                  # mode=min, not max: max exports every intermediate build
+                  # stage, which for three distros held ~3.1 GB of the repo's
+                  # 10 GB Actions cache quota - about 30% of the budget, for a
+                  # job that only runs on push to main. The quota was measured
+                  # over the limit (10.54 GB active), so GitHub was evicting
+                  # LRU, and the ccache entries the build jobs depend on were
+                  # what it evicted. mode=min keeps the final-stage layers,
+                  # which is what cache-from actually reuses here.
+                  cache-to: type=gha,mode=min,scope=${{ matrix.ros_distro }}
```
mfaferek93, line 63: mode=min only exports the final stage, and this is a two-stage build - so the builder stage (apt deps, rosdep, colcon build) rebuilds from scratch on every main push. You free the 3.1 GB, but the publish job gets noticeably slower. If that trade is intentional, worth one sentence in the description; a middle ground is mode=min plus a registry cache for the builder stage only.

.github/workflows/quality.yml
```diff
@@ -298,7 +298,15 @@ jobs:
...
-          CCACHE_MAXSIZE: 500M
+          # 2G, not the 500M the other jobs use. At 500M this job's ccache was
+          # measured at 101-104% of its ceiling with 551-1019 cleanups DURING a
+          # single build - ccache evicting the objects that same build was still
+          # producing - and it never exceeded 18% hits. The instrumented object
+          # set does not fit in 500M and cannot be made to; the -g1 change in
+          # ROS2MedkitSanitizers.cmake shrinks it, this gives it somewhere to
+          # land. Sized against the repo's 10 GB Actions cache quota, which the
+          # docker-publish mode=min change frees room in.
+          CCACHE_MAXSIZE: 2G
```
mfaferek93, line 321: These keys are per-SHA, so no run ever restores a previous cache - every PR starts cold and the size bump alone won't change the 18% hit rate. Needs a restore-keys prefix fallback (ccache-jazzy-asan-) so runs can reuse the latest entry; then the bigger MAXSIZE actually pays off.

.github/workflows/ci.yml
```diff
@@ -258,6 +217,15 @@ jobs:
...
   coverage:
+    # Push to main only. This job owned CI's critical path in 79% of PR runs at
+    # ~40 min, and what a PR loses by not running it is narrow: its Codecov
+    # upload was already gated to push (see the last step), and the source-tree
+    # half of the coverage-scope gate - the half that catches a new package
+    # compiling C++ without opting into instrumentation - runs on every PR in
+    # quality.yml's format-lint job via check_coverage_packages.sh --static-only.
+    # What moves to main-only is the runtime half, which needs .gcda files and
+    # therefore needs this job's build and test run.
+    if: github.event_name == 'push' && github.ref == 'refs/heads/main'
```
mfaferek93, line 247: This drops more than the description says: coverage was the only PR job that built without -DNDEBUG, so after this change no PR build compiles asserts in. A failing assert now only surfaces on main (or in the sanitizer jobs, which also build Release). Either keep one Debug-flavoured PR job or call the loss out in the description - 'no test coverage is removed' isn't quite true as written.

.github/workflows/docker-publish.yml
bburda: You are right that this is not free, and I have put the cost into the description and into
the comment here.

One correction to how big it is. The `COPY src/` layers sit above the `rosdep` + `colcon
build` layer, so any push that touches `src/` already invalidated that layer under
`mode=max` as well. What such a push really loses is the build-deps `apt-get` layer above
it. A push that changes nothing under `src/` loses the whole builder stage, but those are
rare on `main`.

I did not take the middle ground with a separate builder cache. It would give back the
quota this change frees, and that quota is the reason for the change. Would you rather keep
the builder cache and find the space somewhere else?

.github/workflows/quality.yml
bburda: The prefix fallback is already there. `quality.yml` lines 276-277 set `restore-keys:
ccache-jazzy-asan-` under the per-SHA key, and the TSan job has the same at lines 412-413,
so a run does restore the most recent entry instead of starting cold.

The 18% had a different cause. At 500M the job evicted objects during its own build, which
is what the size raise fixes. The runs on this branch show it: ASan went from 18% hits to
99% once it had room, and the job went from 59 minutes to about 33.

Your instinct about the keys was right for a different job, though. `jazzy-test` used the
prefix `ccache-jazzy-`, which also matches `ccache-jazzy-asan-` and the other three, and
`restore-keys` takes the most recent match. It was loading the ASan cache: 2 GB of
instrumented objects it cannot use, 42 hits out of 411 calls, then 219 cleanups as its own
objects fought for the 500M ceiling on top of them. I only saw it after adding a per-build
ccache report. It is fixed the way you suggest here, with a prefix that matches nothing
else.

.github/workflows/ci.yml
bburda: You are right, and the reach is wider than our own three asserts. The flag is
directory-scoped, so `nlohmann/json` routing `JSON_ASSERT` to `assert`, and the vendored
cpp-httplib and dynmsg with their own, were all being compiled out too.

I did not take either of your two options. `ROS2MedkitSanitizers.cmake` now passes
`-UNDEBUG`, so the asserts come back in the two sanitizer jobs. Those already run on every
pull request over a superset of the test selection the coverage job used, so this costs no
CI time, and it does not need a Debug job kept alive just for asserts.

I checked it locally rather than leaving it to CI. The workspace was built `RelWithDebInfo`
with the config flags set to `-O1 -g1` and no `-DNDEBUG`, which is what the sanitizer jobs
compile with minus the instrumentation, and the coverage job's test selection ran over 18
packages. 4571 tests, and no assert fired anywhere. One integration test failed, and it
fails at the same rate with the asserts compiled out, so it is unrelated.

The description now says what is lost instead of claiming nothing is.

QUALITY_DECLARATION.md
```diff
@@ -87,7 +87,7 @@ GitHub Copilot code review is used in addition to human review.
 All pull requests must pass CI before merging:
 
 - **Build & Test job:** Full build + unit/integration tests on Ubuntu Noble / ROS 2 Jazzy, Ubuntu Jammy / ROS 2 Humble, and Ubuntu Resolute / ROS 2 Lyrical. Linter tests on Jazzy only
-- **Coverage job:** Debug build with coverage. Reports are generated for all PRs as artifacts and uploaded to [Codecov](https://codecov.io/gh/selfpatch/ros2_medkit) on pushes to `main`
+- **Coverage job:** Debug build with coverage, run on pushes to `main`. Reports are archived as artifacts and uploaded to [Codecov](https://codecov.io/gh/selfpatch/ros2_medkit). Every pull request is gated on the static coverage-scope check in the Quality workflow, which fails if a package compiles production C++ without opting into coverage instrumentation
 - Linting enforced: `clang-format`, `clang-tidy` via `ament_lint_auto`
```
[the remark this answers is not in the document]
bburda: Reworded. The block now separates what a pull request actually runs from the coverage job,
which only reports on `main`.

Checking the rest of it against the workflows turned up more that was stale: the build is
not "full" because all three distro jobs skip `ros2_medkit_opcua`, clang-tidy is incremental
on a pull request and does not run at all when no C++ changed, the Sphinx and OPC-UA
workflows and two source gates also run on pull requests, and sanitizer instrumentation
does not reach packages that omit the shared module. The section also claimed a merge gate
that no required status check enforces, so it now says that plainly.

CONTRIBUTING.md
```diff
@@ -152,7 +152,7 @@ build.
 
 #### CI/CD
 
-All PRs are tested on Ubuntu 24.04 (Jazzy) with parallel lint + test jobs, plus Humble and Lyrical. Coverage is uploaded to Codecov on push to main. All CI jobs use ccache.
+All PRs are tested on Ubuntu 24.04 (Jazzy), where build and tests run in a single job, plus Humble and Lyrical. Linters and clang-tidy run in the Quality workflow. Coverage is uploaded to Codecov on push to main. All CI jobs use ccache.
```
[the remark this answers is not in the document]
bburda: Corrected. It now names where ccache is configured, which is the CI and Quality workflows
and the OPC-UA workflow's colcon jobs, and says that the Pixi workflow and the Docker image
builds compile the workspace without it. That way a slow build in those two reads as
expected rather than as a cache problem.

mfaferek93 (approved):

PR #593, opened 2026-08-05
fix(gateway): make executor_threads real and report a cancel timeout as a timeout

### LINKED ISSUE #575: [BUG] server.executor_threads has no effect: everything shares one callback group
# Bug report

### Steps to reproduce
1. Start the gateway with `server.executor_threads: 2` (the default).
2. Set a short discovery refresh, e.g. `refresh_interval_ms: 1000`, as the integration tests do.
3. Issue an HTTP request that blocks on a ROS service future, such as cancelling an action execution.
4. Observe how long the response takes while the periodic discovery pass is running.

### Expected behavior
With two executor threads, a service response can be dispatched while another callback is running. `server.executor_threads` means what its name says.

### Actual behavior
Dispatch is serialised regardless of the setting. Every ROS entity the gateway creates lands in the node default MutuallyExclusive callback group: the generic service clients, the action status subscription, and all five timers. A MutuallyExclusive group serialises its callbacks no matter how many executor threads exist, so the second thread is never usable and `server.executor_threads` is decorative.

Two consequences follow. An HTTP thread waiting on a service future can only be unblocked by that one group, and the periodic discovery refresh runs in the same group, so a full discovery pass can sit in front of a response somebody is waiting for. It also makes a class of timeouts indistinguishable from remote failures: a response that arrived but could not be dispatched looks exactly like one that never came.

### Environment
- ROS 2 distro: jazzy and humble
- OS: Ubuntu 24.04 / 22.04

### Additional information
Found while tracing an action-cancel timeout that could not be reproduced under load. The measured cancel round trip is around a millisecond in every configuration tried, including two pinned cores with six competing processes, a 125-node churning graph, and CycloneDDS, while CI showed a five second stall. That gap points at dispatch, not at the wire.

Likely fix: give the blocking client work its own reentrant callback group so responses can be dispatched while other callbacks run.

Related: #576.

### LINKED ISSUE #576: [BUG] Action cancel reports a definitive rejection when it merely timed out
# Bug report

### Steps to reproduce
1. Start an asynchronous action execution through the gateway.
2. Cancel it with `DELETE /apps/{app}/operations/{op}/executions/{id}` under conditions where the CancelGoal response does not arrive within the gateway budget.
3. Read the error body, then read the execution status the gateway reports.

### Expected behavior
A timeout is reported as a timeout, and the gateway state agrees with the `/_action/status` stream it is already subscribed to.

### Actual behavior
The gateway returns HTTP 400 with `x-medkit-ros2-action-rejected`, which states that the action server refused the cancellation. What actually happened is that no answer arrived in time. The goal may well have been accepted for cancellation.

The response shape gives it away: `return_code: 0` with `success: false` is "no answer", not a rejection, and there is no distinct code for it.

On that path the gateway also leaves its tracked status untouched, while the `/_action/status` subscription already knows the real outcome. So the state the gateway reports and the state it has just received disagree, and nothing reconciles them.

### Environment
- ROS 2 distro: humble (observed), jazzy
- OS: Ubuntu 22.04 (CI container)

### Additional information
Worth fixing together: report a timeout as a timeout, and reconcile the cancel outcome against the status stream rather than deciding it from the request future alone.

The cancel budget was raised from 5s to 15s as a stopgap while investigating this. That is not the fix. The measured round trip is around a millisecond in every configuration tried, including two pinned cores under load, a churning 125-node graph, and CycloneDDS, so a five second stall is a lost response rather than a tight budget. The likely underlying cause is tracked in #575.

bburda:
# Pull Request

## Summary

Two related bugs on the action cancel path.

`server.executor_threads` had no effect. Every ROS entity the gateway creates was placed in the node default callback group. That group is mutually exclusive, so callbacks run one at a time, no matter how many executor threads are configured. An HTTP thread waiting for a service or action response could only be unblocked by that one group. The discovery refresh runs in the same group, so a full refresh pass could delay a response that a client was waiting for.

The generic service clients and the three action clients now use a shared reentrant group. The per-action status subscriptions use a shared mutually exclusive group of their own. Order inside one subscription is still kept, but status updates no longer wait behind the default group. Both groups are created once at startup, by a helper in `ros2_common/`. The gate that forbids creating callback groups outside that directory still passes and its allowlist did not grow. Timers and the fault, trigger and `/rosout` subscriptions stay in the default group. Discovery should be serialized against itself, and `refresh_mutex_` already does that.

The second bug: a cancel that timed out was reported as `400 x-medkit-ros2-action-rejected`. That code means the action server refused the cancellation. What really happened is that no answer arrived in time. The goal may have been accepted for cancellation.

`ActionCancelResult` now carries an explicit outcome. The transport sets it on every exit path. One shared mapping is used by `DELETE .../executions/{id}` and by `PUT` with `{"capability": "stop"}`:

- No answer in time: the gateway first reads the status stream it is already subscribed to. If the stream shows the goal as cancelling or cancelled, the cancel is reported as accepted, because it was accepted. If not, the answer is `504` with the standard code `not-responding`, and the message says the outcome is unknown.
- Cancel service not reachable: `503 x-medkit-ros2-action-unavailable`.
- Any other transport failure: `500 x-medkit-ros2-action-unavailable`.
- Execution no longer tracked: `404`. This is the same answer the same request gets one moment later.
- Real rejection from the server: `400 x-medkit-ros2-action-rejected`, not changed.

Two response messages changed. A `503` or `500` on either route now starts with the operation the client asked for and keeps the transport message as the reason: `Stop failed: Cancel service not available`. The same holds for a return code `CancelGoal` does not define. Before, a stop request could get that message worded as a cancel.

The 15 s minimum cancel budget was added earlier as a temporary fix. It is removed here. The budget is `service_call_timeout_sec` again, like the other action RPCs. That parameter had no range check and no documentation. Both are added.

Two more problems on the same path were found while fixing it.

The status subscription used volatile and best effort. The action server offers reliable and transient local. The subscription is also created only after the goal is sent. So a goal that finished quickly could lose its last status message, and nothing could recover it. The subscription now uses the profile that `rcl_action` defines for action clients.

A goal sent while the cleanup timer was removing the subscription for the same action path could end up tracked with no status stream. After this change that state also decides whether a timed out cancel answers 204 or 504.

Two smaller fixes are included, both in code this change already rewrites. The `Location` header of a created execution was built from a fixed apps or components pair. On the areas and functions routes it pointed into the components collection, and it did not use `api_path()`. It is now built from the request path. The `409` answer on the stop route used `invalid-request`. That code is not in the SOVD list for this case, so it now uses `precondition-not-fulfilled`. The same document already uses that code for the same situation. This changes the response body, so it is written here directly.

One change is outside the gateway and not part of either bug. `ros2_medkit_log_bridge` declared `severity_floor` and `max_tracked_nodes` as `int`, while a ROS 2 parameter is `int64`. An out-of-range value narrowed before the range check ran, so a value that should be rejected could pass. Both are read as `int64_t` now and narrowed after the clamp. The cooldown timestamp in the same file is taken by const reference.

The branch was rebased over the base branch while open, and two commits come from that. The tests added here now use `medkit_add_gtest`, which replaced `medkit_set_test_domain`. Without it the package does not configure. The second commit sets `CMAKE_EXPORT_COMPILE_COMMANDS` in the shared linting module, because 13 of the 21 packages exported no compile database and clang-tidy could not analyse them.
---

## Issue

- closes #575
- closes #576

---

## Type

- [x] Bug fix
- [ ] New feature or tests
- [x] Breaking change
- [ ] Documentation only

The breaking part is small. A cancel that answered `400` before can now answer `404`, `500`, `503` or `504`, depending on what happened. The `409` on the stop route uses a different error code. No response that existed before was removed.

---

## Testing

There is a test for each behaviour. Every new test was first run against the old code, to check that it fails, and that it fails for the right reason.

Executor: an integration test points the gateway at an aggregation peer that does not answer. Each discovery pass then stalls in the health check. During the stall the test calls a service backed operation. On `main` the call uses the full 10 s budget and returns `500`. With this change the answer comes back well inside the budget. A second test runs the same operations with `executor_threads: 1`, to show that the reentrant group does not need a second thread. The clamp is tested at both ends of the documented range.

Cancel: a test fixture serves the cancel service directly. It can block, reject, or answer. This covers the timeout, the rejection and the case where the status stream already knows the result. Killing the action server after the goal is accepted covers the unreachable case in an integration test. The QoS fix is covered end to end with an action that finishes quickly. That test failed 2 of 3 runs before the fix and passed 5 of 5 after it.

Build, linters, unit tests, integration tests, clang-tidy and the documentation build all pass. Two integration tests fail. `test_opcua_secured` fails the same way on branches that contain none of this work. The other one is a rosbag test that belongs to a different change.

The generated API description was compared with the merge base. Both gateways were started and `/api/v1/docs` and the entity collections were compared. Nothing was removed. Three responses were added: the `503` and `504` above, and the `409` description. The entity responses are identical.

---

## Checklist

Updated: the executor section of the server configuration page, both operation routes in the REST API page, and the common error code table. That table had four rows with constant names that do not exist in the code. They are corrected here.

### Known, not fixed here

`test_auth_manager.CleanupExpiredTokens` fails about one run in six on my machine. It fails on `main` too. This change does not touch any authentication code. It needs its own issue.

### DIFF
--- a/src/ros2_medkit_gateway/include/ros2_medkit_gateway/compat/generic_client_compat.hpp
+++ b/src/ros2_medkit_gateway/include/ros2_medkit_gateway/compat/generic_client_compat.hpp
@@ -26,6 +26,12 @@
 /// When HAS_GENERIC_CLIENT is false (Humble), GenericServiceClient is a custom class
 /// that replicates GenericClient's behavior using rcl C APIs and the same
 /// rosidl_typesupport_introspection infrastructure available in all distros.
+///
+/// Both paths take the callback group the client's response callback should
+/// be dispatched on (issue #575: the gateway registers its blocking-RPC
+/// clients into a shared Reentrant group so responses are not serialized
+/// behind the node's default MutuallyExclusive group). Passing nullptr falls
+/// back to the node's default group on every distro.
 
 #pragma once
 
@@ -49,15 +55,21 @@
 
 #include <rclcpp/generic_client.hpp>
 
+#include <utility>
+
 namespace ros2_medkit_gateway {
 namespace compat {
 
 using GenericServiceClient = rclcpp::GenericClient;
 
-/// Create a GenericServiceClient (delegates to Node::create_generic_client)
-inline GenericServiceClient::SharedPtr
-create_generic_service_client(rclcpp::Node * node, const std::string & service_name, const std::string & service_type) {
-  return node->create_generic_client(service_name, service_type);
+/// Create a GenericServiceClient (delegates to Node::create_generic_client).
+/// @param group Callback group handling the reply callbacks; nullptr resolves
+///              to the node's default group inside rclcpp.
+inline GenericServiceClient::SharedPtr create_generic_service_client(rclcpp::Node * node,
+                                                                     const std::string & service_name,
+                                                                     const std::string & service_type,
+                                                                     rclcpp::CallbackGroup::SharedPtr group) {
+  return node->create_generic_client(service_name, service_type, rclcpp::ServicesQoS(), std::move(group));
 }
 
 }  // namespace compat
@@ -333,17 +345,22 @@ class GenericServiceClient : public rclcpp::ClientBase {
   std::map<int64_t, Promise> pending_requests_;
 };
 
-/// Create a GenericServiceClient for Humble
-inline GenericServiceClient::SharedPtr
-create_generic_service_client(rclcpp::Node * node, const std::string & service_name, const std::string & service_type) {
+/// Create a GenericServiceClient for Humble.
+/// @param group Callback group handling the reply callbacks; nullptr resolves
+///              to the node's default group inside NodeServices::add_client.
+inline GenericServiceClient::SharedPtr create_generic_service_client(rclcpp::Node * node,
+                                                                     const std::string & service_name,
+                                                                     const std::string & service_type,
+                                                                     rclcpp::CallbackGroup::SharedPtr group) {
   rcl_client_options_t options = rcl_client_get_default_options();
   auto client = std::make_shared<GenericServiceClient>(
       node->get_node_base_interface().get(), node->get_node_graph_interface(), service_name, service_type, options);
 
-  // Register the client with the node's default callback group so the executor
-  // polls it for incoming responses.  Without this, handle_response() is never
+  // Register the client with the given callback group so the executor polls
+  // it for incoming responses. Without this, handle_response() is never
   // called and every future hangs until timeout.
-  node->get_node_services_interface()->add_client(std::dynamic_pointer_cast<rclcpp::ClientBase>(client), nullptr);
+  node->get_node_services_interface()->add_client(std::dynamic_pointer_cast<rclcpp::ClientBase>(client),
+                                                  std::move(group));
 
   return client;
 }
--- a/src/ros2_medkit_gateway/include/ros2_medkit_gateway/compat/generic_client_compat.hpp
+++ b/src/ros2_medkit_gateway/include/ros2_medkit_gateway/compat/generic_client_compat.hpp
@@ -306,23 +306,27 @@ class GenericServiceClient : public rclcpp::ClientBase {
   }
 
   /// Send a request asynchronously. Returns a FutureAndRequestId.
+  ///
+  /// The pending-request mutex is held ACROSS `rcl_send_request`, not just
+  /// around the insertion. The caller is an HTTP thread while `handle_response`
+  /// runs on an executor thread, so a reply to a fast service can arrive before
+  /// this function reaches the insertion. `handle_response` would then find no
+  /// entry, drop the reply, and the caller's future would only end on timeout.
+  /// Sending under the lock makes the reply callback wait for the entry that
+  /// names it. This is what `rclcpp::Client::async_send_request_impl` does with
+  /// its own pending map, and the reason is the same.
   /// @param request Pointer to the deserialized request message (void*)
   FutureAndRequestId async_send_request(const Request request) {
     Promise promise;
     auto future = promise.get_future();
 
-    // Send the request via rcl
     int64_t sequence_number;
+    std::lock_guard<std::mutex> lock(pending_requests_mutex_);
     rcl_ret_t ret = rcl_send_request(get_client_handle().get(), request, &sequence_number);
     if (ret != RCL_RET_OK) {
       rclcpp::exceptions::throw_from_rcl_error(ret, "Failed to send request");
     }
-
-    // Store promise for later fulfillment when handle_response is called
-    {
-      std::lock_guard<std::mutex> lock(pending_requests_mutex_);
-      pending_requests_.emplace(sequence_number, std::move(promise));
-    }
+    pending_requests_.emplace(sequence_number, std::move(promise));
 
     return FutureAndRequestId(std::move(future), sequence_number);
   }
--- a/src/ros2_medkit_gateway/src/http/handlers/operation_handlers.cpp
+++ b/src/ros2_medkit_gateway/src/http/handlers/operation_handlers.cpp
@@ -203,14 +203,9 @@ ErrorInfo make_provider_error(const OperationProviderErrorInfo & info, const std
   return make_plugin_error(info.http_status, info.message, std::move(params));
 }
 
-/// Failure shape produced by `map_cancel_result`. `std::nullopt` from the
-/// mapper means "the cancellation is in progress" and the entry point should
-/// render its success shape (204 for DELETE, 202 for PUT-stop).
-struct CancelFailure {
-  int http_status;
-  const char * error_code;
-  std::string message;
-};
+}  // namespace
+
+namespace detail {
 
 /// Shared outcome mapping for the two cancel entry points (DELETE execution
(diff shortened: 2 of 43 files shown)


src/ros2_medkit_gateway/include/ros2_medkit_gateway/compat/generic_client_compat.hpp
```diff
@@ -333,17 +345,22 @@ class GenericServiceClient : public rclcpp::ClientBase {
...
   auto client = std::make_shared<GenericServiceClient>(
       node->get_node_base_interface().get(), node->get_node_graph_interface(), service_name, service_type, options);
 
-  // Register the client with the node's default callback group so the executor
-  // polls it for incoming responses.  Without this, handle_response() is never
+  // Register the client with the given callback group so the executor polls
+  // it for incoming responses. Without this, handle_response() is never
   // called and every future hangs until timeout.
-  node->get_node_services_interface()->add_client(std::dynamic_pointer_cast<rclcpp::ClientBase>(client), nullptr);
+  node->get_node_services_interface()->add_client(std::dynamic_pointer_cast<rclcpp::ClientBase>(client),
```
mfaferek93, line 366: Now that the client is registered on a group that can run reply callbacks concurrently, the order inside `async_send_request` starts to matter: `rcl_send_request` (L316) returns before the promise is inserted under the mutex (L323). A fast reply can reach `handle_response` first, find no pending entry, drop it, and the caller's future then only ends on timeout. Insert first, erase it if the send fails.

src/ros2_medkit_log_bridge/src/log_bridge_node.cpp
```diff
@@ -65,13 +67,15 @@ void LogBridgeNode::load_parameters() {
   // Default floor is WARN. WARN passes through each node's FaultReporter
   // LocalFilter (threshold/window debounce); ERROR/FATAL bypass it. Raise to 40
   // (ERROR) on chatty / constrained targets to cut volume.
-  const int floor = declare_parameter<int>("severity_floor", kLevelWarn);
+  // ROS parameters are int64; read them as such and narrow only after the
+  // clamp, so an out-of-range value cannot wrap on the way in.
+  const int64_t floor = declare_parameter<int64_t>("severity_floor", static_cast<int64_t>(kLevelWarn));
```
mfaferek93, line 72: This parameter rework is in `ros2_medkit_log_bridge`, unrelated to either fix in the title. No objection to the change itself, it just widens the diff and the revert surface - would rather see it as its own PR.

src/ros2_medkit_gateway/include/ros2_medkit_gateway/compat/generic_client_compat.hpp
bburda: The sequence number only exists after `rcl_send_request` returns, so there is no key to insert before the send, as you suggest. The mutex is now taken before the send and held until the promise is in the map. `rclcpp::Client::async_send_request_impl` uses the same lock scope.

There is no test for it. The window is a few instructions wide, and it only opens if the sending thread is preempted there. I could not force that without adding test-only hooks to the shim.

src/ros2_medkit_gateway/src/http/handlers/operation_handlers.cpp
```diff
@@ -205,6 +205,92 @@ ErrorInfo make_provider_error(const OperationProviderErrorInfo & info, const std
...
+      // the goal finished on its own. Telling the client to watch for
+      // progress would describe something that cannot happen - say what the
+      // gateway already knows instead, in the vocabulary the resource being
+      // named actually answers in (`sovd_status_from_ros2`, not the raw ROS
+      // word: the execution resource never emits "succeeded"/"aborted").
+      if (tracked.has_value() &&
+          (tracked->status == ActionGoalStatus::SUCCEEDED || tracked->status == ActionGoalStatus::ABORTED)) {
+        message += "The execution status resource already reports the goal as " +
+                   sovd_status_from_ros2(tracked->status) + ", so there is nothing left to cancel.";
+      } else {
```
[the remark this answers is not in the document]
bburda: Fixed, and in a third place as well: the branch for a return code `CancelGoal` does not define passed the transport message through unchanged, so a stop request could get `Cancel rejected by action server (return_code 7)`.

Every message is now built from the route's verb. A stop is carried out as an action cancel, so the ROS 2 name stays as the reason after the verb: `Stop failed: Cancel service not available`. New tests in `test_cancel_outcomes.cpp` cover both routes.

src/ros2_medkit_log_bridge/src/log_bridge_node.cpp
bburda: It stays in this pull request, but the description now declares it. `severity_floor` and `max_tracked_nodes` were declared as `int` while a ROS 2 parameter is `int64`, so an out-of-range value narrowed before the range check ran and could pass. Both are read as `int64_t` now and narrowed after the clamp.

The rebase brought a second change of that sort. One commit sets `CMAKE_EXPORT_COMPILE_COMMANDS` in the shared linting module, because 13 of the 21 packages exported no compile database and clang-tidy could not analyse them. The description covers that one too. Each sits in one file and can be reverted alone.

mfaferek93 (approved):

PR #594, opened 2026-08-05
fix(fault_manager): give the fault at a recording-window boundary its own bag

### LINKED ISSUE #574: [BUG] A fault confirmed right after a post-fault recording window gets no rosbag
# Bug report

### Steps to reproduce
1. Configure the fault manager with rosbag snapshots on and a post-fault window, e.g. `snapshots.rosbag.duration_sec: 2.0` and `snapshots.rosbag.duration_after_sec: 0.5`.
2. Report and confirm a fault, and wait for its bag to finalise.
3. Report and confirm a second fault within a few milliseconds of that bag completing.
4. Call `GetRosbag` for the second fault code.

### Expected behavior
The second fault gets a bag, or at minimum a record saying why it could not.

### Actual behavior
No bag and no record. `GetRosbag` reports `success=false` indefinitely and the only trace is a warning in the log:

```
Bag file completed: .../fault_ROSBAG_TEST_001_...
New fault reported: ROSBAG_CLEANUP_TEST          (+1.4 ms)
WARN  Buffer is empty, cannot create bag file
WARN  Failed to create bag file for fault ROSBAG_CLEANUP_TEST
```

`flush_to_bag` returns an empty path when the ring buffer is empty and `on_fault_confirmed` then only warns and returns. No `RosbagFileInfo` row is written and there is no retry, so polling cannot recover a decision made in the first millisecond.

The buffer is empty at that moment by construction: during a post-fault window `message_callback` writes straight to the bag and does not buffer, and `flush_to_bag` has already drained the deque.

### Environment
- ROS 2 distro: reproduced on jazzy and humble
- OS: Ubuntu 24.04 / 22.04 (CI containers), also reproduced locally

### Additional information
This matters most where the feature is supposed to earn its keep. Bursts of correlated faults are the case a black box exists for, and the second fault of a burst is the one that lands in the gap.

The code already accepts the principle for the neighbouring case: `attach_to_active_recording` handles a fault confirming *inside* an active post-fault window. Nothing covers the instant *after* it finalises.

Options, in increasing cost: write a row flagged as having no pre-fault data so the gap is at least visible; keep the recording open across the boundary; or open a post-fault-only bag.

Surfaced when a new package changed test scheduling. The integration tests have been given guards so the suite no longer depends on timing, but those guards only hide it from CI.

bburda:
# Pull Request

## Summary

A fault confirmed right after a post-fault recording window closed got no bag and no record.

During a post-fault window the capture writes messages straight to the bag and stops filling the ring buffer. The flush that started the window already emptied that buffer. So a fault confirmed a few milliseconds after the window closes finds an empty buffer. `flush_to_bag` returned an empty path, the caller only logged a warning, no row was written, and nothing tried again. `GetRosbag` then failed for that fault forever.

This is the case the feature exists for. Faults often arrive in bursts from one root cause, and the second fault of a burst is the one that landed in this gap.

Now, when the buffer is empty and `duration_after_sec > 0`, the capture opens a writer anyway and records only the post-fault window. The second fault of a burst gets its own bag with the data from after the failure. An empty buffer and a real I/O error are now two different results, so a post-fault recording never starts after a failed open. There is no database schema change and no service definition change.

**Which configurations this affects.** The broad topic modes (`all`, `auto` and `entity`, which is the default) subscribe to every topic on the graph, including `/fault_manager/events`. Reporting a fault publishes an event there, so reporting a fault also puts a message in the ring buffer. The next fault of the burst then finds history, not an empty buffer. The empty buffer case belongs mostly to a narrow capture set: `explicit`, a topic list, `config`, or a broad mode with `/fault_manager/events` in `exclude_topics`. The documentation now says this.

## What else is in here

**Recording handover.** Everything that belongs to a recording changes hands in one critical section: the guard, the start time, the writer and the entity filter. If the guard is released first and the writer is taken after it, an incoming confirmation can install its writer in between, and the outgoing finalise then destroys it. Every message is dropped and a row is stored for an empty bag. The entity filter has to be dropped after the writer is taken, not before. Otherwise a subscription callback that already passed the guard reads the empty filter as "write everything" and adds a topic that belongs to no fault to the bag that is being closed.

**Bags and rows stay consistent.** Both finalisation paths use one store-or-discard step. If the metadata store fails, the recording is discarded, because nothing could reach it: bags are looked up by fault code and the quota is computed from rows, so the directory would sit on disk unreachable and uncounted. If the quota sweep fails, the bag and its rows both stay, because the rows are already durable and the sweep is about the whole store. The in-memory backend writes a batch in one step, so a half-stored burst cannot leave rows that name a bag the discard removes. `delete_rosbag_file` now deletes the row before the files, which is what its batch version already did.

**Durations mean what they say.** `duration_sec` is the time the recording was open. It is measured on the monotonic clock and it stops when the writer leaves the write path. Closing the bag and walking the directory for its size are the fault manager's own time, not the recording's. The buffered history a flush may claim is capped at how long the capture has been running, because message ages are wall-clock differences, and a clock step between buffering and flushing would otherwise go into that number in full.

**Latency.** A fault's entity scope is resolved before the recording lock is taken. Resolving it under that lock means a fault store read, whose busy timeout is five seconds, and then two graph queries per topic. Any of that stalling would hold the post-fault window open for as long as the stall.

**Two more problems in the same code.** A fault cleared while its own recording was still being opened kept a row. `on_fault_cleared` removes the code from the recording state only if the guard was already set when the clear arrived, and auto-cleanup finds no row yet, so the finalise inserts one afterwards and nothing removes it. The finalise now asks the store whether each fault is still not cleared. A fault the store does not know keeps its row: unknown is not cleared. Second, past the 32 fault attachment cap, an over-cap fault in `entity` mode got neither a row nor its own data. The cap is meant to withhold only the lookup key, but the code returned before the scope was widened, so the recording never learned that fault's topics and dropped all of its messages.

## Behaviour changes

- `duration_sec` on a stored bag is the real time the recording was open, not the configured windows. It can be longer than `duration_sec + duration_after_sec` when every captured topic goes quiet, because pruning happens when a message arrives.
- The immediate path (`duration_after_sec: 0`) writes its row through the batch call, so a single row goes in a transaction like every other row.
- The post-fault timer is re-armed instead of created again for each recording.

---

## Issue

- closes #574

---

## Type

- [x] Bug fix
- [ ] New feature or tests
- [ ] Breaking change
- [ ] Documentation only

---

## Testing

Each new test was also run against the code without its fix, to check that it fails, and that it fails on the expected assertion.

Tested with a real `RosbagCapture`, a real ROS node, real publishers, a real rosbag2 writer and a real filesystem: the boundary bag and its row on both storage backends; a quota sweep that throws after the rows are committed; a store that fails on the immediate path; a store that throws something that is not a `std::exception`, which the storage interface allows; a finalise that overlaps a scoped boundary recording; the row-before-files delete order; the replaced bag rule for a shared bag; both ends of `max_total_storage_mb`; bag splitting past `max_bag_size_mb`; a fault that attaches to a boundary recording and widens its scope; a fault past the attachment cap that still gets its topics into the bag; a fault cleared while its recording was still opening, which gets no row; repeated confirmations of the fault that already owns the recording, which must not cost a scope resolution each time; and the clock arithmetic behind a recording's origin, including values far enough apart to overflow a signed subtraction.

Tested against a launched `fault_manager_node`: the boundary bag on `sqlite3` and `mcap`, and entity scoping of a boundary recording in the default topic mode. The scoping test reads the bag contents and uses a third node that publishes all the time and never reports a fault, so a recording that stopped scoping cannot pass it.

Tested end to end over the gateway HTTP API: listing and downloading the boundary bag.

**Four things have no test, for the same reason: there is no place to inject the failure.**

The order of the filter clear against the writer handover would need a subscription callback stopped between reading the filter and taking the writer lock.

The duration measurement point moved before the bag close and the directory size walk. Those take tens of milliseconds on a bag a test can create. To get a difference large enough to assert without a flaky test, the bag would need to be about a hundred megabytes, or the directory would need tens of thousands of files. The test next to that change only checks that the measurement does not move past the metadata write, and says so.

The in-memory batch store cannot fail halfway, because the batch is built next to the live map and swapped in. What can be tested there is the replaced bag rule that came with it, and that is tested.

A discard that cannot remove its bag would need `remove_all` to fail. Making the parent directory read-only does that for a normal user but not for root, and the test suite runs as root in CI, so that path has no portable way to fail on purpose.

Build, linters, unit tests, integration tests, clang-tidy and the documentation build pass.

---

## Checklist

The fault manager configuration page, the snapshots tutorial, the package README and the design document describe the recording lifecycle, what decides between a full bag and a post-fault-only bag, and what `duration_sec` means.


src/ros2_medkit_fault_manager/src/rosbag_capture.cpp
```diff
@@ -1098,7 +1257,16 @@ void RosbagCapture::finalize_post_fault_recording() {
...
                   attached.size());
     } catch (const std::exception & e) {
-      RCLCPP_WARN(node_->get_logger(), "Failed to store rosbag metadata for '%s': %s", bag_path.c_str(), e.what());
+      // Without a row nothing can ever reach this bag: retrieval is keyed by fault
+      // code, and quota accounting enumerates rows, so a kept directory would sit
+      // on disk unreachable and uncounted, and the next full storage would not even
+      // be able to evict it. Discard it for the same reason the all-cleared branch
+      // above does.
+      std::error_code ec;
+      std::filesystem::remove_all(bag_path, ec);
```
mfaferek93: This catch assumes nothing can reference the bag yet, but `enforce_storage_limits()` on line 1255 runs after `store_rosbag_files()` has already committed and can still throw (SQLITE_FULL, a failed DELETE, a prepare error). Then `remove_all` deletes a directory that committed fault rows point at, and those rows keep counting toward the quota until GetRosbag notices the files are gone - before this change the path just warned and kept the bag. Moving `enforce_storage_limits()` below the try, or only discarding when the store itself threw, restores the invariant this comment states.

src/ros2_medkit_fault_manager/src/rosbag_capture.cpp
```diff
@@ -1098,7 +1257,16 @@ void RosbagCapture::finalize_post_fault_recording() {
...
+      RCLCPP_WARN(node_->get_logger(),
+                  "Failed to store rosbag metadata for '%s' (%s); discarded the bag, nothing could reference it",
+                  bag_path.c_str(), e.what());
     }
   }
```
mfaferek93: L1273-1276 clears `active_capture_topics_` after `post_fault_timer_mutex_` was already released, so a capture armed in that gap loses its entity filter and records every topic instead of the scoped set. Clear it while the finalize lock is still held, or skip the clear when a new recording has started.

src/ros2_medkit_fault_manager/src/rosbag_capture.cpp
```diff
@@ -351,7 +422,13 @@ void RosbagCapture::on_fault_confirmed(const std::string & fault_code) {
...
     info.file_path = bag_path;
     info.format = config_.format;
-    info.duration_sec = config_.duration_sec;
+    // The span the recording was open, not the configured window: a buffer that
+    // held less than duration_sec of history would otherwise be advertised as a
+    // full one. Exchanged without post_fault_timer_mutex_ here, unlike the finalize
+    // path, and safe for a different reason: no post-roll state was published for
+    // this recording, so no finalise can be running for it, and confirmations are
+    // serialised against each other by the node-level rosbag mutex.
+    info.duration_sec = span_sec_since(recording_started_at_ns_.exchange(0));
```
mfaferek93: The `duration_after_sec == 0` path calls `store_rosbag_file()` at L435 without the try/catch + `remove_all` you just added to the finalize path, so a throwing store still leaves an orphan bag dir here - same failure the PR closes. Probably cleanest as one small store-or-discard helper both paths call.

src/ros2_medkit_fault_manager/src/rosbag_capture.cpp
bburda: Confirmed. `store_rosbag_files()` commits before it returns, so when the sweep throws after it, the rows are already durable and the old code deleted the directory they point at. After that `GetRosbag` fails for those faults, and their bytes still count against `max_total_storage_mb`, which sums rows. The quota never goes down again.

I did not move `enforce_storage_limits()` below the try, because that is not safe here. `finalize_post_fault_recording()` is called from `stop()`, and `stop()` is called from `~RosbagCapture`. A destructor is `noexcept`, so a busy database during shutdown would terminate the process. The sweep now runs after the store, in its own try that only logs. The discard covers a failing store and nothing else. Both handlers also catch exceptions that are not `std::exception`, because the storage interface does not restrict the type.

There is a second half to this. The base `store_rosbag_files()` is a plain loop over single-row writes. A backend that uses it can store part of a burst and then throw, and the discard would then delete a bag that the stored rows point at. The in-memory backend now writes the whole batch in one step, and the base class documents that all-or-nothing is required.

src/ros2_medkit_fault_manager/src/rosbag_capture.cpp
bburda: Agreed, and fixed. The filter now changes hands in the same critical section as the guard, the start time and the writer.

The effect is larger than "records every topic". `entity` is the default mode, and an empty filter means "write everything", so the new recording became a capture of the whole graph.

One detail about the order. Clearing the filter before taking the writer is also wrong. A callback can pass the guard check, then read the filter after it was emptied, take it for "write everything", and write a topic that belongs to no fault into the bag that is being closed. The clear has to happen after the writer is taken out. After that `active_writer_` is null, so a late callback writes nothing.

The old tests could not see this. The race test runs with `topics: all`, where the filter is empty the whole time. The scoping test runs on one thread, so the finalise is over before the next fault confirms. There is a new test that overlaps them.

src/ros2_medkit_fault_manager/src/rosbag_capture.cpp
bburda: Yes, this is a real gap. Both paths go through one helper now. It discards the bag when the store fails, and keeps the bag and the rows when the quota sweep fails.

The immediate path also uses the batch call now, so a single row is written in a transaction like the others.

There was a second problem in the same few lines. The code reset the writer and measured the recording span after that. `unique_ptr::reset()` returns only after `~Writer` has flushed the bag and written `metadata.yaml`, so the close time was added to the recording. Now the writer is taken out under the lock, the span is measured, and the bag is closed after that.

One note about the test double: it only overrode `store_rosbag_files()`, and this path called `store_rosbag_file()`. So it would have kept succeeding here no matter what the test asserted. It fails on both entry points now.

mfaferek93 (approved):

PR #597, opened 2026-08-05
fix(test): give each test its own DDS domain, taken at run time

### LINKED ISSUE #595: [BUG] Test DDS domains map to UDP ports inside the kernel ephemeral range
## Summary

Integration tests get a DDS domain from `ROS2MedkitTestDomain.cmake`. CycloneDDS turns a domain into a UDP discovery port with `7400 + 250 * domain`. Most of the domains we hand out land inside the port range the Linux kernel uses for ephemeral ports, so any process on the machine can be given one of our ports and the whole domain stops working.

When that happens the node dies at startup:

```
ddsi_udp_create_conn: failed to bind to ANY:53150: address in use
[ERROR] [rmw_cyclonedds_cpp]: rmw_create_node: failed to create domain, error Error
Fatal exception in main: failed to initialize rcl node: rcl node's rmw handle is invalid
```

The test then fails with `Exception: Launch stopped before the active tests finished`, which says nothing about the real cause. Every case in the file fails at once, because the fixture never came up.

## The overlap

The kernel ephemeral range on our runners is `32768-60999`. That maps back to domains **102 to 214**.

| package | domain range | domains inside the ephemeral range |
|---|---|---|
| `ros2_medkit_integration_tests` | 130-219 | 85 of 90 |
| `ros2_medkit_linux_introspection` | 130-139 | 10 of 10 |
| `ros2_medkit_opcua` | 220-229 | 0 of 10 |
| `ros2_medkit_gateway` | 30-89 | 0 of 60 |

Only domains `<= 101` and `>= 215` are safe. That is 120 domains in total, and we currently allocate 220.

## How to reproduce

Hold the port with a plain socket, then try to create a node on the matching domain:

```python
import socket, rclpy
s = socket.socket(socket.AF_INET, socket.SOCK_DGRAM)
s.bind(("", 53150))            # 7400 + 250 * 183
rclpy.init(args=["--ros-args"])  # with ROS_DOMAIN_ID=183, RMW=rmw_cyclonedds_cpp
rclpy.create_node("probe")     # fails with the message above
```

Release the port and the same call works.

## Why it was not caught before

Earlier work on test domains fixed collisions **between our own tests** - two tests sharing a domain and seeing each other's nodes. This is a different problem with the same symptom. The port is taken by a process outside our test run, so nothing in our allocation can see it coming. It is rare per run (I found it in 1 of 13 recent Humble jobs), which is why it reads as a flaky test.

## Proposed solution

Move the allocation into the safe band, or stop the kernel from handing out our ports. Two options:

1. Re-carve `ROS2MedkitTestDomain.cmake` so every package draws from `<= 101` or `>= 215`. 120 domains is fewer than the 220 we allocate today, so packages would need smaller pools. ctest runs serially inside a package, so a pool per package rather than a domain per test is enough.
2. Reserve the ports in CI with `net.ipv4.ip_local_reserved_ports` on the container. Smaller change, but it only helps CI, not a developer machine, and it needs checking against Fast-DDS unicast ports too.

Option 1 fixes it everywhere.

## Additional context

Adding a test file to `test/features/` shifts the domain of every test that sorts after it, because the registration uses `file(GLOB ...)` and assigns domains in order. So a test that has been stable for months can start failing after an unrelated test is added, with no change to the test itself.

bburda:
# Pull Request

## Summary

Test DDS domains came from a hand-written table of per-package pools. This replaces the table
with allocation at run time: a test takes a domain when it starts and holds it until it ends.

The first commit fixes the reported bug. Domains were allocated across 1-232. RTPS gives a
domain the UDP slice `[7400 + 250*d, 7400 + 250*d + 249]`, and the kernel hands out ephemeral
ports from `net.ipv4.ip_local_port_range`, 32768-60999 by default. That range maps back to
domains 101-214, so most domains we used could be taken first by any other process on the
machine. The node then died at startup with `failed to bind to ANY:<port>: address in use`,
and the test only said `Launch stopped before the active tests finished`. Test domains now
come from 1-100 and 215-231. Domain 0 is left to the developer shell, and 232 is dropped
because its slice runs past 65535.

The table itself was the second problem. It had to be edited whenever a package outgrew its
pool, the pools had to stay pairwise disjoint, and with 117 domains for more tests than that,
pools wrapped around and two tests of one package shared a domain.

The pools had to be disjoint because colcon runs a separate `ctest` per package, and a CTest
`RESOURCE_LOCK` only binds inside one `ctest` run. An OS-level lock has no such boundary. Two
processes contending for a port do not care which `ctest` started them, so both the table and
the disjointness requirement go away with it.

A test now takes its domain through `domain_coordinator.domain_id` from `ament_cmake_ros`,
which binds a TCP socket on `32768 + domain` and holds it for the caller's lifetime. That
package ships on Humble, Jazzy and Lyrical, so this adds no dependency. Release is the kernel
closing the socket, so a crash and a `SIGKILL` free the domain the same way a normal exit
does. Reuse across a pool is gone with the table, and so is the orphan risk it carried: a node
left behind by a killed test can no longer meet a later test, because the later test only gets
that domain once the lock is free.

The call sites change shape. `ament_add_gtest` plus `medkit_set_test_domain` plus a table
entry becomes `medkit_add_gtest`. There is also `medkit_add_gmock`, `medkit_add_pytest_test`,
`medkit_add_launch_test`, and `medkit_add_wrapped_test` for a test that builds its own command
line. `DOMAINS <n>` asks for several domains at once, and the extras arrive in
`MEDKIT_SECONDARY_DOMAINS`. `medkit_test_needs_no_domain()` marks a test that loads no
middleware.

Adding a test no longer means touching domain allocation anywhere.

### The guard is armed in every package, without the package asking

The failure mode here is silence. A test registered with plain `ament_add_gtest` still passes.
It just runs on domain 0 and shares it with everything else on the machine.

An earlier shape had each package call `medkit_add_domain_allocation_test()`. Four packages
never did, and fourteen of their tests ran on domain 0 with nothing to say so. A guard you have
to remember to arm reproduces the maintenance work this change removes, so
`test_dds_domain_allocation` is now armed from the file ament sources on
`find_package(ros2_medkit_cmake)`, which every package already calls. It reads the generated
CTest properties back and fails on any test whose command does not go through the wrapper,
unless the test is a linter or was marked as needing no domain. A second check,
`test_dds_domain_coverage`, sweeps every package build directory, so a package that never
finds `ros2_medkit_cmake` is still covered.

Measured: 19 gate tests, one per package, 18 of them in packages that ask for nothing.

### Three details that are not obvious

`ros2_medkit_cmake` is declared `NONE`. With no language enabled CMake leaves
`CMAKE_LIBRARY_ARCHITECTURE` empty, so `find_library` never searches the multiarch directory
where Ubuntu keeps libpython. On Humble that makes `launch_testing_ament_cmake` fail inside
`FindPythonLibs`, at configure time, for the whole workspace. The launch test that proves a
launch test gets a domain therefore lives in a package that compiles.

The cross-talk probe starts both peers itself, as concurrent subprocesses that it joins. It
does not rely on two `ctest` tests overlapping, because no `ctest` option guarantees that and
CI does not pass `-j`. It fails unless the peers really did coexist, and it carries its
control: the same two peers forced onto one domain must hear each other. A zero from an
instrument that cannot register a one proves nothing.

`run_test.py` writes a placeholder result file as its first action, so a killed test still
leaves a visible failure. That guarantee starts only once `run_test.py` runs, and the wrapper
does work before it, including a bounded wait for a free domain. A failure in that window used
to leave no result file at all. The wrapper now reserves an equivalent result naming itself as
the cause before anything that can fail.

### Making this safe on a build farm agent

Our CI is not the only place these tests run. `ros2_medkit` is released through rosdistro, and
the farm runs the tests in two job types. Devel jobs run them on all three distros with
`colcon test --executor sequential`, no `-j` and no label filter, so tests run one at a time
and the linters run in the same pass. Humble binarydeb jobs run them too, because Humble's
release build file does not set `package_dependency_behavior`, so `run_package_tests` keeps its
default of true. Jazzy and Lyrical set it to false and run no tests there.

That environment is different from ours in ways that mattered:

- One test held up to 117 lock ports at once, to prove that the excess waits when the band is
  full. Humble binarydeb containers run with `--net=host`, so those would be the agent's own
  ports, on a 4 vCPU machine that also builds other repositories. That test now works against
  a small band it reserved. Peak concurrent wrappers across the suite went from 122 to 12.
- The wait for a free domain was a fixed 180 s, but `ament_add_test` defaults to a 60 s
  timeout. CTest killed the test before the allocator could report, so the error that names
  the cause could only ever appear on 13 of 179 registrations. The wait is now derived from
  each test's own `TIMEOUT`, and `test_dds_domain_allocation` rejects any test whose wait does
  not fit inside its timeout. All 260 wrapped tests can now reach that error.
- `ros2_medkit_test_utils` read `ROS_DOMAIN_ID` with a default of `0`. Domain 0 is the value
  this change exists to avoid, and the guard could not catch it, because such a test is still
  behind the wrapper. It now refuses an unset or zero value, the same way it already did for
  secondary domains.
- A refused `bind` was reported as "all 117 domains are in use". On a machine that denies the
  call, or with the file descriptor limit reached, that is the wrong diagnosis and points at
  the wrong package. Those cases are now separated and named.
- The workspace sweep accepted any sibling directory with a `CTestTestfile.cmake`, so an
  overlay workspace that source-builds a dependency would fail and blame a package we do not
  own. It now only judges packages from this repository.

### Three more, from review

- **A test needing several domains took them one at a time**, and held each one while it waited
  for the next. Two such tests could each hold part of what the other needed. Two integration
  tests ask for four domains, so this was reachable under `ctest -j`. Acquisition is now
  all-or-nothing: the whole set, or nothing held while waiting. That trades a deadlock for a
  bounded livelock, which is paid for by a freshly randomised selector on each attempt and a
  jittered backoff. A fixed interval is what keeps two callers of the same size in step.
- **Nothing read `net.ipv4.ip_local_port_range`.** The band was a constant derived from the
  default range, and on a host tuned to something like `10240 65535`, which is common on CI
  and Kubernetes nodes, most of the band overlaps again. This was a regression: the commit
  this change replaces read that file in five places. The band is now checked against the
  machine that runs the tests, and a band that does not hold is refused rather than narrowed.
  A run quietly reduced to ten domains reads as flaky timeouts, not as a host that needs
  configuring. The refusal names the safe subset so it can be taken on purpose with
  `MEDKIT_TEST_DOMAIN_BAND`. A machine that cannot report its range raises instead of assuming.
- **The stale-result sweep deleted one naming convention.** `colcon test-result` matches none:
  it parses every `.xml` under the build base. Deleting `*.xunit.xml` removed 6 files out of
  177 on a real build tree, because `ament_add_gtest` writes `<target>.gtest.xml`, so a full
  run followed by a single test still reported the previous totals. The sweep now asks colcon
  itself, which also keeps it away from the build inputs: about 230 `.xml` files here are node
  set definitions.

One limit is worth writing down, because it cannot be fixed here. A TCP bind is scoped to a
network namespace. Two containers each get their own, but multicast crosses Docker's default
bridge, so two containers can each hold the same domain and still see each other. This comes
from `domain_coordinator` and applies to upstream's `ament_add_ros_isolated_gtest` in the same
way. It is documented in the module and the design page.

### Shared lint configs now belong to a package

`ament_flake8` and `ament_clang_format` were given config paths like
`${CMAKE_CURRENT_SOURCE_DIR}/../../.clang-format`, which reach the repository root. That works
in a checkout. It cannot work in a binary package build, which unpacks one package. Since
Humble binarydeb runs tests, `flake8` failed there on every build, and nobody saw it because
bloom runs the tests as `dh_auto_test || true`.

The three configs are now files of `ros2_medkit_cmake`, installed next to its CMake modules.
The repository root paths are symlinks into that directory, so editors, pre-commit and
`scripts/clang-tidy-diff.sh` keep working and there is still only one copy of each file.
Resolution goes through `medkit_lint_config()`, which fails with a `FATAL_ERROR` when a config
is missing rather than skipping the check. That failure happens at configure time, which bloom
does not wrap in `|| true`.

A worse case of the same shape was fixed on the way: `.clang-tidy` was installed only if the
root file existed, so a released package shipped none at all. It was masked because
`ENABLE_CLANG_TIDY` defaults to OFF.

---

## Issue

- closes #595

---

## Type

- [x] Bug fix
- [ ] New feature or tests
- [ ] Breaking change
- [ ] Documentation only

---

## Testing

The original failure was reproduced first, on Jazzy with `rmw_cyclonedds_cpp` and domain 183,
whose discovery port is 53150:

- port free: node creation succeeds
- a plain UDP socket holding 53150: `ddsi_udp_create_conn: failed to bind to ANY:53150:
  address in use`, then `rmw_create_node: failed to create domain`, the same text as the issue

The port offsets a domain really binds were measured, not assumed, with three participants on
domain 97: CycloneDDS uses offsets 0, 1, 10, 11 and Fast-DDS uses 0, 10, 11. The code treats
the whole 250-port slice as owned, which is a superset of that and needs no assumption about
how many participants a test starts.

The allocator was then checked against the ways it could be wrong:

| check | result |
|---|---|
| 40 concurrent holders | 40 distinct domains, no duplicate |
| two independent process trees | distinct domains, which a `RESOURCE_LOCK` cannot do |
| a holder killed with `SIGKILL` | domain free immediately, next holder takes it |
| more holders than the band has domains | the excess waits, none doubles up, all served as domains free |
| an outsider holding a lock port | that domain is skipped, allocation still succeeds |
| every domain held | reported as a failure, never a fallback to domain 0 |
| band endpoints 1, 100, 215, 231 | usable |
| `MEDKIT_TEST_DOMAIN_BAND` set to `101`, `232`, `0`, `1-101` | refused, it can narrow the band and never widen it |

Cost is 0.02 ms per test.

Each claim was also broken on purpose to confirm the test that pins it goes red. Exporting one
domain to every test yields
`[isolated_a] CROSS-TALK: heard {'isolated_b': 14937} while holding domain 99 alone`, with the
control still green. Removing the reserved result file turns the runner test red. Removing the
band validation turns the override test red.

One mutation did not discriminate, which is worth writing down: removing the explicit release
in `hold_domains` left both release tests green, because the generator holding the socket is
collected when it goes out of scope and the socket closes anyway. Release is proven by the
`SIGKILL` test, across a real process boundary, not by those two.

### Checked against how the build farm runs the tests

We have no PR jobs on the farm, so the first real farm run would happen after merge. The farm's
own invocation was reproduced here instead. It is not the shape our CI uses.

| check | result |
|---|---|
| `colcon test --executor sequential`, no label filter, whole workspace, built without `--symlink-install` | 19 packages, exit 0, 24 min 37 s |
| tests waiting for a domain under serial execution | none, so nothing holds a domain past its own lifetime |
| the suite inside a `--net=host` container | 7/7 passed, peak 6 lock ports held |
| a real `bloom-generate rosdebian` plus `fakeroot debian/rules build` | all domain and lint config tests pass in the deb tree |
| the same deb tree with the lint config removed | configure fails and names the missing file, instead of skipping the check |
| a squatter holding 113 of the 117 lock ports | `could only hold 4 of 6 domains out of 117 candidates; this machine has almost no free lock ports` |

One result is worth naming. `test_external_app_fault_rollup` fails when it is the only test
running and passes under `-j`. It does that on this branch and on the commit before it, so
serial execution exposes it rather than causing it. It is a separate rosbag issue.

To verify: build the workspace and run `colcon test`. Every package registers
`test_dds_domain_allocation`, which reports any test that does not hold a domain of its own.

---

## Checklist


src/ros2_medkit_cmake/cmake/ROS2MedkitTestDomain.cmake
```diff
@@ -14,90 +14,268 @@
...
+  "ros2_medkit_sovd_service_interface:1-2"
+  "ros2_medkit_param_beacon:3-4"
+  "ros2_medkit_topic_beacon:5-6"
+  "ros2_medkit_graph_provider:7-8"
+  "ros2_medkit_fault_reporter:9-11"
+  "ros2_medkit_log_bridge:12-14"
+  "ros2_medkit_action_status_bridge:15-17"
+  "ros2_medkit_diagnostic_bridge:18-20"
+  "ros2_medkit_fault_manager:21-32"
+  "ros2_medkit_opcua:33-44"
```
mfaferek93: `test_opcua_secured.test.py:71` still hardcodes `ROS_DOMAIN_ID = '229'`, with a comment calling it the free slot in the old 220-229 range. After this move 229 is the first slot of the shared secondary pool that the multi-gateway tests take, so on a full run the secured test's gateway and a peer gateway can share a domain and see each other's `/fault_manager/report_fault`. `test_dds_domain_allocation` won't catch it because the domain is set inside the Python file, not as a CTest ENVIRONMENT entry. Handing it a domain via `medkit_reserve_test_domain` puts it back in the table.

src/ros2_medkit_cmake/scripts/check_test_domains.py
```diff
@@ -0,0 +1,185 @@
...
+        result[prop['name']] = value
+    return result
+
+
+def domain_of(env_entries):
+    """Return the ROS_DOMAIN_ID a test is given, or None."""
+    for entry in env_entries:
+        name, _, value = entry.partition('=')
+        if name == 'ROS_DOMAIN_ID':
+            return int(value)
```
mfaferek93: CTest applies ENVIRONMENT entries in order, so the last `ROS_DOMAIN_ID` wins, but this returns the first. If a test ever ends up with two entries the guard validates a domain the test does not actually run on. `for entry in reversed(env_entries)` fixes it.

src/ros2_medkit_cmake/cmake/ROS2MedkitTestDomain.cmake
```diff
@@ -14,90 +14,268 @@
...
+#
+# Usage:
+#   medkit_reserve_test_domain(_my_domain _my_lock)
+#
+macro(medkit_reserve_test_domain DOMAIN_VAR LOCK_VAR)
+  if(NOT DEFINED _MEDKIT_DOMAIN_POOL)
+    message(FATAL_ERROR "medkit_reserve_test_domain called before medkit_init_test_domains")
+  endif()
+  list(GET _MEDKIT_DOMAIN_POOL ${_MEDKIT_DOMAIN_NEXT} ${DOMAIN_VAR})
+  set(${LOCK_VAR} "medkit_dds_domain_${${DOMAIN_VAR}}")
```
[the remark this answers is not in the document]
bburda: Thanks for looking at this. That code is gone now. `medkit_set_test_domain` and the whole lock scheme around it were removed when the domain stopped being a CTest property: a test now takes its domain at run time and holds an OS-level lock, so there is no `RESOURCE_LOCK` to build a name for.

On the report itself, the guard that would have caught a missing lock was running and green, so I do not think the locks were empty. `DOMAIN_VAR` held a variable name there, not the domain, so the nested expansion was the intended indirection. Either way the file no longer has it.

src/ros2_medkit_integration_tests/CMakeLists.txt
```diff
@@ -129,22 +129,20 @@ if(BUILD_TESTING)
...
-  # medkit_add_launch_test wrapper because each needs a combined
-  # GATEWAY_TEST_PORT + ROS_DOMAIN_ID environment, assigned inline in the loop.
-  medkit_init_test_domains(START 130 END 219)
+  # This package holds more tests than the safe DDS band has domains, so
+  # medkit_set_test_domain wraps around its pool and two tests can end up on the
+  # same domain. That is safe because the same call also puts the domain's
+  # RESOURCE_LOCK on the test, and CTest never runs two tests holding one lock at
+  # the same time - a stale participant from an earlier test cannot leak into a
+  # later one's graph discovery, because the earlier one has finished.
+  # Multi-gateway tests take their EXTRA domains from the shared secondary pool,
```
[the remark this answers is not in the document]
bburda: You were right about the comment, and the trade-off it described is gone too. A killed test can leave orphaned nodes behind, and a `RESOURCE_LOCK` only stops two tests from running on one domain at the same time, so the comment was claiming more than the lock gives.

Domains are no longer reused across a pool at all. Each test holds its domain until it exits, and the next test can only take it once the lock is free, so a node left behind by a killed test cannot meet a later test.

src/ros2_medkit_cmake/cmake/ROS2MedkitTestDomain.cmake
bburda: You were right, and the collision you describe was reachable: 229 became the first slot of the shared secondary pool, and the guard could not see it because the domain was set inside the Python file instead of as a CTest `ENVIRONMENT` entry.

The literal is gone. The test is registered behind the domain wrapper and reads `ROS_DOMAIN_ID` from the environment, and it fails with a clear message when the variable is not set, so running it outside CTest reports itself instead of silently landing on domain 0. It passes that same domain to the gateway it starts.

src/ros2_medkit_cmake/scripts/check_test_domains.py
bburda: Agreed, and `reversed(env_entries)` is what went in. The helpers APPEND, so a caller can add its own entry on top and the guard would have validated a domain the test does not run on.

The fix outlived the code it was written for. The guard no longer checks `ROS_DOMAIN_ID`, because the domain is taken at run time and is not a CTest property any more, but the same reverse scan now reads `MEDKIT_TEST_DOMAINS` for tests that ask for several domains.

src/ros2_medkit_cmake/scripts/medkit_domain.py
```diff
@@ -0,0 +1,397 @@
...
+    if announce is None:
+        def announce(message):
+            print(f'[medkit-domain] {message}', flush=True)
+
+    selector = BandSelector(domains)
+    deadline = time.monotonic() + wait_timeout
+    managers = []
+    acquired = []
+    try:
+        for _ in range(count):
```
mfaferek93: Multi-domain acquisition takes the domains one at a time and holds every partial acquisition while waiting for the rest, so two callers asking for N each can each hold part of what the other needs. Two integration tests use `_MULTI_GATEWAY_DOMAINS 4` (CMakeLists 203 and 223), so under `ctest -j` on a contended band both can stall until the budget runs out and one fails. I reproduced it with a 4-domain band and 2 free: one trial in six ended with the winner at 10.2s and the loser at "no free ROS_DOMAIN_ID after waiting 10s". Acquiring all-or-nothing, or releasing what you hold before waiting, avoids it. Same fix would fix the message at :313, which says all N are in use while the process printing it holds several.

src/ros2_medkit_cmake/scripts/medkit_domain.py
```diff
@@ -0,0 +1,397 @@
...
+import time
+
+import domain_coordinator
+
+#: Where this file and its sibling scripts live, for callers that need to spawn
+#: one of them.
+SCRIPT_DIR = os.path.dirname(os.path.abspath(__file__))
+
+#: Every DDS domain a test may use. See the module docstring for the derivation.
+USABLE_DOMAINS = tuple(range(1, 101)) + tuple(range(215, 232))
```
mfaferek93, line 141: The band arithmetic is right for `32768 60999` (I checked both ends: 100 clears, 101 overlaps, 214 overlaps, 215 clears, 231 clears, 232 runs past 65535), but nothing reads the actual range - `ip_local_port_range` appears only in prose here, in the README and in the design doc. On an agent tuned to something like `10240 65535`, which is common on CI and K8s nodes, domains 12 to 231 overlap again and the multicast bind failure this band exists to prevent comes back as a flake, while code and docs both say the domain is safe. Reading the two integers at import and either deriving the band or failing loudly is cheap.

scripts/test.sh
```diff
@@ -44,6 +49,20 @@ shift 2>/dev/null || true
...
+#
+# `colcon test-result` at the end reports on whatever it finds under build/, and it
+# reads BOTH the ament xunit files and CTest's own Testing/<timestamp>/Test.xml
+# directories. A run that tests one package therefore reports the tallies of every
+# package tested since the last clean, and a preset that selects a handful of tests
+# still prints a total in the thousands. Deleting only the xunit files is not enough:
+# the CTest directories alone are enough to produce a total with no file on disk to
+# back it, which reads as a passing suite that never ran.
+if [ -d build ]; then
+  find build -name '*.xunit.xml' -delete
```
mfaferek93: The comment above is right that deleting only the xunit files is not enough, but this still only takes `*.xunit.xml`. ament_add_gtest writes `<target>.gtest.xml`, and colcon-test-result parses every `*.xml` under the build base, gtest's `<testsuites>` root included. On my build trees the split is 103 gtest vs 70 xunit, and 103 vs 2 in another, so most of the stale files survive. Run the full suite then a single test and the closing summary still reports the previous run's tallies. `find build -path '*test_results*' -name '*.xml' -delete` covers both.

src/ros2_medkit_cmake/scripts/medkit_domain.py
bburda: You are right, and the fix is all-or-nothing acquisition. `_try_take` now takes the whole set or releases everything it managed to take and returns holding nothing.

I picked that over releasing before the wait because it is one path instead of three, and there is no window between the last failed take and the release.

The deadlock becomes a livelock window, where two callers take and release in step. That window is bounded, and it is paid for twice: each attempt gets a fresh randomly started selector, and the backoff is jittered. A fixed interval is what keeps two callers of the same size in lockstep, so the jitter is not decoration.

The message at `:313` was wrong in the way you describe. It now says how many domains it could not take at once, and that it holds none of them while it waits.

The test starts two callers. Each asks for three domains from a band of four. Their first two takes alternate through a barrier, so the case happens on every run instead of one in six. With incremental acquisition put back, that test fails with `NoDomainAvailable`.

src/ros2_medkit_cmake/scripts/medkit_domain.py
bburda: You are right, and it is worse than a missing check. The commit this change replaces read `ip_local_port_range` in five places. I dropped it and kept only the comment that explains the band.

`verify_band()` now reads the file at run time and refuses when the band does not hold on that machine. It does not narrow the band by itself. A run silently reduced to ten domains under `ctest -j` would look like flaky timeouts, not like a host that needs configuring.

The refusal names the sysctl, the slices that overlap, and the subset that is still safe. That subset can be taken on purpose with `MEDKIT_TEST_DOMAIN_BAND`, which is accepted and tested, so that advice can be followed.

It runs in `hold_domains`, so every wrapped test is covered, and `check_test_domains.py` reports it once. That file now imports `medkit_domain` instead of keeping its own copy of the port arithmetic, which is how the old check drifted.

A machine that cannot report its range raises instead of assuming the default. `MEDKIT_TEST_EPHEMERAL_PORT_RANGE` declares it, and when the file is readable the two are merged to the wider one, so a declaration can only make the check stricter.

Checked with real ranges, not with a description. `10240 65535` and `1024 65535` are both refused, through `verify_band`, through `hold_domains`, and end to end through the wrapper, which must not start the command.

scripts/test.sh
bburda: You are right. On my build trees the split was 6 `.xunit.xml` against 171 `.gtest.xml`, so the sweep deleted 6 of 177 files.

The bigger problem is that `colcon test-result` matches no naming convention at all. `colcon_test_result.xunit` parses every `.xml` under the build base and counts the `testsuite` and `testsuites` roots. Any list of patterns we write will drift from it again.

So the sweep now asks colcon itself, with `--delete-yes`. It removes exactly the files colcon would have counted, and it leaves the build inputs alone. We have about 230 `.xml` files that are node set definitions, and a wider `find` would have deleted those. `Testing/` still goes separately, because the ctest side only offers the `TAG` file for deletion.

Measured on the real pipeline. Full unit run reports 4087 tests, then a single test reports 23. With the old sweep put back, the same single test reports 94, so 71 leaked from the previous run.

mfaferek93 (approved):

PR #607, opened 2026-08-11
fix: report every startup parameter that is coerced or refused

### LINKED ISSUE #603: [BUG] clamp_thread_count silently changes a mis-set thread count
# Bug report

`clamp_thread_count()` silently changes a mis-set thread count. The gateway then
runs with a different value than the one in the config file, and nothing in the
logs says so.

The helper is a single line in
`src/ros2_medkit_gateway/include/ros2_medkit_gateway/core/thread_pool_config.hpp`:

```cpp
inline std::size_t clamp_thread_count(int64_t requested, int64_t min_threads, int64_t max_threads) {
  return static_cast<std::size_t>(std::clamp<int64_t>(requested, min_threads, max_threads));
}
```

The comment above it says the clamp exists so a typo cannot drop the pool to zero
or spawn a pathological thread count. That is exactly the case where the operator
needs to be told. Right now the protection works and stays quiet.

The same applies to `clamp_keep_alive_timeout()` in the same header.

### Steps to reproduce
1. Set `server.http_thread_pool_size: 0` in the gateway params file.
2. Start the gateway.
3. Read the startup logs.

### Expected behavior

A warning naming both values, for example:
`server.http_thread_pool_size: requested 0, clamped to 1`.

### Actual behavior

Nothing is logged. The pool is silently sized to the minimum. The config file and
the running process disagree, and there is no way to see it from the outside.

### Environment
- ros2_medkit version: 0.6.0
- ROS 2 distro: all
- OS: all

### Additional information

Suggested fix: make the helper report whether it clamped (return a small struct,
or an out-parameter), and log a warning at each call site with the requested and
the effective value. Call sites are in `src/main.cpp` and `src/http/rest_server.cpp`.

This matters more as more parameters use the helper. A wrong value that is
absorbed without a trace is hard to diagnose in the field, because the symptom
(slow or jittery gateway) is far away from the cause (one line in a config file).

bburda:
# Pull Request

## Summary

A startup parameter set outside its documented range was coerced or replaced and nothing was logged. The config file and the running process then disagreed, with no way to see it from outside. The symptom of a wrong value is a slow or jittery gateway, or a limit that never fires, and it shows up far from the cause.

**Clamped parameters now report the clamp.** Both values are named, in the shape three parameters already used:

```
[WARN] [ros2_medkit_gateway]: sse.max_clients 4096 clamped to 1024
```

This covers `server.executor_threads`, `server.http_thread_pool_size`, `server.keep_alive_timeout_sec`, `sse.max_clients`, `aggregation.max_discovered_peers`, the four `subscription_executor.*` knobs and the five `data_provider.*` knobs.

**Three parameters were checked in the wrong order.** `server.port`, `refresh_interval_ms` and `sse.max_duration_sec` were narrowed from int64 to int and range-checked after that, so a value above INT_MAX wrapped back into the legal band and was taken as valid:

```
server.port:=4294976296   ->   listens on 9000, nothing logged
```

They are validated as int64 first now, and the warning names the refused value rather than the wrapped one.

**Three more wrapped to SIZE_MAX.** `sse.max_subscriptions`, `bulk_data.max_upload_size` and `aggregation.max_discovered_peers` were cast straight to `size_t`, so a negative value became a huge one and the documented limit stopped existing: no HTTP 503 at the subscription cap, no bound on uploads, no cap on discovered peers.

**`topic_sample_timeout_sec` is validated at the read.** Both consumers already fell back to `1.0` on their own, but `DataAccessManager` is in the ROS-neutral core and cannot log, so a value rejected there vanished. Their fallbacks stay as contract guards; the read site is the one that reports.

Two parameters are read a second time in `main.cpp`, only to compare the HTTP pool against the SSE plus cold-wait budget. That read keeps the silent clamp on purpose. One typo should not produce two identical warnings.

`clamp_thread_count` and `clamp_keep_alive_timeout` stay silent, because `gateway_core` links no rclcpp and cannot log. Their comments now say that reporting the clamp is the caller's job.

### Docs

`docs/config/server.rst` and `gateway_params.yaml` claimed every documented range is clamped to the nearest endpoint. That was wrong for about a third of the ranges in the same files, and `server.rst` contradicted itself 320 lines apart. Both now describe the three real behaviours: clamped to the nearest endpoint, refused in favour of the documented default, or rejected by rclcpp so the gateway does not start at all.

`max_parallel_topic_samples` was documented in two places but does not exist in the code. It is replaced by the real `data_provider.max_parallel_samples`. The nine `subscription_executor.*` and `data_provider.*` knobs had no rows in the config docs, so they get a table.

---

## Issue

- closes #603

---

## Type

- [x] Bug fix
- [ ] New feature or tests
- [ ] Breaking change
- [ ] Documentation only

---

## Testing

New integration test `test_startup_param_clamp_warnings` launches four gateways and reads their own process output:

- every clamped parameter below its floor. This gateway also has to serve `/health`, so a floor value that broke request serving would fail the test instead of passing quietly.
- every clamped parameter above its ceiling.
- every clamped parameter at a legal value, which must report nothing at all.
- the refused parameters, including three values above INT_MAX.

It also checks that a parameter read twice at startup is reported once.

To confirm the tests can fail, the changes were reverted with the tests kept:

- reverting the clamp warnings: 26 subtests failed in 0.5 s, each naming its parameter. The three parameters that already warned before this change kept passing.
- making one clamp site warn unconditionally: the in-range gateway test failed with `subscription_executor.max_queue_depth 256 clamped to 256`, which is what a correctly configured gateway would print.
- restoring the narrow-then-check order for `server.port`: the refused-value test failed.

Local runs, all green:

- gateway unit tests: 4032 tests, 0 failures
- integration suite: 878 tests, 0 failures
- sphinx, clang-format, clang-tidy on the changed files: clean

Not covered, and written in the test docstring: `server.executor_threads`, `server.http_thread_pool_size` and `entity_cache.capacity` are tested at the floor and in range, but not above their ceiling. A clamp means the gateway then runs at the ceiling, and a gateway launched past all three was measured at 1299 threads and 1486 MB RSS. The other 14 parameters cover the ceiling direction through the same compare-and-warn code.

---

## Checklist

### DIFF
--- a/docs/config/discovery-options.rst
+++ b/docs/config/discovery-options.rst
@@ -132,6 +132,31 @@ This is useful when running beacon plugins as the sole enrichment source
 When a layer is disabled, its entities are not discovered and its merge
 contributions are skipped. Plugins always run regardless of these flags.
 
+Manifest Loading
+^^^^^^^^^^^^^^^^
+
+.. list-table::
+   :header-rows: 1
+   :widths: 34 8 14 44
+
+   * - Parameter
+     - Type
+     - Default
+     - Description
+   * - ``discovery.manifest_strict_validation``
+     - bool
+     - ``true``
+     - Reject a manifest that fails validation instead of loading the parts of
+       it that happen to parse. Turn it off only to bring up a manifest that is
+       still being written.
+   * - ``discovery.manifest.fragments_dir``
+     - string
+     - ``""``
+     - Directory of manifest fragment YAML files. Every load and reload of the
+       manifest scans it and merges the apps, components and functions it finds
+       on top of the base manifest, so a fleet can ship one base file plus
+       per-robot fragments. Empty disables fragment loading.
+
 Field Groups
 ^^^^^^^^^^^^
 
--- a/docs/config/discovery-options.rst
+++ b/docs/config/discovery-options.rst
@@ -146,9 +146,12 @@ Manifest Loading
    * - ``discovery.manifest_strict_validation``
      - bool
      - ``true``
-     - Reject a manifest that fails validation instead of loading the parts of
-       it that happen to parse. Turn it off only to bring up a manifest that is
-       still being written.
+     - Treat manifest validation WARNINGS as fatal. Validation errors (broken
+       references, circular dependencies, duplicate bindings) always abort the
+       load whatever this is set to, and a manifest is only ever applied whole,
+       so turning this off never loads a partly-valid manifest - it only lets
+       warnings through. Turn it off to bring up a manifest that is still being
+       written.
    * - ``discovery.manifest.fragments_dir``
      - string
      - ``""``
--- a/docs/config/server.rst
+++ b/docs/config/server.rst
@@ -36,8 +36,9 @@ range and the gateway logs both values. This applies to the thread and pool knob
 for these, so the value is refused and the default takes its place. The warning
 names the value that was refused. This applies to ``server.port``,
 ``refresh_interval_ms``, ``discovery.refresh_debounce_ms``,
-``topic_sample_timeout_sec``, ``bulk_data.max_upload_size``,
-``sse.max_subscriptions`` and ``sse.max_duration_sec``:
+``topic_sample_timeout_sec``, ``fault_manager.service_timeout_sec``,
+``bulk_data.max_upload_size``, ``sse.max_subscriptions``,
+``sse.max_duration_sec`` and ``triggers.max_triggers``:
 
 .. code-block:: text
 
@@ -45,13 +46,19 @@ names the value that was refused. This applies to ``server.port``,
 
 **Refused at startup.** ``parameter_service_timeout_sec`` and
 ``parameter_service_negative_cache_sec`` are declared with a floating-point range
-descriptor, so rclcpp itself rejects the value and the gateway does not start:
+descriptor, so rclcpp itself rejects an out-of-range value and the gateway does
+not start:
 
 .. code-block:: text
 
    [ros2_medkit_gateway] Fatal exception in main: parameter
    'parameter_service_timeout_sec' could not be set: Parameter
    {parameter_service_timeout_sec} doesn't comply with floating point range.
+
+One value gets past that descriptor: ``.nan``. The range check compares with
+``<`` and ``>``, and every comparison against NaN is false, so rclcpp reads it as
+in-range. The gateway catches it when it reads the value, warns, and uses the
+default instead.
 
 In every case the point is the same. The config file and the running process must
 not disagree with nothing to show it, because the symptom of a mis-set value (a
@@ -352,7 +359,12 @@ Configure how the gateway connects to the fault manager services and event topic
    * - ``fault_manager.service_timeout_sec``
      - float
      - ``5.0``
-     - Timeout for fault manager service calls such as ``list_faults`` and ``get_snapshots``.
+     - Timeout for fault manager service calls such as ``list_faults`` and
+       ``get_snapshots``. Must be finite and greater than ``0``; anything else
+       is refused with a warning and ``5.0`` is used. Zero or negative would
+       make every fault call report "service not available" instead of waiting,
+       and a non-finite value would make it wait forever, holding an HTTP worker
+       for the life of the process.
    * - ``entity_freeze_frame.enabled``
      - bool
      - ``true``
@@ -664,7 +676,20 @@ Configure condition-based push notifications for resource changes.
      - int
      - ``1000``
      - Maximum number of concurrent triggers across all entities. Returns
-       HTTP 503 when this limit is reached.
+       HTTP 503 when this limit is reached. A value below ``1`` or past what an
+       ``int`` holds is refused with a warning and ``1000`` is used, because the
+       cap is compared as an unsigned count and a negative would remove it.
+   * - ``triggers.on_restart_behavior``
+     - string
+     - ``"reset"``
+     - Behavior on gateway restart for persistent triggers. ``"reset"`` clears
+       all triggers on restart. ``"restore"`` reloads persistent triggers from
+       the storage database.
+   * - ``triggers.storage.path``
+     - string
+     - ``""``
+     - Path to SQLite database for persistent trigger storage. When empty
+       (default), triggers are stored in-memory only and lost on restart.
 
 Fault-trigger rules
 ^^^^^^^^^^^^^^^^^^^
@@ -697,17 +722,6 @@ plugin is loaded; with no plugins these parameters do nothing.
      - ``""``
      - SQLite file the rules are persisted in. Empty keeps them in memory, so
        they are lost on restart.
-   * - ``triggers.on_restart_behavior``
-     - string
-     - ``"reset"``
-     - Behavior on gateway restart for persistent triggers. ``"reset"`` clears
-       all triggers on restart. ``"restore"`` reloads persistent triggers from
-       the storage database.
-   * - ``triggers.storage.path``
-     - string
-     - ``""``
-     - Path to SQLite database for persistent trigger storage. When empty
-       (default), triggers are stored in-memory only and lost on restart.
 
 Example:
 
--- a/docs/config/server.rst
+++ b/docs/config/server.rst
@@ -25,8 +25,8 @@ range and the gateway logs both values. This applies to the thread and pool knob
 (``server.executor_threads``, ``server.http_thread_pool_size``,
 ``server.keep_alive_timeout_sec``, ``sse.max_clients``,
 ``service_call_timeout_sec``, ``logs.buffer_size``, ``entity_cache.capacity``,
-``aggregation.max_discovered_peers``) and to the ``subscription_executor.*`` and
-``data_provider.*`` knobs:
+``aggregation.max_discovered_peers``, ``fault_triggers.poll_interval_ms``) and to
+the ``subscription_executor.*`` and ``data_provider.*`` knobs:
 
 .. code-block:: text
 
@@ -70,8 +70,9 @@ If the gateway behaves as though a parameter were different from what your YAML
 says, its startup log names the parameter.
 
 Note the logger name in the message. Most of these come from
-``ros2_medkit_gateway``, but the HTTP pool and keep-alive lines come from
-``rest_server``.
+``ros2_medkit_gateway``; the three that the REST server reads for itself -
+``server.http_thread_pool_size``, ``server.keep_alive_timeout_sec`` and
+``sse.max_duration_sec`` - come from ``rest_server``.
 
 Network Settings
 ----------------
@@ -276,8 +277,10 @@ Data Access Settings
 
 .. note::
 
-   The defaults listed above match the recommended gateway configuration in
-   ``src/ros2_medkit_gateway/config/gateway_params.yaml``. If
+   The defaults listed above are the ones the code declares. Where the shipped
+   ``src/ros2_medkit_gateway/config/gateway_params.yaml`` sets a different value
+   the row says so - ``topic_sample_timeout_sec`` is the one that differs today.
+   If
    ``topic_sample_timeout_sec`` is out of range, ``DataAccessManager`` and the
    topic transport both fall back to ``1.0``.
 
@@ -592,7 +595,7 @@ configuration.
    * - ``bulk_data.max_upload_size``
      - int
      - ``104857600``
-     - Maximum upload file size in bytes (default: 100 MB). Set to 0 for unlimited.
+     - Maximum upload file size in bytes (default: 100 MB). Set to ``0`` for unlimited. A negative value is refused with a warning and the default is used: it would otherwise wrap to a huge unsigned size, which is the same "no limit" a typo should never buy silently.
    * - ``bulk_data.categories``
      - string[]
      - ``[]``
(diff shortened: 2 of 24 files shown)


src/ros2_medkit_gateway/src/gateway_node.cpp
```diff
@@ -650,15 +693,30 @@ GatewayNode::GatewayNode(const rclcpp::NodeOptions & options) : Node("ros2_medki
...
+  // huge size_t, and the documented "HTTP 503 when this limit is reached" then
+  // never fires. There is no documented ceiling for this knob, so only the
+  // nonsensical low end is refused.
+  const int64_t max_subscriptions_raw = get_parameter("sse.max_subscriptions").as_int();
+  size_t max_subscriptions = 100;
+  if (max_subscriptions_raw < 1) {
+    RCLCPP_WARN(get_logger(), "sse.max_subscriptions %" PRId64 " must be >= 1. Using default 100.",
+                max_subscriptions_raw);
+  } else {
+    max_subscriptions = static_cast<size_t>(max_subscriptions_raw);
```
mfaferek93, line 720: Good fix. The same hole is still open one subsystem over, and it is the identical shape.

`triggers.max_triggers` at :997 does `static_cast<int>(...as_int())` with no check, and `core/managers/trigger_manager.cpp:350` then compares `triggers_.size() >= static_cast<size_t>(config_.max_triggers)`. So `triggers.max_triggers:=-1` gives 18446744073709551615, the cap never fires, and triggers accumulate without bound over the REST API. Startup logs `Trigger subsystem: enabled (max=-1, ...)` and nothing else.

The only reason the sweep missed it is that the `size_t` cast lives in the consumer rather than at the read site. Worth a grep for the same pattern: read site narrows to `int`, consumer casts to `size_t`.

src/ros2_medkit_gateway/src/gateway_node.cpp
```diff
@@ -593,7 +604,23 @@ GatewayNode::GatewayNode(const rclcpp::NodeOptions & options) : Node("ros2_medki
...
+  // contract guards; this is the one place that reports.
+  const auto raw_topic_sample_timeout_sec = get_parameter("topic_sample_timeout_sec").as_double();
+  auto topic_sample_timeout_sec = raw_topic_sample_timeout_sec;
+  // isfinite first, and the range as a positive test. Do NOT rewrite this as
+  // "below 0.1 or above 30.0": every comparison against NaN is false, so that
+  // form hands NaN to both consumers as a sampling timeout.
+  const bool timeout_in_range =
+      std::isfinite(topic_sample_timeout_sec) && topic_sample_timeout_sec >= 0.1 && topic_sample_timeout_sec <= 30.0;
+  if (!timeout_in_range) {
+    RCLCPP_WARN(get_logger(), "topic_sample_timeout_sec %.2f is outside 0.1-30.0. Using default 1.0.",
```
mfaferek93, line 620: The comment above this is the most useful thing in the PR, and it also points at what got missed: three other doubles still have the negated form or no check at all.

`fault_manager.service_timeout_sec` (declared :98, default 5.0) is never validated. `ros2/transports/ros2_fault_service_transport.cpp:50` does `std::max(timeout.count(), 0.0)` - and `std::max` returns `(a < b) ? b : a`, so with `a` = NaN the comparison is false and NaN comes straight back out. `wait_for_service(duration<double>(NaN))` then duration_casts NaN to nanoseconds (UB; x86-64 `cvttsd2si` gives INT64_MIN), and `wait_for_service_nanoseconds` treats any non-positive timeout as wait-forever. With the fault manager down the first `GET /faults` pins a cpp-httplib worker permanently, and six of them exhaust the default pool.

`.nan` is a plain YAML value here - rcl's parser has an explicit case for it. `-1.0` is the quiet version: coerced to 0.0 at :50, after which every fault RPC returns "service not available" with nothing logged.

Also worth a look: `docs/config/server.rst:46-48` says the `FloatingPointRange` descriptor makes rclcpp reject out-of-range values on the two `parameter_service_*` floats. True for ordinary values, but rclcpp's range check compares with `<` and `>` and returns success when `step == 0.0`, and every comparison against NaN is false - so NaN should slip through there too. I could not verify that against rclcpp source (not installed here), so treat it as worth checking rather than established.

src/ros2_medkit_gateway/src/http/rest_server.cpp
```diff
@@ -155,10 +167,17 @@ RESTServer::RESTServer(GatewayNode * node, const std::string & host, int port, c
...
-    RCLCPP_WARN(node_->get_logger(), "sse.max_duration_sec must be > 0, using default 3600");
-    max_duration_sec = 3600;
+  // Validated as int64 before narrowing: narrowing first wraps a value above
+  // INT_MAX into a small positive one that passes the check, so 4294967300
+  // would silently cap every subscription at 4 seconds. The warning names the
+  // requested value, otherwise the operator cannot tell which value was refused.
+  const int64_t max_duration_raw = node_->get_parameter("sse.max_duration_sec").as_int();
+  int max_duration_sec = 3600;
+  if (max_duration_raw < 1 || max_duration_raw > std::numeric_limits<int>::max()) {
+    RCLCPP_WARN(node_->get_logger(), "sse.max_duration_sec %" PRId64 " must be > 0. Using default 3600.",
```
mfaferek93: The value is named, which is the point of the PR, but the reason is wrong. The branch is `< 1 || > INT_MAX`, and the message says `must be > 0`.

`sse.max_duration_sec:=4294967300` - the exact value this branch was added for - prints `sse.max_duration_sec 4294967300 must be > 0`. The operator is told a number that plainly is greater than zero must be greater than zero, and never learns the real constraint. `must be in [1, 2147483647]` would do it.

Worth fixing before merge because `test_startup_param_clamp_warnings.test.py:158-159` asserts that fragment verbatim, so the wording is now pinned. The sibling sites you wrote get it right: `gateway_node.cpp:676` says "is negative. Must be >= 0".

Minor, same line: this logs to `node_->get_logger()` while the two clamps at :104 and :110 use `rclcpp::get_logger("rest_server")`. Three coercion reports from one constructor under two logger names, and `docs/config/server.rst:33` shows operators `[rest_server]:` for exactly this class of message.

src/ros2_medkit_integration_tests/test/features/test_startup_param_clamp_warnings.test.py
```diff
@@ -0,0 +1,402 @@
...
+REJECTED_PARAMS = [
+    ('server.port', 4294976296,
+     'Invalid port 4294976296. Must be between 1024-65535. Using default 8080.'),
+    ('refresh_interval_ms', 4294967396,
+     'Invalid backstop refresh interval 4294967396ms'),
+    ('sse.max_duration_sec', 4294967300,
+     'sse.max_duration_sec 4294967300 must be > 0'),
+    ('discovery.refresh_debounce_ms', 70000,
+     'Invalid discovery.refresh_debounce_ms 70000ms'),
+    ('topic_sample_timeout_sec', 60.0,
```
mfaferek93: `60.0` is the only value fed here, and the old form `t < 0.1 || t > 30.0` rejects it identically with the same message. So revert gateway_node.cpp:617-618 to the old comparison and this whole file stays green while NaN reaches `Ros2TopicTransport` and `DataAccessManager` as a sampling timeout.

That is the headline fix of the head commit, and right now the only thing guarding it is the code comment at :614-616 saying clang-tidy suggests exactly that rewrite.

One caveat before adding a NaN row here: I am not sure NaN survives the launch path. launch_ros serialises params to YAML, PyYAML writes `float('nan')` as `.nan`, and `rcl_yaml_param_parser` resolves doubles with `strtod`, which does not accept it. Unverified either way. If it does not round-trip, lifting the range test into a free function in a header and unit-testing NaN/inf/0.05/60.0 there is cheaper - same shape `clamp_thread_count` already has.

src/ros2_medkit_action_status_bridge/src/action_status_bridge_node.cpp
```diff
@@ -105,11 +107,14 @@ ActionStatusBridgeNode::~ActionStatusBridgeNode() {
 }
 
 void ActionStatusBridgeNode::load_parameters() {
-  const int aborted_severity =
-      static_cast<int>(declare_parameter<int>("aborted_severity", ros2_medkit_msgs::msg::Fault::SEVERITY_ERROR));
+  // Read as int64 and checked before narrowing. A ROS parameter is an int64, so
+  // declaring this as int narrowed first and a value above INT_MAX wrapped back
+  // into [0,3] and was accepted in silence: 4294967298 becomes 2.
+  const int64_t aborted_severity =
```
mfaferek93, line 113: These two fixes have no test anywhere. `grep -rn 'aborted_severity\|dedup_capacity'` across every test file in the repo returns nothing.

`4294967298` used to narrow to `2` and pass silently as ERROR, which is exactly the class the PR closes elsewhere and pins in the new integration test. Two of the five fixed sites are unpinned, so a regression here is invisible. Even one row in the existing table would cover it.

docs/config/discovery-options.rst
```diff
@@ -132,6 +132,31 @@ This is useful when running beacon plugins as the sole enrichment source
...
+
+.. list-table::
+   :header-rows: 1
+   :widths: 34 8 14 44
+
+   * - Parameter
+     - Type
+     - Default
+     - Description
+   * - ``discovery.manifest_strict_validation``
```
mfaferek93, line 146: This describes something the flag does not do.

`discovery/manifest/manifest_manager.cpp:72-78` fails on validation errors unconditionally, with `strict` false as well. And `manifest_ = std::move(loaded)` at :89 only runs after every check passes, so "loading the parts that happen to parse" never happens in either mode. Same shape in `load_manifest_from_string`.

The flag's only effect is whether validation *warnings* are fatal (:84-87). Second sentence about bringing up a manifest that is still being written is right; the first one is what needs rewording.

docs/config/server.rst
```diff
@@ -547,6 +665,38 @@ Configure condition-based push notifications for resource changes.
...
+     - How often the rules are evaluated. Floored at ``50`` ms so the loop
+       cannot busy-spin; a lower value is clamped with a warning. Read only when
+       the engine actually starts, so with no plugin loaded a mis-set value is
+       neither applied nor reported - it does nothing either way.
+   * - ``fault_triggers.storage.path``
+     - string
+     - ``""``
+     - SQLite file the rules are persisted in. Empty keeps them in memory, so
+       they are lost on restart.
    * - ``triggers.on_restart_behavior``
```
mfaferek93: These two rows belong to the table above. The new `Fault-trigger rules` heading at :669 was spliced between `triggers.max_triggers` (:663) and the remaining `triggers.*` rows, so `triggers.on_restart_behavior` here and `triggers.storage.path` at :706 are now rows of the `fault_triggers` list-table, sitting right under `fault_triggers.storage.path`.

Two different subsystems - `gateway_node.cpp:213-216` declares `triggers.*` for TriggerManager, :219-221 declares `fault_triggers.*` for the rule engine. The Triggers table is left with two of its four rows, and the reference now reads as if `triggers.storage.path` and `fault_triggers.storage.path` were siblings.

src/ros2_medkit_gateway/config/gateway_params.yaml
```diff
@@ -5,6 +5,15 @@
...
 #   - Launch files: parameters=[...]
 #   - This YAML file
+#
+# Ranges: where a parameter below documents a valid range, that range is
+# enforced when the value is read. An out-of-range value does not abort startup
+# and does not silently take effect either - it is coerced to the nearest
+# endpoint and the gateway logs a warning naming both the requested and the
+# effective value, e.g.
+#   [WARN] server.http_thread_pool_size 0 clamped to 1
+# So if the gateway behaves as though a parameter were different from what this
```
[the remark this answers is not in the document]
bburda: Changed the example to show the logger, so it matches what the log really prints. I also removed the parameter list from this header. The same list lived here and in the config page and the two had already drifted, so now only the config page carries it.

src/ros2_medkit_gateway/src/http/rest_server.cpp
```diff
@@ -94,10 +95,20 @@ RESTServer::RESTServer(GatewayNode * node, const std::string & host, int port, c
...
+  const auto http_thread_pool_size = clamp_thread_count(requested_pool_size, 1, 1024);
+  if (static_cast<int64_t>(http_thread_pool_size) != requested_pool_size) {
+    RCLCPP_WARN(rclcpp::get_logger("rest_server"), "server.http_thread_pool_size %" PRId64 " clamped to %zu",
+                requested_pool_size, http_thread_pool_size);
+  }
+  const auto requested_keep_alive = node_->get_parameter("server.keep_alive_timeout_sec").as_int();
+  const auto keep_alive_timeout_sec = clamp_keep_alive_timeout(requested_keep_alive, 1, 3600);
+  if (static_cast<int64_t>(keep_alive_timeout_sec) != requested_keep_alive) {
+    RCLCPP_WARN(rclcpp::get_logger("rest_server"), "server.keep_alive_timeout_sec %" PRId64 " clamped to %ld",
+                requested_keep_alive, static_cast<long>(keep_alive_timeout_sec));
```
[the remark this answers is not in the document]
bburda: Right, `std::time_t` is not guaranteed to fit in `long`. Both the warning and the info line now cast to `int64_t` and use `PRId64`.

src/ros2_medkit_gateway/src/gateway_node.cpp
bburda: Confirmed and fixed. It is read as `int64` and checked before narrowing now, refused outside `[1, 2147483647]` with a warning that names the value, falling back to 1000. The test covers both ends: `-1` and a value above `INT_MAX`. I also grepped for the shape you described and found the same thing in the log bridge, so that one is in this branch too.

src/ros2_medkit_gateway/src/gateway_node.cpp
bburda: Both parts confirmed. `fault_manager.service_timeout_sec` is now `isfinite` plus a positive range test, with a warning. The test uses `+inf`, not NaN, because NaN already fails `> 0` on its own while `+inf` passes it and is caught only by `isfinite`.

On the descriptor question you marked as unverified: I checked it on a running gateway. Jazzy accepts NaN through `FloatingPointRange` and starts with `timeout=nans`; Lyrical refuses it and the process exits 1. So the finiteness check is added at the read for the distros that let it through, and the docs now describe the split instead of stating one distro's behaviour as universal.

src/ros2_medkit_gateway/src/http/rest_server.cpp
bburda: Fixed. The message now says `must be in [1, 2147483647]`, and the test fragment was updated with it. The warning also moved to the `rest_server` logger, and the note in the config page that tells operators which logger to grep now lists all three lines that come from it.

src/ros2_medkit_integration_tests/test/features/test_startup_param_clamp_warnings.test.py
bburda: You were right that `60.0` left the fix unpinned. NaN does survive the launch path: I passed `float('nan')` through `launch_ros` and the gateway logged it, so no unit-test detour was needed. NaN is a row now, and the ordinary out-of-range value sits on a different gateway, so the finiteness guard and the range are pinned separately. I also added a value below `0.1`, which nothing was testing before.

src/ros2_medkit_action_status_bridge/src/action_status_bridge_node.cpp
bburda: Added. The integration test launches a second bridge with both values above `INT_MAX` and asserts each warning names the value that was supplied. Reverting either fix fails it, and the log then shows `aborted_severity=2`, which is the wrap you described.

docs/config/discovery-options.rst
bburda: Correct, the row described something the flag does not do. Validation errors abort the load whatever it is set to, and the manifest is only ever applied whole, so the flag only decides whether warnings are fatal. The row says that now.

docs/config/server.rst
bburda: Fixed. `triggers.on_restart_behavior` and `triggers.storage.path` are back in the Triggers table, and the fault-trigger section now starts after that table and its example.

mfaferek93 (approved):

PR #608, opened 2026-08-11
Do not register launch tests in a system package build, and let the macro own their properties

### LINKED ISSUE #602: [BUG] All 70 launch tests fail in the Humble binary package build
# Bug report

All 70 launch tests of `ros2_medkit_integration_tests` fail in the Humble binary package build on the ROS build farm. They fail on every build. The package is still published, because bloom runs the tests as `dh_auto_test || true`, so the result is ignored.

Only Humble is affected. Its `release-build.yaml` does not set `package_dependency_behavior`, so `run_package_tests` stays at its default value `true`. Jazzy and Lyrical set it to `false`. Those two build with `-Pnocheck` and `-DBUILD_TESTING=OFF` and run no tests in that job.

### Steps to reproduce
1. Open the Humble binarydeb job for `ros2_medkit_integration_tests`:
   https://build.ros2.org/job/Hbin_uJ64__ros2_medkit_integration_tests__ubuntu_jammy_amd64__binary/lastBuild/consoleText
2. Search the console output for `No module named`.
3. Search for `tests passed`.

### Expected behavior

The launch tests run and pass, like they do under `colcon test`.

### Actual behavior

Every launch test fails in under a second:

```
0% tests passed, 70 tests failed out of 70
```

They all fail the same way. The log has 140 lines like this:

```
launch_test.py: error: No module named 'ros2_medkit_test_utils'
```

The deb is built and published anyway. This is the reason:

```
override_dh_auto_test:
	dh_auto_test || true
```

### Possible cause

This part is a guess. I did not reproduce it.

`ros2_medkit_test_utils` is a Python package inside `ros2_medkit_integration_tests`. It is installed by `ament_python_install_package(ros2_medkit_test_utils)` in `src/ros2_medkit_integration_tests/CMakeLists.txt`. The install itself works. The farm log shows `ros2_medkit_test_utils-0.6.0-py3.10.egg-info` being created.

In the package build the tests run through `make test`. That happens before the install step. At that point nothing adds the package directory to `PYTHONPATH`. Under `colcon test` the environment hooks do that. That would explain why the same tests pass locally.

### Environment
- ros2_medkit version: 0.6.0-1
- ROS 2 distro: Humble. Jazzy and Lyrical run no tests in this job.
- OS: Ubuntu 22.04 jammy, amd64, in the build farm container

### Additional information

There is a second problem of the same kind in `ros2_medkit_cmake`. It calls `ament_flake8` with a config file path that points outside the package. A binary package build unpacks only that one package, so the path cannot resolve there. The same job type hides it the same way.

One question to answer with this issue: do we want the tests in the Humble binary package jobs to work, or do we want to switch them off like Jazzy and Lyrical do? Right now they run and always fail, and nobody sees it.

bburda:
# Pull Request

## Summary

A launch test resolves its own package through the ament index, which only an installed prefix carries. A system package build runs its test step before the install step, so every launch test in this repo failed in the Humble binarydeb job with `package '<name>' not found, searching: ['/opt/ros/humble']`. Humble is the only supported distro whose binarydeb job still runs package tests; every distro from Iron onward disables them. Because bloom runs that step as `dh_auto_test || true`, the failures could not stop the build, so what shipped was a published log full of errors that said nothing about the package.

The reported cause was incomplete. Adding `ros2_medkit_test_utils` to `PYTHONPATH` moves the failure from the import to `PackageNotFoundError`, because the launch tests also resolve their own executables through the ament index. The suite structurally needs an install space, so it is not registered where there is none.

Skipping registration alone is not enough either, and that is why the macro changed. Callers named the test afterwards to set labels and environment, and `set_tests_properties` on a test that does not exist is a hard CMake error, not a no-op. That would have replaced a noisy log with a failed configure, in five packages. `medkit_add_launch_test` now takes `LABELS`, `ENV` and `ARGS` and applies them itself, and every call site across seven packages moved onto it. This also removes a trap the tree documented in several comments: a plain `set_tests_properties(... ENVIRONMENT ...)` after the macro dropped the domain the macro had set, leaving the test on domain 0.

The guard needs two signals, because either alone has a false positive:

- the install prefix is the ROS distribution root, which also matches `colcon build --install-base /opt/ros/<distro>`, and
- `CMAKE_INSTALL_LOCALSTATEDIR` is `/var`, which debhelper always passes and no colcon build does. The value matters: `include(GNUInstallDirs)` defines the same variable as a relative `var`.

gtests and pytest tests are deliberately untouched. That chroot is the only place these packages are exercised against a minimal dependency closure, and it has already earned its keep: it exposed a missing runtime dependency in the fault manager, which is fixed separately.

---

## Issue

- closes #602

---

## Type

- [x] Bug fix
- [x] New feature or tests
- [ ] Breaking change
- [ ] Documentation only

---

## Testing

**The guard is proven by running a real cmake configure, not by matching text.** A test configures every package that registers launch tests, with the distribution prefix and `CMAKE_INSTALL_LOCALSTATEDIR=/var`, and requires the configure to succeed, the guard's message to appear, no launch test to be registered and at least one non-launch test to remain. Injecting a property call on a launch test by name makes it fail, including in the mixed-case spelling `Set_Tests_Properties(...)`, which is valid CMake.

`ros2_medkit_graph_watchdog` cannot be a dependency of the integration test package without a cycle, so it runs the same check in its own suite rather than being covered by accident. Neither test can skip: registration is decided in CMake, and where the test is registered it asserts unconditionally.

**The migration is behaviour-preserving.** `ros2_medkit_integration_tests` registers the same 91 tests before and after, with the same labels, the same `GATEWAY_TEST_PORT` values from 9100 to 9970 in steps of 10, and the four per-test timeout overrides unchanged at 180, 300, 300 and 300 seconds.

**Distros.** The probe projects declare a language, because a project with no language enabled reaches deprecated `FindPythonLibs` through `python_cmake_module` on Humble, which this repo documents in `ros2_medkit_cmake/design/index.rst`. That is verified by the Humble job, not locally.

**What this cannot prove.** That the binarydeb log is actually clean. The next Humble binarydeb run after a release is the check; everything here shows that in a build configured the same way, the configure succeeds and no launch test is registered.

---

## Note for reviewers

`medkit_add_launch_test` callers must now pass `LABELS`, `ENV` and `ARGS` to the macro. Naming a launch test afterwards to set properties is not supported, and a test will fail the build if it reappears.

---

## Checklist

### DIFF
--- a/src/ros2_medkit_cmake/README.md
+++ b/src/ros2_medkit_cmake/README.md
@@ -12,7 +12,7 @@ build acceleration, and centralized linting configuration across all packages.
 | `ROS2MedkitCoverage.cmake` | gcov/lcov instrumentation behind `-DENABLE_COVERAGE=ON` |
 | `ROS2MedkitLinting.cmake` | Shared lint configs via `medkit_lint_config()`, plus the opt-in clang-tidy gate (CI runs `run-clang-tidy` instead) |
 | `ROS2MedkitSanitizers.cmake` | ASan / TSan / UBSan behind `-DSANITIZER=asan,ubsan` |
-| `ROS2MedkitTestDomain.cmake` | Per-package `ROS_DOMAIN_ID` allocation for test isolation |
+| `ROS2MedkitTestDomain.cmake` | Per-test `ROS_DOMAIN_ID` allocation for test isolation - an OS-level socket lock taken at test start and released when the process ends, not a per-package table |
 | `ROS2MedkitWarnings.cmake` | Shared warning flags and vendored-code relaxations |
 
 ### ROS2MedkitCompat
@@ -196,6 +196,103 @@ For a test that builds its own command line and so cannot take an ament test run
 `medkit_add_wrapped_test(<name> [DOMAINS <n>] COMMAND <cmd...>)`, which puts
 `medkit_run_with_domain.py` in front of the command instead.
 
+#### A launch test's properties belong to the call that registers it
+
+`medkit_add_launch_test(<name> <file> [TIMEOUT n] [DOMAINS n] [LABELS "a;b"] [ENV "K=V" ...]
+[ARGS "foo:=bar" ...])` is the only supported way to give a launch test a label or an
+environment variable:
+
+```cmake
+medkit_add_launch_test(test_multi test/test_multi.test.py TIMEOUT 300 DOMAINS 4
+  LABELS "integration" ENV "FOO=bar" "BAZ=qux" ARGS "foo:=bar")
+```
+
+Setting them afterwards - `set_tests_properties(test_multi PROPERTIES LABELS
+...)` or `set_property(TEST test_multi APPEND PROPERTY ENVIRONMENT ...)` - is
+not supported, and not merely discouraged: see the next section for why the
+build can register no launch test at all, and a property call that names a
+test CMake never registered is a hard configure error, not a no-op. `ENV` is
+applied after the domain the macro already set for the test, by appending, so
+a caller's own environment can never overwrite `MEDKIT_TEST_DOMAINS` - an
+`ENV` entry that sets the literal key `MEDKIT_TEST_DOMAINS=` fails the
+configure with `FATAL_ERROR` naming `DOMAINS` as the way to set it, rather
+than silently landing behind the macro's own entry and winning at run time
+the way a plain `set_tests_properties(... PROPERTIES ENVIRONMENT ...)` used
+to. This is a literal string match, not a generator-expression evaluation:
+it does not chase down a value spelled to avoid the literal key (e.g.
+`ENV "$<1:MEDKIT_TEST_DOMAINS=99>"`), which nobody writes by accident and
+which a blanket rejection of `$<` would also catch legitimate uses of.
+`LABELS` overrides `add_launch_test()`'s own default of `"launch_test"`, the same
+override callers used to apply by hand. `ARGS` forwards extra launch
+arguments straight to `add_launch_test()`, same as it always did - it has its
+own keyword rather than falling through as an unrecognised argument, because
+`LABELS` being multi-value (to survive `ENV`'s/`LABELS`'s own semicolon-split
+values, see the `.cmake` comment) would otherwise absorb an `ARGS` that had
+no keyword of its own instead of erroring on it.
+
+This does not reach `medkit_add_gtest`, `medkit_add_gmock`,
+`medkit_add_pytest_test` or `medkit_add_wrapped_test`: those are always
+registered, so `MEDKIT_TEST_DOMAINS` is appended to their `ENVIRONMENT` and a
+caller adding entries of its own still uses
+`set_property(TEST ... APPEND PROPERTY ENVIRONMENT ...)` afterwards, same as
+before. A plain `set_tests_properties(... PROPERTIES ENVIRONMENT ...)` drops
+the domain there too, and the check below says so.
+
+#### Launch tests are not registered in a system package build
+
+`medkit_add_launch_test()` registers nothing when two conditions both hold -
+the install prefix is the ROS distribution root, AND
+`CMAKE_INSTALL_LOCALSTATEDIR` is the absolute path `/var` - and prints why
+once per directory instead:
+
+```
+ros2_medkit: launch tests are not registered in this build. Installing into
+/opt/ros/humble is a system package build, whose test step runs before the
+install step, so a launch test cannot resolve this package through the ament
+index. gtests and pytest tests still run.
+```
+
+A launch test resolves the executables and the Python helper package of the
+package under test through the ament index, which only an installed prefix
+carries. Installing straight into the ROS distribution root is what a system
+package build does - the ROS build farm's binarydeb job - and that job runs
+its test step *before* the install step, so the package it is testing is not
+yet on `AMENT_PREFIX_PATH`. Every launch test then fails on resolution rather
+than on anything the test set out to check. Humble is the only distro whose
+binarydeb job still runs package tests at all; every distro from Iron onward
+disabled that step (`ros2/ros_buildfarm_config` PR 298). Because bloom runs it
+as `dh_auto_test || true`, the failures cannot stop the build, so what used to
+reach the build farm's log was every launch test failing for a reason that
+said nothing about the package.
+
+The install prefix alone is not a safe signal: `colcon build --install-base
+/opt/ros/<distro>` also installs straight into the distribution root, but its
+test step runs *after* its own install step, so its launch tests work fine.
+`CMAKE_INSTALL_LOCALSTATEDIR` is the second signal that tells the two apart -
+debhelper's `cmake` buildsystem always passes it
+(`-DCMAKE_INSTALL_LOCALSTATEDIR=/var`), and no `colcon build` ever does. The
+check is on the *value*, not merely `DEFINED`: a package that
+`include(GNUInstallDirs)`s defines `CMAKE_INSTALL_LOCALSTATEDIR` itself, to
+the relative `var` (no leading slash), which is a different value from what
+debhelper passes - `DEFINED` alone cannot tell those apart, so the guard
+requires the absolute `/var`. No in-tree package includes `GNUInstallDirs`
+today, so this is hardening against a real default rather than a fix for an
+observed failure. Both comparisons (the prefix and `CMAKE_INSTALL_LOCALSTATEDIR`)
+are exact (`STREQUAL`), so a trailing slash is stripped from each before
+comparing rather than trusted to match bloom's spelling.
+
+The "once per directory" of the printed message is a macro-scoped CMake
+variable, not a `CACHE` entry: a package that calls `medkit_add_launch_test`
+many times in one `CMakeLists.txt` sees the explanation once, and a different
+`CMakeLists.txt` in the same build - a different package, or a second
+`add_subdirectory()` - prints its own.
+
+gtests and pytest tests are untouched on purpose. That chroot is the only
+place our packages are exercised against a minimal dependency closure, and it
+is where the missing rosbag2 sqlite3 storage plugin was found - taking that
+coverage away along with the launch tests would trade one blind spot for
+another.
+
 Only domains **1-100 and 215-231** are drawn from. RTPS gives a domain the UDP slice
 `[7400 + 250 * d, 7400 + 250 * d + 249]`, and the kernel hands out ephemeral ports from
 `net.ipv4.ip_local_port_range` (32768-60999 by default), which covers domains 101-214. If an
@@ -237,7 +334,13 @@ part of what the other needs is a deadlock that only ends when both wait budgets
 `MEDKIT_TEST_DOMAINS` is appended to the test's `ENVIRONMENT`, so a caller adding its own
 entries must use `set_property(TEST ... APPEND PROPERTY ENVIRONMENT ...)`. A plain
 `set_tests_properties(... PROPERTIES ENVIRONMENT ...)` afterwards drops it, and the check
-below says so.
+below says so. This is the general rule for `medkit_add_gtest`, `medkit_add_gmock`,
+`medkit_add_pytest_test` and `medkit_add_wrapped_test` - all four are always registered, so a
+call naming the test afterwards is safe, just easy to get wrong in the APPEND direction.
+**A launch test is the one exception**: it is not always registered (see "Launch tests are not
+registered in a system package build" above), so a `set_tests_properties` or `set_property`
+call naming it afterwards is not a subtler bug to get right, it is unsupported - use
+`medkit_add_launch_test`'s own `LABELS` and `ENV` parameters instead.
 
 The failure this scheme has to guard against is silent: a test registered with plain
 `ament_add_gtest` or `add_launch_test` does not fail, it runs on domain 0 and sees every node
--- a/src/ros2_medkit_cmake/README.md
+++ b/src/ros2_medkit_cmake/README.md
@@ -213,15 +213,18 @@ not supported, and not merely discouraged: see the next section for why the
 build can register no launch test at all, and a property call that names a
 test CMake never registered is a hard configure error, not a no-op. `ENV` is
 applied after the domain the macro already set for the test, by appending, so
-a caller's own environment can never overwrite `MEDKIT_TEST_DOMAINS` - an
-`ENV` entry that sets the literal key `MEDKIT_TEST_DOMAINS=` fails the
-configure with `FATAL_ERROR` naming `DOMAINS` as the way to set it, rather
-than silently landing behind the macro's own entry and winning at run time
-the way a plain `set_tests_properties(... PROPERTIES ENVIRONMENT ...)` used
-to. This is a literal string match, not a generator-expression evaluation:
-it does not chase down a value spelled to avoid the literal key (e.g.
-`ENV "$<1:MEDKIT_TEST_DOMAINS=99>"`), which nobody writes by accident and
-which a blanket rejection of `$<` would also catch legitimate uses of.
+a caller spelling the literal key `MEDKIT_TEST_DOMAINS=` cannot overwrite it -
+an `ENV` entry that does fails the configure with `FATAL_ERROR` naming
+`DOMAINS` as the way to set it, rather than silently landing behind the
+macro's own entry and winning at run time the way a plain
+`set_tests_properties(... PROPERTIES ENVIRONMENT ...)` used to. This is a
+literal string match, not a generator-expression evaluation, so the promise
+covers only that plain spelling: it does not chase down a value spelled to
+avoid the literal key (e.g. `ENV "$<1:MEDKIT_TEST_DOMAINS=99>"`), which still
+collides at run time exactly as if this check did not exist. Nobody writes
+that by accident, and a blanket rejection of `$<` would also catch legitimate
+uses of it in some other `ENV` entry, so this check only promises to catch
+the plain spelling.
 `LABELS` overrides `add_launch_test()`'s own default of `"launch_test"`, the same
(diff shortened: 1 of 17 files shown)


src/ros2_medkit_cmake/cmake/ROS2MedkitTestDomain.cmake
```diff
@@ -146,6 +146,59 @@ macro(medkit_add_pytest_test TEST_NAME TEST_PATH)
...
+# today, so this is hardening against a real GNUInstallDirs default rather
+# than a fix for an observed failure.
+#
+# The prefix alone would also catch `colcon build --install-base /opt/ros/jazzy`,
+# whose test step runs after its install step and whose launch tests work fine.
+# If this ever stops matching the build farm the guard simply does not fire and we
+# are back to a noisy log, which is the safe direction to be wrong in.
+function(_medkit_launch_tests_can_register _output_var)
+  set(${_output_var} TRUE PARENT_SCOPE)
+  if(NOT DEFINED ENV{ROS_DISTRO} OR NOT DEFINED CMAKE_INSTALL_LOCALSTATEDIR)
```
mfaferek93: This precondition is the one that breaks the whole guard. `$ENV{ROS_DISTRO}` is a third signal the PR description does not mention, and it is missing exactly where the guard has to work: the binarydeb chroot is `FROM ubuntu:jammy`, not a ROS image.

Verified in `ros:humble`:
- `setup.sh` and `local_setup.sh` contain zero references to `ROS_DISTRO`; the export lives only in `share/ros_environment/environment/1.ros_distro.sh`
- `ros-humble-ros-workspace`, which owns `setup.sh`, has no `Depends` line at all
- `ros-humble-ros-environment` appears in none of this repo's apt dependency closures (36 of 39 declared deps resolve, 0 pull it). Positive control: `ros-humble-ros-core` does pull it.

End-to-end, with `ros_environment` moved aside and bloom's exact flag order, run under a cleared environment because the chroot has no ROS variables preset:

```
env -i PATH=... sh -c '. /opt/ros/humble/setup.sh; cmake ... \
  -DCMAKE_INSTALL_LOCALSTATEDIR=/var -DCMAKE_INSTALL_PREFIX=/opt/ros/humble'

-- ROS_DISTRO=[]        CAN_REGISTER=[TRUE]    <- launch tests still registered
-- ROS_DISTRO=[humble]  CAN_REGISTER=[FALSE]   <- control, ros_environment present
```

Heads up on how this hides: run the same probe in `ros:humble` without clearing the environment and you get FALSE and it looks fine, because the image sets `ROS_DISTRO` as a Docker `ENV`. The chroot does not.

So the binarydeb log is unchanged, plus two new failures land there: `test_launch_test_guard_system_build` now registers (guard says it can) and fails its own `assert distro, 'ROS_DISTRO is unset'` in both packages.

Dropping the env dependency keeps the two-signal property:

```cmake
if(NOT DEFINED CMAKE_INSTALL_LOCALSTATEDIR)
  return()
endif()
...
if(_malt_prefix MATCHES "^/opt/ros/[^/]+$" AND _malt_localstatedir STREQUAL "/var")
```

`colcon build --install-base /opt/ros/<anything>` still passes, since it never sets `CMAKE_INSTALL_LOCALSTATEDIR=/var`.

src/ros2_medkit_cmake/test/test_launch_test_guard.py
```diff
@@ -0,0 +1,455 @@
...
+        '-B', str(build),
+        f'-DCMAKE_INSTALL_PREFIX={install_prefix}',
+    ]
+    if localstatedir is not None:
+        cmake_args.append(f'-DCMAKE_INSTALL_LOCALSTATEDIR={localstatedir}')
+    # Override, not replace: cmake still needs the inherited AMENT_PREFIX_PATH /
+    # CMAKE_PREFIX_PATH to resolve find_package(ament_cmake) and
+    # find_package(launch_testing_ament_cmake) at all. Only ROS_DISTRO itself is
+    # pinned, so every probe imitates a chosen distro regardless of who runs it.
+    env = dict(os.environ)
```
mfaferek93, line 260: Every probe forces `ROS_DISTRO` into the child env, so all eight guard cases pin it as present, which is the one state that does not hold in the job this guard targets (see my comment on `ROS2MedkitTestDomain.cmake:191`). `cmake_guard.py:49` inherits it implicitly instead, and both system_build call sites go further and assert it is set (`test_launch_test_guard_system_build.py:154` and `:222`, graph_watchdog copy `:59` and `:86`).

Net effect: the suite proves the guard fires when `ROS_DISTRO` is set, and nothing covers the case where it is not.

One case that deletes `ROS_DISTRO` from the child env and still expects suppression fails today and passes with the `MATCHES` form. That is the assertion that would have caught this before the build farm did.

src/ros2_medkit_cmake/test/test_launch_test_guard.py
```diff
@@ -0,0 +1,455 @@
...
+    The comparison is STREQUAL, which is exact, so the guard normalises the
+    prefix before comparing it rather than relying on the caller to spell it
+    exactly as bloom does.
+    """
+    result, build = configure_probe(
+        tmp_path,
+        LAUNCH_PROBE,
+        '/opt/ros/jazzy/',
+        {'probe_launch.test.py': '# empty launch test file\n'},
+        localstatedir='/var',
```
mfaferek93: This probe cannot fail. CMake normalises `CMAKE_INSTALL_PREFIX` itself before the project sees it, so the trailing slash is gone before your regex runs, and the docstring's claim that "the guard normalises the prefix before comparing it" is not what happens.

Measured in `ghcr.io/selfpatch/ros2_medkit-jazzy`, cmake 3.28.3:

```
-DCMAKE_INSTALL_PREFIX=/opt/ros/jazzy/      ->  [/opt/ros/jazzy]   already stripped
-DCMAKE_INSTALL_LOCALSTATEDIR=/var/         ->  [/var/]            untouched
```

Same on 3.22.6 and 4.4.2, including a mutant run with `ROS2MedkitTestDomain.cmake:195` replaced by a plain `set()`: guard fires either way. So :195 is dead code and :196 is the live one, and :196 has no coverage at all.

One word here fixes it: `localstatedir='/var/'`. That makes this the only test of the normalisation that can actually stop the guard firing, and it protects the line from someone reasonably concluding both regexes are redundant after noticing the prefix one is.

src/ros2_medkit_integration_tests/ros2_medkit_test_utils/cmake_guard.py
```diff
@@ -0,0 +1,109 @@
...
+    )
+
+    guarded_tests = registered_test_names(guarded_build_dir)
+    suppressed = baseline_tests - guarded_tests
+    assert suppressed, (
+        f'{name}: nothing was suppressed under system-package conditions - the guard did not '
+        f'remove any test that the baseline configure registered:\n'
+        f'baseline={sorted(baseline_tests)}'
+    )
+    assert guarded_tests, (
```
mfaferek93: `assert guarded_tests` only proves something survived, not "exactly the launch tests and nothing else" as the docstring at :79-82 says.

Traced for `ros2_medkit_log_bridge`: if the guard also swallowed gtests, baseline `{test_log_bridge, test_integration, lint..., test_dds_domain_allocation}` becomes guarded `{lint..., test_dds_domain_allocation}` - `suppressed` non-empty, `guarded_tests` non-empty, subset holds, so both system_build tests still pass.

The intended backstop, `test_launch_test_guard.py:335-353`, probes only a pytest test even though its docstring at :341 names gtests as the thing being protected. A gtest probe alongside `PYTEST_PROBE` closes it.

src/ros2_medkit_integration_tests/CMakeLists.txt
```diff
@@ -194,67 +222,78 @@ if(BUILD_TESTING)
...
+  # after the demo nodes it monitors are already publishing (see the test's
+  # module docstring). The default 120s feature timeout does not leave enough
+  # headroom for that startup delay plus the 30-90s polling budgets its cases
+  # need, so it is widened to match the scenario-test timeout.
+  #
+  # test_graph_provider_stale and test_graph_provider_sse also drive a real
+  # greenwave_monitor node through the same 10s launch-ordering delay plus
+  # warmup as test_graph_provider_greenwave above, and their staleness/SSE
+  # polling budgets are similarly generous - both widened to match.
+  set(_MEDKIT_TEST_TIMEOUT_OVERRIDES
```
mfaferek93, line 251: This swaps a hard error for a silent no-op, and it is the one place in the migration where a check was traded away rather than moved.

Before, `set_tests_properties(test_graph_provider_sse PROPERTIES TIMEOUT 300)` on a name that does not exist is a configure failure. Now a stale entry in this list just never matches at :261-262 and nothing says so. Rename or move `test_graph_provider_sse.test.py` and its budget silently drops from 300 back to the glob default of 120, resurfacing as an unexplained timeout - the exact failure the comment block above was written to prevent.

Cheap guard: after both loops, check every even-indexed name in the list was consumed and `message(FATAL_ERROR)` otherwise.

src/ros2_medkit_cmake/README.md
```diff
@@ -237,7 +334,13 @@ part of what the other needs is a deadlock that only ends when both wait budgets
...
 entries must use `set_property(TEST ... APPEND PROPERTY ENVIRONMENT ...)`. A plain
 `set_tests_properties(... PROPERTIES ENVIRONMENT ...)` afterwards drops it, and the check
-below says so.
+below says so. This is the general rule for `medkit_add_gtest`, `medkit_add_gmock`,
+`medkit_add_pytest_test` and `medkit_add_wrapped_test` - all four are always registered, so a
+call naming the test afterwards is safe, just easy to get wrong in the APPEND direction.
+**A launch test is the one exception**: it is not always registered (see "Launch tests are not
+registered in a system package build" above), so a `set_tests_properties` or `set_property`
+call naming it afterwards is not a subtler bug to get right, it is unsupported - use
+`medkit_add_launch_test`'s own `LABELS` and `ENV` parameters instead.
```
mfaferek93: This reads as "launch tests cannot have properties set afterwards, full stop", but the macro owns only two: `LABELS` at `ROS2MedkitTestDomain.cmake:327` and `ENVIRONMENT` at :332-334. There is no route through `medkit_add_launch_test` for `RESOURCE_LOCK`, `RUN_SERIAL`, `WILL_FAIL`, `DEPENDS` or `SKIP_RETURN_CODE`, and the README names no alternative.

The need is live in this repo - `graph_watchdog/CMakeLists.txt:233,254` already puts `RESOURCE_LOCK` on gtests, and a launch test that needs one has nowhere to go.

The obvious escape hatch works and stays correct under both guard states, verified by configure: `if(TEST t) set_property(TEST t APPEND PROPERTY RESOURCE_LOCK lock) endif()` yields `RESOURCE_LOCK "lock"` alongside the macro's own properties. Worth one sentence here saying that is the supported shape for anything the macro does not own, otherwise a caller reading the only spec concludes it is impossible.

src/ros2_medkit_integration_tests/package.xml
```diff
@@ -42,6 +42,32 @@
...
 
+  <!-- Configured for real, not linked against, by
+       test/test_launch_test_guard_system_build.py, which is what actually
+       needs these declared: a package used at test time must be a
+       test_depend even when nothing here #includes or find_package()s it at
+       build time. ros2_medkit_gateway and ros2_medkit_fault_manager are
+       already declared above; these four were resolving only by luck,
+       through ros2_medkit_gateway's own exec_depend chain.
+
+       ros2_medkit_graph_watchdog is the seventh package that test file
```
mfaferek93: Both claims in this comment are stale. `LAUNCH_TEST_PACKAGES` in `test_launch_test_guard_system_build.py:96-103` has exactly six entries, and :92-95 says graph_watchdog is deliberately not listed - so it is not "the seventh package that test file configures".

And the tail at :64-65 says the resolvability probe "reports it by name": `resolvable_packages` iterates only those six, so it never sees graph_watchdog at all. Its coverage lives in the graph_watchdog copy of the test instead.

The build-order reasoning above is still right and worth keeping; it is only the two sentences about what the test does that need correcting.

src/ros2_medkit_cmake/cmake/ROS2MedkitTestDomain.cmake
bburda: You are right, and I reproduced it on both versions of the module before changing anything.

With `ROS_DISTRO` removed from the environment, the guard as it was pushed returns `CAN_REGISTER=TRUE`, and with it present it returns `FALSE`. So it never fired in the chroot and the log would have stayed as it was. I also checked the three facts your comment rests on: `setup.sh` and `local_setup.sh` have no reference to `ROS_DISTRO`, the export lives only in `ros_environment`, and no `package.xml` here declares that package.

The guard now reads only CMake variables, in the form you suggested: `CMAKE_INSTALL_LOCALSTATEDIR` must be defined and equal to `/var`, and the prefix must match `^/opt/ros/[^/]+$`. `colcon build --install-base` still registers, because colcon does not pass that variable. The comment above the function now says why the environment is not read, so the next person does not put it back.

src/ros2_medkit_cmake/test/test_launch_test_guard.py
bburda: ## 3768011430 | mfaferek93 | test_launch_test_guard.py:235

Agreed. Every probe pinned the one state that does not hold in the target job, so the suite proved the guard works where it was never the question.

There is now a case that deletes `ROS_DISTRO` from the child environment and expects suppression. Against the guard as pushed it fails with `assert 'probe_launch_test' not in ['probe_launch_test']`; against the new one it passes. Both `test_launch_test_guard_system_build.py` copies also stopped asserting the variable is set and use a literal prefix instead.

src/ros2_medkit_cmake/test/test_launch_test_guard.py
bburda: ## 3768011436 | mfaferek93 | test_launch_test_guard.py:329

Confirmed on this machine, CMake 3.28.3: a trailing slash on `CMAKE_INSTALL_PREFIX` is already gone before the project runs, and one on `CMAKE_INSTALL_LOCALSTATEDIR` is not. I did not repeat it on 3.22 or 4.4.

The prefix normalisation is deleted as dead code and the probe now passes `localstatedir='/var/'`, so it covers the line that can actually stop the guard firing. The comment records why one of the two needs stripping and the other does not.

src/ros2_medkit_integration_tests/ros2_medkit_test_utils/cmake_guard.py
bburda: ## 3768026295 | mfaferek93 | cmake_guard.py:102

You are right that the assertion only proves something survived. A guard that also swallowed gtests would pass it.

There is now a gtest probe next to the pytest one, so the claim that the guard covers launch tests only is checked for the kind of test the docstring actually names.

src/ros2_medkit_integration_tests/CMakeLists.txt
bburda: ## 3768026298 | mfaferek93 | CMakeLists.txt:246

Agreed, and this was the one place where the migration removed a check instead of moving it.

After both loops the list is now checked: any entry that matched no discovered test name stops the configure with `FATAL_ERROR` naming the entry. A renamed or deleted test file fails the build instead of quietly dropping that test's budget back to the glob default.

src/ros2_medkit_cmake/README.md
bburda: ## 3768026304 | mfaferek93 | README.md:343

Correct, the macro owns `LABELS` and `ENVIRONMENT` and nothing else, and I confirmed `RESOURCE_LOCK` is already used on gtests in the watchdog package.

The README now scopes the claim to those two properties and documents the shape for everything else, including your `if(TEST t) set_property(TEST t APPEND PROPERTY ...) endif()`, which stays correct whether the guard fired or not.

src/ros2_medkit_integration_tests/package.xml
bburda: ## 3768026309 | mfaferek93 | package.xml:53

Both sentences were stale, thank you for catching it. The list has six entries and the watchdog package is covered by its own copy of the test, not by this one.

The two sentences are corrected and the build-order reasoning above them is kept as it was.

src/ros2_medkit_fault_manager/src/rosbag_capture.cpp
```diff
@@ -163,41 +166,38 @@ RosbagCapture::RosbagCapture(rclcpp::Node * node, FaultStorage * storage, const
...
+    const std::string other_format = (config_.format == "sqlite3") ? "mcap" : "sqlite3";
+    if (auto other_err = storage_probe_(other_format)) {
       RCLCPP_WARN(node_->get_logger(),
-                  "No usable rosbag storage backend: '%s' (%s) and sqlite3 (%s) both failed to load; black-box "
-                  "rosbag capture disabled (freeze-frame snapshots still work)",
-                  config_.format.c_str(), reason.c_str(), truncate_reason(*sqlite_err).c_str());
+                  "No usable rosbag storage backend: '%s' (%s) and '%s' (%s) both failed to load; install "
+                  "rosbag2_storage_mcap and rosbag2_storage_default_plugins. Black-box rosbag capture disabled "
+                  "(freeze-frame snapshots still work)",
+                  config_.format.c_str(), reason.c_str(), other_format.c_str(), truncate_reason(*other_err).c_str());
```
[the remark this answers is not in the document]
bburda: ## 3761098463 | Copilot | rosbag_capture.cpp:188

This file is no longer part of this pull request. The fault manager changes moved to a separate one so that the launch-test change can be reviewed on its own.

The point is valid and will be handled there: the "no usable backend" warning should use the same helper as the other one, so an operator gets an install command instead of two ROS package names.

mfaferek93 (approved):

PR #609, opened 2026-08-12
fix(gateway): read discovery configuration from the documented 'config:' key

### LINKED ISSUE #529: [BUG] Manifest top-level config: block is silently ignored (config read only from discovery:)
# Bug report

### Steps to reproduce
1. Write a manifest with a top-level `config:` block that sets `unmanifested_nodes`:
```yaml
config:
  unmanifested_nodes: ignore
```
(some example manifests place these keys under `config:`.)
2. Run the gateway in hybrid discovery mode, with runtime nodes that are not listed in the manifest.

### Expected behavior
`unmanifested_nodes: ignore` hides the runtime nodes that are not in the manifest, so they do not show up in the tree.

### Actual behavior
The setting has no effect. The unmanifested nodes still appear as Apps (the policy stays at its default `warn`). The cause: `ManifestParser::parse_config` is only called with `root["discovery"]` (manifest_parser.cpp:126-127), so a top-level `config:` block is never read. The key only works when it is placed under `discovery:`. A manifest that uses `config:` is silently ignored, and in hybrid mode the extra runtime nodes leak into the entity tree.

### Environment
- ros2_medkit version: 0.6.0
- ROS 2 distro: Jazzy

### Additional information
Two ways to fix it: read config from `root["config"]` as well, or update the example manifests and the schema to use `discovery:`. It would also help to log a warning when a manifest has a top-level key the parser does not read (like `config:`), so the mistake is visible instead of silent.

bburda:
# Pull Request

## Summary

The manifest's top-level `config:` block was never read. The parser looked only at an
undocumented `discovery:` key, so the block used by the schema, all seven shipped example
manifests, the demo manifests and the integration tests was ignored without a word in the log,
and every setting inside it stayed at its default.

This reads `config:` and keeps `discovery:` working as a deprecated alias, so a manifest written
against the source rather than the schema does not break. When both appear, `config:` wins and
both facts are logged.

**Closing the silent-failure class that hid this**

- An unrecognised top-level key is reported instead of vanishing.
- A `config:` block, or a value inside it, of the wrong YAML kind is reported and the default
  applies, instead of throwing and costing the whole manifest. A key with no value still means
  "not set".
- An unrecognised `unmanifested_nodes` value becomes a validation warning, so
  `discovery.manifest_strict_validation` decides whether it rejects the manifest.

**Making the three documented settings do what they say**

- `unmanifested_nodes` reaches the merge pipeline. `error` keeps the gateway serving and reports
  the undeclared nodes on `GET /health`; a startup abort was only ever promised in prose, never
  implemented, and a diagnostic gateway going down when the system misbehaves is the wrong
  trade.
- `inherit_runtime_resources: false` keeps a linked app to the resources it declares, including
  when its id matches the bound node's name.
- `allow_manifest_override: false` demotes the manifest layer's field groups; an explicitly
  configured per-group merge policy still wins.

All three now survive a manifest reload instead of freezing at startup.

**A merge defect found on the way**

An entity's `source` was merged as ordinary metadata content. Provenance is what decides whether
the orphan filter may delete an entity, so a field-group policy must not be able to reassign it:
with the manifest layer demoted for metadata, a declared entity was retagged and then deleted.
This is reachable on `main` today through the gateway parameters alone, without any of the
changes here. `source` is now filled only when absent and is never reassigned by a policy.

**Health contract**

`warnings` and `warning_schema_version` are emitted whenever the gateway can produce a warning,
not only when aggregation is active, and both are required in the schema. Warning objects keep
identifier domains apart: `entity_ids` carries addressable SOVD entity ids for every code, and
ROS node names live in the new `ros_node_fqns`. `discovery.linking` reports the configured
policy, so a monitor can branch on it without parsing prose.

That invariant is enforced on the one path that could break it. The `leaf_id_collision` warning
publishes a Component id announced by a remote peer, and nothing validated it. It is now checked
before it is published and withheld with a warning log when it is not addressable; the warning's
`message` still names the component, so only the link that would not have worked is dropped.

**Tooling**

Integration feature tests are registered with `CONFIGURE_DEPENDS`. Without it a test file added
to an existing build tree is never registered and the suite reports success without running it -
measured here: three of this branch's four new test files were silently skipped, and the same
suite went from 1036 to 1091 tests once the glob was re-evaluated.

For the same class of reason, `scripts/clang-tidy-diff.sh` now considers staged and working-tree
changes as well as committed ones. It was narrowing its file list to committed work only, so the
pre-commit stage its own header describes analysed nothing and exited 0.

**Deletion is decided by the owning layer, not by a string**

A discovery plugin may set its own `source` tag, and two shipped plugins do - `beacon` and, on
its authenticated path, `opcua`. Neither matched the `"plugin"` prefix the orphan filter checks,
so `unmanifested_nodes: ignore` erased those entities. The filters now consult the layer that
owns an entity, which the merge already tracks, instead of the string. No `source` value
changes: one capability is gated on the literal `"node"`, the value is served as
`x-medkit.source` and feeds identity provenance, and the `opcua` tag carries a deliberate trust
distinction that must not be collapsed.

This also makes true a promise the documentation already made - that `ignore` spares apps from
the manifest, the asset inventory or a plugin.

**Two things to know before merging**

1. **Behaviour changes on upgrade.** Anyone whose manifest sets `unmanifested_nodes: ignore` has
   had it ignored until now. It starts taking effect, so their entity tree gets smaller -
   runtime-discovered nodes that are not declared in the manifest disappear. In this repository
   that affects the `multi_ecu_aggregation` demo, whose documented entity counts shift.
2. **Release ordering.** `warnings` and `warning_schema_version` move from optional to required
   in the schema. A client regenerated from this spec would reject a response from a gateway
   that omits them, so the client regeneration belongs with the gateway release, not before it.
   The OpenAPI component `HealthAggregationWarning` is also renamed `HealthWarning`; the wire
   shape is unchanged, but a regenerated client sees a different schema name.
3. **The two new manifest notices are advisory in this release.** An unrecognised
   `unmanifested_nodes` value (R014) and a malformed configuration block (R015) are logged, not
   raised to the validator, and become validation warnings in the next release. Without that
   grace period a manifest carrying, say, `unmanifested_nodes: Warn` - which loaded fine before
   this change because the block was never read - would be refused under the default strict
   validation, and hybrid discovery would silently degrade to `runtime_only`.

---

## Issue

- closes #529

---

## Type

- [x] Bug fix
- [ ] New feature or tests
- [x] Breaking change
- [ ] Documentation only

---

## Testing

Unit suite 124/124. Full integration suite 1091 tests, 0 failures, on an otherwise idle machine.

Four new integration test files drive a live gateway rather than asserting on log text:

- the documented key and the deprecated alias produce the same entity tree, and a node that
  joins *after* the gateway is up stays out of it under `ignore`
- a malformed configuration block loads with a warning under non-strict validation and is
  rejected under strict
- provenance holds across repeated discovery passes, sampled until the entity cache has actually
  advanced rather than after a fixed sleep
- the policy surface on `/health`, the inheritance gate, and the override precedence, each with
  a control gateway so a difference is attributable to the setting

API differential against `main`, captured from two running gateways: 113 lines of schema change,
no routes added, removed or altered. The schema change is the `HealthWarning` rename, the new
required `ros_node_fqns`, `warnings` and `warning_schema_version` becoming required, and enum
constraints on `code` and `unmanifested_policy`.

The eight existing integration tests whose manifests set `unmanifested_nodes: ignore` were
re-read, not just re-run: each was checked for whether it now passes for a weaker reason, and
two that pinned the old warning-contract version were updated.

Reviewers can verify the core fix directly:

```
ros2 run ros2_medkit_gateway gateway_node --ros-args \
  -p discovery.mode:=hybrid -p discovery.manifest_path:=<manifest with config: unmanifested_nodes: ignore>
curl -s localhost:8080/api/v1/apps | jq '.items[].id'
```

Before this change the policy has no effect; after it, undeclared runtime nodes are absent.

---

## Checklist

### DIFF
--- a/docs/config/manifest-schema.rst
+++ b/docs/config/manifest-schema.rst
@@ -22,6 +22,7 @@ A manifest file has the following top-level structure:
      description: string
 
    config:                # Optional - discovery behavior settings
+                          # (deprecated alias for this key: "discovery:")
      unmanifested_nodes: string
      inherit_runtime_resources: boolean
      allow_manifest_override: boolean
@@ -32,6 +33,17 @@ A manifest file has the following top-level structure:
    apps: []               # Optional - app definitions
    functions: []          # Optional - function definitions
    scripts: []             # Optional - pre-defined script entries
+
+.. note::
+
+   **Unknown top-level keys are ignored, and the gateway says so.** Any
+   top-level key outside the set the parser reads
+   (``manifest_version``, ``metadata``, ``config``, ``discovery``, ``areas``,
+   ``components``, ``assets``, ``apps``, ``functions``, ``scripts``,
+   ``capabilities``) is skipped, and the gateway logs a warning naming the key
+   and listing the ones it does know. If a whole block of your manifest seems
+   to have no effect, that log line is the first place to look - a misspelled
+   or misplaced top-level key is the usual cause.
 
 manifest_version (Required)
 ~~~~~~~~~~~~~~~~~~~~~~~~~~~
@@ -76,6 +88,26 @@ config (Optional)
 
 Discovery behavior configuration.
 
+Every setting in this block describes how the manifest is combined with what
+runtime discovery finds, so none of them does anything in ``manifest_only``
+(which parses the block and then runs without the merge pipeline that reads
+it) or in ``runtime_only`` (which has no manifest at all).
+
+Within ``hybrid`` mode they differ in what else they need.
+``unmanifested_nodes`` and ``inherit_runtime_resources`` are applied by the
+runtime linker, which only exists when the runtime layer is enabled.
+``allow_manifest_override`` is applied to the manifest layer's own policies,
+so it takes effect whenever the manifest layer is built - though with no
+runtime layer to merge against there is nothing for it to change.
+
+.. note::
+
+   ``config:`` is the canonical top-level key. ``discovery:`` is accepted as a
+   **deprecated alias** for the same block: it is read, the gateway logs a
+   deprecation warning naming it, and it will be removed in a future release.
+   A manifest that declares both keys uses ``config:`` and ignores
+   ``discovery:``; both facts are logged.
+
 .. list-table::
    :header-rows: 1
    :widths: 25 15 60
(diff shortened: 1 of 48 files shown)


src/ros2_medkit_gateway/src/discovery/manifest/manifest_parser.cpp
```diff
@@ -201,18 +266,110 @@ ManifestMetadata ManifestParser::parse_metadata(const YAML::Node & node) const {
...
+    if (!policy_node.IsScalar()) {
+      notices.push_back({"R015", "Manifest setting 'unmanifested_nodes' must be a string, but is a " +
+                                     std::string(yaml_kind_name(policy_node)) + "; the default ('warn') applies."});
+    } else if (const std::string policy = policy_node.as<std::string>(); !policy.empty()) {
+      config.unmanifested_nodes = ManifestConfig::parse_policy(policy);
+      // parse_policy is total - every unrecognised string maps to WARN - so at
+      // this call site a typo is indistinguishable from a deliberate "warn".
+      // Detect it by round-tripping the parsed value rather than by keeping a
+      // second copy of the valid-value list here.
+      if (ManifestConfig::policy_to_string(config.unmanifested_nodes) != policy) {
```
mfaferek93, line 392: R014 is a *validation* notice, and that makes this an upgrade break under the shipped default.

`manifest_manager.cpp:162-164` turns validation notices into `add_warning`, `:85-88` is `if (strict && has_warnings()) return false`, and `manifest_strict_validation` defaults true (`gateway_node.cpp:183`, `gateway_params.yaml:253`). Hybrid then degrades to `runtime_only` at `discovery_manager.cpp:72-84`; manifest-only fails init outright.

So a manifest carrying `unmanifested_nodes: Warn` (capital W) or `inherit_runtime_resources: 0` loaded fine before this PR, because the whole `config:` block was unread, and after it the manifest is refused and the declared SOVD entity tree disappears. The operator changed nothing but the version.

None of the seven shipped examples trips it - all lower-case and valid - so CI stays green and only customer manifests find it. The code comments are honest that these are validation notices; the PR body isn't.

Either a release note, or make R014/R015 log-only for one cycle and promote them next release.

src/ros2_medkit_gateway/src/discovery/manifest/manifest_parser.cpp
```diff
@@ -122,9 +155,23 @@ Manifest ManifestParser::parse_string(const std::string & yaml_content) const {
     manifest.metadata = parse_metadata(root["metadata"]);
   }
 
-  // Parse discovery config
-  if (root["discovery"]) {
-    manifest.config = parse_config(root["discovery"]);
+  // Parse discovery config. `config:` is the documented, canonical key.
+  // `discovery:` is the key the parser originally read and is kept working as
+  // a deprecated alias; when both are present, `config:` wins.
+  const bool has_config = static_cast<bool>(root["config"]);
```
mfaferek93: `static_cast<bool>` is true for a key with no value, so an empty `config:` beats a populated `discovery:` and nothing is applied.

```yaml
config:
discovery:
  unmanifested_nodes: error
```

Measured with yaml-cpp 0.8 in the jazzy image: `config` gives `bool=1, IsNull=1, IsDefined=1, IsMap=0`. So `has_config` wins here, `parse_config_block` returns defaults on the `IsNull` branch at :288-289 whose comment says "present but empty == absent", and the `else if` at :170 never runs. Policy stays `warn`.

Then :166-168 logs "'config:' is used and 'discovery:' is ignored", which reads to the operator as confirmation their block took effect. Neither block did. Same for `config: {}`.

The precedence test treats null as present, the parse path treats it as absent - one of the two needs to move. `root["config"].IsDefined() && !root["config"].IsNull()` here would do it. No test covers this combination.

docs/config/manifest-schema.rst
```diff
@@ -85,20 +117,184 @@ Discovery behavior configuration.
...
+A node is *unmanifested* (an *orphan*) when it is present in the ROS graph and
+no manifest app binds to it.
+
+.. list-table::
+   :header-rows: 1
+   :widths: 25 75
+
+   * - Value
+     - Behaviour
+   * - ``ignore``
```
mfaferek93, line 146: This promise does not hold for a plugin that sets its own `source`.

`core/discovery/merge_types.hpp:174-176` protects `"manifest"`, `"inventory"` and anything with the `"plugin"` prefix. But `plugin_layer.cpp:72-83` only defaults `source` to `"plugin"` when the provider left it empty, and its comment explicitly invites a custom tag ("e.g. a protocol tag like 'opcua'"). Two shipped plugins take it:

- `beacon_entity_mapper.cpp:86` - `new_app.source = "beacon"`, used by both `param_beacon` and `topic_beacon`
- `opcua_plugin.cpp:732` - `comp.source = opcua_identity_trusted(client_config_) ? "opcua" : "plugin"`

Neither starts with `"plugin"`, so neither is protected. With `config: {unmanifested_nodes: ignore}` - reachable through the documented key for the first time because of this PR - beacon apps have no `bound_fqn`, survive the linker filter, and are then erased by the `hide_orphans` sweep. Every beacon-discovered app silently vanishes.

The OPC UA line is the sharper version: turning on a trusted OPC UA identity changes whether the component is protected from deletion. That coupling can't be intended.

Either protect anything a plugin layer contributed regardless of tag, or have `PluginLayer` namespace custom tags as `plugin:<tag>` so the prefix check keeps working.

src/ros2_medkit_gateway/src/http/handlers/health_handlers.cpp
```diff
@@ -70,14 +77,43 @@ http::Result<dto::Health> HealthHandlers::get_health(const http::TypedRequest &
 
       auto linking = dm->get_linking_result();
       if (linking) {
+        const auto manifest_config = dm->get_manifest_config();
```
mfaferek93, line 80: This puts a manifest-load-scoped lock on the liveness path.

`get_manifest_config()` reaches `ManifestManager::get_config()` (`manifest_manager.cpp:235-236`), which takes `mutex_`. `load_manifest` holds that same mutex from entry (`:34-35`) through `parse_file`, `apply_fragments` (directory scan plus a parse per fragment), `apply_inventory_csv` and validation.

Checked against the parent commit: `/health` previously touched `ManifestManager` zero times.

So a plugin calling `notify_entities_changed` on a deployment with a populated fragments dir parks every concurrent `/health` until the load finishes, and a container liveness probe with a short timeout can restart the gateway during a routine reload. The handler's own comment a couple of blocks down still says "atomic reads only; safe to serve from /health even under load".

Cheapest fix: have `refresh_manifest_config()` stash the policy in an atomic, or cache it next to `cached_result_` under `result_mutex_` - it is refreshed on the same path already.

src/ros2_medkit_gateway/src/discovery/discovery_manager.cpp
```diff
@@ -109,6 +109,50 @@ void DiscoveryManager::apply_layer_policy_overrides(const std::string & layer_na
   }
 }
 
+// The manifest layer's policies come from two sources that must be applied in
+// this order, at boot AND after every reload. Shared so the two paths cannot
+// drift: a reload that applied them in the other order would silently invert
+// R11a's precedence and let the blanket flag beat an explicit deployment
+// parameter.
+void DiscoveryManager::apply_manifest_layer_policies(discovery::ManifestLayer & manifest_layer) {
```
mfaferek93, line 117: This function and `refresh_manifest_config()` below it have no test at all, and they are the new half of the reload path.

`test_health_handlers.cpp:932` is the only test in the repo that calls `trigger_reload()`, and it covers `unmanifested_nodes` only. `inherit_runtime_resources` and `allow_manifest_override` do not appear in that file, and there is no reload test in the integration suite.

Specifically untested:
- the `else` branch calling `reset_policies_to_defaults()`, which exists solely for reload (your own comment: "A reload can flip the flag back on")
- the `set_manifest_config()` push that carries `inherit_runtime_resources` across a reload
- the ordering invariant your comment calls load-bearing ("a reload that applied them in the other order would silently invert R11a's precedence")

Two more cases in `HealthLinkingTest` next to the existing reload test would close it: flip `allow_manifest_override` false->true and re-check the hosts union, flip `inherit_runtime_resources` true->false and re-check `/apps/<id>/data`.

src/ros2_medkit_integration_tests/test/features/test_manifest_config_key.test.py
```diff
@@ -0,0 +1,526 @@
...
+DOCUMENT_MATCH_TIMEOUT_SEC = 15.0
+
+
+def _assert_documents_match(test, fetch_left, fetch_right, *, what):
+    """Assert two served documents converge, reporting the diff if they do not."""
+    deadline = time.monotonic() + DOCUMENT_MATCH_TIMEOUT_SEC
+    left = right = None
+    while time.monotonic() < deadline:
+        try:
+            left, right = fetch_left(), fetch_right()
```
mfaferek93: This helper can pass without fetching anything.

Python builds the whole right-hand tuple before unpacking, so if either fetch raises, neither name is rebound and both keep the `None` from :219:

```
left = right = None
try: left, right = ok(), boom()
except RuntimeError: pass
-> left=None right=None, assertEqual passes
```

`_collection_items` calls `raise_for_status()`, so any 5xx is a `RequestException` and is swallowed at :225-226. If the right-hand gateway is down or erroring for the whole 15 s window, :228 is `assertEqual(None, None)` and the test is green. For `test_entity_tree_matches_the_canonical_key` the right-hand side is the canonical `config:` gateway, i.e. the thing under test, and the docstring at :209-212 promises this specifically cannot happen.

Same logic byte-for-byte at `test_unmanifested_policy.test.py:290-305`, so a fix needs to land in both (or the helper needs to move into `ros2_medkit_test_utils`).

Tracking `last_error` and failing when no successful pair was ever fetched is enough.

src/ros2_medkit_gateway/include/ros2_medkit_gateway/dto/registry.hpp
```diff
@@ -82,8 +82,8 @@ using AllDtos =
                // Auth domain DTOs
                AuthCredentials, AuthTokenResponse, AuthRevokeRequest, AuthRevokeResponse,
                // Health / Root domain DTOs
-               HealthDiscoveryLinking, HealthDiscovery, HealthAggregationWarning, Health, VersionInfoVendor,
-               VersionInfoEntry, XMedkitVersionInfo, VersionInfo, RootCapabilities, RootAuth, RootTls, RootOverview,
+               HealthDiscoveryLinking, HealthDiscovery, HealthWarning, Health, VersionInfoVendor, VersionInfoEntry,
```
mfaferek93, line 85: `HealthAggregationWarning` -> `HealthWarning` renames a published `components/schemas` entry, and nothing in `docs/` mentions it (grep finds nothing).

The wire change is additive and you signalled it properly with `warning_schema_version` 1->2. But the version covers the wire shape, not schema component names, so a client generated against the OpenAPI document breaks on regeneration with no note telling the author why.

One line in `docs/api/rest.rst` or the changelog is enough. Flagging because everything else about this surface in the PR is carefully versioned, so the omission looks accidental rather than decided.

src/ros2_medkit_gateway/src/discovery/manifest/manifest_parser.cpp
bburda: You are right about the upgrade path. R014 and R015 are log-only now. They do not reach the validator, so a manifest with `Warn` or `0` still loads under the default strict setting. They become validation warnings in the next release. This is written in the schema docs and in the PR description.

src/ros2_medkit_gateway/src/discovery/manifest/manifest_parser.cpp
bburda: Reproduced, including the log line that tells the operator the opposite of what happened. The precedence check now uses the same present-test as the block parser, so a null `config:` falls through to the alias and no collision notice is logged. A table test covers all eight combinations of the two keys.

docs/config/manifest-schema.rst
bburda: Changed at the root you pointed at. Deletion no longer reads the `source` string. It uses the layer that owns the entity, which the merge already records. No `source` value changed, so the trust distinction on the OPC UA path stays. One correction from tracing it: only Apps are lost. Components and Areas are dropped only when the namespace matches, and we could not confirm the OPC UA Component case without a live plugin.

src/ros2_medkit_gateway/src/http/handlers/health_handlers.cpp
bburda: Confirmed, and it came from this branch. `/health` did not touch ManifestManager before. The policy is now a snapshot without a lock, refreshed on the reload path. With the lock put back, a test shows `/health` blocked for 63 ms while a 180 ms load ran.

src/ros2_medkit_gateway/src/discovery/discovery_manager.cpp
bburda: Correct, only `unmanifested_nodes` was covered. I added the two cases you listed: `allow_manifest_override` false to true, which hits the reset branch, and `inherit_runtime_resources` true to false, checked on `/apps/<id>/data`.

src/ros2_medkit_integration_tests/test/features/test_manifest_config_key.test.py
bburda: Fixed, and checked three ways. The old helper passes against a dead port without comparing anything. The new one fails there with a clear error. On live ports it passes. The helper moved to the shared test utils and both call sites use it.

src/ros2_medkit_gateway/include/ros2_medkit_gateway/dto/registry.hpp
bburda: It was an omission, not a decision. Added a note in the REST docs next to the version 2 contract.

mfaferek93 (dismissed):

mfaferek93 (approved):

Issue #251, opened 2026-03-05
[BUG] Buildfarm integration tests fail - rosbag2 sqlite3 storage plugin unavailable on Jazzy/Noble


mfaferek93:
# Bug report

### Steps to reproduce
1. Buildfarm triggers `Jdev__ros2_medkit__ubuntu_noble_amd64` build #2
2. Compilation passes, integration tests run
3. Tests `test_scenario_bulk_data_download` (3/3 fail) and `test_scenario_fault_inspection` (1/6 fail) error out

### Expected behavior

Integration tests pass on buildfarm the same way they pass locally / in CI.

### Actual behavior

Fault manager fails to initialize rosbag capture:
```
[ERROR] [rosbag2_storage]: Could not load/open plugin with storage id 'sqlite3'
[INFO] [rosbag2_storage]: Available read-write storage plugins: 'mcap'
[ERROR] [fault_manager]: Failed to write bag file: No storage could be initialized
```

Downstream effect: bulk data download tests fail with "No rosbag found" and fault inspection test fails with "GET /apps/lidar_sensor/faults not available after 30.0s".

Build log: https://build.ros2.org/job/Jdev__ros2_medkit__ubuntu_noble_amd64/2/console

### Root cause

Default rosbag format is hardcoded to `sqlite3` in two places:
- `snapshot_capture.hpp:55` - `std::string format{"sqlite3"};`
- `fault_manager_node.cpp:646` - `declare_parameter<std::string>("snapshots.rosbag.format", "sqlite3")`

On Jazzy/Noble buildfarm, `rosbag2_storage` no longer bundles the sqlite3 plugin. Only `mcap` is available. The code in `rosbag_capture.cpp:682-686` incorrectly assumes sqlite3 is always available.

Additionally, `rosbag2_storage_mcap` is only declared as `<test_depend>` in `package.xml`, not as a runtime `<depend>`.

### Fix

1. Change default format from `sqlite3` to `mcap` in `snapshot_capture.hpp` and `fault_manager_node.cpp`
2. Move `rosbag2_storage_mcap` from `<test_depend>` to `<depend>` in `package.xml`
3. Update `rosbag_capture.cpp` - remove "sqlite3 is always available" assumption

### Environment
- ros2_medkit version: 0.3.0 (main, commit 4c02bd4)
- ROS 2 distro: Jazzy Jalisco
- OS: Ubuntu Noble 24.04 (buildfarm)

Issue #606, opened 2026-08-11
Release v0.7.0


bburda:
## Summary

Cut release **0.7.0**. 0.6.0 shipped 2026-06-22. Since that tag `main` has 59 merged PRs and every `package.xml` is still at 0.6.0. This issue tracks the documentation work, the changelogs, the bugs that have to go in first, and the pre-tag checklist.

## Highlights since 0.6.0

**New packages**

- `ros2_medkit_fault_detection` - shared, protocol-agnostic fault detection for gateway plugins: threshold, status bit and fault-code enum in one header-only evaluator (#486). Also added to the Docker image workspace (#494).
- `ros2_medkit_graph_watchdog` - gateway plugin that raises a fault for graph problems nothing reports today. Three detectors so far: QoS mismatch (#571), orphan (#578) and parameter drift (#580).

**OPC UA and PLC**

- Config-less discovery. A read-only network scan finds the server (#509), `auto_browse` walks the address space and builds the SOVD tree from it (#510), native Alarms and Conditions become faults with no configuration (#511), and identity, writability and fault triggers are read from the device itself (#544).
- Identity from the device: nameplate fields (#492) and `order_code` (#499).
- `PLC_COMMS_LOST` fault when the connection drops (#507), alarms routed to separate faults by message substring (#506), `severity` accepted as an alias in `event_alarms` (#501), non-condition events dropped from the alarm path (#552).
- Security and reconnect hardening, replay after reconnect, multi-alarm handling, secure profile (#485). Fixed a data race between the REST and poll threads that crashed the gateway (#519).

**Faults**

- Append-only, hash-chained audit log of fault state transitions, off by default (#487).
- Freeze-frame: a compact JSON snapshot is persisted when a fault is confirmed (#491), the faulting entity's own data is captured with no configuration (#538), faults already raised at startup get one (#563), and so does an entity serving last known values (#565).
- Black box: a fault confirmed inside an active post-roll window keeps its bag (#561), and a fault landing on a recording-window boundary now gets its own bag (#594).
- Debounce counter clamped, confirmed and healed status latched (#484). A PASSED event no longer re-dates a fault, and real SSE loss is counted (#573).
- Faults from plugin-provided entities are visible in list, detail and freeze-frame through fault-scope ownership (#503), and an external Component owns its fault manager faults (#530).
- `FaultReporter` can be built from an `rclcpp_lifecycle::LifecycleNode` (#556).

**Assets and manifest**

[...]

PR #610, opened 2026-08-12
Declare the rosbag storage plugins, default to mcap, and serialize the plugin loader

### LINKED ISSUE #619: [BUG] Installed fault manager has no rosbag storage backend, and the plugin loader can segfault
# Bug report

Two defects in black-box rosbag capture, both found while cleaning up the launch-test noise in the Humble binarydeb job.

### Steps to reproduce

**1. No storage plugin in the install closure**

1. Install `ros2_medkit_fault_manager` from a binary package, or build it in a container that has only its declared dependencies, so nothing else pulls in a rosbag2 storage plugin.
2. Set `snapshots.rosbag.enabled: true`.
3. Start the fault manager and trigger a fault.

**2. SIGSEGV in the storage plugin loader**

1. In one process, run two threads that each open and then close a `rosbag2_cpp::Writer` on the same storage id, in a loop.
2. Repeat for a few hundred iterations.

### Expected behavior

1. A bag file is written for the fault.
2. The loop finishes.

### Actual behavior

**1.** No bag is written. The node logs that no storage backend is usable and disables rosbag capture, so a feature the README leads with is silently off on a stock install.

The package declares `rosbag2_storage` and `libsqlite3-dev`. `rosbag2_storage` is only the plugin interface, and `libsqlite3-dev` is the C library the fault database uses. Neither is a storage backend. The two actual plugin packages, `rosbag2_storage_mcap` and `rosbag2_storage_default_plugins`, are not declared at all. Development machines and CI runners have them installed for other reasons, which is why this stayed invisible; the Humble binarydeb job is the only place the package meets a minimal dependency closure.

**2.** The process dies in `rcutils_unload_shared_library` with the instruction pointer at zero.

`rosbag2_cpp::Writer` loads its storage plugin inside `open()`, and `~Writer` can unload it. The state both touch is `class_loader`'s process-global library registry, so every writer of a given format shares one `rcutils_shared_library_t`. A close racing an open puts two threads into `dlclose` on the same handle: the winner closes it and zeroes the struct, the loser's `dlclose` returns "shared object not open" and then calls the now-null `allocator.deallocate`.

Measured with a program that does nothing but open/close loops on one format, four threads by 200 iterations: 20 of 20 runs failed on `mcap` and 20 of 20 on `sqlite3`. It belongs to neither backend, and it reproduces without any of our code, so the bug is upstream in `class_loader`. What we can do here is stop entering it.

### Environment

- ros2_medkit version: 0.6.0
- ROS 2 distro: reproduced on humble, jazzy and lyrical
- OS: Ubuntu 22.04 / 24.04 / 26.04

### Additional information

While fixing the first defect the default bag format is going back to `mcap`. It was moved to `sqlite3` earlier on the premise that the sqlite3 plugin always ships, and that premise is false for the same reason as above. `mcap` is what rosbag2 has defaulted to since Iron, and Foxglove and Lichtblick open it without a conversion step.

That is a breaking change: an installation that upgrades without setting `snapshots.rosbag.format` starts producing `.mcap` bags instead of `.db3`, and the bulk-data download filename changes with it.

bburda:
# Pull Request

## Summary

Two defects, both surfaced by the same build farm job once its launch-test noise was out of the way.

**1. Black-box rosbag capture had no storage plugin in its install closure.**

`rosbag2_storage` is the plugin interface and `libsqlite3-dev` is the C library the fault database uses. Neither provides a storage backend. An installed fault manager therefore found no backend, logged a warning and disabled black-box capture, which is a feature the README leads with. The Humble binarydeb job is the only place these packages meet a minimal dependency closure, which is why this had been invisible.

The default format returns to `mcap`. That is what rosbag2 has defaulted to since Iron and what Foxglove and Lichtblick open without a conversion step. This restores the default this project chose at 0.4.0; it was moved to `sqlite3` in a later fix on the stated premise that sqlite3 always ships, and that premise was false for the same reason as above. Both plugins are now exec and test dependencies, and the runtime stage of the image installs them, so the published image can write the format it advertises.

Storage resolution is symmetric now: whichever format is configured, if its plugin fails to load the other is probed and used, naming the missing package. Capture disables only when neither loads. Before, a configured `sqlite3` that failed to load disabled capture even when mcap was usable, and blamed a broken base install.

Both warnings name the package an operator installs (`ros-$ROS_DISTRO-rosbag2-storage-mcap`), not the ROS package name, which differs by the distro prefix and by underscore-versus-hyphen.

**2. A SIGSEGV in the storage plugin loader.**

`rosbag2_cpp::Writer` loads its storage plugin inside `open()`, and `~Writer` can unload it. The racing state is `class_loader`'s process-global library registry, so the crash is a double unload of one shared handle: the winner calls `dlclose` and zeroes the `rcutils_shared_library_t`, the loser's `dlclose` returns `shared object not open` and it then calls the now-null `allocator.deallocate`. The faulting frame is `rcutils_unload_shared_library`.

This is a bug in `class_loader`, not in this package; what changes here is that this package stops entering it. It reproduces on the merge base, and on both backends, so it is neither new nor caused by the default change above. It is worth reporting upstream, and a minimal reproduction exists that does not involve this code.

---

## Issue

Closes #619

---

## Type

- [x] Bug fix
- [x] New feature or tests
- [x] Breaking change
- [ ] Documentation only

---

## Testing

**The default format is proven on disk, not from a reported field.** A launch test starts a fault manager with no `snapshots.rosbag.format` parameter, triggers a fault, and requires a `.mcap` file in the resulting bag directory. A unit case pins the struct default separately, and a manifest test pins both plugin declarations, including that neither carries a `condition` attribute that would silently drop them on one distro.

**The loader race, measured.** A two-thread program that does nothing but open and close a writer on one storage id: 0 of 20 runs clean at baseline, on mcap and on sqlite3 alike. With the plugin mutex, 140 of 140 runs clean over 112000 operations per backend. A control with `open()` outside the lock still leaked failures, which is why it is inside. A regression test crosses two captures' plugin traffic in one process; a per-thread lock makes it crash, and a reversed lock order makes it hang, which ctest reports as a named timeout.

**The install hints are checked against the log, not against the source.** A test captures the rcutils output around the constructor for all three storage outcomes and requires the apt package name in each message. Two mutations were run against it: putting the ROS package names back in the disable warning, and dropping the package name from the fallback hint. Both fail the test.

**Cost.** The plugin mutex is held across a close: about 0.37 ms for an empty bag and about 1.1 ms for a 256 MB one, which is the largest a single flush can produce. `writer_mutex_` is no longer held across an open or a close, so a close cannot stall the post-roll callbacks.

**Suites.** `ros2_medkit_fault_manager` and the full lint set clean, locally, on this branch rebased on current `main`.

**What this cannot prove.** The dependency half is proven by the next Humble binarydeb run, not by a green suite on machines that have both plugins installed anyway.

---

## Note for reviewers

**Breaking:** an installation that upgrades without setting `snapshots.rosbag.format` starts producing `.mcap` bags instead of `.db3`, and the bulk-data download filename changes with it. Anything reading the old extension directly has to follow.

One test's measurement was corrected alongside: `test_rosbag_integration`'s helper counted messages arriving at the test node's own subscription and assumed the fault manager had buffered a comparable amount, but the fault manager buffers nothing while a post-fault recording is in flight. The assertion was right and the instrument feeding it was not.

---

## Checklist

### DIFF
--- a/src/ros2_medkit_fault_manager/src/fault_manager_node.cpp
+++ b/src/ros2_medkit_fault_manager/src/fault_manager_node.cpp
@@ -1114,7 +1114,7 @@ SnapshotConfig FaultManagerNode::create_snapshot_config() {
     config.rosbag.qos_match = declare_parameter<bool>("snapshots.rosbag.qos_match", true);
 
     config.rosbag.lazy_start = declare_parameter<bool>("snapshots.rosbag.lazy_start", false);
-    config.rosbag.format = declare_parameter<std::string>("snapshots.rosbag.format", "sqlite3");
+    config.rosbag.format = declare_parameter<std::string>("snapshots.rosbag.format", "mcap");
     config.rosbag.storage_path = declare_parameter<std::string>("snapshots.rosbag.storage_path", "");
 
     int64_t max_bag_size = declare_parameter<int64_t>("snapshots.rosbag.max_bag_size_mb", 50);
--- a/src/ros2_medkit_fault_manager/test/test_rosbag_capture.cpp
+++ b/src/ros2_medkit_fault_manager/test/test_rosbag_capture.cpp
@@ -119,9 +119,27 @@ TEST_F(RosbagCaptureTest, ConstructorWithDisabledRosbag) {
 }
 
 // @verifies REQ_INTEROP_088
+TEST_F(RosbagCaptureTest, DefaultFormatIsMcap) {
+  // mcap is what Foxglove and Lichtblick open without conversion, and what
+  // rosbag2 itself defaults to since Iron. A silent drift back to sqlite3 would
+  // hand operators a black box their viewer cannot read.
+  RosbagConfig defaults;
+  EXPECT_EQ(defaults.format, "mcap");
+}
+
+// @verifies REQ_INTEROP_088
 TEST_F(RosbagCaptureTest, ConstructorFallsBackOnUnknownFormat) {
-  // An unknown/unavailable format must NOT terminate the node; it degrades to
-  // sqlite3 (always shipped with rosbag2) and capture stays enabled.
+  // An unknown format string must NOT terminate the node. It normalises to
+  // "sqlite3" first, then goes through the same real probe as an explicit
+  // sqlite3 configuration would, so this test uses the real probe on
+  // purpose - unlike the crash-safety tests below, which inject a double.
+  // That means the resolved format is not fixed: on a host with only mcap
+  // installed and not sqlite3 (the buildfarm chroot before
+  // rosbag2_storage_default_plugins lands as a real dependency), resolution
+  // falls back to mcap instead, which is the fallback being symmetric doing
+  // exactly its job, not a failure. The contract this test is actually about
+  // is that capture stays enabled and resolves to one of the two real
+  // backends, not "invalid_format" - not which one that happens to be here.
   auto rosbag_config = create_rosbag_config();
   rosbag_config.format = "invalid_format";
   auto snapshot_config = create_snapshot_config();
@@ -129,7 +147,8 @@ TEST_F(RosbagCaptureTest, ConstructorFallsBackOnUnknownFormat) {
   EXPECT_NO_THROW(rb = std::make_shared<RosbagCapture>(node_.get(), storage_.get(), rosbag_config, snapshot_config));
   ASSERT_NE(rb, nullptr);
   EXPECT_TRUE(rb->is_enabled());
-  EXPECT_EQ(rb->config().format, "sqlite3");
+  EXPECT_TRUE(rb->config().format == "sqlite3" || rb->config().format == "mcap")
+      << "resolved format was '" << rb->config().format << "', neither a real backend";
 }
 
 // The crash-safety branches below force the storage probe via an injected double,
@@ -157,6 +176,29 @@ TEST_F(RosbagCaptureTest, ConfiguredFormatUnavailableFallsBackToSqlite3) {
 }
 
 // @verifies REQ_INTEROP_088
+TEST_F(RosbagCaptureTest, ConfiguredSqlite3UnavailableFallsBackToMcap) {
+  // The symmetric case of the fallback above: a known format (sqlite3) whose
+  // plugin is unavailable degrades to mcap and capture stays enabled. Neither
+  // backend is privileged - whichever is configured falls back to the other,
+  // not always to sqlite3.
+  auto rosbag_config = create_rosbag_config();
+  rosbag_config.format = "sqlite3";
+  auto snapshot_config = create_snapshot_config();
+  RosbagCapture::StorageProbeFn probe = [](const std::string & f) -> std::optional<std::string> {
+    if (f == "sqlite3") {
+      return std::string("simulated: sqlite3 plugin not found");
+    }
+    return std::nullopt;  // mcap usable
+  };
+  std::shared_ptr<RosbagCapture> rb;
+  EXPECT_NO_THROW(
+      rb = std::make_shared<RosbagCapture>(node_.get(), storage_.get(), rosbag_config, snapshot_config, probe));
+  ASSERT_NE(rb, nullptr);
+  EXPECT_TRUE(rb->is_enabled());
+  EXPECT_EQ(rb->config().format, "mcap");
+}
+
+// @verifies REQ_INTEROP_088
 TEST_F(RosbagCaptureTest, NoUsableBackendDisablesCaptureWithoutCrashing) {
   // When neither the configured format nor sqlite3 is usable, capture self-disables
   // and the node keeps running (no throw out of the constructor).
@@ -175,8 +217,10 @@ TEST_F(RosbagCaptureTest, NoUsableBackendDisablesCaptureWithoutCrashing) {
 
 // @verifies REQ_INTEROP_088
 TEST_F(RosbagCaptureTest, Sqlite3BaselineUnavailableDisablesCapture) {
-  // When the always-shipped sqlite3 baseline itself fails to load, capture
-  // self-disables rather than crashing the node.
+  // When sqlite3 is configured and neither it nor its mcap fallback loads,
+  // capture self-disables rather than crashing the node - the symmetric case
+  // of NoUsableBackendDisablesCaptureWithoutCrashing above, starting from
+  // sqlite3 instead of mcap.
   auto rosbag_config = create_rosbag_config();
   rosbag_config.format = "sqlite3";
   auto snapshot_config = create_snapshot_config();
--- a/src/ros2_medkit_fault_manager/test/test_rosbag_capture.cpp
+++ b/src/ros2_medkit_fault_manager/test/test_rosbag_capture.cpp
@@ -14,14 +14,18 @@
 
 #include <gtest/gtest.h>
 
+#include <algorithm>
 #include <atomic>
 #include <chrono>
+#include <cstdarg>
 #include <cstdint>
+#include <cstdio>
 #include <cstdlib>
 #include <filesystem>
 #include <fstream>
 #include <limits>
 #include <memory>
+#include <mutex>
 #include <optional>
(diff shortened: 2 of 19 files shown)


src/ros2_medkit_fault_manager/src/fault_manager_node.cpp
```diff
@@ -1114,7 +1114,7 @@ SnapshotConfig FaultManagerNode::create_snapshot_config() {
     config.rosbag.qos_match = declare_parameter<bool>("snapshots.rosbag.qos_match", true);
 
     config.rosbag.lazy_start = declare_parameter<bool>("snapshots.rosbag.lazy_start", false);
-    config.rosbag.format = declare_parameter<std::string>("snapshots.rosbag.format", "sqlite3");
+    config.rosbag.format = declare_parameter<std::string>("snapshots.rosbag.format", "mcap");
```
mfaferek93, line 1117: This is the breaking bit, and no CHANGELOG is touched by any of the three commits (`git diff --name-only 77efd37c..ee73ba86 | grep -i changelog` is empty). `Forthcoming` exists and is populated at `CHANGELOG.rst:5-7`, so the section is being kept current otherwise.

Worth doing here rather than at release time, because this is the third flip of this same default and the second one with no entry:

- `d9e03491` (closes #251) moved it TO mcap because the buildfarm chroot had no sqlite3 plugin, and promoted `rosbag2_storage_mcap` to `depend`.
- `e526ce48` moved it BACK to sqlite3 and demoted mcap to `test_depend`, re-breaking #251. It shipped under 0.6.0 as "Made rosbag capture enablement crash-safe (#430)" with no mention of the default. That omission is how it got through.

There is also a record to correct while you are in there. `fault_manager/CHANGELOG.rst:28` and `gateway/CHANGELOG.rst:137` already say, under 0.4.0, "Default rosbag storage format changed from ``sqlite3`` to ``mcap``". At the merge base that was false: this line read `"sqlite3"` and `snapshot_capture.hpp:69` read `format{"sqlite3"}`. So the released changelog has been claiming mcap for two releases while the code shipped sqlite3. This PR makes it true again.

Operator-visible effect with no note anywhere: bags land as `.mcap` instead of `.db3`, and the release notes are the only migration surface.

src/ros2_medkit_fault_manager/test/test_rosbag_capture.cpp
```diff
@@ -118,18 +237,37 @@ TEST_F(RosbagCaptureTest, ConstructorWithDisabledRosbag) {
...
+  // backends, not "invalid_format" - not which one that happens to be here.
   auto rosbag_config = create_rosbag_config();
   rosbag_config.format = "invalid_format";
   auto snapshot_config = create_snapshot_config();
   std::shared_ptr<RosbagCapture> rb;
   EXPECT_NO_THROW(rb = std::make_shared<RosbagCapture>(node_.get(), storage_.get(), rosbag_config, snapshot_config));
   ASSERT_NE(rb, nullptr);
   EXPECT_TRUE(rb->is_enabled());
-  EXPECT_EQ(rb->config().format, "sqlite3");
+  EXPECT_TRUE(rb->config().format == "sqlite3" || rb->config().format == "mcap")
```
mfaferek93: This can't fail. Resolution structurally only ever ends in one of those two strings or with capture disabled, so the assertion is a tautology.

Mutation traced: delete the normalization at `rosbag_capture.cpp:255-258`. With `format = "invalid_format"` the probe fails at :260, `other_format` computes to `"sqlite3"` at :262 (because `"invalid_format" != "sqlite3"`), sqlite3 probes clean, :283 sets the format to `"sqlite3"`, capture stays enabled. Both assertions here still pass.

That leaves the rule you state at `rosbag_capture.cpp:249` and repeat in the pytest docstring - "An unknown format string falls back to sqlite3" - with no coverage anywhere. The two paths are operator-visibly different too: normalization warns "Unknown rosbag storage format '%s'; using 'sqlite3'", whereas the fallback warns "is unavailable ... Falling back" and runs the bogus name through `storage_plugin_package()`.

The reason given for relaxing it at :255-258 ("on a host with only mcap installed and not sqlite3") doesn't hold either - the same binary already hard-requires both plugins at :1625, and this PR adds both as `test_depend` so that it can.

Inject a probe returning `std::nullopt` for everything and `EXPECT_EQ(rb->config().format, "sqlite3")` becomes exact and environment-independent, same shape as every other case in this file.

src/ros2_medkit_fault_manager/test/test_rosbag_capture.cpp
```diff
@@ -14,12 +14,18 @@
...
+#include <cstdio>
+#include <cstdlib>
 #include <filesystem>
 #include <fstream>
 #include <limits>
 #include <memory>
+#include <mutex>
 #include <optional>
 #include <sstream>
 #include <string>
```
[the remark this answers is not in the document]
bburda: You are right, the file was reaching that header only through `rclcpp.hpp`. Added `#include "rcutils/logging.h"` next to it. The test I took the capture class from declares it the same way.

src/ros2_medkit_fault_manager/test/test_rosbag_default_format.test.py
```diff
@@ -0,0 +1,362 @@
...
+import tempfile
+import time
+import unittest
+
+from launch import LaunchDescription
+import launch.actions
+import launch_ros.actions
+import launch_testing.actions
+import launch_testing.markers
+import rclpy
```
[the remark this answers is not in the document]
bburda: ## 3774605305 | Copilot | test_rosbag_default_format.test.py:34

Removed.

One correction on the reason you gave: flake8 does not flag this and did not flag it here. `import a.b` binds the name `a`, and `launch_testing` is used further down the file, so pyflakes counts the binding as used. The import is still dead weight, so it is gone.

It is a copy-paste pattern rather than a convention. Nine launch tests in this repo import `launch_testing.markers` and one of them uses it. I also checked that nothing depended on it indirectly: `launch_testing` exports `asserts` from its own `__init__`, so the `assertExitCodes` call below does not need the import.

src/ros2_medkit_fault_manager/src/fault_manager_node.cpp
bburda: ## 3776852936 | mfaferek93 | fault_manager_node.cpp:1117

I checked each part of what you traced and it holds. `Forthcoming` is populated, both `ros2_medkit_fault_manager` and `ros2_medkit_gateway` claim the mcap default under 0.4.0, and at the merge base `snapshot_capture.hpp:69` reads `format{"sqlite3"}`. So the released changelog has been claiming mcap for two releases while the shipped default was sqlite3.

We fill changelogs in the release branch, at release time, rather than per PR, so I am not adding the entry here. To make sure it is not lost the way it was last time, both items are now written into the 0.7.0 release issue: the entry for the default going back to mcap, and the decision on the 0.4.0 lines, which is either correcting them in place or recording the reversal under 0.7.0.

https://github.com/selfpatch/ros2_medkit/issues/606

src/ros2_medkit_fault_manager/test/test_rosbag_capture.cpp
bburda: ## 3776852952 | mfaferek93 | test_rosbag_capture.cpp:269

You are right, and I ran the mutation you traced. With the normalisation deleted the old case still passed for exactly the reason you give: the probe for `invalid_format` fails, `other_format` computes to `sqlite3`, and resolution lands on the same value normalisation would have produced. The documented rule had no coverage anywhere.

The case now injects a probe that reports every backend usable, so normalisation is the only thing that can move the format, and it asserts `sqlite3` exactly. With the normalisation deleted it now fails at line 273.

The reason I had given for relaxing it does not hold either, as you say: both plugins are `test_depend` now, and the parametrised bag test in the same binary already requires both, so the mcap-only host it was written for cannot occur.

mfaferek93 (approved):

PR #618, opened 2026-08-13
Describe occurrence_count as the edge counter it is

### LINKED ISSUE #617: [BUG] occurrence_count is documented as counting every FAILED event; it counts episodes
# Bug report

### Steps to reproduce
1. Report the same `fault_code` from the same `source_id` with `event_type: 0 (FAILED)` three
   times, one second apart.
2. `GET /api/v1/functions/<id>/faults`.
3. Read `Fault.msg`, `src/ros2_medkit_msgs/README.md` or `docs/api/messages.rst` to find out what
   the field you just read is supposed to mean.

### Expected behavior

The documented meaning of `occurrence_count` matches what the fault manager computes.

### Actual behavior

`occurrence_count` stays at 1 while `last_occurred` advances on every report.

That behaviour is correct and intended. Commit `67bdb9cb` ("fault_manager: count occurrences on
edges, not per report") made the counter count episodes. It increments only on a real edge: a new
fault, or a new FAILED after the fault was CLEARED. Tests hold it in place, including
`FaultStorageTest.SameSourceMultipleReportsConfirms` (three FAILED from one source, expects 1) and
`ClearedFaultCanBeReactivated` (expects 2).

The documentation was never updated and now describes a different field:

| Location | Claim | Status |
|---|---|---|
| `src/ros2_medkit_msgs/msg/Fault.msg:45` | "Total number of FAILED events aggregated across all sources" | wrong |
| `src/ros2_medkit_msgs/README.md:27` | "Total FAILED events aggregated across all sources" | wrong |
| `docs/api/messages.rst:42` | repeats the same comment | wrong |
| `src/ros2_medkit_fault_manager/README.md:187-188` | "`occurrence_count` and `last_occurred` still update meanwhile" | contradicts `fault_storage.hpp:40-41`, and the header is the correct one |
| `src/ros2_medkit_fault_manager/src/fault_manager_node.cpp:662` | `EVENT_UPDATED` described as "data updated (occurrence_count, sources, etc.)" | the counter is not updated on that path |

A reader who trusts any of the first three will size a fault's activity using a field that answers
a different question, and the docs will not correct them, because the edge-counting rule is only
written in source comments (`fault_storage.hpp:40-41`, `:89`, `fault_storage.cpp:187-190`,
`sqlite_fault_storage.cpp:456-458`).

### Environment
- ROS 2 distro: Jazzy
- OS: Ubuntu 24.04

### Additional information

Write sites, for whoever fixes the wording: `src/ros2_medkit_fault_manager/src/fault_storage.cpp:130`
(new fault) and `:162-163` (reactivation after CLEARED), with the SQLite equivalents at
`sqlite_fault_storage.cpp:573` and `:459-462`. The FAILED branch for an existing fault that is not
CLEARED deliberately leaves it alone (`fault_storage.cpp:187-190`).

Since the field answers "how many episodes", the docs should say that. It is also worth saying
what does answer "how many reports", which today is only the event stream.

bburda:
# Pull Request

## Summary

`occurrence_count` counts occurrence edges: the first FAILED event, and each FAILED that arrives
while the fault is CLEARED. A repeated FAILED inside one occurrence does not change it and
advances `last_occurred` instead. That has been the behaviour since "fault_manager: count
occurrences on edges, not per report", and tests hold it in place, but the documentation was
never updated.

Eight places described it wrongly, in three groups.

**Three said it counts every FAILED event.** `Fault.msg`, the `ros2_medkit_msgs` README and the
API reference. A reader who trusts them will size a fault's activity with a field that answers a
different question.

**Three listed it as something `fault_updated` carries.** `FaultEvent.msg`, the gateway README
and the comment in the report handler. A re-report inside one occurrence emits `fault_updated`
but does not touch the counter.

**Two contradicted the storage header outright.** The fault_manager README and design doc said
`occurrence_count` and `last_occurred` both keep updating during the latch window, while
`FaultStorage`'s own header says `last_occurred` does and `occurrence_count` does not. Both now
use the header's wording, so the three descriptions of that window read the same.

`first_occurred` turned out to be wrong in the same way, and is fixed here too. It is reset when a
FAILED event reactivates a CLEARED fault, so it marks the start of the latest occurrence, not the
first report ever. Both backends do that deliberately (`fault_storage.cpp:155-158`,
`sqlite_fault_storage.cpp:437-441`, written back at `:485-492`) and both pin it in tests
(`test_fault_manager.cpp:402`/`423`, `test_sqlite_storage.cpp:480`/`501`). It matters more after
this PR rather than less: with `occurrence_count` now trustworthy as an edge counter, the obvious
next step is to pair the two into a rate, and that arithmetic is wrong precisely because the
timestamp moved with the last edge.

No behaviour changes. The only change under `src/` that is not a document is one comment.

---

## Issue

- closes #617

---

## Type

- [ ] Bug fix
- [ ] New feature or tests
- [ ] Breaking change
- [x] Documentation only

---

## Testing

`ros2_medkit_msgs` builds clean from an empty cache, so both edited `.msg` files still parse and
generate. `docs/api/messages.rst` and `fault_manager/design/index.rst` parse under docutils. The
`Fault.msg` block copied into `messages.rst` is byte-identical to the file again.

For a reviewer the useful check is the claim itself rather than the build. The write sites are
`fault_storage.cpp:130` (new fault) and `:162-163` (reactivation after CLEARED), with the SQLite
equivalents at `sqlite_fault_storage.cpp:573` and `:459-462`. The FAILED branch for an existing
non-CLEARED fault deliberately leaves the counter alone at `fault_storage.cpp:187-190`.
`FaultStorageTest.SameSourceMultipleReportsConfirms` and `ClearedFaultCanBeReactivated` pin both
directions.

I could not compile `ros2_medkit_fault_manager` locally, because the image I had available is a
runtime image without `nlohmann_json`. The change there is a comment inside an existing branch,
so CI is the check.

---

## Checklist

Notes on the last three: nothing breaks, since no field, type or behaviour changes. No tests were
added because nothing asserts these strings and the behaviour they now describe is already
covered by the tests named above. The documentation is the change.

One thing this does not fix. `docs/api/messages.rst` is a hand-made copy of twelve interface
files inside `code-block:: text`, which is how this drift happened. A `literalinclude` would need
start markers in every `.msg` and `.srv`, since they all open with the licence header, so it is a
larger change than this issue and I left it alone. Worth deciding separately.

### DIFF
--- a/src/ros2_medkit_msgs/README.md
+++ b/src/ros2_medkit_msgs/README.md
@@ -24,7 +24,7 @@ Core fault data model representing an aggregated fault condition with AUTOSAR DE
 | `first_occurred` | builtin_interfaces/Time | When fault was first reported |
 | `last_occurred` | builtin_interfaces/Time | When fault last occurred (FAILED events only) |
 | `last_passed` | builtin_interfaces/Time | When fault last reported PASSED (zero = never) |
-| `occurrence_count` | uint32 | Total FAILED events aggregated across all sources |
+| `occurrence_count` | uint32 | Times this fault has occurred, counted on edges (first FAILED, then each FAILED that arrives while CLEARED). Repeats within one occurrence do not increment it |
 | `status` | string | Current status (see STATUS_* constants) |
 | `reporting_sources` | string[] | List of source identifiers that reported this fault |
 
--- a/src/ros2_medkit_msgs/README.md
+++ b/src/ros2_medkit_msgs/README.md
@@ -21,7 +21,7 @@ Core fault data model representing an aggregated fault condition with AUTOSAR DE
 | `fault_code` | string | Global fault identifier (e.g., "MOTOR_OVERHEAT") |
 | `severity` | uint8 | Severity level (use SEVERITY_* constants) |
 | `description` | string | Human-readable description |
-| `first_occurred` | builtin_interfaces/Time | When fault was first reported |
+| `first_occurred` | builtin_interfaces/Time | When the current occurrence started; reset when a FAILED event reactivates a CLEARED fault, so it moves with `occurrence_count` |
 | `last_occurred` | builtin_interfaces/Time | When fault last occurred (FAILED events only) |
 | `last_passed` | builtin_interfaces/Time | When fault last reported PASSED (zero = never) |
 | `occurrence_count` | uint32 | Times this fault has occurred, counted on edges (first FAILED, then each FAILED that arrives while CLEARED). Repeats within one occurrence do not increment it |
--- a/docs/api/messages.rst
+++ b/docs/api/messages.rst
@@ -28,7 +28,9 @@ Core fault data model representing an aggregated fault condition.
    # Human-readable description of the fault condition
    string description
 
-   # Timestamp when this fault was first reported
+   # Timestamp when the current occurrence started (reset when a FAILED event
+   # reactivates a CLEARED fault, so it moves with occurrence_count instead of
+   # marking the fault's first report ever)
    builtin_interfaces/Time first_occurred
 
    # Timestamp when this fault last occurred (FAILED events only; a PASSED
--- a/docs/api/messages.rst
+++ b/docs/api/messages.rst
@@ -38,7 +38,10 @@ Core fault data model representing an aggregated fault condition.
    # Timestamp of the last PASSED event reported for this fault (zero = never)
    builtin_interfaces/Time last_passed
 
-   # Total number of FAILED events aggregated across all sources
+   # Number of times this fault has occurred, counted on edges: one for the first
+   # FAILED event, and one more each time a FAILED event arrives while the fault is
+   # CLEARED. Repeated FAILED events within one occurrence do not increment it;
+   # last_occurred is what advances on those.
    uint32 occurrence_count
 
    # Current fault status (PREFAILED, PREPASSED, CONFIRMED, HEALED, CLEARED)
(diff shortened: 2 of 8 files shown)


src/ros2_medkit_msgs/README.md
```diff
@@ -24,7 +24,7 @@ Core fault data model representing an aggregated fault condition with AUTOSAR DE
 | `first_occurred` | builtin_interfaces/Time | When fault was first reported |
```
mfaferek93: Row above is wrong in the same way, and the code says so out loud.

Both backends reset `first_occurred` on every reactivation, deliberately:

```cpp
// fault_storage.cpp:155-158
// FAILED event reactivates the fault - reset debounce counter and first-seen:
// this is a new outage cycle, not a continuation of the one that just cleared.
state.first_occurred = timestamp;
```

`sqlite_fault_storage.cpp:438-441` does the same and writes it back in the UPDATE at :485-488. So after five occurrences this is the start of the fifth, not the first. The backends agree, so the behaviour is intended and only the wording is off.

It matters more after this PR, not less: `occurrence_count` is now trustworthy as an edge counter, so a reader will pair the two to get a rate ("five occurrences since first_occurred") and that arithmetic is wrong precisely because the timestamp moved with the last edge. Same failure mode your PR body describes for the counter.

Same sentence in two more places, both outside this diff: `Fault.msg:35` and `docs/api/messages.rst:31`.

Something like "when the current occurrence started; reset when a CLEARED fault is reactivated" would match the code. Entirely fair to defer this since #617 is scoped to `occurrence_count` - flagging it because it is a one-line fix in a table you are already editing, and an audit that stops one row short tends not to get revisited.

bburda: Good catch! Fixed!

mfaferek93 (approved):

Issue #621, opened 2026-08-16
Pass the hardware_id of the diagnostic message as the source id in the diagnostic bridge


nnarain:
## Summary

In the systems I work on, we are still primarily on ROS 1 (I know, I know).

This means most of our nodes are in ROS 1. We bridge various topics to the ROS 2 system and as mentioned in #527 we have an effectively equivalent method of reporting fault codes through ros diagnostics. So I'm primarily relying on the diagnostic stream for the initial stages of integrating medkit into our robots.

When setting up the diagnostic bridge on our systems, I found that everything is reported under the `/diagnostic_bridge` application. Which makes sense in the current implementation. However if I'm defining components/apps for the hardware on our robots such as lidars, cameras, etc, then I'd expect to see the fault codes we have defined appear under "Faults" tab in the web ui for those specific devices.

---

## Proposed solution (optional)

Pass the `hardware_id` of the diagnostic message through as the source id in the diagnostic bridge.

Now it's not necessarily the case that the `hardware_id` will match the node name but this feels like a reasonable incremental step (and I have thoughts on additions to this). And it's pretty easy to make the hardware ID node name.

---

## Additional context (optional)

I'll have a PR up momentarily :)

PR #622, opened 2026-08-16
Pass diagnostic hardware_id as the source ID in the diagnostic bridge

nnarain:
# Pull Request

## Summary

Following the same pattern as in the log bridge. I'm setting up multiple fault reporters in the diagnostic bridge based on the hardware_id in the diagnostic message.

This does assume the `hardware_id` is the fully qualified node name, in order for it to be mapped in to the correct entity. But it think it's a reasonable step forward.

Additional things might be:

- adding a mapping from hardware_id to node name (either for this node, or in the manifest)

---

## Issue

- #621 

---

## Type

- [ ] Bug fix
- [x] New feature or tests
- [ ] Breaking change
- [ ] Documentation only

---

## Testing

I added two tests. And tested it locally using the web ui.

image

image

---

## Checklist

### DIFF
--- a/src/ros2_medkit_diagnostic_bridge/src/diagnostic_bridge_node.cpp
+++ b/src/ros2_medkit_diagnostic_bridge/src/diagnostic_bridge_node.cpp
@@ -117,10 +117,13 @@ std::string DiagnosticBridgeNode::source_id_for(const diagnostic_msgs::msg::Diag
     return status.hardware_id;
   }
 
-  RCLCPP_WARN_THROTTLE(
-      get_logger(), *get_clock(), 10000,
-      "Diagnostic '%s' hardware_id '%s' is not a slash-containing node ID, using bridge source_id '%s'",
-      status.name.c_str(), status.hardware_id.c_str(), fqn.c_str());
+  // Not RCLCPP_WARN_THROTTLE: this method is const, so get_clock() yields a const
+  // Clock, and Clock::now() is non-const on Humble. Once per process is also the
+  // right cadence here - the throttle keeps one static per call site, so it would
+  // name a single diagnostic every 10s and never the others.
+  RCLCPP_WARN_ONCE(get_logger(),
+                   "Diagnostic '%s' hardware_id '%s' is not a slash-containing node ID, using bridge source_id '%s'",
+                   status.name.c_str(), status.hardware_id.c_str(), fqn.c_str());
   return fqn;
 }
 
--- a/src/ros2_medkit_diagnostic_bridge/src/diagnostic_bridge_node.cpp
+++ b/src/ros2_medkit_diagnostic_bridge/src/diagnostic_bridge_node.cpp
@@ -16,6 +16,8 @@
 
 #include <algorithm>
 #include <cctype>
+#include <cinttypes>
+#include <limits>
 
 #include "ros2_medkit_msgs/msg/fault.hpp"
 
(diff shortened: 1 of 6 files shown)


src/ros2_medkit_diagnostic_bridge/src/diagnostic_bridge_node.cpp
```diff
@@ -60,18 +60,45 @@ void DiagnosticBridgeNode::load_parameters() {
...
+  auto it = reporters_.find(source_id);
+  if (it != reporters_.end()) {
+    return it->second.get();
+  }
+
+  // Multiple FaultReporters on one node are safe: FaultReporter guards
+  // parameter declaration with has_parameter().
+  auto reporter = std::make_unique<ros2_medkit_fault_reporter::FaultReporter>(this->shared_from_this(), source_id);
+  auto * raw = reporter.get();
+  reporters_[source_id] = std::move(reporter);
```
mfaferek93: This map only grows. The log bridge has the same map but caps it: max_tracked_nodes (default 512), oldest one evicted, and a tracked_reporter_count() so a test can check it. Worth copying that part too. 

Each reporter holds a service client, and drivers often put serial numbers or device paths in hardware_id, so on a long-running node this can add up.

src/ros2_medkit_diagnostic_bridge/src/diagnostic_bridge_node.cpp
```diff
@@ -60,18 +60,45 @@ void DiagnosticBridgeNode::load_parameters() {
...
+  // Use a mutable clock copy - Humble's RCLCPP_WARN_THROTTLE requires non-const Clock.
+  rclcpp::Clock clock(*get_clock());
+  RCLCPP_WARN_THROTTLE(get_logger(), clock, 10000, "Diagnostic '%s' has empty hardware_id, using bridge source_id '%s'",
+                       status.name.c_str(), get_fully_qualified_name());
+  return get_fully_qualified_name();
+}
+
 void DiagnosticBridgeNode::process_diagnostic(const diagnostic_msgs::msg::DiagnosticStatus & status) {
+  const std::string source_id = source_id_for(status);
+  auto * reporter = reporter_for(source_id);
```
mfaferek93: This line runs before `map_to_fault_code`. So a diagnostic that maps to nothing, and gets skipped two lines below, still creates a reporter. Moving this under the `fault_code.empty()` check would keep ignored diagnostics free.

src/ros2_medkit_diagnostic_bridge/src/diagnostic_bridge_node.cpp
```diff
@@ -60,18 +60,45 @@ void DiagnosticBridgeNode::load_parameters() {
...
+  if (reporter == nullptr) {
```
mfaferek93: This can't happen, right? `make_unique` throws instead of returning null.

mfaferek93 (commented):
Good direction! Everything landing under `/diagnostic_bridge` is a real problem, and source attribution is the right fix.

Two small things first: the reporter map needs the log bridge's cap, and the reporter should be created after the mapping check.

The bigger question is the hardware_id assumption you already noted. `diagnostic_updater` usually puts a serial number or none there, not a node name. 
When that happens the fault gets a source that matches no entity. It then disappears from every per-entity faults list and stays only in the global /faults, which has no entity_id or href to click through. Right now those faults at least land under `/diagnostic_bridge`, which is a real entity. So on many systems this would swap "shown in the wrong place" for "not shown anywhere".

The manifest mapping you suggested solves that. Until it exists, it may be safer to keep the current behaviour by default and put the hardware_id path behind a parameter, so nobody loses faults after an upgrade.

This part is hard to see from the bridge as the scope rule lives in the gateway, far from what you were touching 😅

src/ros2_medkit_diagnostic_bridge/src/diagnostic_bridge_node.cpp
nnarain: `get()` doesn't throw.

https://en.cppreference.com/cpp/memory/unique_ptr/get

nnarain:
Ya I'll put this feature behind a flag

src/ros2_medkit_diagnostic_bridge/src/diagnostic_bridge_node.cpp
mfaferek93: ahh, true

src/ros2_medkit_diagnostic_bridge/src/diagnostic_bridge_node.cpp
```diff
@@ -60,17 +73,57 @@ void DiagnosticBridgeNode::load_parameters() {
...
+  return raw;
+}
+
+size_t DiagnosticBridgeNode::tracked_reporter_count() {
+  std::lock_guard<std::mutex> lock(reporters_mutex_);
+  return reporters_.size();
+}
+
+std::string DiagnosticBridgeNode::source_id_for(const diagnostic_msgs::msg::DiagnosticStatus & status) const {
+  if (use_hardware_id_as_source_id_ && !status.hardware_id.empty() && status.hardware_id.find('/') != std::string::npos) {
```
mfaferek93: Only thing left red in CI. clang-format wants a wrap here.

src/ros2_medkit_diagnostic_bridge/src/diagnostic_bridge_node.cpp
```diff
@@ -60,17 +73,57 @@ void DiagnosticBridgeNode::load_parameters() {
...
+
+  const std::string fqn = get_fully_qualified_name();
+
+  // Use a mutable clock copy - Humble's RCLCPP_WARN_THROTTLE requires non-const Clock.
+  rclcpp::Clock clock(*get_clock());
+  RCLCPP_WARN_THROTTLE(
+      get_logger(), clock, 10000,
+      "Diagnostic '%s' hardware_id '%s' is not a slash-containing node ID or attribution is disabled, using bridge "
+      "source_id '%s'",
+      status.name.c_str(), status.hardware_id.c_str(), fqn.c_str());
```
mfaferek93: With the flag off, which is the default, the `if` above is always false, so every diagnostic reaches this warning. People who upgrade and change nothing get a warning every 10 seconds about a feature they never turned on. When the flag is on it is worth one line, and then you don't need the clock copy at all :)

mfaferek93 (commented):
one more minor and one small CI fail

mfaferek93 (approved):

PR #623, opened 2026-08-16
Keep more than one rosbag per fault, and stop losing the evidence around it

### LINKED ISSUE #620: Is there some way to retain more than one bag per fault_code?
## Summary

I'm trying to configure the fault manager to keep more than one rosbag capture for repeated
occurrences of the same fault. Is there a way to define a max number of bags per fault_code
(e.g. keep the last N) before older ones for that code are deleted? Right now it seems only
one bag is ever kept per `fault_code`.

## Environment

- ROS 2 distro: **Humble** (Ubuntu 22.04.5 LTS)
- `ros-humble-ros2-medkit-fault-manager`: `0.6.0-1jammy.20260726.134439`
- Storage backend: `sqlite`

## Current behavior

Every time a fault **re-confirms** (e.g. `PREFAILED → CONFIRMED` again after healing, or
`CONFIRMED` while occurrence data updates), `RosbagCapture` writes a new bag and calls
`FaultStorage::store_rosbag_file()`. That function deletes the previously stored bag for the
same `fault_code` before recording the new one — so only the **most recent** bag per fault
code is ever retained, regardless of `max_bag_size_mb` / `max_total_storage_mb`:

```cpp
// fault_storage.cpp — InMemoryFaultStorage::store_rosbag_file()
void InMemoryFaultStorage::store_rosbag_file(const RosbagFileInfo & info) {
  // Delete existing bag file if present (prevent orphaned files on re-confirm)
  auto it = rosbag_files_.find(info.fault_code);
  if (it != rosbag_files_.end()) {
    if (it->second.file_path != info.file_path) {
      std::filesystem::remove_all(it->second.file_path, ec);   // <-- previous bag for this code is deleted here
    }
  }
  rosbag_files_[info.fault_code] = info;
}
```

The SQLite backend does the same via `DELETE FROM rosbag_files WHERE fault_code = ?` in
`SqliteFaultStorage::delete_rosbag_file()`, invoked the same way on each new capture for that
code.

Importantly, **this is separate from `auto_cleanup`**: `auto_cleanup` only deletes a bag when
the fault is *cleared* (`RosbagCapture::on_fault_cleared`). The one-bag-per-fault_code deletion
above happens unconditionally on every re-confirmation, so setting `auto_cleanup: false` does
not help retain history.

## Steps to reproduce

1. Enable rosbag capture with the config below (attached in full).
2. Trigger a fault that confirms, heals, and re-confirms multiple times (e.g. a flapping
   `SYSTEM_CPU`/sensor fault with `healing_enabled: true`).
3. Observe the log: each re-confirmation logs `Bag file completed: <path>` at a new
   `fault_{CODE}_{timestamp}` path, and the previous bag directory for that same fault code is
   gone from disk.

## Expected / desired behavior

A way to retain a bounded **history** of bags per fault_code instead of only the latest — e.g.
a `snapshots.rosbag.max_bags_per_fault` (or similar) parameter, analogous to how
`max_per_fault` already works for JSON snapshots. Something like:

```yaml
rosbag:
  enabled: true
  max_bags_per_fault: 5   # keep the last 5 bags for a given fault_code; delete oldest beyond that
```

so a fault that recurs (heals and re-confirms) leaves a trail of captures instead of only ever
showing the most recent occurrence.

## Config file used

```yaml
# Snapshot Capture Configuration
#
# This file configures which topics to capture when specific faults are confirmed.
# The captured data provides debugging context at the moment of fault occurrence.
#
# Topic resolution priority:
#   1. fault_specific - exact fault code match (highest priority)
#   2. patterns - regex pattern match on fault code
#   3. default_topics - fallback for all other faults (lowest priority)
#
# Usage:
#   ros2 run ros2_medkit_fault_manager fault_manager_node \
#     --ros-args -p snapshots.config_file:=/path/to/snapshots.yaml

fault_specific:
  MOTOR_OVERHEAT:
    - /joint_states
    - /motor/temperature
    - /motor/current
  BATTERY_LOW:
    - /battery_state
    - /power/consumption
  LIDAR_COMM_FAILURE:
    - /perception/lidar/scan
    - /perception/lidar/diagnostics

patterns:
  "MOTOR_.*":
    - /joint_states
    - /cmd_vel
  "SENSOR_.*":
    - /diagnostics
    - /sensor_data
  ".*_COMM_FAILURE":
    - /diagnostics

default_topics:
  - /diagnostics
  - /rosout

rosbag:
        enabled: true                  # Enable rosbag recording
        duration_sec: 5.0              # Pre-fault buffer duration
        duration_after_sec: 1.0        # Post-fault recording duration
        topics: "config"               # Topic selection: "entity" (default), "config", "all", "explicit"
        # include_topics: []             # Additional topics to include
        # exclude_topics: []             # Topics to exclude
        exclude_sensor_topics: false   # Auto-exclude image/points/depth/compressed in broad modes
        lazy_start: false              # Start recording on first fault
        format: "sqlite3"              # Storage format
        qos_match: true                # Match each topic's publisher QoS

        # Where the flushed bags are written. Empty ("") falls back to the system temp
        # dir (/tmp); set an explicit path to keep them. Bags are named
        # fault_{FAULT_CODE}_{timestamp}/. The node creates the directory if needed.
        storage_path: "/mnt/persistent/instor/bags/fault_bags"

        max_buffer_mb: 256             # Ring-buffer RAM cap
        max_bag_size_mb: 100           # Max total storage
        max_total_storage_mb: 1000

        # Keep bags after their fault is cleared (default true deletes them on ClearFault).
        auto_cleanup: false            # Auto-delete old bags
```

(applied via `--ros-args -p snapshots.rosbag.*` per the file's own note, with `enabled: true`,
`storage_path`, and `auto_cleanup: false` set as shown above.)

mfaferek93:
Closes #620.

Only the newest rosbag was ever kept per fault code. That was enforced by the schema, not by a retention policy, so no configuration could change it: a fault that kept coming back held on to its freeze frame and threw away every recording but the last.

The fix separates three things the old code treated as one:

- a **link** - some fault claiming some recording - keyed by `(fault_code, file_path)`, one row
- a **recording** - one bag on disk - identified by `file_path`, named publicly by `recording_id`
- the **bytes**, which belong to `file_path`

Rows are per link, bytes are per path, and a bag is unlinked when its last row goes. The reverse relation already worked: a burst writes one row per attached fault, all pointing at one bag, so the sharing logic and the quota needed no changes. This adds the mirror image.

### Retention

`snapshots.rosbag.max_bags_per_fault`, default 1, which is what the code did before: a new recording replaces the old one. Shipping at parity means a regression report is about the plumbing rather than the policy. Past the cap the oldest goes, matching the direction the storage quota already evicts in.

It is more a fairness knob than a depth knob. `max_total_storage_mb` is the real disk bound, and a big per-fault cap lets one flapping fault eat the budget and push out everyone else's black box. Docs say so.

Not consistent with `snapshots.max_per_fault`, which rejects new snapshots once full, and deliberately so. Rejecting a new recording would mean a technician standing next to a machine that is faulting right now downloads a bag from three days ago.

### URLs

`/bulk-data/rosbags/{id}` takes a recording id now. Our own docs and the SSE payload tell clients to build that URL from a fault code, so an id that names no recording is retried as one and serves that fault's newest bag. Old addresses keep working, hence `feat` and not `feat!`.

### Four bugs found on the way

None were reachable while a fault could only hold one recording:

- quota eviction deleted by fault code, so evicting one bag would have wiped every recording of every fault it touched, up to 33 in one burst
- `get_rosbag_file` had no `ORDER BY` and would have served an arbitrary recording
- the two stale-row self-heals deleted a fault's whole history when one bag went missing from disk
- `delete_rosbag_file` and `delete_rosbag_files` read only the first `file_path`, so deleting a fault removed every row but left all but one bag on disk, unreachable and still counted against a quota that sums rows

### Migration

SQLite cannot drop a column-level `UNIQUE` and `CREATE TABLE IF NOT EXISTS` does nothing to an existing database, so this is the documented table rebuild. It runs on first open, is idempotent, and decides whether it is needed by looking at the schema (`PRAGMA index_list`, origin `'u'`) rather than a version counter - fresh, migrated and already-migrated databases all answer correctly with no bookkeeping.

Uniqueness is a `CREATE UNIQUE INDEX` on `(fault_code, file_path)` now. Keyed on the path because that is what identifies the thing on disk: if two bags ever produced the same basename, keying on `recording_id` would silently replace one, while keying on the path degrades the same collision to a mislabeled download.

### Tests

6871 pass. Each suite below was also run with the fix reverted, to check it goes red for the right reason.

`test_rosbag_storage_parity.cpp` is new: 20 cases, each run against both storage backends from one body. The in-memory backend enforced one-per-fault independently, so `storage_type: memory` could have kept the old behaviour with nothing failing. It caught two real divergences - the delete-path bug above, and `INSERT OR REPLACE` renumbering a refreshed row so the backends ordered a tie group differently.

`test_rosbag_history.test.py` drives confirm, acknowledge, confirm against SQLite, plus cap eviction and both ways of addressing a bag. Set `max_bags_per_fault` to 1 and its first case fails, which is the reported bug.

`test_rosbag_history_download.test.py` does the same over HTTP through the gateway, off a real failure: `lidar_sensor` checks `min_range >= max_range` and reports itself, the test only moves the parameters. It asserts two descriptors with different ids, different downloaded bytes, both `bulk_data_uri`s resolving, and the fault-code URL still serving the newest.

Migration coverage builds a legacy database by hand and checks that no `origin='u'` index survives, every row does, ids keep their order, `recording_id` is backfilled, and opening twice more changes nothing.

### Three more places evidence went missing

Found while checking the above against a running box, and fixed here because
shipping the retention change without them leaves a fault whose evidence does
not line up.

**A recording reached clients with no capture time.** The fault manager has
always stamped one, but only the freeze-frame branch of the transport forwarded
it, so every bag arrived without one and the UIs rendered "N/A". Harmless while
a fault held one recording; with several the timestamp is the only field that
places an occurrence.

**A capture was stored in part.** A confirmation writes one snapshot row per
topic and the cap counted rows, rejecting the tail once full - so a capture
straddling the limit landed with some topics present and the rest silently
absent, which on the wire reads exactly like "that topic was not publishing".
Captures are now written as a unit and the cap drops whole ones, oldest first.
A capture larger than the cap is kept whole rather than torn.

**Acknowledging a fault deleted the history it was told to keep.**
`auto_cleanup` defaults on and deletes every recording of the fault, so raising
`max_bags_per_fault` to collect a trail bought nothing: the first
acknowledgement wiped it. With a history configured the cap governs retention;
at the default of one, nothing changes. `clear_fault` deleted the value
snapshots on the same edge, so the new `snapshots.retain_on_clear` keeps those
too - both survive an acknowledgement or neither does, rather than a fault being
left holding bags whose readings are gone.

Each has a test that goes red when its fix is reverted, on both storage backends
where the storage layer is involved.

### Consumers

Four needed changes, each with its own PR: ros2_medkit_web_ui#95, platform#270 (fleet_ui), ros2_medkit_foxglove_extension#21, ros2_medkit_mcp#24. Both web UIs were checked in a browser against a gateway holding two real bags. The service type hashes change, so the gateway and fault manager deploy together.


src/ros2_medkit_gateway/src/http/handlers/bulkdata_handlers.cpp
```diff
@@ -112,6 +112,97 @@ std::string rosbag_recording_id(const std::string & file_path) {
...
+  if (rosbag_data.contains("fault_codes") && rosbag_data["fault_codes"].is_array()) {
+    auto codes = rosbag_data["fault_codes"].get<std::vector<std::string>>();
+    if (!codes.empty()) {
+      return codes;
+    }
+  }
+  // No list on the wire: a peer that predates the field, where the addressed id
+  // was the fault code itself. Authorizing against it reproduces the old check.
+  return {requested_id};
+}
```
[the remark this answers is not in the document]
mfaferek93: Hardened in 31a1cad7, though not for the stated reason. `get_rosbag` goes straight to the ROS transport, which builds `fault_codes` from a `std::vector<std::string>`, so the elements cannot be anything but strings today and the conversion cannot throw. The function does claim to tolerate a peer that predates the field, so it now converts element by element and skips anything that is not a string - that makes the comment true rather than aspirational.

src/ros2_medkit_fault_manager/test/test_rosbag_storage_parity.cpp
```diff
@@ -0,0 +1,413 @@
...
+/// therefore runs against both backends from one body.
+
+#include <gtest/gtest.h>
+
+#include <cstdlib>
+#include <filesystem>
+#include <fstream>
+#include <memory>
+#include <string>
+#include <vector>
```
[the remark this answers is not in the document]
mfaferek93: Correct, `std::sort` at line 307 with no `<algorithm>`. Added in 31a1cad7.

src/ros2_medkit_fault_manager/test/test_rosbag_storage_parity.cpp
```diff
@@ -0,0 +1,413 @@
+// Copyright 2025 mfaferek93
```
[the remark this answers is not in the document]
mfaferek93: Right, and inconsistent with the other two files this PR adds, which are 2026. Fixed in 31a1cad7.

src/ros2_medkit_fault_manager/src/fault_storage.cpp
```diff
@@ -339,21 +351,70 @@ void InMemoryFaultStorage::set_max_snapshots_per_fault(size_t max_count) {
...
+      const size_t before = updated.size();
+      updated.erase(std::remove_if(updated.begin(), updated.end(),
+                                   [&fault_code, oldest](const SnapshotData & s) {
+                                     return s.fault_code == fault_code && s.capture_id == oldest;
+                                   }),
+                    updated.end());
+      // One capture can exceed the cap on its own. Dropping every older set and
+      // still being over means the cap is smaller than this fault's topic count:
+      // keep the newest capture whole rather than tearing it.
+      if (updated.size() == before) {
```
bburda: `found` is true whenever this fault has any row, so the `remove_if` above always deletes something. That means `updated.size() == before` is never true and this `break` never runs. When the newest capture alone is larger than the cap, the loop deletes that capture too and the fault is left with zero snapshots. SQLite does not do this, because its trim has `AND capture_id <> ?2`.

So with `storage_type: memory`, the default `max_per_fault: 10` and a capture set of 11 topics, every freeze frame is lost. Could you skip the newest capture here as well? `ACaptureLargerThanTheCapIsKeptWholeRatherThanTorn` checks this rule, but only in the SQLite fixture.

src/ros2_medkit_fault_manager/src/rosbag_capture.cpp
```diff
@@ -590,6 +591,19 @@ void RosbagCapture::on_fault_cleared(const std::string & fault_code) {
     return;
   }
 
+  // Acknowledging a fault must not destroy a history the operator asked to keep.
+  // auto_cleanup deletes EVERY recording of the fault, so with a cap above one the
+  // first acknowledgement wiped the whole trail that max_bags_per_fault had just
+  // been raised to collect - one default silently cancelling another. When a
+  // history is configured the cap governs retention and acknowledgement leaves the
+  // evidence alone; at the default cap of 1 this is exactly the old behaviour.
+  if (config_.max_bags_per_fault != 1) {
```
bburda: This makes `auto_cleanup` do nothing for every cap except 1, including 0. Two things about that.

The config docs say "Leave this ``false`` when raising ``max_bags_per_fault``, or acknowledging a fault discards the history that was just kept". That cannot happen any more. Could you update the docs or this guard?

Second, `finalize_post_fault_recording` did not get the same rule. Its `wants_a_row` lambda still looks only at `config_.auto_cleanup`. With `max_bags_per_fault: 3` and `auto_cleanup: true`, if the fault is acknowledged while its post-roll still runs, `wants_a_row` returns false, `rows` stays empty, and line 1584 deletes the bag from disk. Could `wants_a_row` check the cap too? No test covers this: both new unit tests use `duration_after_sec = 0.0`, and the launch test sets `auto_cleanup: False`, which returns above this line.

src/ros2_medkit_fault_manager/src/snapshot_capture.cpp
```diff
@@ -147,15 +148,21 @@ void SnapshotCapture::capture(const std::string & fault_code) {
 
   // Accumulate a compact dict of captured topic values alongside the per-topic snapshots.
   nlohmann::json freeze_frame = nlohmann::json::object();
+  // One id for the whole capture, minted here so every row of this confirmation
+  // shares it. Monotonic rather than a clock read: the rows are written seconds
+  // apart under load and a wall clock is not guaranteed to move forward.
+  static std::atomic<int64_t> capture_seq{0};
```
bburda: This counter starts at 0 in every process. After a restart the next capture gets id 1, while rows already in SQLite have higher ids. `store_snapshots` then deletes `MIN(capture_id)` and protects `MAX(capture_id)`, so it deletes the capture that was just written and keeps the old ones.

Example: `max_per_fault: 10`, 3 topics, 3 captures already stored. After a restart the next capture writes 3 rows with id 1, the row count is 12, and the trim drops the whole id 1 group, including the new rows. `get_snapshots` also returns the old captures first. Could you read `MAX(capture_id)` from storage once at construction and continue from there? `retain_on_clear` makes stored rows the normal case.

src/ros2_medkit_fault_manager/src/snapshot_capture.cpp
```diff
@@ -166,6 +173,14 @@ void SnapshotCapture::capture(const std::string & fault_code) {
   RCLCPP_INFO(node_->get_logger(), "Captured %zu/%zu snapshots for fault '%s'", captured_count, topics.size(),
               fault_code.c_str());
 
+  // One write for the whole capture. Storing topic by topic let the per-fault cap
+  // reject the tail of a capture, leaving a freeze frame with holes and nothing
+  // saying which values were dropped.
+  for (auto & row : rows) {
+    row.capture_id = capture_id;
+  }
+  storage_->store_snapshots(rows);
```
bburda: Could you check the fault status again here before storing? `capture()` runs on a pool worker, `clear_fault` runs on the service thread, and `capture_topic_on_demand` waits up to `snapshots.timeout_sec` for each silent topic.

Before this change rows were written one by one, so a clear in the middle removed what was already stored. Now nothing is written yet when `DELETE FROM snapshots WHERE fault_code = ?` runs, so the whole set is stored after it, on a fault that is already CLEARED. Nothing removes those rows later, so on SQLite they survive a restart.

src/ros2_medkit_fault_manager/src/fault_manager_node.cpp
```diff
@@ -159,6 +194,7 @@ FaultManagerNode::FaultManagerNode(const rclcpp::NodeOptions & options) : Node("
   // Apply snapshot limit to storage
   if (max_snapshots > 0) {
     storage_->set_max_snapshots_per_fault(static_cast<size_t>(max_snapshots));
+    storage_->set_retain_snapshots_on_clear(retain_snapshots_on_clear);
```
bburda: Could you move this line out of the `if (max_snapshots > 0)` block? `snapshots.max_per_fault: 0` is documented as unlimited. With that value the setter is never called, so `snapshots.retain_on_clear: true` has no effect and clearing a fault still deletes its snapshots. The two settings are independent.

src/ros2_medkit_gateway/src/http/handlers/bulkdata_handlers.cpp
```diff
@@ -342,24 +405,37 @@ http::Result<http::BinaryResponse> BulkDataHandlers::download(const http::TypedR
...
+    // own downloadable copy of it before - so this grants nothing new, it only
+    // renames the door. Tested with the shared boundary-aware matcher: the
+    // transport's get_fault(code, source) check is a raw prefix match, so app id
+    // "plc" would otherwise claim the assets of "plc_line1".
     auto source_filters = get_source_filters(entity);
     std::set<std::string> scope(source_filters.begin(), source_filters.end());
-    auto fault_result = fault_mgr->get_fault(fault_code, "");
-    if (!fault_result.success || !faults::fault_in_source_scope(fault_result.data, scope)) {
+
+    const auto attached_codes = detail::rosbag_attached_fault_codes(rosbag_result.data, bulk_data_id);
```
bburda, line 463: On the fallback path the id from the URL is no longer checked against the entity. Could you keep that check when the fault manager resolved the id as a fault code?

Example: `FAULT_A` (source `app_a`) and `FAULT_B` (source `app_b`) confirm in one burst and share one recording. `GET /apps/app_a/bulk-data/rosbags/FAULT_B` returned 404 before, because `FAULT_B` is not in the scope of `app_a`. Now the fault manager finds no recording with that name, falls back to the fault branch, and returns the shared recording with `fault_codes = [FAULT_A, FAULT_B]`. `any_of` passes on `FAULT_A` and the request returns 200.

The bytes are the same ones `app_a` can already download under its own id, so no new data is exposed. But the 200 tells the caller that another fault code shares a recording with one of its own faults. `build_sovd_fault_response` rejects mixed-source faults at line 806 for that reason. The union looks right when the id was a recording id. Only the fallback case also needs the requested code in scope.

src/ros2_medkit_gateway/src/http/handlers/bulkdata_handlers.cpp
```diff
@@ -112,6 +112,106 @@ std::string rosbag_recording_id(const std::string & file_path) {
...
+    }
+    if (recording_id.empty()) {
+      continue;  // nothing addressable - a row with neither id nor path
+    }
+
+    const std::string fault_code = row.value("fault_code", "");
+    int64_t created_at_ns = 0;
+    if (auto it = faults_by_code.find(fault_code); it != faults_by_code.end()) {
+      const double first_occurred = it->second.value("first_occurred", 0.0);
+      created_at_ns = static_cast<int64_t>(first_occurred * 1'000'000'000);
```
bburda: Taking the date from the fault's `first_occurred` gives every recording of one fault the same `creation_date`. For an acknowledged fault they all get `1970-01-01T00:00:00.000Z`, because `list_faults` uses `include_cleared = false` by default and the fault is then missing from `fault_map`. The rows are still listed, because `clear_fault` only changes the status.

This PR is what makes an acknowledged fault keep its recordings, and the date is the field that separates one occurrence from another. Could `ListRosbags.srv` also return the `created_at_ns` that storage already holds, and the descriptor use that? The new e2e test reaches this state, since it acknowledges after each occurrence and then reads the listing, but it does not check `creation_date`.

src/ros2_medkit_fault_manager/src/sqlite_fault_storage.cpp
```diff
@@ -243,26 +243,236 @@ void SqliteFaultStorage::initialize_schema() {
...
+    std::string error = err_msg ? err_msg : "Unknown error";
+    sqlite3_free(err_msg);
+    throw std::runtime_error("Failed to begin recording_id backfill: " + error);
+  }
+  try {
+    for (const auto & [row_id, file_path] : pending) {
+      SqliteStatement update(db_, "UPDATE rosbag_files SET recording_id = ? WHERE id = ?");
+      update.bind_text(1, rosbag_recording_id(file_path));
+      update.bind_int64(2, row_id);
+      update.step();
```
bburda: Could you check this `step()` result and throw, like the other `step()` calls in this file? On `SQLITE_BUSY` or `SQLITE_FULL` the error is lost, the `COMMIT` below still reports success, and rows keep `recording_id = ''`.

`delete_rosbag_recording` and `drop_rosbag_link` have no guard for an empty id, so `DELETE FROM rosbag_files WHERE recording_id = ''` matches every such row of every fault, and `evict_bags_over_quota` calls it with whatever `recording_by_path` holds. Could you also return early for an empty id in both functions? Then one bad row cannot cause a wide delete.

src/ros2_medkit_fault_manager/src/sqlite_fault_storage.cpp
```diff
@@ -849,12 +1121,15 @@ std::vector<SnapshotData> SqliteFaultStorage::get_snapshots(const std::string &
...
+      "SELECT fault_code, topic, message_type, data, captured_at_ns, capture_id FROM snapshots WHERE fault_code "
       "= ?";
   if (!topic_filter.empty()) {
     sql += " AND topic = ?";
   }
-  sql += " ORDER BY captured_at_ns DESC";
+  // capture_id before the timestamp: the rows of one capture are written seconds
+  // apart under load and their timestamps interleave with a neighbouring capture's,
+  // so ordering by time alone splits a set the reader then cannot regroup.
+  sql += " ORDER BY capture_id DESC, captured_at_ns DESC";
```
bburda, line 1142: No reader uses this order today, and the two backends do not agree on it. `InMemoryFaultStorage::get_snapshots` does not sort at all and returns insertion order, so there the oldest capture comes first and here the newest does.

The only reader, `handle_get_snapshots`, puts all captures into one topic map with last-write-wins, and takes `captured_at` from the newest row. On SQLite that returns the values of the oldest kept capture together with the timestamp of the newest one. This was already true before, but the change from reject-new to keep-newest means the reported capture now moves forward after every confirmation. Could the reader keep only `snapshots.front().capture_id`, and the in-memory backend sort the same way?

src/ros2_medkit_fault_manager/src/sqlite_fault_storage.cpp
```diff
@@ -243,26 +243,236 @@ void SqliteFaultStorage::initialize_schema() {
...
+
+  // Order matters: drop the legacy constraint before adding the column, so the
+  // rebuild only ever has to copy the old column set.
+  migrate_rosbag_files_drop_unique();
+  migrate_rosbag_files_add_recording_id();
+
+  // Indexes last, so they serve a fresh table and a rebuilt one alike.
+  //
+  // idx_rosbag_files_path is capability, not tuning: path_referenced() and
+  // path_shared_with_other_fault() scan on file_path on every delete, against a
```
bburda: `path_shared_with_other_fault()` has no callers left after this change, since all three call sites are removed in this diff. So this comment names a function that no longer scans anything. Could you remove the function together with the mention, or leave only `path_referenced()` here? `drop_rosbag_link()` is the same case: added to the interface, implemented in both backends and tested, but no production code calls it.

docs/config/fault-manager.rst
```diff
@@ -324,10 +325,33 @@ Capture continuous rosbag recordings around fault events.
...
+
+   ``max_bags_per_fault`` is a **fairness** knob, not a depth knob.
+   ``max_total_storage_mb`` is the real disk bound and eviction across it is
+   global and oldest-first, so a fault that flaps often enough will consume the
+   budget and push out every other fault's black box. Raise the per-fault cap
+   when you need the history of a specific intermittent fault; raise the total
+   budget with it if other faults still need theirs.
+
+   The cap keeps the newest recordings and evicts the oldest. It deliberately
+   does not match ``snapshots.max_per_fault``, which rejects new snapshots once
```
bburda: This note says `snapshots.max_per_fault` rejects new snapshots when full, but this PR changes it to keep-newest. `store_snapshots` now removes whole capture sets, oldest first. The table row at line 213 still says the old rule, and the comment on `set_max_rosbags_per_fault` in `fault_storage.hpp` calls the rosbag cap "deliberately the opposite of set_max_snapshots_per_fault's reject-new". Could you update all three? As written, the note explains a difference between the two caps that no longer exists.

docs/config/fault-manager.rst
```diff
@@ -324,10 +325,33 @@ Capture continuous rosbag recordings around fault events.
      - Maximum total storage for all rosbags (MB). A recording shared by a
        burst of faults counts once towards the total, and eviction removes a
        whole burst's bag at a time (oldest first).
+   * - ``rosbag.max_bags_per_fault``
+     - ``1``
+     - How many recordings one fault code keeps. Past the cap the oldest is
+       unlinked, so the default reproduces the historical behaviour exactly: a
+       new recording replaces the previous one. ``0`` means unlimited, bounded
+       only by ``max_total_storage_mb``. ``3`` is a reasonable value for a fault
```
bburda, line 343: Could you also mention `snapshots.recapture_cooldown_sec` here? It is 60 s by default, it gates rosbag capture as well, and the entry for a fault is removed only in `handle_clear_fault`. So a fault that returns faster than one minute still gets one recording, whatever this cap is set to. That is the case this sentence tells the operator to raise the cap for.

docs/config/fault-manager.rst
```diff
@@ -324,10 +325,33 @@ Capture continuous rosbag recordings around fault events.
      - Maximum total storage for all rosbags (MB). A recording shared by a
        burst of faults counts once towards the total, and eviction removes a
        whole burst's bag at a time (oldest first).
+   * - ``rosbag.max_bags_per_fault``
```
bburda, line 338: Could you add a row for `snapshots.retain_on_clear` to this table? It is a new parameter and it decides whether acknowledging a fault keeps its snapshots. The note below about keeping the evidence consistent depends on it, but the parameter is not documented in this file or in the snapshots tutorial.

docs/api/rest.rst
```diff
@@ -1371,7 +1371,7 @@ Download a specific bulk-data file.
 
 .. code-block:: bash
 
-   curl -O -J http://localhost:8080/api/v1/apps/motor_controller/bulk-data/rosbags/550e8400-e29b-41d4-a716-446655440000
+   curl -O -J http://localhost:8080/api/v1/apps/motor_controller/bulk-data/rosbags/fault_MOTOR_OVERHEAT_1738664999000
```
bburda, line 1377: Could you update the listing section above as well? Line 1337 shows a UUID as `id`, but it is a recording id now. Line 1343 shows `x-medkit.fault_code`, which is now the `fault_codes` array. The text at lines 1352-1355 says "each fault gets its own descriptor with the full bag size", and this PR changes that to one descriptor per recording. Line 1367 still says `Content-Disposition: attachment; filename="FAULT_CODE.mcap"`, but the filename is `<recording_id>.<format>` now. The tutorial page was updated, so the two pages now describe the same response in different ways.

src/ros2_medkit_fault_manager/config/snapshots.yaml
```diff
@@ -79,6 +79,13 @@ default_topics:
 # `default_topics` from this YAML file. Rosbag settings must be configured
 # via ROS 2 parameters (--ros-args -p snapshots.rosbag.*) or launch files.
 
+# Keep a fault's value snapshots when it is acknowledged (default: false)
+# Off is the historical behaviour: clearing a fault deletes them. Turn it on
+# together with rosbag.max_bags_per_fault, or acknowledging leaves the fault
+# holding recordings whose matching readings are gone - evidence that no longer
+# lines up. Growth stays bounded by max_per_fault either way.
+retain_on_clear: false
```
bburda: Nothing reads this key. `load_snapshot_config_from_yaml` reads only `fault_specific`, `patterns` and `default_topics`. The real setting is the ROS parameter `snapshots.retain_on_clear`. An operator who edits it here gets no effect and no warning. Could you remove it, or turn it into a comment that points to the parameter, the way the rosbag block below does?

src/ros2_medkit_msgs/srv/ListRosbags.srv
```diff
@@ -32,6 +32,10 @@ bool success
 string[] fault_codes
 
 # Paths to the rosbag files/directories.
+# Parallel to fault_codes. Rows of one burst repeat the same recording id, which is
+# how a client groups the entries that serve the same bytes.
+string[] recording_ids
```
bburda, line 37: The comment `# Paths to the rosbag files/directories.` now sits above `recording_ids`, and `file_paths` on line 39 has no comment left. Could you move that line back down to `file_paths`?

src/ros2_medkit_fault_manager/src/fault_storage.cpp
```diff
@@ -445,37 +542,144 @@ void InMemoryFaultStorage::store_rosbag_files(const std::vector<RosbagFileInfo>
...
 
-  auto it = rosbag_files_.find(fault_code);
-  if (it == rosbag_files_.end()) {
+  // Newest, matching the SQLite backend's ORDER BY created_at_ns DESC, id DESC.
+  const RosbagRow * best = nullptr;
+  for (const auto & row : rosbag_files_) {
+    if (row.info.fault_code != fault_code) {
+      continue;
+    }
+    if (best == nullptr || std::tie(best->info.created_at_ns, best->seq) < std::tie(row.info.created_at_ns, row.seq)) {
```
bburda, line 579: Could you add `#include <tuple>` at the top of this file? `std::tie` is used at six places added in this change, and neither this file nor `fault_storage.hpp` includes `<tuple>`. It builds today only because `rclcpp/rclcpp.hpp` includes it indirectly. That can break on one distro after an unrelated header change.

src/ros2_medkit_fault_manager/test/test_rosbag_history.test.py
```diff
@@ -0,0 +1,477 @@
...
+        self.assertEqual(by_id.recording_id, newest.recording_id)
+
+    def test_03_the_cap_evicts_the_oldest_recording(self):
+        """Past max_bags_per_fault the oldest goes, newest are kept."""
+        fault_code = 'CAPPED_FAULT'
+
+        recordings = []
+        seen = set()
+        for _ in range(MAX_BAGS_PER_FAULT + 1):
+            self.assertTrue(self._clear_fault(fault_code).success or True)
```
bburda: `x or True` is always true, so this line checks nothing and a failed clear is not reported here. It would appear later as a confusing timeout saying the re-confirmation reused the previous recording. Could you assert `.success` directly?

src/ros2_medkit_fault_manager/test/test_rosbag_storage_parity.cpp
```diff
@@ -0,0 +1,414 @@
...
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+/// Rosbag retention parity between the two storage backends.
+///
+/// A fault holding several recordings was enforced by the SQLite schema before, and
+/// separately by a std::map in the in-memory backend. Two enforcement points means two
+/// chances to diverge, and `storage_type: memory` silently keeping the old behaviour
+/// would be invisible to a suite that only exercises SQLite. Every assertion here
+/// therefore runs against both backends from one body.
```
bburda, line 21: Could you extend this suite to the snapshot side too? The same reason applies: `store_snapshots`, the snapshot cap and `set_retain_snapshots_on_clear` also got a separate implementation in each backend in this PR. They do differ today: the in-memory cap deletes the newest capture where SQLite keeps it, and `get_snapshots` returns the two orders reversed. `ACaptureLargerThanTheCapIsKeptWholeRatherThanTorn` and `ClearFaultKeepsSnapshotsWhenEvidenceIsRetained` would fail here as TYPED_TESTs.

src/ros2_medkit_fault_manager/src/fault_storage.cpp
mfaferek93: Right on both counts, and worse than the `break` being dead: since `found` is always true the loop runs until the fault has zero rows, so `storage_type: memory` lost every freeze frame. Fixed in 3881bac4 by computing the newest capture up front and excluding it from the eviction candidates, which is what SQLite's `AND capture_id <> ?2` does.

Negative control: with the exemption removed, the new `ACaptureLargerThanTheCapIsKeptWholeRatherThanTorn` parity test fails on the in-memory backend with `rows.size()` 0 vs 5, and passes on SQLite. So the divergence you describe reproduces exactly.

src/ros2_medkit_fault_manager/src/rosbag_capture.cpp
mfaferek93: Both parts fixed in 8e1f5656. The two places now ask one predicate, `keeps_history()`, which is `max_bags_per_fault != 1` and so covers 0 as well: `on_fault_cleared` and `wants_a_row` cannot disagree about the same fault any more. Docs updated in 06f18b3c - the `auto_cleanup` row now says it has no effect once a history is configured, instead of telling operators to turn it off.

You were right that no test covered it. Added `AcknowledgingDuringThePostRollKeepsAConfiguredHistory` with `duration_after_sec = 0.3` and the fault cleared in the store while the window is open. With the old lambda it fails on an empty `get_rosbag_files`.

src/ros2_medkit_fault_manager/src/snapshot_capture.cpp
mfaferek93: Confirmed and fixed in 3881bac4. It is worse than ordering: `newest_capture` is `MAX(capture_id)` read back from the table, so after a restart the protected set is an old one while `MIN(capture_id)` is the capture just written, which is then the first thing trimmed.

`FaultStorage::get_max_capture_id()` is new in both backends and `SnapshotCapture` seeds its counter from it at construction. The counter moved from a function-local static to a member for that. Negative control: without the seeding, `CaptureIdContinuesFromWhatStorageAlreadyHolds` fails with `actual: 1 vs 41`.

src/ros2_medkit_fault_manager/src/snapshot_capture.cpp
mfaferek93: Fixed in 8e1f5656, with one qualifier. `capture()` now re-reads the fault before storing and skips the batch when the store reports CLEARED, mirroring `wants_a_row` including the "absent is not cleared" rule.

The qualifier is `retain_on_clear`: when it is on, the acknowledgement never promised the readings were gone, so the batch is stored. `FaultStorage::retains_snapshots_on_clear()` is new for that. Two tests cover both directions.

src/ros2_medkit_fault_manager/src/fault_manager_node.cpp
mfaferek93: Fixed in 8e1f5656 - the setter is now outside the block, since the cap and retention are independent questions and 0 is documented as unlimited.

Covered by `UnlimitedSnapshotRetentionTest.RetentionAppliesWithNoPerFaultCap`, which builds the node with `max_per_fault: 0` and `retain_on_clear: true` and asserts snapshots survive the clear. It fails on the old wiring.

src/ros2_medkit_gateway/src/http/handlers/bulkdata_handlers.cpp
mfaferek93: Agreed, and the distinction you draw is the right one: the union is correct when the id really was a recording id, and only the fallback needs the requested code in scope. Fixed in 06f18b3c.

The compatibility path is now detected by `detail::rosbag_resolved_by_fault_code()` - the resolved `recording_id` differs from the requested segment - and on that path the requested code must pass `fault_in_source_scope` before the union check runs. A peer that sends no recording id is deliberately not treated as the fallback, because the attached-codes helper already reduces to the pre-620 check there. Three tests pin the three cases.

src/ros2_medkit_gateway/src/http/handlers/bulkdata_handlers.cpp
mfaferek93: Fixed in 06f18b3c. `ListRosbags.srv` now returns `created_at_ns` per row and the descriptor uses it, falling back to the fault only for a row that predates the field.

You are right that the e2e reached this state without checking it. Added `test_02_every_recording_is_dated_by_its_own_capture`, which reads `creation_date` from the bulk-data listing after the fault has been acknowledged twice and asserts the dates are distinct and not 1970. Reverting the change makes it fail with exactly the epoch assertion.

src/ros2_medkit_fault_manager/src/sqlite_fault_storage.cpp
mfaferek93: Both fixed in 3881bac4, though the second one partly evaporated.

The backfill `step()` now throws on anything other than `SQLITE_DONE`, so a dropped `SQLITE_BUSY` cannot leave `recording_id = ''` behind a successful COMMIT.

`delete_rosbag_recording` returns 0 on an empty id in both backends. `drop_rosbag_link` got the same guard and then the whole function went away with your next comment, so there is nothing left to guard there.

src/ros2_medkit_fault_manager/src/sqlite_fault_storage.cpp
mfaferek93: Fixed in two places. `InMemoryFaultStorage::get_snapshots` now sorts by `capture_id DESC, captured_at_ns DESC` to match the SQL (3881bac4), and `handle_get_snapshots` keeps only `snapshots.front().capture_id` (8e1f5656) so the topic map is one capture rather than a blend of several under a single `captured_at`.

`GetSnapshotsReturnsTheNewestCaptureFirst` runs against both backends, which is the assertion that was missing.

src/ros2_medkit_fault_manager/src/sqlite_fault_storage.cpp
mfaferek93: Both removed in 3881bac4, declarations, implementations and the mention in that comment. `drop_rosbag_link` went with them, including its tests - `delete_rosbag_files` already unlinks a fault's rows in bulk, so it was API without a caller.

docs/config/fault-manager.rst
mfaferek93: All three updated in 06f18b3c. The note now says the two caps evict in the same direction, the table row for `max_per_fault` describes whole-capture-set eviction with the newest set exempt, and the comment on `set_max_rosbags_per_fault` no longer claims to be the opposite of anything. The difference the note used to explain stopped existing when snapshot eviction changed, and I did not go back to the docs. Thanks for catching it.

docs/config/fault-manager.rst
mfaferek93: Documented in 06f18b3c, but I did not change the behaviour, and I would like your view on that.

You are right that the cooldown gates the whole capture job, bags included, so a fault returning sooner than 60 s keeps one recording whatever the cap says - which is the flapping fault the cap exists for. My reasoning for leaving it: the cooldown is real protection against a capture storm, and quietly weakening it for one config combination seemed worse than making the interaction visible.

So the docs now say it explicitly, and 8e1f5656 logs a warning at startup when `max_bags_per_fault != 1` and the cooldown is non-zero. If you think the cap should override the cooldown instead, say so and I will change it.

docs/config/fault-manager.rst
mfaferek93: Added in 06f18b3c, with a note that it is independent of `max_per_fault` - which is worth spelling out precisely because of the wiring bug you found two comments up.

docs/api/rest.rst
mfaferek93: All four updated in 06f18b3c: the example `id` is a recording id, `x-medkit.fault_code` is now the `fault_codes` array, the paragraph says one descriptor per recording rather than per fault, and `Content-Disposition` shows `<recording_id>.mcap`. Also added a line saying one fault code can appear on several descriptors told apart by `creation_date`, since that is the shape this change introduces.

src/ros2_medkit_fault_manager/config/snapshots.yaml
mfaferek93: Removed in 06f18b3c and replaced with a comment pointing at the ROS parameter, the same way the rosbag block below does it. An operator editing a key that nothing reads is worse than no key at all.

src/ros2_medkit_msgs/srv/ListRosbags.srv
mfaferek93: Fixed in 06f18b3c. `file_paths` has its comment back and `recording_ids` got its own.

src/ros2_medkit_fault_manager/src/fault_storage.cpp
mfaferek93: Added in 3881bac4. Six uses of `std::tie` resting on a transitive include through `rclcpp.hpp` is exactly the kind of thing that breaks on one distro for an unrelated reason.

src/ros2_medkit_fault_manager/test/test_rosbag_history.test.py
mfaferek93: Fixed in 06f18b3c. The first pass has nothing to acknowledge yet, so the assertion is now conditional on the occurrence index and checks `.success` with the response message on failure. Verified it is not vacuous: an early version referenced a field that does not exist on `ClearFault_Response` and the suite failed loudly rather than passing.

src/ros2_medkit_fault_manager/test/test_rosbag_storage_parity.cpp
mfaferek93: Done in 3881bac4. `SnapshotRetentionParityTest` is a second typed suite in the same file, sharing the backend factory, with six cases: your two, plus whole-set eviction order, `get_snapshots` ordering, `get_max_capture_id`, and the default clear-drops-snapshots behaviour.

It earned its place immediately - it is what fails on the in-memory trim and on the ordering divergence, both of which were invisible to a suite that only covered rosbags.

docs/tutorials/snapshots.rst
```diff
@@ -700,20 +703,26 @@ Rosbag files are downloaded via SOVD bulk-data endpoints.
 
    curl http://localhost:8080/api/v1/apps/motor_controller/bulk-data/rosbags
 
+One item per **recording**, not per fault. A burst of correlated faults shares a
+single recording and appears once, with every fault it covers listed in
+``x-medkit.fault_codes``. A fault that confirmed several times contributes one
+item per recording it kept (see ``max_bags_per_fault`` below).
```
bburda: This points the reader to `max_bags_per_fault` below, but the page has no row for it. The key appears only here and in the snapshot list above. Could you add a row for `snapshots.rosbag.max_bags_per_fault` to the rosbag config table, so the pointer has a target?

bburda (commented):
Three more things on this page, outside this diff, that I think also need a change:

1. The `snapshots.max_per_fault` row in the snapshot config table still says that new snapshots are rejected when the limit is reached. This change makes it keep-newest and evict whole capture sets instead. The same row in `docs/config/fault-manager.rst` was updated. Could you update this one too?

2. The note near the top of the page says snapshots are deleted when a fault is cleared, and the troubleshooting section repeats it. Both are only true with the default now. With `snapshots.retain_on_clear: true` the snapshots stay. That parameter also has no row in the snapshot config table on this page. Could you add the row and make the two sentences conditional?

3. The `snapshots.rosbag.auto_cleanup` row tells the operator to set it to `false` to keep bags. After this change `auto_cleanup` does nothing at all when `max_bags_per_fault` is not 1. The same row in `docs/config/fault-manager.rst` already says this. Could you match it here?

docs/tutorials/snapshots.rst
mfaferek93: Added the row in d14f293d - default 1, 0 = unlimited, oldest evicted first, with the fairness-vs-depth note. Also fixed the pointer: the table sits above that section, so it now says "in the configuration table above".

bburda (approved):

PR #625, opened 2026-08-19
Detect a node that dies when nothing else is left to report it

### LINKED ISSUE #624: A node dies and nothing is left to report it
Part of #570.

## Summary

A node stops. The process exits, or it is killed, or its container restarts and it never
comes back. Everything that node provided stops with it. Nothing reports this.

The node cannot report it itself: the thing that would send the fault is the thing that
died. The other detectors do not see it either. `qos_mismatch` needs both sides present to
compare their profiles. `orphan` looks at a topic with no subscriber, and a dead publisher
takes its topic with it. `lifecycle_expectation` only watches nodes the operator listed as
required-active, and it hands absence over to the presence class on purpose. This issue is
that presence class.

`GRAPH_NODE_DISAPPEARED` already exists as a fault code in the package. No detector raises
it today.

## Proposed solution

Track every app the reliability gate has armed, and raise `GRAPH_NODE_DISAPPEARED` when an
armed node stops being alive for longer than a grace period.

These points decide whether the detector is useful or only noisy.

**Alive is not the same as present in the snapshot.** In runtime discovery a dead node
leaves the entity snapshot, so the two look the same. In manifest and hybrid discovery the
manifest keeps the App in the snapshot and only clears its online flag. A detector that
counted membership would make a manifest node immortal. It has to read the online flag.

**A deactivated lifecycle node is not dead.** Its process is alive and only its state
changed. That belongs to `GRAPH_NODE_INACTIVE`, and raising here too would report one
problem twice under two codes.

**The key has to be stable.** An app id is recomputed on every sweep. It is the bare node
name unless two nodes collide on it right now. On a multi-robot graph the id flips when a
same-named sibling starts or stops, so a live node held under its old id would be reported
as gone. The fully qualified name does not move.

**The grace period is wall clock, not ticks.** The entity cache is rebuilt on a graph
event, debounced to about one refresh per second. Every tick between two refreshes sees the
same snapshot, so one absent cache generation is counted again on each tick. At a fast
tick, a tolerance of two ticks is shorter than one refresh cycle, and a single stale
generation becomes an immediate ERROR. The tolerance needs a floor in milliseconds, and it
has to say so when it raises the tick count.

**Short-lived CLI nodes must not be tracked.** `ros2 topic echo` and the other CLI verbs
each start a real node called `_ros2cli_<pid>`. Discovery has no hidden-node filter, so a
few seconds of `ros2 topic echo` is enough to arm one, and Ctrl-C then leaves a permanent
death behind. Every invocation gets a new pid, so they add up.

**A new death behind an old one: a known limitation, not solved here.** A tracked entry never
ages out, so one departure nobody cares about (a closed rviz2, a calibration node that
finished) leaves the fault confirmed. While it stands, a second node's death is added to the
fault's description but arrives as an update to an already-confirmed record: no state
transition, no freeze frame, no rosbag of its own.

Acknowledging the fault between the two deaths avoids this - the fault manager treats a report
after an acknowledgement as a new occurrence, counts it, and gives it its own evidence. So the
limitation is precisely: a second death while the first fault is still outstanding does not get
its own evidence.

This is not solved in this change, and the two obvious ways to solve it were measured and
rejected rather than skipped:

- Sending a clear and a raise back to back does nothing. A single opposite-direction report
  cannot move the fault manager's hysteresis latch, so the raise lands on the already-confirmed
  branch and emits an update, which is the behaviour being complained about.
- Calling the clear service instead is worse than not doing it. It deletes the per-topic
  readings captured for that fault, and by default it also clears every symptom the correlation
  engine attributes to the fault - so an operation meant to preserve evidence destroys it.

A correct fix belongs in the fault manager, as an operation that re-confirms a record without
clearing it. Anyone picking that up should start there rather than in the detector.

## Also in scope: suppression

#570 asks for suppression to be opt-in and explicit, never a side effect. This is the first
detector that needs it, so the shared part lands here.

A suppressor answers one question: should this key be left out of the fault. Two of them
are needed now:

- an operator allowlist, for nodes whose departure is expected
- clean lifecycle shutdown, for a node that went through its own shutdown transition
  instead of dying

A suppressor also has to say whether its veto is durable. A durable veto (the node is on
the allowlist, the node shut itself down cleanly) means the bookkeeping for that node can
be reclaimed. A veto that can lift later must never reclaim anything, because then the real
fault is lost for good once the condition ends.

Nothing is suppressed unless the operator names it in the configuration.

bburda:
# Pull Request

## Summary

A node dies and nothing reports it. The node cannot report it itself, and the other detectors
cannot see it: `qos_mismatch` needs both sides present to compare profiles, `orphan` looks at a
topic whose publisher took the topic with it, and `lifecycle_expectation` only watches nodes the
operator listed as required-active.

This adds `node_death`, which raises `GRAPH_NODE_DISAPPEARED` for a node that was alive and armed
and stopped being alive, together with the suppression framework the umbrella issue asks to be
opt-in and explicit.

**Alive is not membership in the entity snapshot.** In runtime discovery a dead node leaves the
snapshot, so the two look the same; in manifest and hybrid discovery the manifest keeps the App
and only clears its online flag, so a detector counting membership would make a manifest node
immortal. Tracking is keyed on the stable fully qualified name rather than the app id, because an
id is recomputed each sweep and gains a namespace prefix once a bare-name collision exists, so a
live node held under its old id would be reported as gone.

**The grace period is wall clock, not ticks.** The entity cache is rebuilt on a debounced graph
event, so every tick between two refreshes sees the same snapshot and one absent cache generation
is counted again on each of them. At a fast tick a two-tick tolerance is shorter than one refresh
cycle, so the window has a floor in milliseconds and says so when it raises the configured value.

**Suppression is opt-in.** Nothing is suppressed unless `suppress` names it. A configured
`allowlist` that `suppress` does not name has no effect and warns that it has none. Two
suppressors ship: an operator allowlist, and clean lifecycle shutdown. A suppressor also declares
whether its veto is durable, because only a durable veto may reclaim tracker bookkeeping - a veto
that can lift later would otherwise lose a real fault for good once the condition ends.

## Also in this change: the absence boundary

`GRAPH_NODE_INACTIVE` says a required node is not active. `GRAPH_NODE_DISAPPEARED` says a node is
gone. Both can be true of one node, they have different repairs, and both standing at once is
correct - this change does not remove that.

What changed is narrower. Sustained absence used to mature a violation that had not yet been
reported, so a node observed non-active for a few ticks and then killed acquired an inactive fault
built entirely from evidence gathered after it could no longer be observed. That rule existed
because no presence detector existed to own departures. Now one does, so absence continues a
violation that has already matured and no longer creates one that has not.

With one exception, which is the part worth reviewing closely: the reliability gate refuses to arm
a managed node that never reads `active`, and `node_death` only tracks armed nodes. For such a
node the handover has no receiver, so absence may still mature its violation - otherwise a
required node that comes up unconfigured and dies is reported by nothing at all. The split is
keyed on whether the node was ever armed, latched rather than recomputed, because a managed node
legitimately reads not-armed again after an ordinary deactivate while `node_death` is still
tracking it.

## Known limitation

A second node's death, while the fault from the first is still outstanding, is added to the
fault's description but arrives as an update to an already-confirmed record: no state transition,
no freeze frame and no recording of its own. Acknowledging the fault between the two deaths avoids
this, because a report arriving after an acknowledgement is counted as a new occurrence and gets
its own evidence.

Two ways of forcing it from the detector were measured and rejected rather than skipped. Sending a
clear and a raise back to back does nothing: a single opposite-direction report cannot move the
fault manager's hysteresis latch, so the raise lands on the already-confirmed branch. Calling the
clear service instead is worse than not doing it, because it deletes the per-topic readings
captured for that fault and by default clears every symptom the correlation engine attributes to
it. A correct fix belongs in the fault manager, as an operation that re-confirms a record without
clearing it. This is documented in the package README and design doc.

## Also in this change: presence ownership rests on knowledge

Two detectors divide one job, and until now the division was decided by a fact the plugin did not
have. `GRAPH_NODE_DISAPPEARED` is raised for a node the reliability gate has armed; the gate arms a
managed node whose lifecycle state reads `active`, and it also arms one whose state has never been
measured, because it deliberately treats an unknown state as "do not gate" so that a node nobody
can read is not silenced across every other detector. Those two answers are not the same fact, and
the second was being latched permanently into the flag `lifecycle_expectation` consults before it
lets absence mature a violation. A required node that never configured could therefore be reported
by the presence detector and, by the same latch, not reported by the only detector that could still
have described what was wrong with it.

The division now asks a second, stricter question, and the gate itself is unchanged. Ownership is
`earned` when the node is not managed at all or its label reads `active`; it is withheld while the
state is merely unread and the watcher still has re-seed attempts left; and once those attempts are
spent it is `provisional` rather than earned - the node is owned, because otherwise nothing would
ever report its departure, but that grant yields the moment a real label finally arrives. A grant
made on knowledge latches, because a later deactivate does not hand the node back. A grant made on
ignorance does not, because the reason for it stops being true the instant somebody else can report
the node.

One consequence is worth stating rather than leaving to be found: a node that departs during its own
warmup is reported by nobody. That is the bringup-quiesce trade-off behaving as designed, it
predates this change, and it is written down where the boundary is described.

## Also in this change: the sanitizer test budget

The sanitizer jobs walk the workspace one package at a time under a single 45-minute budget for
the whole step. This package's end-to-end suite runs 24 minutes of that under instrumentation, up
from 10:38 on `main`, so the step stopped fitting: the TSan run was cut inside the package that
follows this one, and ASan finished the same run with 20 seconds to spare. No single scenario is
responsible - the slowest is 106 s and the rest is a 30-50 s tail across 73 targets, because these
tests wait on grace windows, respawn delays and restart loops rather than compute.

`quality.yml` now tests this package in a job of its own, once per sanitizer, and `ci.yml` does the
same over humble and lyrical, where the same cap applies to the step rather than to the job: lyrical
was cut inside this package and humble finished with three minutes to spare. The sweeps skip it. The job builds only the chain up to the package, which still includes the
integration-test package the scenarios launch from, and restores the matching sweep's ccache
instead of saving a second copy of the same objects. Instrumentation, the timeout multiplier and
the sanitizer options are unchanged from the sweep it left, and the job is not path-filtered,
because the plugin drives the gateway, the fault manager and discovery, so the changes most likely
to break it are outside its own tree.

The scenarios' own wall-clock budgets now scale with `MEDKIT_TEST_TIME_SCALE`, which the sanitizer
jobs already set to the factor they apply to every declared CTest timeout. A budget asserted inside
a test is invisible to that rewrite, so an instrumented graph that is slow to forget a departed node
blows one and the failure reads as a detector that never reported. Only give-up bounds scale: poll
intervals, enforced respawn delays and the sustained-observation windows keep their values, and the
scale is 1.0 whenever the variable is unset, so the normal jobs are unchanged.

---

## Issue

- closes #624

---

## Type

- [ ] Bug fix
- [x] New feature or tests
- [ ] Breaking change
- [ ] Documentation only

---

## Testing

The tests were written and merged before the detector, so every scenario was watched failing for
the right reason first, and the suite is what the implementation had to satisfy rather than a
description of it.

**73 ctest targets, all green, of which 42 are end-to-end** against a real gateway, a real fault
manager and a real ROS graph. 23 of those scenarios are new here.

The end-to-end set covers: a process that exits, its return and the clearing that follows, a
managed node that deactivates but keeps running, a manifest node that never came online, CLI-
convention node names, two nodes sharing a bare name in different namespaces, a fast tick against
a stale cache generation, a node restarting repeatedly across a window, a gateway restart with a
death outstanding, each suppressor including the inert-allowlist warning, pruning that must not
heal a fault that is still true, and each side of the absence boundary.

Every scenario that asserts an ABSENCE gates on the plugin being armed first and uses a helper
that fails when the fault surface is unreachable, so it cannot pass against a stack where the
plugin never loaded. Two such helpers are added here, each with a test that points it at a surface
which dies mid-window and fails if the helper still passes.

The whole suite was also run with the plugin path pointed at a file that does not exist: every
scenario fails, naming the load failure.

Reviewers can check the boundary with `ctest -R "node_death_boundary_e2e"`, the suppressors with
`ctest -R "node_death_suppression_e2e"`, and the tracker's bound with
`ctest -R "test_node_liveness_tracker"`.

No route is added or changed. The only public surface difference is additive content inside the
existing `x-medkit-watchdog` payload, which gains a `node_death` block.

---

## Checklist

### DIFF
--- a/src/ros2_medkit_plugins/ros2_medkit_graph_watchdog/README.md
+++ b/src/ros2_medkit_plugins/ros2_medkit_graph_watchdog/README.md
@@ -1393,16 +1393,33 @@ the manifest declared - rather than a gap here.
 override lands after `ON_ERROR_FAILURE`/`ON_ERROR_ERROR` out of `errorprocessing` - the
 standard way a driver reports a hardware fault it cannot recover from, which is precisely the
 death an operator needs reported, not silenced. Without an observed transition history the
-departure is unclassified and stays reported - the safe direction. This mechanism is
+departure is unclassified and stays reported - the safe direction. The suppressor itself is
 stateless by construction: a detector's `configure()` (where the suppressor chain is built)
 runs before any per-tick context exists, so it stores nothing at construction and reads the
-gate fresh on every call. Its retention window is not this detector's own to size, either:
-before this detector's own `configure()` has run, the plugin predicts the same
-`prune_ticks_` (`max(prune_grace, miss_grace + 1)`) this detector will compute, then sizes
-the lifecycle watcher's departed-node retention from it with enough margin that a
-clean-shutdown label is still cached at this detector's own reclaim tick, one tick to
-spare - see `GraphWatchdogPlugin::compute_departed_retention_ticks()` for the exact
-arithmetic, which is larger than `prune_ticks_` alone.
+gate fresh on every call.
+
+**The verdict is latched; the evidence is not.** `node_death` remembers that a durable
+suppressor vetoed a given departed key and stops re-asking for as long as that key stays
+tracked and absent. This is not an optimisation. The departed-lifecycle label the suppressor
+reads is retained for a bounded number of ticks, while the veto has to hold all the way to
+`prune()`'s reclaim - which itself needs the veto UNBROKEN for `prune_ticks` consecutive
+ticks. Re-asking every tick made those two windows race, and the retention clock starts at
+the wrong event to win it: it is anchored when the node's lifecycle SERVICES leave the graph,
+while the reclaim tick counts from when its App does, one or more sweeps later. Once the
+label expired mid-veto the suppressor abstained, the streak reset to zero, and with the label
+gone for good the key could never be suppressed again - a cleanly shut-down node named in
+`GRAPH_NODE_DISAPPEARED` for the life of the process. Latching is sound because that is what
+`durable()` means: such a verdict never lifts for a given key, so remembering it cannot
+diverge from asking again. Only positive verdicts from durable suppressors are latched, and
+the entry is dropped the moment the key is present again, so one departure's verdict is never
+carried into the next.
+
+The retention window is still not this detector's own to size: before its `configure()` has
+run, the plugin predicts the same `prune_ticks_` (`max(prune_grace, miss_grace + 1)`) this
+detector will compute and sizes the lifecycle watcher's departed-node retention from it. With
+the verdict latched, all that window has to outlive is the FIRST tick suppression is evaluated
+on; it is left sized to the reclaim tick anyway, as slack against the services-before-App lead
+above - see `GraphWatchdogPlugin::compute_departed_retention_ticks()`.
 
 **Bounded by evidence, not by age.** This detector is zero-config over every armed App in
 the graph - a strictly larger scope than `lifecycle_expectation`'s named `require_active`
@@ -1430,6 +1447,26 @@ sustained pressure from still-maturing churn the departed set may therefore exce
 exists to prevent in the first place. `tracking_saturated` reports exactly this: the departed
 set still exceeds `tracked_node_cap` even after collapsing every confirmed death it safely
 could, because immature entries alone account for the excess.
+
+**What collapsing costs, and why it is not fixed here.** Collapsing DISCARDS the identity: the
+fold is one-way, `collapsed_dead_count_` only ever grows within a tracker's life, and nothing
+decrements it when a collapsed node comes back. Two consequences follow, and both are real.
+The description stops naming those nodes and carries an unattributable `and N more` instead;
+and because the aggregated fault clears only when its dead set is EMPTY, the synthetic
+collapsed entry keeps it non-empty, so once anything has been collapsed
+`GRAPH_NODE_DISAPPEARED` cannot clear again for the life of the process - it ends by operator
+acknowledge (`DELETE /api/v1/apps/graph_watchdog/faults/GRAPH_NODE_DISAPPEARED`) or by a
+restart, never by the graph recovering.
+
+Neither is repairable from inside this tracker. Decrementing on return needs the identity the
+collapse threw away, and keeping identities is exactly the unbounded state the cap exists to
+prevent; decaying the count by time would heal a fault by waiting, which this package refuses
+everywhere else (an occurrence is closed by an operator, not by the passage of time). What
+makes it acceptable is what collapsing is allowed to touch: only a MATURED entry, whose death
+has already been reported by name. No departure becomes unreported by being collapsed - the
+fault stays raised, and it is the attribution and the self-healing that are lost, not the
+warning. Reaching the condition at all needs more than `tracked_node_cap` (512 by default)
+distinct dead identities at once, which is a churn problem in its own right.
 
 Unlike `lifecycle_expectation`'s identical-looking cap, a PRESENT/armed key here is never the
 thing evicted to free a slot. `LifecycleExpectationTracker` carries two clocks per node, so a
--- a/src/ros2_medkit_plugins/ros2_medkit_graph_watchdog/README.md
+++ b/src/ros2_medkit_plugins/ros2_medkit_graph_watchdog/README.md
@@ -96,7 +96,8 @@ built from it.
 |---|---|
 | Peer-aggregated apps (`app.source` starting `peer:`) | A peer app carries no ROS binding of its own, so `effective_fqn()` is empty for every one of them; tracking them would collapse an entire peer fleet onto one `""` key, and one online peer app would mask every other peer app's departure. |
 | `_ros2cli_<pid>` nodes | Every `ros2` CLI invocation (`ros2 topic echo`, `ros2 param get`, ...) spins up a real, short-lived node under this prefix - rcl's own hidden-node naming convention, matched structurally rather than guessed at. Letting these through would accumulate one permanently "dead" entry per CLI invocation for the life of the gateway. |
-| An App that never comes online | A key is admitted to tracking only once it has been armed by the reliability gate at least once. A manifest App whose binding never starts is never armed and so can never be falsely called dead; a node still inside its `warmup_cycles` window gets the identical protection. Once tracked, though, a node's continued life-or-death judgement rests on PRESENCE alone, not on staying armed - a tracked node that later goes lifecycle-inactive is not thereby mistaken for dead. |
+| An App that never comes online | A key is admitted to tracking only once the reliability gate has said this detector OWNS its departure - armed at least once, and on one of the two GROUNDS ownership can rest on (a lifecycle state read as `active`, or no lifecycle to read at all; the third case, a state nobody could read, is admitted PROVISIONALLY and is described in the row below). A manifest App whose binding never starts is never armed and so can never be falsely called dead; a node still inside its `warmup_cycles` window gets the identical protection. Once tracked, though, a node's continued life-or-death judgement rests on PRESENCE alone, not on staying armed - a tracked node that later goes lifecycle-inactive is not thereby mistaken for dead. |
+| A managed App whose lifecycle state has not been read YET | Not tracked, and only for as long as that is still true. The gate's own raise permission is answered PERMISSIVELY for a managed node whose `GetState` has not answered (`LifecycleWatcher::node_ok()`) - without that, `qos_mismatch`, `orphan` and `param_drift` would all fall silent about a node whose lifecycle service is broken. Permission is not knowledge, though, so tracking is withheld while a measurement may still arrive. `LifecycleWatcher` charges a GetState re-seed budget per node, and only for a read that actually ran, so "still asking" is a fact rather than a guess. Once that budget is spent the node is tracked here after all: nothing will ASK again, and `lifecycle_expectation` looks at nothing unless an operator named the node in `require_active` (empty by default), so withholding it any longer would mean the death is reported by nobody. That admission is PROVISIONAL, not final. The `~/transition_event` subscription outlives the seed budget, so a label can still arrive without being asked for, and one that reads non-active says the node was the lifecycle detector's all along - so the presence detector hands it back on the spot, while it is still alive. Waiting for the budget alone would only have delayed the wrong report, not prevented it. |
 
 Composable/component nodes hosted in one process die together: if the container process
 dies, every node it hosted leaves the snapshot on the same tick, and all of them are folded
@@ -720,8 +721,8 @@ node's SETTLED observation had started, and resets nothing:
 | Settled observation | What sustained absence does |
 |---|---|
 | `inactive`, ALREADY reported under `GRAPH_NODE_INACTIVE` | the streak continues exactly as it did on its last present tick, so the fault stays raised and keeps naming a node that is no longer there |
-| `inactive`, not yet past `grace`, node WAS armed at some point | the streak is HELD - neither advanced (no fault is born while nobody can observe the node) nor erased (a node that returns still inactive resumes rather than re-earning `grace`). The presence detector could report this exact departure instead (`node_death` tracks any node the reliability gate has armed at least once), so this detector does not need to. |
-| `inactive`, not yet past `grace`, node was NEVER armed | the streak keeps climbing on absence exactly as it would on a present tick. `node_death` only ever tracks a node the gate has armed, so a node that never reached that bar is structurally invisible to it no matter what happens afterwards - if this detector held the streak too, the departure would be reported by nothing at all. |
+| `inactive`, not yet past `grace`, node the presence detector OWNED at some point | the streak is HELD - neither advanced (no fault is born while nobody can observe the node) nor erased (a node that returns still inactive resumes rather than re-earning `grace`). The presence detector could report this exact departure instead (`node_death` tracks any node the reliability gate has admitted for presence ownership at least once), so this detector does not need to. |
+| `inactive`, not yet past `grace`, node the presence detector NEVER owned | the streak keeps climbing on absence exactly as it would on a present tick. `node_death` only ever tracks a node the gate has admitted for ownership AND that is online, so a node that never reached that bar - one that never reached `active`, or one whose ROS binding never started - is structurally invisible to it no matter what happens afterwards. If this detector held the streak too, the departure would be reported by nothing at all. |
 | unread label, or no tracked lifecycle | the unmeasured clock keeps climbing under that same cause, so the node is eventually reported under `GRAPH_NODE_UNREADABLE` / `GRAPH_NODE_NOT_MANAGED` |
 | `active` | nothing. Anything the node had started but not corroborated is released, the entry becomes idle and is reclaimed silently |
 
@@ -781,17 +782,22 @@ Where the presence class CAN also report the same departure, that is no longer t
 thing standing between such a node and invisibility: `GRAPH_NODE_DISAPPEARED` (this
 package's `node_death` detector) independently reports it, whether or not the node was
 ever measured not-active first. But `node_death` only ever tracks an App once the
-reliability gate has armed it, and the gate refuses to arm a MANAGED node that is not
-`active` (`LifecycleWatcher::node_ok()`) - so a `require_active` node that never reaches
-`active` is never armed, is never tracked by `node_death`, and its departure can never
-raise `GRAPH_NODE_DISAPPEARED`, whatever kills it. The tracker therefore splits on whether
-a node was EVER armed, read from the reliability gate itself rather than guessed from the
-observed label: once armed, a below-`grace` streak that goes absent is simply HELD -
-maturing it would raise `GRAPH_NODE_INACTIVE` from ticks gathered while nobody could
-observe the node, a fault its presence never earned, and `GRAPH_NODE_DISAPPEARED` is there
-to report the departure instead. A node that was NEVER armed gets no such backstop -
-`GRAPH_NODE_DISAPPEARED` structurally cannot report it - so absence keeps advancing its
-streak exactly as a present tick would, maturing it within the same bound an armed node's
+reliability gate says it OWNS that App's departure, and ownership needs the node's lifecycle
+state either KNOWN not to be a managed-non-active one (no managed record at all, or a record
+reading `active`) or asked for as often as it ever will be. So a `require_active` node that
+never reaches `active` is never tracked by `node_death`, and neither is an app that is not
+online. A managed node whose `GetState` has not answered is withheld only while the watcher is
+still asking: the gate would let it raise throughout (`LifecycleWatcher::node_ok()` treats an
+unread label as permission, deliberately), and once the re-seed budget is spent the node is
+`node_death`'s after all, because nothing else would ever report it. The tracker therefore
+splits on
+whether the presence detector EVER owned a node, read from the reliability gate itself rather
+than guessed from the observed label: once owned, a below-`grace` streak that goes absent is
+simply HELD - maturing it would raise `GRAPH_NODE_INACTIVE` from ticks gathered while nobody
+could observe the node, a fault its presence never earned, and `GRAPH_NODE_DISAPPEARED` is
+there to report the departure instead. A node the presence detector never owned gets no such
+backstop - `GRAPH_NODE_DISAPPEARED` structurally cannot report it - so absence keeps advancing
+its streak exactly as a present tick would, maturing it within the same bound an owned node's
 UNMEASURED clock already has (`grace` + `absence_grace` + 1 ticks). An ALREADY-matured
 streak still continues through absence exactly as before, for either kind of node: two
 codes standing at once for the same node (`GRAPH_NODE_INACTIVE` because it was measured
@@ -1364,6 +1370,21 @@ the lifecycle watcher's cached label for the departed node's LAST observed trans
 | `unconfigured` | No, deliberately - it is also the resting state of a node whose `configure()` failed or that never activated at all, and suppressing on it would hide exactly that startup failure. |
 | Any other label, or no departure on record at all | No (abstains). |
 
+**Which bindings the mechanism can see.** The departed-lifecycle record it reads is written
+when a tracked app's `get_state` path stops appearing in `App::services`, which is what the
+lifecycle watcher builds its managed set from. `App::is_online` plays no part in that, and the
+two do not have to coincide: a manifest-declared App outlives the node it binds, keeping its id
+and fqn while only its runtime-derived collections empty out. That case still works - the
+services go with the node, so the departure is recorded - because a manifest cannot declare
+services at all: `ManifestParser::parse_app` reads `id`, `name`, `description`,
+`is_located_on`, `depends_on`, `tags`, `external` and `ros_binding`, and nothing else, so an
+app's services can only ever come from a live runtime match. The case the mechanism genuinely
+cannot cover is `discovery.inherit_runtime_resources: false`, which restores the manifest's own
+(empty) collections on every sweep: such an app is never lifecycle-tracked in the first place,
+so it has no lifecycle label, `node_ok()` never gates it, and there is nothing for this
+suppressor to classify. That is the documented meaning of the flag - the app exposes only what
+the manifest declared - rather than a gap here.
+
 `finalized` needs the extra corroboration because it is also where a node's `on_error`
 override lands after `ON_ERROR_FAILURE`/`ON_ERROR_ERROR` out of `errorprocessing` - the
 standard way a driver reports a hardware fault it cannot recover from, which is precisely the
@@ -1424,7 +1445,7 @@ present but not active, `GRAPH_NODE_DISAPPEARED` says a node is gone, and an ope
 each has a different repair - reactivate it, or find out why it left. A node CONFIRMED
 `GRAPH_NODE_INACTIVE` that then departs keeps that fault (its description switches to say
 the node has since left the graph) AND now also raises `GRAPH_NODE_DISAPPEARED`, since this
-detector tracks every armed App independently of whatever `lifecycle_expectation` thinks of
+detector tracks every App it owns independently of whatever `lifecycle_expectation` thinks of
 it.
 
 What changed is narrower than that, and it only applies to a node this detector could ever
@@ -1432,17 +1453,19 @@ have tracked. Before this detector existed, `lifecycle_expectation`'s own violat
 let a SUSTAINED absence mature an unconfirmed (below-`grace`) streak into a confirmed
 `GRAPH_NODE_INACTIVE` - the only way a node stuck in a restart loop, never observed long
 enough to cross `grace` while present, would ever be reported at all. Where this detector
-can independently report the same departure - which needs the node to have been ARMED at
-least once, since `node_death` only ever tracks an App the reliability gate has armed, and
-the gate refuses to arm a MANAGED node that is not `active` - that absence-alone maturity is
-gone: absence CONTINUES a violation that has already matured past `grace` (the fault stays
-raised, still naming the node), but no longer CREATES one that has not. A node that was
-briefly non-active and then died is reported as gone (`GRAPH_NODE_DISAPPEARED`) rather than
-also acquiring an inactive fault born from ticks gathered while nobody could observe it. A
-`require_active` node that never reaches `active` is never armed, is never tracked here, and
-its departure can never raise `GRAPH_NODE_DISAPPEARED` - for that one node,
+can independently report the same departure - which needs the presence detector to have OWNED
+the node at least once, since `node_death` only ever tracks an App the reliability gate has
+admitted for ownership, and ownership of a MANAGED node is EARNED only by a lifecycle state
+read as `active` (a state nobody could read earns nothing: it is granted provisionally and given
+up again the moment a label arrives) - that absence-alone maturity is gone: absence CONTINUES a violation that has
+already matured past `grace` (the fault stays raised, still naming the node), but no longer
+CREATES one that has not. A node that was briefly non-active and then died is reported as gone
+(`GRAPH_NODE_DISAPPEARED`) rather than also acquiring an inactive fault born from ticks
+gathered while nobody could observe it. A `require_active` node that never reaches `active` is
+never owned, is never tracked here, and its departure can never raise
+`GRAPH_NODE_DISAPPEARED`; the same holds for an app that is not online. For those,
 `lifecycle_expectation` keeps the older behaviour: absence still matures a below-`grace`
-streak, because it is structurally the only detector that will ever get to report it. See
+streak, because it is structurally the only detector that will ever get to report them. See
 `lifecycle_expectation`'s own "Absence continues the last CORROBORATED observation, it never
 erases" section above for the mechanism.
 
@@ -1484,6 +1507,31 @@ Acknowledging between the two deaths avoids this: `DELETE
 /api/v1/apps/graph_watchdog/faults/GRAPH_NODE_DISAPPEARED` closes the first occurrence, so the
 second node's next report reactivates a CLEARED record instead of updating a CONFIRMED one - a
 genuine new occurrence, with its own `occurrence_count` and its own capture.
+
+**A fault outlives a gateway restart even if the node comes back, and only an acknowledgement
+closes it.** Within one process this detector heals on return: the node reappears, the dead set
+empties, a PASSED flows and the fault heals. Across a restart it does not, in either direction.
+The restarted instance may clear `GRAPH_NODE_DISAPPEARED` only once it has ITSELF put a FAILED
+on the wire for it, because until then it cannot tell "the node came back" from "I never saw
+that node" - and clearing on the second reading is the ungated heal the guard exists to
+prevent. So a fault CONFIRMED before a restart stays CONFIRMED afterwards whether or not the
+node returned, and with the sqlite backend it survives every reboot until
+`DELETE /api/v1/apps/graph_watchdog/faults/GRAPH_NODE_DISAPPEARED` acknowledges it. That is the
+acknowledge model doing what it is for, not an accident: an occurrence is closed by an
+operator, not by the passage of a restart.
+
+Re-seeding the tracker from the fault store at startup would change it, and today it cannot be
+done. `/fault_manager/list_faults` would tell this detector that a `GRAPH_NODE_DISAPPEARED`
+record is outstanding and that it is among its reporting sources - but not WHICH nodes it
+names, which is the only thing that would make a fresh instance's silence meaningful. The fault
+manager keeps one record per `fault_code`, and this detector folds every dead key into that one
+record's description, capped at `kMaxDescriptionChars` with the remainder collapsed into a
+count, so the key list is not recoverable even by parsing prose. `lifecycle_expectation` has the
+same boundary and resolves it differently (a bounded hold, then the clear flows) because its
+`require_active` set is enumerable from config alone; this detector is zero-config over the
+whole graph and has no such set. The integration test
+`RestartedInstanceNeverClearsAFaultItDidNotRaiseEvenAsTheNodeReturns` pins the behaviour so it
+cannot change silently.
 
 **Test tiers.**
 
@@ -1534,7 +1582,7 @@ genuine new occurrence, with its own `occurrence_count` and its own capture.
    once the pressure passes; and a cap-forced collapse stays raised through a genuine
    recovery but clears once reconfigured, mirroring the ungated-clear guard's own reconfigure
    test for an ordinary (uncollapsed) fault.
-3. **E2e**, three files, each launching its own real gateway + fault_manager + demo-node
+3. **E2e**, four files, each launching its own real gateway + fault_manager + demo-node
    stack:
    - `test/e2e/test_node_death_e2e.test.py` (ten scenarios): raise and name the node; clear
      once it returns; no heal with the fault manager's healing disabled; a lifecycle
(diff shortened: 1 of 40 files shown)


src/ros2_medkit_plugins/ros2_medkit_graph_watchdog/src/detectors/node_death_detector.cpp
```diff
@@ -0,0 +1,601 @@
...
+          armed.insert(key);
+          break;
+        case PresenceOwnership::kProvisional:
+          armed.insert(key);
+          break;
+        case PresenceOwnership::kNone:
+          // Never earned, and no longer owned: hand it back while it is still alive. A key
+          // that WAS earned keeps its admission - a node that ran and then deactivated is
+          // still a death when it stops running.
+          if (earned_.count(key) == 0) {
```
mfaferek93, line 351: This hand-back fires for every node that dies and returns, not only for a provisional grant. During the absence `earned_` is pruned (line 341), so on the return tick the re-warming entity answers `kNone` from `allows_raise()` and the key is released from the tracker. If the node dies again before it re-arms (crash loop faster than `warmup_cycles`), the second death is never reported - and since `ever_raised_` is already true and `report.dead` is now empty, the detector emits PASSED every tick while the node is dead again, exactly the false heal the tracker doc rules out.

I reproduced it: die -> raise -> one-tick return -> die again gives 0 new FAILED and 10 false PASSED over 10 ticks, and `tracked_count` drops 2 -> 1 on the return tick. The e2e restart-loop scenarios miss it because they wait for `wait_until_watchdog_armed` before every kill, and the integration tests mostly run `warmup_cycles=0`.

Suggest handing back only when the gate's answer comes from a measured non-active label (or the key is positively known to have been admitted on kProvisional grounds), never on plain `!allows_raise()` during re-warmup. A crash-looping node with uptime shorter than `warmup_cycles` is otherwise reported once, healed, and then silent forever.

src/ros2_medkit_plugins/ros2_medkit_graph_watchdog/src/detectors/node_death_detector.cpp
```diff
@@ -0,0 +1,601 @@
...
+    // good proxy: any App elsewhere in the graph - the gateway's own node included - tends
+    // to arm within a tick or two of startup, which would make tracker_.tracked_count() go
+    // non-zero long before the SPECIFIC node a stored, outstanding fault names has been
+    // re-observed at all. A guard keyed on that count would open on an unrelated node's
+    // arming and clear a fault this process has not actually re-verified - exactly the
+    // silent, ungated heal this guard exists to rule out. report.dead's own emptiness says
+    // nothing that specific either (every dead key is merged into one aggregate), which is
+    // why the guard is keyed on whether THIS instance has raised at all, not on what it
+    // currently knows about.
+    if (report.dead.empty() && !ever_raised_) {
```
mfaferek93, line 499: The `ever_raised_` guard covers the restart-and-still-gone quadrant that restart_rebaseline pins, but the fourth quadrant is unhandled: gateway restarts, then the node RETURNS. The fresh instance tracks the node (present, armed, misses 0), `report.dead` stays empty, `ever_raised_` stays false, so no PASSED is ever sent - I verified a fresh detector emits zero PASSED over 20 ticks of re-observing the returned node healthy. With the default sqlite backend the pre-restart CONFIRMED `GRAPH_NODE_DISAPPEARED` then survives every reboot and only a manual DELETE closes it, even though this instance has re-observed the exact named node. The common shape is a robot reboot after a crash: everything comes back green except a permanently CONFIRMED death fault.

lifecycle_expectation heals its analogous restart case with a bounded hold (restart_departed pins the heal), so the asymmetry will surprise operators. If permanent-until-acknowledged is the intended DTC-like semantics here, please pin it with a test and say so in the README/restart_rebaseline docstring - the real fix probably needs the detector to adopt outstanding `GRAPH_NODE_DISAPPEARED` keys from the store at startup, which the fire-and-forget client cannot do today.

src/ros2_medkit_plugins/ros2_medkit_graph_watchdog/include/ros2_medkit_graph_watchdog/lifecycle_shutdown_suppressor.hpp
```diff
@@ -0,0 +1,77 @@
...
+/// Stateless by design: a detector's own `configure()` (where the suppressor chain is built)
+/// runs before any DetectorContext exists, so this class stores nothing at construction and
+/// reads the gate fresh on every call instead.
+class LifecycleShutdownSuppressor : public Suppressor {
+ public:
+  bool suppresses(const std::string & fqn, const DetectorContext & ctx) const override {
+    if (ctx.gate == nullptr) {
+      return false;  // not yet wired (e.g. a bare-context test) -> nothing to read, abstain
+    }
+    const auto departed = ctx.gate->departed_lifecycle_state_of(fqn);
```
mfaferek93, line 53: The departed-lifecycle record this reads is only written when the app's binding leaves `current_paths` in LifecycleWatcher::update(), i.e. when the get_state service disappears from `App::services`. node_death's liveness is `is_online`, which is a different predicate: a manifest/hybrid app with `inherit_runtime_resources=false` and a declared lifecycle service keeps its get_state path in the snapshot after the node dies, so the watcher never records a departure while node_death reports the key dead. Result: with `suppress: ["lifecycle"]`, a clean SHUTDOWN of such a manifest node is still reported as GRAPH_NODE_DISAPPEARED. Fail-open, so no false heal - but the suppression the operator opted into silently never applies, and the durable-veto prune path never reclaims the key either. Worth either recording a departure when `is_online` flips false for a tracked id, or a line in the clean-shutdown README section scoping the mechanism to runtime-discovered bindings. Flagging from reading the code (merge_pipeline.cpp restores manifest-declared services after the merge; the watcher's `current_paths` loop does not consult `is_online`) - I did not build a manifest-app repro for this one.

src/ros2_medkit_plugins/ros2_medkit_graph_watchdog/src/detectors/node_death_detector.cpp
bburda: You are right, and the trace matches what I found. During the absence the key is pruned from the earned set. On the return tick the gate then answers the same way for a node that is re-warming and for a node measured not active, so the key was handed back to a node that was only warming up.

The gate now returns four grounds instead of true or false: nothing known yet, measured and not ours, provisional, and earned. Only "measured and not ours" takes a key away. Re-warming does not.

Working on that fix surfaced a second path into the same fault. A dying managed node loses its lifecycle services one sweep before its App entry goes. With no lifecycle record the node reads like a plain node, which is an earned ground, so a handed-back key walked back in on the tick before it died. A key that was handed back is now re-admitted only on a label that reads active, never on a lost record.

Both are covered now: an e2e that kills, waits for the return, and kills again without waiting for arming, and an integration test that drives the sweep order for the second one.

src/ros2_medkit_plugins/ros2_medkit_graph_watchdog/src/detectors/node_death_detector.cpp
bburda: Confirmed, and the behaviour stays, with a test to hold it.

I checked whether the detector can find out at startup which nodes an outstanding GRAPH_NODE_DISAPPEARED names. It cannot. The fault message carries one record per fault code, and this detector puts every dead node into that single description, with the rest collapsed into a count. So a fresh instance cannot tell "this node came back" from "I never saw this node". Clearing on the second reading would be an ungated clear, which two existing tests forbid.

So this is pinned instead of changed: a restarted detector sends no PASSED over 20 ticks while the node is healthy again, and dropping the guard turns that test red. The README now says it next to the acknowledge model, including why lifecycle_expectation can use a bounded hold and this one cannot. Its required set comes from config and can be listed. This detector watches the whole graph with no config.

src/ros2_medkit_plugins/ros2_medkit_graph_watchdog/include/ros2_medkit_graph_watchdog/lifecycle_shutdown_suppressor.hpp
bburda: I checked both halves. You are right about the first one. LifecycleWatcher::update() keys on the get_state path in App::services and never reads is_online.

The second half does not hold. A manifest app cannot declare services. The manifest parser reads id, name, translation_id, description, is_located_on, depends_on, tags, external and ros_binding, and nothing else. Services can only come from a live runtime match, and that match is gone the sweep the node dies. So the departure is recorded and the suppression applies.

Added a watcher test for that exact shape: the App stays with its id and fqn, is_online is false, the services are gone, and the departure record must be written.

One real consequence is now in the README clean-shutdown section. With discovery.inherit_runtime_resources set to false the app carries no services on any sweep, so it is never lifecycle-tracked and there is nothing for the suppressor to classify.

src/ros2_medkit_plugins/ros2_medkit_graph_watchdog/src/detectors/lifecycle_expectation_detector.cpp
```diff
@@ -308,9 +320,40 @@ class LifecycleExpectationDetector : public Detector {
...
+        // the presence class had a node it had already handed back - the same silence in the
+        // opposite direction.
+        //
+        // ANDed with is_online because node_death skips an app that is not online before it
+        // ever consults the gate, and the gate has no notion of online-ness: a manifest app
+        // whose node never started is armed, carries no lifecycle record, and would otherwise
+        // read as owned by a detector that will never look at it. The remaining filters
+        // node_death applies need no mirror here: a peer-aggregated app and one with no binding
+        // both have an empty fqn and are already skipped above.
+        const bool armed = app.is_online && presence_ownership(ctx.gate, app.id) == PresenceOwnership::kEarned;
```
mfaferek93, line 353: `armed = app.is_online && presence_ownership(...) == kEarned` mirrors the gate but not node_death's admission rule, and the two now disagree on one sweep. A dying managed node loses its get_state service one sweep before its App goes (the shape c5563770 guards against in node_death), the watcher drops the record, and the gate answers kEarned for a node measured `unconfigured` a tick earlier. That latches `ever_armed` (tracker.hpp:556), so absence HOLDS the below-grace streak (tracker.hpp:659) instead of maturing it - while node_death never admitted the key (disowned on every measured tick, and the released-until-active rule at node_death_detector.cpp:318 refuses the kEarned sweep).

Net effect: a `require_active` node that sat `unconfigured` and then died raises nothing, forever, and the INACTIVE clear for every other node stays withheld. On main it matured within grace + absence_grace + 1 ticks.

I reproduced it with both detectors on one gate: `require_active: ["/victim"]`, grace 30; 10 present ticks labelled unconfigured; one present tick with no lifecycle record; 60 absent ticks -> 0 GRAPH_NODE_INACTIVE, 0 GRAPH_NODE_DISAPPEARED. The same sequence without the drop sweep gives 37 INACTIVE FAILED. Cleanest fix is probably not to re-derive ownership here at all and ask node_death what it tracks (`NodeLivenessTracker::known_keys()` is already a seam); short of that, `armed` needs the same released-until-active rule, so a kEarned resting on "no record" never latches for a node this tracker already measured non-active.

src/ros2_medkit_plugins/ros2_medkit_graph_watchdog/src/detectors/node_death_detector.cpp
```diff
@@ -0,0 +1,662 @@
...
+          // between the label reaching the status route and the release, against a ~1 s entity
+          // cache refresh. The window cannot be closed by deciding at REPORT time instead: by
+          // then the key has departed, `earned_` has been pruned for it (that prune is what
+          // bounds this map against identity churn), and the departed record carries only the
+          // LAST label - which reads "inactive" both for a node that was earned and then
+          // deactivated, which must be reported, and for one that was only ever provisional,
+          // which must not. Telling those apart at report time needs per-key ownership history
+          // that survives the departure, which is the unbounded state the prune exists to
+          // prevent. The window is stated in the README rather than papered over.
+          if (earned_.count(key) == 0) {
```
mfaferek93, line 351: The hand-back here only checks `earned_`, and `earned_` is pruned to the live graph every tick (390-392), so a node this detector has already REPORTED dead loses its earned ground during the outage. When it respawns, a lifecycle node's first GetState reads `unconfigured` (nothing re-drives it without a lifecycle manager) - that is kDisowned with `earned_` empty, so the key is released, the dead set empties and PASSED heals the fault. If it crashes again, node_death is not tracking it, and lifecycle_expectation holds the streak because `ever_armed` survives across incarnations (tracker.hpp:556), so the `!ever_armed` advance at :659 never runs. One report, a heal, then silence for the rest of the crash loop - the case the README says this detector most exists for. README:724 justifies the hold with "node_death tracks any node the gate has admitted at least once", which does not hold across a departure.

Reproduced twice: plain node_death (tracked 2 -> 1 on the respawn, 1 PASSED, FAILED count 1 before and after the second death) and with lifecycle_expectation on the same gate and `require_active: ["/victim"]` (PASSED x2 on respawn, DISAPPEARED 1/1, INACTIVE 0 after 60 absent ticks). Suggest not handing back a key whose tracker entry is currently carrying a death (expose `NodeLivenessTracker::is_departed(key)` and keep it on kDisowned while set), and re-deriving `ever_armed` for a returning incarnation the way node_death re-derives `earned_`.

src/ros2_medkit_plugins/ros2_medkit_graph_watchdog/test/e2e/test_node_death_suppression_e2e.test.py
```diff
@@ -0,0 +1,788 @@
...
+            _poll_apps_absent(PORT, KILLED_NODE, timeout=DEPARTURE_TIMEOUT_SEC),
+            f'{KILLED_NODE} never left GET /apps after SIGTERM')
+
+        fault = poll_faults(PORT, FAULT_CODE, timeout=RAISE_TIMEOUT_SEC)
+        if fault is None:
+            self.fail(
+                f'{FAULT_CODE} never raised after {KILLED_NODE} (active, killed outright) '
+                'exited - a departure lifecycle_expectation has nothing to say about must still '
+                'be named')
+
```
mfaferek93, line 644: This row never reaches the suppressor. CLEAN_NODE is launched with `auto_activate=False` (:345), polled as `unconfigured` (:605), driven UNCONFIGURED_SHUTDOWN to finalized and only then killed - it is never `active`, so `presence_ownership()` answers kDisowned on every measured tick and node_death never admits the key. It never enters `report.dead`, so LifecycleShutdownSuppressor is never consulted, and `forbidden=[CLEAN_NODE]` stays true with `suppress: ["lifecycle"]` removed or the suppressor deleted. That contradicts the docstring at :55-57 / :93-95 and the README test-tier claim; the only real-stack proof of the mechanism is vacuous (the integration tests through the seam are fine). To exercise it the node has to be measured active first, then go through deactivate/cleanup/shutdown with the watcher seeing the transitions, then be killed. Verified by reading reliability_gate.cpp:244-246 and node_death_detector.cpp:333-355 against the row.

src/ros2_medkit_plugins/ros2_medkit_graph_watchdog/test/e2e/test_node_death_e2e.test.py
```diff
@@ -0,0 +1,1492 @@
...
+    gate still skips the comparison the same way a single-sample None always did.
+    """
+    deadline = time.monotonic() + timeout
+    status = None
+    last_count = None
+    streak_start = None
+    while time.monotonic() < deadline:
+        now = time.monotonic()
+        status = watchdog_detector_status(port, detector_id)
+        count = status.get('tracked_count') if status is not None else None
```
mfaferek93, line 652: `_poll_stable_tracked_count` crashes when the first sample is None: `last_count` and `streak_start` start as None, `watchdog_detector_status()` returns None on any RequestException, slow first GET or non-200, so on that poll `count != last_count` is False and the `elif` evaluates `now - streak_start` with `streak_start` still None - TypeError. The docstring promises "None on every poll counts as stable at None"; the code does the opposite. On a loaded TSan runner one slow first GET turns `ros2cli_ignored` (and the suppression file's prune row, same helper at test_node_death_suppression_e2e.test.py:458-466) into an ERROR unrelated to the detector. Seeding `streak_start = time.monotonic()` before the loop fixes both copies.

src/ros2_medkit_plugins/ros2_medkit_graph_watchdog/test/e2e/test_node_death_e2e.test.py
```diff
@@ -0,0 +1,1492 @@
...
+        )
+
+    def test_02_a_gateway_restart_does_not_manufacture_a_clear(self, gateway_node, target_node):
+        old_pid = gateway_node.process_details['pid']
+        os.kill(old_pid, signal.SIGTERM)
+        self.assertTrue(
+            _wait_until_port_is_down(PORT, timeout=60.0),
+            f'the gateway (pid {old_pid}) kept answering after SIGTERM - nothing restarted')
+        self.assertTrue(
+            wait_until_watchdog_armed(PORT, timeout=90.0),
```
mfaferek93, line 1420: These two are the only give-up budgets in the file not multiplied by `TIME_SCALE` (the module says at :199-204 that every one is), while test_lifecycle_expectation_e2e.test.py:1111,1120 scales the identical restart drive. Under the x3 sanitizer rewrite the re-arm after a full gateway respawn gets a third of the budget the first arming got. Related race in the boundary file: `B2_GRACE = 15` and `B6_GRACE = 15` (test_node_death_boundary_e2e.test.py:207,210) give a 16-tick = 3.2 s nominal window from DEACTIVATE to maturity, and the row spends the label poll, the tracked-nodes poll, `_fault_present_now` and the ~1.1 s post-kill snapshot tail inside it. Lose it in B2 and `assert_fault_absent_throughout(INACTIVE)` goes red blaming the detector; lose it in B6 and INACTIVE raises via the ordinary present path, so the row is green without the `!ever_armed` absence branch ever running. Scale the two timeouts and raise the two graces (40-50, or `* TIME_SCALE`). Reading only - I did not run the e2e locally.

src/ros2_medkit_plugins/ros2_medkit_graph_watchdog/src/detectors/node_death_detector.cpp
```diff
@@ -0,0 +1,662 @@
...
+          // active" is, "there is no managed record at all" is not. For a key this detector has
+          // already handed back that difference decides everything, because a dying managed
+          // node produces the second shape on its way out - its lifecycle services leave the
+          // snapshot a sweep before the App does, the watcher drops the record, and a node with
+          // no record reads exactly like a plain one. Re-admitting on that would let the act of
+          // dying erase the measurement that disowned the node, and the death would be reported
+          // by the detector the label had already disqualified. So a released key waits for the
+          // graph to say ACTIVE again; the disappearance of what disowned it is not news that
+          // the node became ours.
+          if (released_.count(key) != 0 && !reads_active(ctx.gate, app.id)) {
```
mfaferek93, line 318: The released-until-active rule is right for the dying-node sweep, but it has no exit for a node that stays alive. A key in `released_` is only re-admitted on a label reading active, and `released_` is only pruned on absence - so a node that was disowned (inactive) and then genuinely stops being managed (drops its lifecycle services while alive - the droppable_lifecycle_node shape the presence e2e already treats as real) answers kEarned with no record on every later sweep and is never re-admitted, and nothing else reports its eventual death. Not a regression against main, but the README's "What this does NOT close" list does not mention it. Reproduced: present as a plain node for 20 sweeps after the disown, tracked_count stays 1, 0 FAILED after it dies. Either expire `released_` after the watcher's departed-retention window, or document the shape.

src/ros2_medkit_plugins/ros2_medkit_graph_watchdog/design/graph_watchdog.rst
```diff
@@ -866,9 +958,340 @@ overlapping - plus the shared-watcher seam the re-bind behaviour lives in:
...
+at construction and reads the gate fresh on every call - and its retention window is not
+this detector's own to size: before this detector's own ``configure()`` has run, the
+plugin predicts the same ``prune_ticks_`` (``max(prune_grace, miss_grace + 1)``) this
+detector will compute, then sizes the lifecycle watcher's departed-node retention from it
+with enough margin that a clean-shutdown label is still cached at this detector's own
+reclaim tick, one tick to spare
+(``GraphWatchdogPlugin::compute_departed_retention_ticks()`` - the resulting window is
+larger than ``prune_ticks_`` alone).
+
+**Bounded by evidence, not by age - and why this cap, unlike the sibling's, cannot
```
mfaferek93, line 1097: This section still describes the old cap design: idle (present) entries evicted first, collapsing always frees room so a newcomer is never refused, `tracking_saturated` "never observed true here", and it cites `NodeLivenessTrackerCap.SaturationNeverFiresBecauseEveryKeyIsEitherIdleOrCollapsible`. The code and the README say the opposite - a present entry is never evicted (node_liveness_tracker.hpp:85-96), an immature departed entry is never collapsed (:98-108), `tracking_saturated` is true whenever immature entries alone exceed the cap (`enforce_departed_cap()`), and the test that exists is `AtZeroMissGraceEveryDepartedEntryIsInstantlyMaturedSoSaturationNeverFires`. Worth bringing in line with README "Bounded by evidence, not by age".

src/ros2_medkit_plugins/ros2_medkit_graph_watchdog/README.md
```diff
@@ -1199,6 +1315,330 @@ still fit inside the 480-char cap (`3 * 150 + 2 * 2 = 454 <= 480`).
...
+acknowledge model doing what it is for, not an accident: an occurrence is closed by an
+operator, not by the passage of a restart.
+
+Re-seeding the tracker from the fault store at startup would change it, and today it cannot be
+done. `/fault_manager/list_faults` would tell this detector that a `GRAPH_NODE_DISAPPEARED`
+record is outstanding and that it is among its reporting sources - but not WHICH nodes it
+names, which is the only thing that would make a fresh instance's silence meaningful. The fault
+manager keeps one record per `fault_code`, and this detector folds every dead key into that one
+record's description, capped at `kMaxDescriptionChars` with the remainder collapsed into a
+count, so the key list is not recoverable even by parsing prose. `lifecycle_expectation` has the
```
mfaferek93, line 1529: "not recoverable even by parsing prose" overstates it. The description is this detector's own deterministic text - `node <fqn> disappeared (N missed cycles)` joined by `; ` - persisted and updated on every FAILED (ReportFault.srv), and `/fault_manager/list_faults` returns it. For every record whose description was not capped (one dead node - the common shape) the key IS recoverable; only the collapsed remainder is not, and those heal nowhere today anyway. Pinning the behaviour for this PR is a fair call (a ListFaults client at startup is its own change), but the sentence should say the detector chooses not to parse rather than that it cannot, and the operational consequence deserves one plain sentence: after any reboot that follows a node death the operator must DELETE GRAPH_NODE_DISAPPEARED by hand, even though everything is green.

src/ros2_medkit_plugins/ros2_medkit_graph_watchdog/src/detectors/lifecycle_expectation_detector.cpp
```diff
@@ -308,9 +320,40 @@ class LifecycleExpectationDetector : public Detector {
...
+        // does not actually hold - either because the watcher has not finished asking about it,
+        // or because it asked, gave up, and then a label arrived and took the node back.
+        //
+        // kEarned only, and the two exclusions are for different reasons.
+        //
+        // A PROVISIONAL grant is deliberately not enough: the tracker latches this value for
+        // the node's whole life, and a provisional grant is the one node_death itself gives up
+        // the moment a real label arrives. Latching it here would leave this detector believing
+        // the presence class had a node it had already handed back - the same silence in the
+        // opposite direction.
```
mfaferek93, line 345: Related to the hold this block explains: the once-per-episode withheld-clear warning at :569 (outside this diff) still says "that hold releases as soon as the node reads active or its streak passes grace". For an ever-armed node that has DEPARTED neither can happen until it returns (README:724 and tracker.hpp:664-666 say the hold is unbounded), so the only diagnostic an operator gets tells them to wait for something that is not coming. In my run the same line also promised a GRAPH_NODE_NOT_MANAGED report "after 60 ticks", which the absence path zeroes (tracker.hpp:636), so it never fires either. The tracker already knows `is_departed`; saying "until the node returns" for that case would be honest.

src/ros2_medkit_plugins/ros2_medkit_graph_watchdog/include/ros2_medkit_graph_watchdog/lifecycle_expectation_tracker.hpp
```diff
@@ -132,6 +137,20 @@ struct LifecycleMatch {
   std::string entry;  ///< the require_active entry that matched
   std::string fqn;    ///< the node's stable App::effective_fqn()
   std::optional<std::string> state;
+  /// Whether the presence detector will OWN this node's departure. Narrower than "may
+  /// raise": it needs the node's lifecycle state either known not to be a managed-non-active
+  /// one or asked for as often as it ever will be, so a managed node whose state has not
```
mfaferek93, line 142: This doc contradicts the detector. Lines 142-144 say the flag "answers true again once the watcher has stopped asking"; the detector deliberately passes `== kEarned` only and explains at lifecycle_expectation_detector.cpp:339-345 why kProvisional must not latch. Lines 147-151 say defaulting to true "keeps the already-shipped absence handling" - the shipped handling on main was advance-on-absence, and `true` selects the new HOLD path, so an unwired caller silently loses a reporting path rather than keeping one. Fix the text, and consider defaulting to false (the direction that never drops a report).

src/ros2_medkit_plugins/ros2_medkit_graph_watchdog/src/graph_watchdog_plugin.cpp
```diff
@@ -144,31 +183,42 @@ int GraphWatchdogPlugin::compute_departed_retention_ticks(const nlohmann::json &
   // configure() has not run yet at the point set_context() needs this value, so its
   // own kDefaultMissGrace/clamping cannot be reused directly here.
   int node_death_miss_grace = kDefaultNodeDeathMissGrace;
```
mfaferek93, line 185: The per-detector keys now read int64 before the range check - good - but the plugin-scope keys at :154-171 (`tick_interval_ms`, `warmup_cycles`, `prune_grace`) still `get<int>()` before validating, the exact narrow-then-validate wrap this PR fixes below. `prune_grace: 4294967296` becomes 0, `resolve_node_death_prune_grace()` faithfully injects 0, both sides agree, nothing warns. Consistency is preserved, the operator's value is not. Same int64 read as here would close it.

src/ros2_medkit_plugins/ros2_medkit_graph_watchdog/test/e2e/test_presence_ownership_e2e.test.py
```diff
@@ -0,0 +1,1244 @@
...
+    def test_settled_ignorance_is_reported_by_presence_and_by_lifecycle(self, target_node):
+        self.assertTrue(
+            wait_until_watchdog_armed(PORT, timeout=ARM_TIMEOUT_SEC, app_id=UNREADABLE_NODE),
+            f'graph_watchdog never reported {UNREADABLE_NODE} armed - the permissive gate answer '
+            'this row is about was never reached, so nothing below would prove anything')
+        self.assertTrue(
+            wait_until_faults_endpoint_live(PORT, timeout=FAULTS_LIVE_TIMEOUT_SEC),
+            'GET /faults never answered 200 - nothing below would prove anything')
+
+        entity = _poll_watchdog_entity(PORT, UNREADABLE_NODE, '', timeout=LABEL_TIMEOUT_SEC)
```
mfaferek93, line 663: This row (and the one at :734) polls only for an empty label and then kills, relying on the comment's tick-ordering argument that the GetState budget has been spent; the provisional row in the same file reads `measurement_pending=False` off the route first. If the gateway ever delivers the App one refresh before its service endpoints, the kill lands on kUnclaimed, the key is never admitted, and the red reads as a detector regression. Passing `measurement_pending=False` to `_poll_watchdog_entity` at both sites makes the precondition explicit.

src/ros2_medkit_plugins/ros2_medkit_graph_watchdog/test/e2e/test_node_death_suppression_e2e.test.py
```diff
@@ -0,0 +1,788 @@
...
+# No default on purpose - see harness-consuming siblings' identical rationale: a default makes
+# this file FAIL OPEN. A KeyError is loud.
+SCENARIO = os.environ['WATCHDOG_E2E_SCENARIO']
+PORT = get_test_port()
+
+FAULT_CODE = 'GRAPH_NODE_DISAPPEARED'
+_NODE_DEATH_PREFIX = 'plugins.graph_watchdog.detectors.node_death'
+_LIFECYCLE_PREFIX = 'plugins.graph_watchdog.detectors.lifecycle_expectation'
+
+# Same fast cadence and miss_grace convention test_node_death_e2e.test.py uses: comfortably past
```
mfaferek93, line 146: `MISS_GRACE = 20` is explained here as chosen to sit past the wall-clock floor, but it is injected only in `prune_no_false_heal` (:318); the other four scenarios run on the default 2, floored to 14 with a startup WARN, so the timing arithmetic in their docstrings is computed from a grace the gateway does not use. Moving it into the base `detector_params` like test_node_death_e2e.test.py:314 does keeps the docstrings honest.

src/ros2_medkit_plugins/ros2_medkit_graph_watchdog/src/detectors/lifecycle_expectation_detector.cpp
bburda: Fixed, and the first attempt was not enough. We first made lifecycle read node_death's tracked set instead of the gate. That closed your sequence and opened another: a provisional admission entered the set, the latch caught it, node_death later released the key, and the node was again reported by nobody. The latch is gone now. There is no `ever_armed` and no `armed` field; lifecycle reads current membership every tick. A latch cannot express the other ways a key stops being owned, so removing it was simpler than correcting it.

src/ros2_medkit_plugins/ros2_medkit_graph_watchdog/src/detectors/node_death_detector.cpp
bburda: Fixed. `earned_` was pruned during the outage, so the respawn read `unconfigured`, the key was handed back and PASSED healed the fault. It is now kept for as long as the tracker keeps the key. Both of your reproductions have tests.

src/ros2_medkit_plugins/ros2_medkit_graph_watchdog/src/detectors/node_death_detector.cpp
bburda: Fixed. A released key expires after `miss_grace + 1` record-less present ticks. A later check found the window collapsed across an offline flicker, because the cleanup pruned against online nodes only, so an offline tick erased both the release and the counter. The counter is held across the gap now, and only observed present ticks advance it.

src/ros2_medkit_plugins/ros2_medkit_graph_watchdog/design/graph_watchdog.rst
bburda: Fixed, all of it, including the test name that does not exist. A present entry is never evicted, an immature departed entry is never collapsed, and saturation is true when immature entries alone exceed the cap.

src/ros2_medkit_plugins/ros2_medkit_graph_watchdog/README.md
bburda: Fixed. It says the detector does not parse its own description, not that it cannot. The operational sentence is there too: after a reboot that follows a node death the operator has to delete the fault by hand even though everything is green.

src/ros2_medkit_plugins/ros2_medkit_graph_watchdog/src/detectors/lifecycle_expectation_detector.cpp
bburda: Fixed. A departed entry gets its own sentence saying the hold lasts until the node returns, pinned by a log test.

src/ros2_medkit_plugins/ros2_medkit_graph_watchdog/test/e2e/test_node_death_suppression_e2e.test.py
bburda: Fixed, and it found a real defect. Making the row reach the mechanism was the easy part. Once it did, it failed under TSan with both nodes named. The cause: the departed lifecycle record expires while the veto is still needed. Retention is 81 ticks counted from the services leaving, the reclaim is 81 misses counted from the App leaving, so the two windows meet with one tick of margin and any lead between those two events consumes it. Then the veto lifts, the record is gone, and a cleanly shut-down node stays named for the life of the process. A durable suppressor's yes is now remembered per departed key, so the record is needed once. Your comment is what made this visible.

src/ros2_medkit_plugins/ros2_medkit_graph_watchdog/include/ros2_medkit_graph_watchdog/lifecycle_expectation_tracker.hpp
bburda: Fixed, and your point about the default was sharper than it looked. Five tracker tests were relying on it. The field and the flag are both gone, so there is nothing left to default.

src/ros2_medkit_plugins/ros2_medkit_graph_watchdog/test/e2e/test_node_death_suppression_e2e.test.py
bburda: Fixed. It is in the base parameters, so all five scenarios use the value their docstrings describe.

src/ros2_medkit_plugins/ros2_medkit_graph_watchdog/test/e2e/test_presence_ownership_e2e.test.py
bburda: Fixed at both sites. They read `measurement_pending=False` from the route instead of arguing from tick order.

src/ros2_medkit_plugins/ros2_medkit_graph_watchdog/src/graph_watchdog_plugin.cpp
bburda: Fixed for all three keys. They read the wide integer before the range check. The test fails when `get<int>()` is put back.

src/ros2_medkit_plugins/ros2_medkit_graph_watchdog/test/e2e/test_node_death_e2e.test.py
bburda: Fixed for the two budgets, and the whole file is scaled now, twelve constants and about a hundred inline deadlines. The ones left unscaled are listed with reasons. The `_blink` arithmetic is one of them: it measures the detector's own window, so scaling it would destroy the claim the row makes.

src/ros2_medkit_plugins/ros2_medkit_graph_watchdog/test/e2e/test_node_death_e2e.test.py
bburda: Fixed in both copies. `streak_start` is seeded before the loop, so a slow first GET counts as stable at None the way the docstring says.

mfaferek93 (approved):

Issue #612, opened 2026-08-13
[BUG] A timeout is never reported as a timeout


bburda:
# Bug report

### Steps to reproduce
1. Run an aggregator with a peer. On the peer, define an entity with an operation that takes
   longer than `aggregation.timeout_ms` (default 2000 ms).
2. `POST /api/v1/components/<id>/operations/<op>/executions` on the aggregator, then send the
   same request directly to the owning peer.
3. On a single gateway, execute an operation whose own work takes longer than
   `service_call_timeout_sec` (default 10 s).
4. Request a fanned-out collection such as `GET /api/v1/faults` while one peer is slow enough to
   exceed the fan-out budget.

### Expected behavior

A request that ran out of time says so. It returns a timeout status, or at least a message that
names the budget that was exceeded. A timeout can be told apart from a peer that is really down,
and a partial answer can be told apart from a complete one.

### Actual behavior

Three paths, three different ways of not reporting a timeout.

**1. Peer forward: reported as an unavailable peer.**

```
aggregator  -> 502 in 2.002822 s
  {"error_code": "vendor-error",
   "message": "Peer '<name>' at http://<host>:<port> is unavailable",
   "vendor_code": "x-medkit-peer-unavailable"}
owning peer -> 200 in 4.936800 s   (the full operation result)
```

The peer was alive at that moment. Probed from inside the aggregator container at the same time:
`/api/v1/health` returned 200 and `/api/v1/components/<id>` returned 200. The message reports a
peer as down while that peer answers the same request in 4.9 s.

**2. Fan-out: reported as success.** A peer that times out during a fanned-out collection gives
`HTTP 200` with `x-medkit.partial = true` and `x-medkit.failed_peers`. No status change, no
error code, no message.

**3. Local operation: reported as a generic failure.** An operation that exceeds
`service_call_timeout_sec` returns `500` with `"message": "Service call failed"`.

### Environment
- ros2_medkit version: main (a2683796), gateway vendor_info.version 0.6.0, SOVD API 1.0.0
- ROS 2 distro: Jazzy
- OS: Ubuntu 24.04

### Additional information

**Why the peer forward cannot tell the two cases apart.** `httplib::Result::operator bool()` is
a null check (`src/vendored/cpp_httplib/httplib.h:1053`). The forward path branches on `!result`
at `src/core/aggregation/peer_client.cpp:637` and never reads `result.error()`. So
[...]

Issue #611, opened 2026-08-13
[BUG] GET /faults/stream on an aggregating gateway returns 200 and streams nothing


bburda:
# Bug report

### Steps to reproduce
1. Run an aggregator gateway with one or more peers. Each peer runs on its own ROS domain with
   its own `fault_manager`. Give the aggregator its own domain and no local fault producers.
2. Open `GET /api/v1/faults/stream` on the aggregator and on a peer that has live faults, in the
   same time window.
3. Trigger faults on the peer.

### Expected behavior

The aggregator relays the fault events of its peers, the same way `GET /api/v1/faults` already
fans out to peers. If it cannot do that, it should not advertise a route it cannot serve, and
should return an explicit error instead.

### Actual behavior

The aggregator returns `200 OK` with `Content-Type: text/event-stream` and then sends only
`:keepalive` comments. Measured against the owning peer in the same 60 s window:

```
aggregator  ->  0 events, 1 ":keepalive"
owning peer -> 10 events
```

Repeated in a separate 20 s window: aggregator 0 events, owning peer 84 events.

A 200 with an open but silent stream looks exactly like "there are no faults". A client that
trusts it will report the system as healthy. In a deployment where the aggregator is the only
port an operator can reach, this is the only fault stream available.

### Environment
- ros2_medkit version: main (a2683796), gateway vendor_info.version 0.6.0, SOVD API 1.0.0
- ROS 2 distro: Jazzy
- OS: Ubuntu 24.04

### Additional information

`SSEFaultHandler` only subscribes to its own ROS domain. The constructor at
`src/ros2_medkit_gateway/src/http/handlers/sse_fault_handler.cpp:68` creates one subscription to
`<namespace>/fault_manager/events`. The file contains no reference to `aggregat`, `peer` or
`fan_out`. An aggregating gateway runs on its own domain, with its own `fault_manager` that no
producer reports to.

`GET /faults` does fan out (`src/http/handlers/fault_handlers.cpp:449`). The stream does not.

`SSEStreamProxy` already exists and looks complete. It is declared at
`include/ros2_medkit_gateway/core/aggregation/stream_proxy.hpp:81`, implemented in
`src/core/aggregation/sse_stream_proxy.cpp`, and compiled into `gateway_core`. It is never
constructed outside `test/test_stream_proxy.cpp`. The symbol has no other reference in
production code, and the file history is two commits, both moves and Rule-of-Five boilerplate.

[...]

Issue #528, opened 2026-07-13
[BUG] Aggregator times out fanning out large resource payloads to peers


bburda:
# Bug report

### Steps to reproduce
1. Run an aggregator gateway that fans out to one or more peer gateways.
2. On a peer gateway, add an App bound to a node that publishes a large topic (for example an occupancy grid `/map`, or a costmap `costmap_raw`).
3. Send `GET /api/v1/apps/<id>/data/<resource>` for that large resource through the aggregator. Then send the same GET straight to the peer gateway that owns it.

### Expected behavior
The aggregator returns the same data as the owning gateway. It can be slower, but it should not time out.

### Actual behavior
The request through the aggregator times out for large resources. The same GET on the owning gateway returns in under one second. Small resources go through the aggregator fine. Only large payloads fail. The data is correct and reachable, only the aggregator fan-out fails.

### Environment
- ros2_medkit version: 0.6.0
- ROS 2 distro: Jazzy
- OS: Ubuntu 24.04

### Additional information
This looks like a fixed HTTP timeout or a size limit on the aggregator fan-out, reached only when the peer response is large. A bigger or configurable fan-out timeout, or streaming the peer response instead of holding it all in memory, would probably fix it.

bburda:
Confirmed on main at `a2683796`. The same 2000 ms budget also affects a second case that is not
about large payloads: an operation executed through the aggregator.

`aggregation.timeout_ms` (`src/gateway_node.cpp:240`) is used as both the connect and the read
timeout in `PeerClient::ScopedClient` (`src/core/aggregation/peer_client.cpp:813-814`), and
`PeerClient` holds a single `int timeout_ms_` that both construction sites fill from
`config_.timeout_ms` (`src/aggregation/aggregation_manager.cpp:219,275`). There is no separate
budget for `POST .../executions`, so a synchronous ROS 2 service call on a peer gets 2 s end to
end, while the peer's own budget for the same call is `service_call_timeout_sec`, default 10 s.
We measured an operation that returned in 4.9 s on the owning peer and 502 in 2.002822 s through
the aggregator.

Two more things worth handling in the same fix:

- The write timeout is never set anywhere in the gateway, so it stays at the cpp-httplib default
  of 5 s (`src/vendored/cpp_httplib/httplib.h:41-46`) and does not follow
  `aggregation.timeout_ms`. Raising `timeout_ms` alone will not raise it.
[...]

PR #626, opened 2026-08-20
fix(aggregation): one addressing model for an aggregating entity's resources

### LINKED ISSUE #613: [BUG] The aggregator refuses or 404s entity work its peers can serve
# Bug report

### Steps to reproduce
1. Run an aggregator with a peer. Configure a scripts backend on the peer only.
2. `GET /api/v1/apps/<id>/scripts` on the aggregator, then the same request on the owning peer.
3. For a merged entity, meaning an Area or Function with contributors on more than one gateway:
   `GET /api/v1/areas/<id>/operations` on the aggregator, then
   `POST /api/v1/areas/<id>/operations/<op-owned-by-a-peer>/executions`.

### Expected behavior

The aggregator serves what its peers can serve, or it does not advertise it. An operation that
appears in a list can be executed through the gateway that listed it.

### Actual behavior

**Scripts.** Same path, two gateways, seconds apart:

```
aggregator  -> 501  {"error_code": "not-implemented", "message": "Scripts backend not configured"}
owning peer -> 200  {"_links": {...}, "items": [{...}]}
```

The message describes the aggregator's own backend, but the entity it names belongs to a peer
that has one. So the answer is wrong about whose configuration is missing.

**Operations on a merged entity.** `GET .../operations` returns 200 and includes operations
contributed by peers. `POST .../operations/<that-op>/executions` returns
`404 operation-not-found`. The aggregator lists operations it will not execute.

### Environment
- ROS 2 distro: Jazzy
- OS: Ubuntu 24.04

### Additional information

**Scripts: the backend check runs before the peer-forward check.**

```cpp
// src/ros2_medkit_gateway/src/http/handlers/script_handlers.cpp:128-130
if (!script_mgr_ || !script_mgr_->has_backend()) {
  return tl::unexpected(make_error(501, ERR_NOT_IMPLEMENTED, "Scripts backend not configured"));
}
...
auto entity_result = ctx_.validate_entity_for_route(req, entity_id);   // :138, never reached
```

The same order appears in all eight script handlers. The checks are at lines 128, 179, 264, 312,
358, 442, 496 and 549, and the matching `validate_entity_for_route` calls are at 138, 189, 283,
331, 377, 469, 523 and 576.

`ScriptManager` is always constructed (`src/gateway_node.cpp:947`), so the routes are always
registered, but `backend_` stays null unless this gateway has its own scripts directory or a
plugin ScriptProvider.

This makes `/scripts` the only entity resource collection whose peer path cannot be reached at
all. Every other collection at least reaches `validate_entity_for_route` and is proxied for a
routed remote leaf (`src/http/handlers/handler_context.cpp:315-320`).

**Operations: the list fans out, the execute does not.** `list_operations` merges peer items
unconditionally (`src/http/handlers/operation_handlers.cpp:407`). `create_execution` calls
`validate_entity_for_route` (`:545`), which only proxies for a routed remote leaf, then resolves
the operation against the local cache and returns `404 operation-not-found` (`:628-631`).

Merged Areas, merged Functions and hierarchical parent Components are deliberately kept out of
the routing table (`design/aggregation.rst:276-282`), so `is_remote` is false for them and
neither mechanism applies.

The execution lifecycle cannot be routed even for genuinely remote entities. `list_executions`
(`:727`), `get_execution` (`:796`), `cancel_execution` (`:852`) and `update_execution` (`:915`)
call `validate_entity_id` instead of `validate_entity_for_route`, so they never proxy. An
execution id created on a peer returns 404 on the aggregator.

The same one-way pattern applies to `{entity}/data`, `{entity}/configurations` and
`{entity}/logs`. The list is merged from peers, but the single-item read and write
(`data_handlers.cpp:349,465`, `config_handlers.cpp:451,565`, `log_handlers.cpp:219,252`) only
reach a peer through `validate_entity_for_route`, never for a merged entity.

**Capabilities are not filtered by peer aggregation.** `EntityCapabilities::for_type`
(`src/core/models/entity_capabilities.cpp:19-109`) is a static per-type table. It does have an
`aggregated_collections_` set with an `is_aggregated()` accessor
(`entity_capabilities.cpp:127-128`), but that describes aggregation from sub-entities inside one
gateway, for example a Function collecting data from its hosted Apps, as its own doc comment says
(`include/ros2_medkit_gateway/core/models/entity_capabilities.hpp:88-94`). It is not connected to
`AggregationManager` or to peer fan-out, and `is_aggregated()` has no caller outside the class.

Proxied peer detail responses are copied as they are
(`src/core/aggregation/peer_client.cpp:660-670`), with only five headers allowed through and the
body never rewritten. Capability hrefs are host-relative
(`src/core/http/handlers/capability_builder.cpp:80-83`). So a client reads the peer's `scripts`
href out of a proxied response, follows it back to the aggregator, and gets the 501. This
overlaps #515, which covers the per-entity side of the same problem.

bburda:
## Summary

An aggregating gateway lists resources that belong to members on other gateways, but it served every request for them locally. A data read looked for the peer's topic on its own ROS graph. An execution called the peer's service there. A configuration resolved against nodes only this gateway can see. All three failed while the member and its gateway were healthy.

This branch gives every entity and every resource collection one addressing model:

- An item id carries its member only when more than one member provides it. Ambiguity is decided from the declared tree, not from one response, so a listing and an execution cannot disagree about the same id.
- A member-qualified request is served by the member's own route on the gateway that owns it. Data, operations, executions and configurations share one dispatch point. A member this gateway owns is still served here.
- Availability is a separate question. It is carried by the entities a request can be addressed to, which are Apps and Components. An Area or a Function groups members and has no availability of its own.
- An entity a peer declared in its manifest stays in the tree when that peer goes quiet, keeps the items it last reported, and answers `504 not-responding`. An entity the peer only discovered at runtime disappears.
- Where one provider exposes the same short name at two ROS paths, those operations are addressed by their path. Every id that resolves today is unchanged.

Defects found while building this and fixed here:

- A partial read of a peer was published as a complete picture. A Function whose detail request failed was merged with no members, and that memberless copy replaced the retained declaration, so it outlived the peer.
- A `504 not-responding` from a nested collection was read as a failed request, so one unreachable member behind a peer froze that peer's whole view until it was restarted.
- `X-Client-Id` was not forwarded, so a lock on a peer-owned entity could not even be taken through an aggregating gateway.
- The `Location` header of a forwarded execution was removed by the response header allowlist, so clients were not pointed at the gateway that owns the goal.
- The orphan sweep deleted a just-restored persistent trigger before discovery had reported its entity. That also deletes the row from the store, so a persistent trigger did not survive a gateway restart under load and nothing said so.
- A cyclic subscription accepted a resource path for a collection that is streamed whole, then discarded it and streamed everything.

## Issue

- closes #613

Not covered by this PR and still open: #528 and #612 (timeout handling), #611 (fault stream on an aggregating gateway).

## Type

- [x] Bug fix
- [x] New feature or tests
- [ ] Breaking change
- [ ] Documentation only

## Testing

New end-to-end coverage drives two and three gateway topologies over HTTP, kills a peer mid-run, and asserts on payloads and on which gateway answered, not on status codes alone.

Measured on Jazzy:

- gateway unit suite: 124 tests, 0 failures (`ctest -j2 -LE "linter|integration"`)
- gateway linters: 9 tests, 0 failures
- integration package: 102 tests, 0 failures
- full workspace build: 0 compiler warnings

Aggregation feature specs grew from 0 to 49 cases in `test_grouping_entity_aggregation`, plus new specs for aggregator-only configurations, lock identity across a hop, and persistent trigger restore before discovery.

Two integration tests in `ros2_medkit_graph_provider` failed once each under parallel load during verification (`test_graph_provider_greenwave`, `test_graph_provider_stale`). Both are timing assertions, both pass repeatedly in isolation, and this branch changes no production code in that package.

## Checklist

### DIFF
--- a/docs/api/rest.rst
+++ b/docs/api/rest.rst
@@ -678,8 +678,33 @@ What this means for a request:
   does not provide - which is what tells an absent item apart from an item that
   exists and currently carries no data. A member half followed by nothing names
   no item and is ``404`` as well.
-- Reads are permissive: ``GET`` of a bare id returns the first match rather
-  than refusing, which is the behaviour every existing client depends on.
+- ``GET /{entity}/operations/{id}`` and
+  ``GET /{entity}/operations/{id}/executions`` refuse exactly what the execution
+  refuses, with the same body. Reading an operation under an id that names
+  several of them would describe one without saying which, and the same id is a
+  ``400`` the moment the caller runs it. The collection never offers such an id,
+  so only a stale one arrives, and it leaves with ``parameters.operation_ids``.
+  An unambiguous bare id reads and lists exactly as before.
+
+Listing the Executions of an Operation
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+``GET /{entity}/operations/{id}/executions`` resolves ``{id}`` by the rule
+above, so a member-qualified id and a ROS-path id both work, and the answer has
+three distinct forms:
+
+- an **action** returns the goals it holds, newest first;
+- a **service** returns ``200`` with an empty ``items`` array. A service call
+  completes inside its own ``POST`` and leaves no execution resource, so its
+  collection exists and is permanently empty. Whether an operation can ever have
+  executions is read from ``asynchronous_execution`` on the operation itself;
+- an id that names **no operation** is ``404 operation-not-found``, and an
+  unknown member half is ``404 resource-not-found`` naming that half - so a typo
+  is never answered as an operation that simply has not been run.
+
+Goals live on the gateway that sent them, so an id naming a peer-owned member is
+dispatched to that member's own route exactly as the ``POST`` was. A goal
+started through an aggregating entity is therefore listed through it too.
 
 Where a Member-Qualified Request is Served
 ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
@@ -933,7 +958,9 @@ Execute Operations
       }
 
 ``GET /api/v1/components/{id}/operations/{operation_id}/executions``
-   List all executions for an operation.
+   List all executions for an operation. Actions return their goals; a service
+   returns an empty ``items`` array, because a service call leaves no execution
+   resource behind. An id naming no operation is ``404``.
 
 ``GET /api/v1/components/{id}/operations/{operation_id}/executions/{execution_id}``
    Get execution status and result.
--- a/docs/api/rest.rst
+++ b/docs/api/rest.rst
@@ -746,6 +746,28 @@ before. This applies to ``GET`` and ``PUT`` of a single ``/data`` item, to
 ``POST`` of an ``/operations`` execution, and to ``GET``, ``PUT`` and
 ``DELETE`` of a single ``/configurations`` item.
 
+**An id needs no member half to reach its owner.** A ``/data`` item one member
+provides keeps its bare id - qualification follows ambiguity, not aggregation -
+so the bare topic path is the id the collection hands back, and it is dispatched
+by the member the tree records as providing that topic:
+
+.. code-block:: text
+
+   GET /api/v1/functions/vehicle_health/data/chassis%2Fbrakes%2Fpressure
+
+is answered, when every member providing ``/chassis/brakes/pressure`` belongs to
+one peer, by
+
+.. code-block:: text
+
+   GET /api/v1/apps/pressure_sensor/data/chassis/brakes/pressure
+
+on that peer. A topic a member this gateway runs provides is sampled here,
+unchanged; a topic whose providers are spread across gateways names no single
+place, and the local graph answers it as before. ``/operations`` resolves a bare
+id the same way - the operation is resolved first, and the member owning its ROS
+path is where the execution is sent.
+
 ``/configurations`` keeps its own id scheme, ``<app_id>:<param_name>``, and the
 member half is the app id. Because nothing on the owning gateway is aggregating,
 the parameter is addressed there by its bare name:
--- a/docs/api/rest.rst
+++ b/docs/api/rest.rst
@@ -1924,11 +1924,17 @@ Subscriptions are temporary - they do not survive server restart.
 
 **Supported collections:**
 
-- ``data`` - Topic data (requires a resource path, e.g. ``/data/temperature``)
-- ``faults`` - Fault list (resource path optional, e.g. ``/faults`` or ``/faults/fault_001``)
-- ``configurations`` - Parameter values (resource path optional)
-- ``logs`` - Application log entries from ``/rosout``
-- ``x-*`` - Vendor extensions (e.g. ``x-medkit-graph``)
+- ``data`` - Topic data. Requires a resource path naming the topic, e.g. ``/data/temperature``
+- ``faults`` - Fault list. Streamed as a whole; no resource path
+- ``configurations`` - Parameter values. Streamed as a whole; no resource path
+- ``logs`` - Application log entries from ``/rosout``. Streamed as a whole; no resource path
+- ``x-*`` - Vendor extensions (e.g. ``x-medkit-graph``). Streamed as a whole unless the
+  plugin registering the sampler declares that it narrows its payload to a named resource
+
+A collection that is streamed as a whole delivers every item of that collection on
+every tick. A resource URI naming a single item of such a collection is refused with
+400 ``x-medkit-invalid-resource-uri`` rather than accepted and answered with the whole
+collection.
 
 **Interval values:**
 
@@ -1956,7 +1962,8 @@ Subscriptions are temporary - they do not survive server restart.
 
    - ``resource`` (string, required): Full SOVD resource URI to observe
      (e.g. ``/api/v1/apps/{id}/data/{topic}``, ``/api/v1/apps/{id}/faults``,
-     ``/api/v1/functions/{id}/x-medkit-graph``)
+     ``/api/v1/functions/{id}/x-medkit-graph``). The URI ends at the collection
+     unless that collection's sampler narrows its payload to a named resource
    - ``protocol`` (string, optional): Transport protocol. Only ``"sse"`` supported. Default: ``"sse"``
    - ``interval`` (string, required): One of ``fast``, ``normal``, ``slow``
    - ``duration`` (integer, required): Subscription lifetime in seconds.
@@ -1965,7 +1972,9 @@ Subscriptions are temporary - they do not survive server restart.
    **Error responses:**
 
    - **400** ``invalid-parameter`` - Invalid interval, duration <= 0, or duration exceeds max
-   - **400** ``x-medkit-invalid-resource-uri`` - Malformed resource URI or path traversal
+   - **400** ``x-medkit-invalid-resource-uri`` - Malformed resource URI, path traversal,
+     ``data`` without a topic path, or a resource path on a collection that is streamed
+     as a whole
    - **400** ``x-medkit-entity-mismatch`` - Resource URI references different entity than route
    - **400** ``x-medkit-collection-not-supported`` - Entity doesn't support the collection
    - **400** ``x-medkit-collection-not-available`` - No data provider registered for collection
--- a/docs/api/rest.rst
+++ b/docs/api/rest.rst
@@ -645,6 +645,20 @@ consider it unique. In practice:
   every contributor is named in ``member_ids``. A path is qualified only when
   two gateways each contribute an item under it.
 
+**When one member carries the short name twice.** An operation's wire id is the
+last segment of its ROS path, so ``left/calibrate`` and ``right/calibrate`` on
+one node are two operations called ``calibrate``. The member half names that
+same member for both copies and separates nothing, so those items take the ROS
+path, leading slash stripped, as their item half::
+
+   robot/left/calibrate                       # on the App itself
+   primary_calibration:robot/left/calibrate   # on an entity that aggregates it
+
+The form is decided per provider: a short name its own provider carries once
+keeps that short name, whatever another provider does with the same name. The
+split at the first colon is unchanged, because a ROS path carries no colon, and
+the path form is the one ``/data`` already uses for a topic.
+
 What this means for a request:
 
 - A bare id that names one item works, on every route. Every client that sends
@@ -654,7 +668,11 @@ What this means for a request:
   one member provides is refused with ``400 invalid-request``, naming the
   qualified form and listing the members in ``parameters.member_ids``. Running
   whichever member was walked first without saying which one ran is the defect
-  this removes.
+  this removes. A short name that ONE member carries at two ROS paths is
+  refused the same way, listing those paths in ``parameters.ros2_paths``.
+  Either refusal carries ``parameters.operation_ids``: the ids that do address
+  what collided, as the collection lists them, so the client sends one back
+  rather than deriving it.
 - A qualified id is accepted on the single-item routes. A member half that
   names no member of the entity is ``404``, and so is an item half that member
   does not provide - which is what tells an absent item apart from an item that
@@ -816,7 +834,9 @@ Operations Endpoints
 
 Execute ROS 2 services and actions. Operation ids follow
 :ref:`member-qualified-ids`: a short name that only one member exposes is used
-bare, and one that several expose is addressed ``<member_id>:<operation>``.
+bare, one that several expose is addressed ``<member_id>:<operation>``, and one
+that a single member exposes at two ROS paths is addressed by the path itself,
+without its leading slash.
 
 List Operations
 ~~~~~~~~~~~~~~~
--- a/docs/api/rest.rst
+++ b/docs/api/rest.rst
@@ -2513,6 +2513,16 @@ configuration:
 
 Non-persistent triggers are always cleared on restart.
 
+Restore happens once, while the gateway starts. The number of triggers it put
+back is logged, so a restart that restored fewer than expected is visible in
+the gateway log rather than only in a later 404.
+
+The gateway also removes triggers whose entity has left discovery. A restored
+trigger is exempt from that until its entity has been discovered at least once:
+immediately after a restart nothing has been discovered yet, and an entity that
+has merely not been reported yet has not disappeared. A restored trigger whose
+entity never appears stays listed and can be deleted through the API.
+
 Fault Triggers (threshold rules)
 --------------------------------
 
--- a/docs/api/rest.rst
+++ b/docs/api/rest.rst
@@ -804,6 +804,14 @@ retained entity:
 ``status: "offline"``. Availability of an entity and health of a peer are
 separate questions and are reported separately.
 
+``x-medkit.available`` is emitted **only when false**, so an absent field means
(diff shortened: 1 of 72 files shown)


docs/api/rest.rst
```diff
@@ -608,10 +608,225 @@ Functions
...
+   primary_calibration:robot/left/calibrate   # on an entity that aggregates it
+
+The form is decided per provider: a short name its own provider carries once
+keeps that short name, whatever another provider does with the same name. The
+split at the first colon is unchanged, because a ROS path carries no colon, and
+the path form is the one ``/data`` already uses for a topic.
+
+What this means for a request:
+
+- A bare id that names one item works, on every route. Every client that sends
```
mfaferek93, line 680: This claim doesn't hold for /data on a peer-aggregated entity: the collection fan-out lists a peer-only topic bare, but a bare id is always served locally (address_data_item leaves member_id empty, dispatch_to_member returns kServeLocally), and since peer_client never pulls topics into App, there is no owner to dispatch to - the local sample then 404s with x-medkit-ros2-topic-unavailable. Operations solve exactly this via ops.owner_by_path lookup before dispatch; data has no equivalent, so either scope this bullet to routes where it's true or give /data the same owner-based dispatch. Same sentence repeated in src/ros2_medkit_gateway/README.md:301.

src/ros2_medkit_gateway/src/core/managers/trigger_manager.cpp
```diff
@@ -657,79 +682,124 @@ void TriggerManager::set_on_removed(OnRemovedCallback callback) {
...
+                                                                     ChangeType::UPDATED);
+                                                  });
+        if (handle) {
+          topic_handles_[trigger_id] = std::move(handle);
+        } else {
+          // Subscribe failed during persistent-trigger restore (e.g. the
+          // topic disappeared between shutdown and restart, or rclcpp threw
+          // inside TriggerTopicSubscriber). Queue the trigger for retry on
+          // the next refresh tick instead of leaving it active-but-silent.
+          unresolved_data_triggers_.push_back({trigger_id, entity_id, resource_path, std::chrono::steady_clock::now()});
```
mfaferek93: The entity_seen shield keeps a restored data trigger alive for as long as discovery takes, but both subscription-recovery paths still give up after ~60s: the subscriber's pending queue (kPendingTimeoutSec=60 in trigger_topic_subscriber.hpp, and that is the branch actually taken here - subscribe() defers instead of failing when the topic is absent, so this unresolved_data_triggers_ path only fires when rclcpp throws) and unresolved_timeout_ on this queue. If the entity shows up later than that - the exact scenario this PR fixes - the trigger survives the sweep but no subscription is ever re-created, so it reports active and never fires. Consider re-arming resolution when entity_seen flips true, or holding both retry budgets open while entity_seen is false.

src/ros2_medkit_gateway/src/http/handlers/operation_handlers.cpp
```diff
@@ -404,10 +700,46 @@ http::Result<dto::Collection<dto::OperationItem>> OperationHandlers::list_operat
...
+  std::unordered_set<std::string> paths_from_peers;
   for (auto & item : fan_out.items) {
+    if (item.x_medkit.has_value() && item.x_medkit->ros2.has_value()) {
+      const auto & ros2 = *item.x_medkit->ros2;
+      auto path = ros2.service.value_or(ros2.action.value_or(std::string{}));
+      if (!path.empty()) {
+        paths_from_peers.insert(std::move(path));
+      }
+    }
+    qualify_from_declared_tree(item);
```
mfaferek93, line 736: When both gateways declare an app with the same id, the merge renames the peer's copy to `<peer>__<id>`, but fan-out items still carry `member_ids` in the peer's own naming - so `qualify_from_declared_tree` here stamps the peer's operation with the local app's name, and the collection ends up with two items both called `shared_sensor:calibrate` while `resolve_operation` (owner_by_path match) always picks the local copy. The correctly-named retained item (`<peer>__shared_sensor:calibrate`) is exactly the one dropped at line 727 when the peer answers, so the healthy path shows the wrong id and the peer's copy becomes unaddressable. The peer's `member_ids` need translating through the collision rename (routing table) before qualifying - same pattern applies to the `qualify_ambiguous_ids` call in data_handlers.cpp:445.

src/ros2_medkit_gateway/src/http/handlers/operation_handlers.cpp
```diff
@@ -404,10 +700,46 @@ http::Result<dto::Collection<dto::OperationItem>> OperationHandlers::list_operat
...
   }
+
+  if (!fan_out_suppressed) {
+    for (auto & item : retained_from_peers) {
+      const auto & ros2 = *item.x_medkit->ros2;
+      auto path = ros2.service.value_or(ros2.action.value_or(std::string{}));
+      if (!path.empty() && paths_from_peers.count(path) > 0u) {
+        continue;  // the owner answered for itself, which is the better copy
+      }
+      item.x_medkit->available = false;
```
mfaferek93: In the mixed shape (function declared only on this gateway with a peer-owned member, exactly what MIXED_FUNCTION in test_aggregator_only_configurations builds), get_peer_contributors returns empty for the function, fan_out_collection early-returns, and this loop then stamps every retained peer-owned operation available:false even though the owning peer is healthy - the dto contract says false means the provider is not answering, and dispatch_to_member happily executes the same id. The fallback needs to distinguish "no peer contributes this entity, so fan-out never ran" from "fan-out ran and the owner did not answer", e.g. by checking the member's reachability instead of inferring it from paths_from_peers; a /operations assertion on MIXED_FUNCTION would pin this down, since the existing healthy-path tests only cover the both-sides-declared function.

src/ros2_medkit_integration_tests/CMakeLists.txt
```diff
@@ -232,10 +237,22 @@ if(BUILD_TESTING)
     test_peer_aggregation
     test_cross_ecu_fanout
     test_daisy_chain_aggregation
+    test_grouping_entity_aggregation
```
mfaferek93, line 245: test_grouping_entity_aggregation is missing from _MEDKIT_TEST_TIMEOUT_OVERRIDES, so its 49 cases run under the feature glob's 120 s ctest kill - yet its own budgets exceed that: setUpClass alone may poll 3x60 s, z1 polls up to 60 s after killing the peer, and five cases park 15 s action goals (order=30 at 2 Hz) that later cases drain through 30-60 s waits. A slow runner gets the whole file SIGKILLed with no test name in the output, which is exactly why test_graph_provider_* (smaller budgets) got 300 here - give this suite the same.

src/ros2_medkit_integration_tests/test/features/test_grouping_entity_aggregation.test.py
```diff
@@ -0,0 +1,2405 @@
...
+            f'two suppressed member requests took {elapsed:.1f}s, which is a bounce, not a hop',
+        )
+
+    # ------------------------------------------------------------------ R10
+    # A PEER THAT STOPS ANSWERING.
+    #
+    # Everything above runs against a live pair. These run after it, because
+    # unittest orders methods alphabetically within a class and these are the
+    # only ones prefixed `test_z`. The peer is killed exactly once, by the
+    # first of them, and the rest read the aggregator afterwards.
```
mfaferek93, line 2457: The z-block proves a silent peer's declared entities are retained and flipped unavailable, but nothing anywhere (this suite, daisy chain, or the unit tests) ever brings a peer back: `mark_unreachable` is one-way, so recovery depends entirely on a later successful fetch superseding the retained declaration, and that path is unobserved - a regression that kept replaying the retained copy would permanently poison the tree after one transient outage. Since aggregation.rst explicitly promises "when a peer recovers ... it is automatically re-included", a kill-then-restart case (the gate-started-process pattern from test_triggers_restore_before_discovery would work) asserting available flips back to true and runtime entities re-merge would pin the other half of this contract.

src/ros2_medkit_integration_tests/test/features/test_grouping_entity_aggregation.test.py
```diff
@@ -0,0 +1,2405 @@
...
+        to be able to name both, because every addressing rule below uses the
+        leaf id as its qualifier. If the merge collapses them, a compound id
+        built from that qualifier names two entities at once and the ambiguity
+        the compound form exists to remove has simply moved up one level.
+        """
+        response = requests.get(f'{PRIMARY_URL}/apps', timeout=10)
+        self.assertEqual(response.status_code, 200)
+        ids = [item.get('id') for item in response.json().get('items', [])]
+
+        matching = [i for i in ids if i == COLLIDING_LEAF or i.endswith(f'__{COLLIDING_LEAF}')]
```
mfaferek93, line 698: COLLIDING_LEAF never appears in the member half of any compound id: this R1 test checks /apps listing and detail 200, but nothing asserts the merged Function's member_ids carry secondary_gateway__shared_sensor, and no read ever sends shared_sensor:<item> or the renamed form, so the __-strip in forward_request is only unit-tested, never driven E2E. That hides a real inconsistency: merge_apps renames the colliding App, but merge_functions dedups the peer's hosts list verbatim (entity_merger.cpp:113-119), so the renamed copy is never a member of the merged Function - contradicting the new comment in the configurations-reset handler that claims the merge makes this impossible. A member-qualified read through MERGED_FUNCTION on both shared_sensor copies right after this test would pin it.

src/ros2_medkit_gateway/src/openapi/capability_generator.cpp
```diff
@@ -258,12 +275,35 @@ nlohmann::json CapabilityGenerator::generate_resource_collection(const ResolvedP
...
+    }
+    const auto owner_of = [&ops](const std::string & full_path) -> std::string {
+      auto owner = ops.owner_by_path.find(full_path);
+      return owner != ops.owner_by_path.end() ? owner->second : std::string{};
+    };
+    const auto addressed_by_path = http::operation_paths_addressed_by_path(ops);
+    const auto operation_id_of = [&](const std::string & name, const std::string & full_path) {
+      return qualified_item_id(http::operation_item_half(name, full_path, addressed_by_path), short_name_counts[name],
+                               owner_of(full_path));
+    };
```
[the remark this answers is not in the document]
bburda: I do not think this one applies. `qualified_item_id` returns the item id unchanged when the owner is empty, and `owner_by_path` is only populated for an aggregated entity, so a leaf entity never gets a member half. The path form is already used, through `operation_item_half`.

src/ros2_medkit_gateway/src/core/openapi/route_registry.cpp
```diff
@@ -493,8 +497,12 @@ nlohmann::json RouteRegistry::to_openapi_paths() const {
...
               {"app_id", "The app identifier"},
               {"function_id", "The function identifier"},
-              {"data_id", "The data item identifier (ROS 2 topic name)"},
-              {"operation_id", "The operation identifier"},
+              {"data_id",
+               "The data item identifier (ROS 2 topic name), or 'member_id:topic' when more than one member of "
+               "the entity contributes an item under that name"},
+              {"operation_id",
+               "The operation identifier (service or action short name), or 'member_id:operation' when more than "
+               "one member of the entity exposes that name"},
```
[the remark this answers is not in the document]
bburda: Correct, fixed. The description now names the ROS path form as well.

docs/api/rest.rst
bburda: You were right, the sentence was too strong. A peer never described its topics, so a merged peer App had none and a bare id had no owner to dispatch to. Peers now report their apps' topics, so the owner map covers them and a bare id is served on the gateway that owns it. The sentence is kept and made true instead of scoped down.

The listing holds a peer copy back and emits it only when the fan-out did not carry that path, otherwise the same topic was listed twice.

src/ros2_medkit_gateway/src/core/managers/trigger_manager.cpp
bburda: Confirmed, and it was worse than one branch being taken. Restore runs inside the constructor and the subscription executor is wired later, so at restore time the deferred branch is always taken and the clock always starts before the graph can answer.

The budget now starts when discovery first reports the entity. While the entity has never been seen the resolution keeps retrying for as long as the record is kept, and the wait is logged. `kPendingTimeoutSec` and `unresolved_timeout_` are unchanged at 60.

src/ros2_medkit_gateway/src/http/handlers/operation_handlers.cpp
bburda: Fixed. One correction to the description: the renamed id was already executable. The problem was that the collection never offered it, so a client following the listing ran the local copy twice.

A fan-out item's member is now resolved to the id the merge gave its owner, both in the item and in the id itself. `data_handlers.cpp` had the same hole and got the same fix.

The collision could not be driven through operations at all before this, because both sides bound that app to nodes with no services. There is a demo node now that serves one and names its own node in the answer, which is the only thing on the wire that separates two copies that both return 200.

src/ros2_medkit_gateway/src/http/handlers/operation_handlers.cpp
bburda: Confirmed and fixed. `available` is now read from the member instead of from whether the fan-out ran.

I picked member reachability over "did the fan-out run" because the fan-out also does not run when the peer is dead, so it cannot separate the two cases, and the DTO already defines the field in terms of the member. `MIXED_FUNCTION` has an assertion for this now.

src/ros2_medkit_integration_tests/CMakeLists.txt
bburda: Agreed, added at 300. On our runner the file takes about 38 s, but the internal budgets can reach past 120 s and then the whole file is killed with no test name, which is the part that matters.

src/ros2_medkit_integration_tests/test/features/test_grouping_entity_aggregation.test.py
bburda: Added as its own file, since this suite leaves its peer dead. The peer is killed and then replaced through a gate file, so the return happens when the case asks for it.

It checks the flag going false and back, the runtime member disappearing and merging again, a read that answered 504 answering with the peer's payload, and no retained copy left beside the live one. No gateway defect: the tree recovers. A mutation that latches the unhealthy state fails four of the eight cases.

One thing worth knowing: the two signals do not clear together. `available` clears on the first refresh after the health check passes, `is_online` is the peer's own account of its graph and follows one or two refreshes later.

src/ros2_medkit_integration_tests/test/features/test_grouping_entity_aggregation.test.py
bburda: First half is right and is covered now, the renamed member is driven end to end.

The second half does not hold. The hosts are remapped before `merge_functions` runs, in `AggregationManager::fetch_and_merge_peer_entities`, using the routing table that `merge_apps` builds. The comment above it says the merge order is deliberate for exactly this reason. So the renamed copy is a member of the merged Function.

mfaferek93 (approved):

PR #629, opened 2026-08-20
fix(fault_manager): keep the near-miss series when a fault is cleared

### LINKED ISSUE #630: [BUG] Clearing a fault destroys the record of how often it nearly happened
# Bug report

The fault manager keeps no history of how often a fault code nearly confirmed, and what is
captured for a fault is deleted when the fault is acknowledged.

### Steps to reproduce

1. Configure debounce so a fault needs more than one report to confirm, for example
   `confirmation_threshold: -4`.
2. Report the same fault code FAILED three times. The fault goes to PREFAILED and the debounce
   counter moves from -1 to -3. Nothing is written per report, only the counter is updated in
   place.
3. Report FAILED once more so the fault confirms, then call `~/clear_fault` on it.
4. Look in the fault database for any trace of the three reports that happened before
   confirmation.

### Expected behavior

There is a record of how often the fault code approached confirmation. One entry per report
that moved the debounce counter, and the record stays after the fault is acknowledged.

### Actual behavior

There is no such record. Two behaviours combine:

- the debounce counter is updated in place, so each report overwrites the previous state and no
  series is ever built
- `clear_fault` runs `DELETE FROM snapshots WHERE fault_code = ?`, so what was captured for the
  fault is removed at acknowledgement

Both storage backends behave this way, `SqliteFaultStorage` and `InMemoryFaultStorage`.

### Why this matters

This is data being destroyed, not a feature that is missing. Any later analysis of recurring
faults, for example how often a code approaches confirmation and whether that rate is growing,
needs this series. It cannot be rebuilt afterwards. Every day a deployment runs without it, more
of that history is lost for good.

### Environment

- ROS 2 distro: Jazzy (the code is shared, so Humble and Lyrical behave the same)
- OS: Ubuntu 24.04

### Additional information

Source: `src/ros2_medkit_fault_manager/src/sqlite_fault_storage.cpp`, the `DELETE FROM snapshots`
in `clear_fault` and the in-place counter update in `report_fault_event`.

Any fix should bound how much is kept per fault code, so the database does not grow without a
limit, and the bound should be configurable instead of a fixed number in the code.

bburda:
### Why

Nothing in the fault manager recorded how often a fault code nearly confirmed. Two behaviours
combined to destroy that information. The debounce counter was updated in place, so every report
overwrote the previous state and no series was ever built. And `clear_fault` deleted what had been
captured for the fault, so acknowledging the fault removed the rest.

This is data being destroyed, not a feature that was missing. Any later analysis of recurring
faults, for example how often a code approaches confirmation and whether that rate is growing,
needs this series, and it cannot be rebuilt after the fact. Every day a deployment runs without
it, more of that history is gone for good. That is why this is a fix and not an enhancement.

### What a near miss is

A FAILED report that moves the debounce counter but does not confirm the fault. A PASSED report
moves the counter in the healing direction, so it is not a near miss. The predicate lives in one
place, `is_near_miss()`, so both storage backends agree.

### The near-miss series

1. New `near_misses` table, one row appended per near miss, never updated in place. Each row holds
   the timestamp, the counter after the report, the confirmation threshold it was measured
   against, the severity, the reporting source, and the fault status the report left behind.
2. `resulting_status` is what makes the series readable. The HEALED latch holds the status the
   whole way from the healing threshold down to the confirmation threshold, so every report on the
   way back into a fault that **does** confirm also moves the counter without confirming. Entries
   recording `PREFAILED` are approaches from a resting state; entries recording `HEALED` are a
   counter walking back down under the latch. Without the field the two are indistinguishable and
   the question the series exists for is not answerable.
3. Entries are kept and evicted in **arrival order**, not by timestamp. Reporters carry their own
   clocks, so a report can arrive with a timestamp behind one already stored; ordering by timestamp
   would let such a report evict itself the moment it was written.
4. `near_miss.max_per_fault`, default 200, 0 means unlimited, evicting the **oldest** entries.
   That is the opposite of `snapshots.max_per_fault`, which keeps the earliest, because a series
   that stops growing at boot cannot show whether the rate is changing. Applying the bound also
   trims what is already stored, and the node warns with the number of entries that dropped,
   because a mistyped value deletes history that cannot be recovered.
5. `clear_fault` no longer touches the series, and neither does the startup reclassification of
   HEALED faults, which is the second place that removed captured data.
6. The fault row and the near-miss row commit together, so a failure on the append cannot leave
   the debounce counter advanced with the near miss missing. Only FAILED reports take that
   transaction: a PASSED report writes at most one row, and taking the writer lock for it would
   make a heal heartbeat fail with `SQLITE_BUSY` where before it could not.
7. `InMemoryFaultStorage` keeps the same series, and a test drives one report sequence through both
   backends and compares what they store.

### Snapshot retention, now that the switch exists on main

`snapshots.retain_on_clear` landed separately, so this branch no longer adds it. What it does add
is the handling that switch still needs.

8. The startup reclassification of HEALED faults deleted their snapshots regardless of the
   setting. Retention therefore held only until the next restart, which then removed exactly what
   it was set to keep, and the backends disagreed: the in-memory one kept the snapshots, SQLite
   dropped them. Both follow the switch now.
9. `~/get_fault` served the freeze-frame only when no snapshots remained, which was the signal
   that acknowledgement had removed them. With retention on they never run out, so the frame - the
   state at the most recent confirmation - stayed hidden behind snapshots of earlier occurrences.
10. `~/get_snapshots` returns one entry per topic and now serves the **newest** capture of that
    topic. It let the last row processed win, and the backends return these in opposite orders, so
    SQLite served the oldest value under the newest `captured_at`.

### Two other corrections along the way

The SQLite backend brings a stored debounce counter back into the reporting config's band before
applying a report and the in-memory one did not. Per-entity threshold overrides make that
reachable in one process, and it offset the whole series and the report at which a fault confirms.

### Before enabling anything

With the default `confirmation_threshold: -1` the first FAILED report confirms the fault at once,
so there is never a near miss to store. The series only fills where debounce is configured.

With `snapshots.retain_on_clear` on, read `snapshots.max_per_fault` as a cap for the whole life of
the database rather than for one fault cycle. It rejects new snapshots and keeps the earliest, and
the rejection is silent, so a code that has reached the cap records nothing on later occurrences.
Raise it, or set it to 0, if you need snapshots from every occurrence.

The near-miss bound is per fault code, not per database. Fault codes are unbounded in cardinality,
so the bound caps what any single code costs, not the total.

### Known limitation

There is no service or REST surface for the series. It is read through `FaultStorage` or from the
database file, so nothing can query it operationally yet and no end-to-end test can exercise it.
Adding one needs a new service definition, which is deliberately out of scope here.

---

## Issue

- closes #630

---

## Type

- [x] Bug fix
- [x] New feature or tests
- [ ] Breaking change
- [ ] Documentation only

---

## Testing

Full package suite: 739 tests, 0 errors, 0 failures (unit, integration and linters).

```bash
./scripts/test.sh all --packages-select ros2_medkit_fault_manager
```

Near-miss series, SQLite: appended and not overwritten, the confirming report is not counted,
survives `clear_fault`, continues across a reactivation, survives closing and reopening the
database, a report after reopen extends the series in order, a PASSED report is not counted, a
CRITICAL immediate confirm is not counted, `confirmation_threshold: -1` records nothing, a FAILED
report under the HEALED latch is counted and is distinguishable from an approach, each entry
carries its own threshold and severity and source and status, arrival order wins over timestamp
order, the bound keeps the newest, a bound of 1 works, the bound is per fault code, 0 is unlimited,
`SIZE_MAX` is unlimited rather than emptying the table, applying a smaller bound trims and reports
how many it dropped, applying an unlimited bound keeps everything, an unknown code returns empty,
a PASSED report on an unknown fault writes nothing, the table and the status column are both
created on a database from an older build, and the series survives HEALED reclassification.

The in-memory backend has the same contract covered, and one parity test drives an identical
sequence with mixed per-entity thresholds through both backends and compares the stored series.

Snapshot retention through the startup reclassification: kept when configured, dropped by default,
unrelated faults untouched, and the same contract on both backends.

Read path, driven through the real `~/get_snapshots` and `~/get_fault` services: the newest
snapshot of a topic wins, and the freeze-frame stays visible behind retained snapshots.

Node level: `near_miss.max_per_fault` reaches storage, the default is 200, a negative value falls
back to the default instead of becoming `SIZE_MAX`, 0 gives unlimited, snapshots are deleted on
clear by default, `snapshots.retain_on_clear` reaches storage, and it is applied before the startup
reclassification rather than after it.

Every behaviour above was checked by mutation: reverting each fix in turn fails only the tests that
encode it, and flipping either default fails exactly the default-behaviour tests.

---

## Checklist

- [x] Breaking changes are clearly described (there are none: the parameter defaults to current
      behaviour, and the schema changes are additive with a migration)
- [x] Docs were updated if behavior or public API changed (package README,
      `docs/config/fault-manager.rst`, `docs/tutorials/snapshots.rst`, default parameters file)

### DIFF
--- a/docs/config/fault-manager.rst
+++ b/docs/config/fault-manager.rst
@@ -83,8 +83,16 @@ CONFIRMED - the fault nearly happened. PASSED reports move the counter in the he
 (the fault receding) and are not near misses.
 
 The fault manager appends one entry per near miss to a per-fault-code series, holding the
-timestamp, the counter value after the report, the confirmation threshold, the severity and the
-reporting source. The series is **retained when the fault is cleared**, because acknowledging one
+timestamp, the counter value after the report, the confirmation threshold, the severity, the
+reporting source and the fault status the report left behind.
+
+The status matters when reading the series. The HEALED latch holds the status all the way from the
+healing threshold down to the confirmation threshold, so reports on the way back into a fault that
+does confirm are also near misses by the definition above. Entries recording ``PREFAILED`` are
+approaches from a resting state; entries recording ``HEALED`` are a counter walking back down under
+the latch. The recorded confirmation threshold belongs to the reporting source, while the counter
+is shared by all sources of that fault code, so with per-entity thresholds it is not on its own the
+distance to confirmation. The series is **retained when the fault is cleared**, because acknowledging one
 fault cycle must not erase how often that code approached confirmation across cycles.
 
 .. code-block:: yaml
--- a/docs/config/fault-manager.rst
+++ b/docs/config/fault-manager.rst
@@ -75,6 +75,45 @@ The fault manager uses AUTOSAR DEM-style debounce filtering to prevent fault fla
    For immediate fault confirmation (no debounce), set ``confirmation_threshold: 0``.
    Faults with ``SEVERITY_CRITICAL`` always bypass debounce regardless of this setting.
 
+Near-Miss Retention
+~~~~~~~~~~~~~~~~~~~
+
+A **near miss** is a FAILED report that moved the debounce counter without the fault ending up
+CONFIRMED - the fault nearly happened. PASSED reports move the counter in the healing direction
+(the fault receding) and are not near misses.
+
+The fault manager appends one entry per near miss to a per-fault-code series, holding the
+timestamp, the counter value after the report, the confirmation threshold, the severity and the
+reporting source. The series is **retained when the fault is cleared**, because acknowledging one
+fault cycle must not erase how often that code approached confirmation across cycles.
+
+.. code-block:: yaml
+
+   fault_manager:
+     ros__parameters:
+       near_miss:
+         max_per_fault: 200                # Entries kept per fault code (0 = unlimited)
+
+.. list-table::
+   :header-rows: 1
+   :widths: 30 12 58
+
+   * - Parameter
+     - Default
+     - Description
+   * - ``near_miss.max_per_fault``
+     - ``200``
+     - Near-miss entries retained per fault code. When the bound is reached the **oldest**
+       entries are evicted - the opposite of ``snapshots.max_per_fault``, which keeps the
+       earliest, because a series frozen at boot says nothing about whether the rate of near
+       misses is changing. Set to 0 for unlimited, accepting growth with the reporting rate.
+
+.. note::
+
+   The series lives in the ``near_misses`` table of the fault database and is read through the
+   storage API (``FaultStorage::get_near_misses``). A database written by an earlier build gains
+   the table on first open. There is no service or REST surface for it yet.
+
 Per-Entity Thresholds
 ~~~~~~~~~~~~~~~~~~~~~
 
--- a/docs/config/fault-manager.rst
+++ b/docs/config/fault-manager.rst
@@ -112,9 +112,9 @@ fault cycle must not erase how often that code approached confirmation across cy
    * - ``near_miss.max_per_fault``
      - ``200``
      - Near-miss entries retained per fault code. When the bound is reached the **oldest**
-       entries are evicted - the opposite of ``snapshots.max_per_fault``, which keeps the
-       earliest, because a series frozen at boot says nothing about whether the rate of near
-       misses is changing. Set to 0 for unlimited, accepting growth with the reporting rate.
+       entries are evicted, the same direction as ``snapshots.max_per_fault`` and the rosbag cap:
+       a series frozen at boot says nothing about whether the rate of near misses is changing.
+       Set to 0 for unlimited, accepting growth with the reporting rate.
 
 .. note::
 
--- a/src/ros2_medkit_fault_manager/README.md
+++ b/src/ros2_medkit_fault_manager/README.md
@@ -180,9 +180,10 @@ reporter emitting a stream of distinct codes still grows the table; the bound ca
 code costs, not the total.
 
 Retention is **bounded per fault code** by `near_miss.max_per_fault` (default 200), evicting the
-**oldest** entries first. That is deliberately the opposite of the snapshot limit's keep-earliest
-rule: a series frozen at boot says nothing about whether the rate is changing. Set it to 0 for
-unlimited, accepting that the database then grows with the reporting rate.
+**oldest** entries first. That is the same direction as `snapshots.max_per_fault` and the rosbag
+cap, and for the same reason: a series frozen at boot says nothing about whether the rate is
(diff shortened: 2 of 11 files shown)


docs/config/fault-manager.rst
```diff
@@ -75,6 +75,53 @@ The fault manager uses AUTOSAR DEM-style debounce filtering to prevent fault fla
...
+   :header-rows: 1
+   :widths: 30 12 58
+
+   * - Parameter
+     - Default
+     - Description
+   * - ``near_miss.max_per_fault``
+     - ``200``
+     - Near-miss entries retained per fault code. When the bound is reached the **oldest**
+       entries are evicted - the opposite of ``snapshots.max_per_fault``, which keeps the
```
mfaferek93: The "opposite of `snapshots.max_per_fault`, which keeps the earliest" claim is no longer true on this branch. Since the keep-newest change (df421024, an ancestor here), both backends evict whole capture sets oldest-first and keep the newest - I ran both with `set_max_snapshots_per_fault(2)` and three capture sets, and sets 2 and 3 survive while set 1 is evicted, in memory and SQLite alike. Line 409 of this same file already documents the snapshot cap as keep-newest, so the two sections contradict each other. As written, an operator sizing `snapshots.max_per_fault` under `retain_on_clear` would conclude the newest evidence gets rejected once the cap fills, which is exactly backwards. The same stale claim is in the fault_manager README (line 183) and the `record_near_miss_locked` trim comment in sqlite_fault_storage.cpp (1357-1358) - worth fixing all three in one pass. The near-miss eviction rule itself needs no change, it is aligned with the snapshot cap, not opposite to it.

bburda: You are right. The snapshot cap moved to keep-newest in `df421024` and I left the old wording behind. I checked both backends before changing anything: with a cap of 2 and three capture sets, sets 2 and 3 survive in memory and in SQLite, so "keeps the earliest" was simply wrong.

Fixed in `3687c341`, all three places you listed:

- `docs/config/fault-manager.rst` - the `near_miss.max_per_fault` table entry
- `README.md` - the retention paragraph
- the trim comment in `sqlite_fault_storage.cpp`

All three now say the same thing as the rosbag cap section further down in this file: oldest evicted, newest kept, same direction for all three caps.

The eviction rule itself is unchanged, as you said. Only the reasoning was stale, and it read backwards for anyone sizing `snapshots.max_per_fault` under retention.

mfaferek93 (approved):

PR #636, opened 2026-08-24
fix: five test failures on main, three timing races and two RoboStack-only

### LINKED ISSUE #635: [BUG] Five test failures on main: three timing races, two RoboStack-only
# Bug report

Five test failures on `main`, in three different CI workflows. They have separate causes, so this is a single report for the set rather than one per symptom.

### Steps to reproduce

1. Push to `main` and let CI run.
2. Read the **job** conclusion of the RoboStack job, not the run conclusion. That job is `continue-on-error`, so the run reports success even when the job fails.
3. Re-run the sanitizer jobs a few times. The first three failures below are timing dependent and do not appear on every run.

### Expected behavior

All jobs pass.

### Actual behavior

**1. `test_aggregator_only_configurations`, ASan job**

```
AssertionError: 'remote_calibration:calibrate' not found in {'calibrate': {...
  'x-medkit': {'entity_id': 'mixed_health', 'member_ids': ['local_calibration'], ...}}}
: the mixed Function does not offer its peer-owned operation: ['calibrate']
```

A peer-owned member is in this gateway's tree before what it offers is. The App arrives with the peer's entity list, and its operations arrive by a second read the aggregator makes per App. The suite starts as soon as the Apps have merged, so a test that addresses a peer-owned operation can run inside that gap and see only the local half.

**2. `test_ros2_parameter_transport`, TSan job**

```
WARNING: ThreadSanitizer: data race
  Write of size 8 by main thread:  operator delete(void*, unsigned long)  <- test body
  Previous atomic read of size 1 by thread T66:
    rclcpp::executors::ExecutorNotifyWaitable::is_ready(rcl_wait_set_s const&)
  Mutex created at: rclcpp::Executor::trigger_entity_recollect(bool)
```

The test body adds a node to an executor that a fixture thread is already spinning. `add_node` rebuilds the executor's entity collection while the spin thread reads it through the notify waitable. The same shape is in 15 test bodies across three files, so the report names whichever test happened to run and reads as a defect in that one.

Note: the Jazzy headers document `add_node` and `remove_node` for the case where the executor is waiting in another thread, and `trigger_entity_recollect` is documented as thread safe. This may be an rclcpp defect rather than misuse, but the tests need to stop doing it either way.

**3. `test_merge_provenance`, Humble job**

```
AssertionError: False is not true : the window must contain samples from before
and after the change; it changed at sample 0 of 12
```

The sampling window reads a generation counter, starts a late node, then takes its first sample with no delay in front of it. Nothing orders those two, so the first recorded sample can already be on the far side of the change. Jazzy and Lyrical pass on the same commit. There is no distro specific code path here.

**4. `test_rosbag_storage_dependency`, RoboStack job, both distros**

```
ERROR collecting src/ros2_medkit_fault_manager/test
  launch_testing/pytest/hooks.py:193: in pytest_pycollect_makemodule
E ModuleNotFoundError: No module named 'test_rosbag_history'
ERROR: found no collectors for .../test/test_rosbag_storage_dependency.py
```

`launch_testing`'s collection hook imports every `*.test.py` file pytest considers. The module name keeps the dot, so `test_rosbag_history.test.py` becomes `test_rosbag_history.test` and the import asks for a package that does not exist. The error is raised against the directory, so it takes down the collection of every plain pytest file in it.

This has been failing since 2026-08-13. It first named a different sibling. Seven launch tests were later moved to `test/integration` and one was left behind, so the same failure came back naming the file that stayed.

**5. `test_aggregation_manager`, RoboStack job, both distros**

```
httplib.h:5860: void httplib::Server::stop(): Assertion `svr_sock_ != INVALID_SOCKET' failed.
(return code -6)
```

`Server::stop()` asserts the listening socket is still valid whenever the server reports itself running, but it is the listen thread that clears the running flag, and only once it comes back out of `accept()`. Three tests stop the mock peer mid-scope and the destructor stops it again, so the second stop can see a running server whose socket has already been taken away. A build with `NDEBUG` drops the assertion, which is why this only shows under RoboStack.

### Environment

- ros2_medkit version: `main`
- ROS 2 distro: Jazzy, Humble, and the RoboStack environments for both
- OS: Ubuntu 24.04 and 22.04 CI images

### Additional information

The RoboStack job has been red since 2026-08-13 while its run reported success. Worth deciding separately whether that job should keep `continue-on-error`.

bburda:
# Pull Request

## Summary

Fixes the five failures reported in the issue below. They have separate causes, so each is a separate commit.

**Three timing races**

- `test_merge_provenance` took its first sample after starting the late node with no delay in front of it, so the first recorded sample could already be past the change. Sample 0 is now taken before the node starts and its generation is the baseline. Sample 1 still follows the change with no sleep, so the observation at the moment of the change is kept. The deadline for the advance now runs from the change, not from the first baseline read.
- `test_aggregator_only_configurations` started as soon as the peer's Apps had merged, but the operations of a peer-owned member arrive by a second read. The read now retries until the operation is there and returns the last response either way, so the test still fails on its own assertion. The three waits in that file also take the shared discovery budget, so they scale with `MEDKIT_TEST_TIME_SCALE`.
- 15 test bodies added a node to an executor that a fixture thread was already spinning. They now use a `SpinningExecutor` that takes its nodes up front, starts its thread after them, and cancels and joins before any node leaves. On Jazzy `cancel()` clears the same flag `spin()` sets on entry, so the helper repeats the cancel until the worker is provably out, and it catches what `spin()` throws instead of letting it terminate the binary.

**Two RoboStack-only failures**

- `test_rosbag_history.test.py` sat next to a plain pytest file, and `launch_testing`'s collection hook then failed the whole directory. It moves to `test/integration` with the seven launch tests already there.
- Three tests stopped the mock peer server mid-scope and the destructor stopped it again. The stop now goes through the fixture, which issues it once. The join stays in the destructor, so this does not trade the abort for a hang.

Two smaller things: a check in `ros2_medkit_cmake` now fails when any directory holds both a launch test and a plain pytest file, because that shape has caused the collection failure twice. And the rosbag descriptor list is built field by field, which clears two build warnings.

---

## Issue

- closes #635

---

## Type

- [x] Bug fix
- [ ] New feature or tests
- [ ] Breaking change
- [ ] Documentation only

---

## Testing

Local, on Jazzy: 5292 tests across `ros2_medkit_gateway`, `ros2_medkit_fault_manager`, `ros2_medkit_graph_watchdog`, `ros2_medkit_integration_tests` and `ros2_medkit_cmake`, 0 errors, 0 failures. Gateway builds with no warnings. The RoboStack smoke script passes locally.

Two of the fixes have a falsifying check, run both ways:

- Putting `test_rosbag_history.test.py` back next to the plain pytest file reproduces the CI error word for word, including `found no collectors`. Moving it gives 6 passed. Checked with the interpreter from a RoboStack environment.
- The new layout check reports exactly the offending directory when the file is put back, and passes when it is not.

Two things I could not verify locally, so please read the CI result for them rather than my word:

- TSan does not start on my machine, it aborts on `unexpected memory mapping` before any test runs. The sanitizer job is the first real run of that fix.
- I have no RoboStack build locally, only the Python interpreter from an existing environment. The httplib fix rests on the mechanism and on the fact that `server_->stop()` now has one caller, not on a local reproduction. I did not reproduce that abort here even with 200 repeats under single core pinning, which fits: the assertion needs the listen thread to still be inside `accept()`.

Worth knowing when reading the RoboStack result: that job is `continue-on-error`, so a red job still shows a green run. It has been red since 2026-08-13.

---

## Checklist


src/ros2_medkit_integration_tests/test/features/test_aggregator_only_configurations.test.py
```diff
@@ -333,6 +337,28 @@ def _items(self, entity_path, collection):
...
+        offers is: the App arrives with the peer's entity list, and its
+        operations arrive by a second read the aggregator makes per App. In
+        between, the member is listed and offers nothing, which is the same
+        shape as a member that genuinely has none.
+
+        The last read is returned either way, so a collection that never gains
+        the item fails on the caller's own assertion and message rather than on
+        a timeout that says only that something did not happen.
+        """
+        deadline = time.monotonic() + DISCOVERY_TIMEOUT
```
mfaferek93: The retry itself is right - it waits on the real condition and returns the last read so the caller's assertion fires. But it pushes this file's worst case past its ctest budget. DISCOVERY_TIMEOUT is 60s (x scale); setUpClass alone is already 3 of them (two _wait_for_apps + _wait_until_merged = 180s), and this retry adds up to a fourth. The feature glob assigns TIMEOUT 120, so on a genuinely broken merge the file dies on ctest timeout with no test name in the output - the exact case the _MEDKIT_TEST_TIMEOUT_OVERRIDES block exists for. Its neighbours (test_peer_recovery, test_triggers_restore_before_discovery) are already listed at 300; this test isn't. An entry there, or a shorter budget for this retry, closes it. Happy path unaffected - 5/5 local runs, ~15s each.

bburda: You are right, and it turned out to be worse than the comment says. I checked before changing anything: setUpClass was already three discovery budgets, so 180s against the glob default of 120s, before this PR touched the file. A broken merge could already die on a ctest timeout with no test name. The retry stacked a fourth budget on top of a file that was over budget already.

So I did both things you offered, because neither one alone closes it. A shorter budget still leaves the three class gates above 120s. The override alone leaves the retry waiting far longer than the thing it waits for.

- The retry now uses `PEER_OPERATION_TIMEOUT`, 15s times the scale, instead of the discovery budget. When it runs, the class gates have already established that the peer Apps are online and merged here, so what is left is one aggregation refresh copying that App's operations across, and the gateways run that refresh every second under test.
- `test_aggregator_only_configurations` is now in `_MEDKIT_TEST_TIMEOUT_OVERRIDES` at 300, next to the neighbours you named.

Worst case drops from about 240s to about 195s against a 300s budget. I confirmed `TIMEOUT "300"` in the generated CTestTestfile rather than assuming the override applied.

mfaferek93 (approved):

PR #638, opened 2026-08-25
fix(aggregation): separate peer time budgets, report timeouts as timeouts, relay peer fault streams

### LINKED ISSUE #528: [BUG] Aggregator times out fanning out large resource payloads to peers

### LINKED ISSUE #611: [BUG] GET /faults/stream on an aggregating gateway returns 200 and streams nothing

### LINKED ISSUE #612: [BUG] A timeout is never reported as a timeout

bburda:
# Pull Request

## Summary

Three defects in peer aggregation. They share `aggregation.timeout_ms` and the peer
fan-out path, and the same files carry all three, so they are fixed together.

### 1. One budget for every kind of peer call

`PeerClient` had a single `timeout_ms`. It was used as the connect timeout and as
the read timeout at the same time. So reading a peer's entity list and waiting for
a peer to do real work were given the same time.

The result: an operation on a peer had 2 s end to end, while the peer's own budget
for the same call is `service_call_timeout_sec`, 10 s by default. A large resource
did not fit in 2 s either.

Raising `timeout_ms` could not fix the other half of this. `set_write_timeout` was
never called anywhere in the gateway. The write budget stayed at the cpp-httplib
default of 5 s and followed no config value at all.

There are four budgets now:

| Budget | Key | Default | Applies to |
|---|---|---|---|
| connect | `aggregation.timeout_ms` | 2000 | every peer call |
| read, metadata | `aggregation.timeout_ms` | 2000 | discovery, health, collection fan-out |
| read, forward | `aggregation.forward_timeout_ms` (new) | 15000 | a forwarded request |
| write | `aggregation.write_timeout_ms` (new) | 5000 | every peer call |

Connect stays short everywhere. A peer that cannot finish a TCP handshake is out of
reach, whatever the request asks for. The forward budget is set above a peer
gateway's own service budget, because that is what a forward waits for.

All three keys are read as int64 and clamped before they are narrowed to int. Before
this, a value above `INT_MAX` wrapped back into the legal range and passed the check.
The range is 100 to 600000 ms and every clamp is logged. If `forward_timeout_ms` is
below `timeout_ms`, it is raised to it with its own warning, because a forwarded
request must not be cut off earlier than a listing.

### 2. A timeout was not reported as a timeout

`httplib::Result::operator bool()` is only a null check. The forward path branched on
`!result` and never read `result.error()`. A read timeout, a refused connection and a
cancelled call all gave the same `502` and the message "is unavailable". That message
said the peer was down while the same peer answered the same request in 4.9 s.

The fan-out path stored one boolean per peer and threw away the error string it had
just built. So a timeout, a refused connection, a peer returning 500, a body over the
size limit and unparseable JSON were the same entry in `failed_peers`.

A local operation returned `500 "Service call failed"` whether it timed out or failed
to serialize its request. The action send-goal path did not even build a message:
"Send goal timed out" named neither the action nor the budget.

`ERR_NOT_RESPONDING` was already reserved for this case and was returned from one
place only. It now covers all three:

| Case | Before | After |
|---|---|---|
| forward, out of time | `502` peer-unavailable | `504` `not-responding`, names the budget |
| forward, nothing there | `502` peer-unavailable | same, and the reason is named |
| fan-out, any failure | `partial` + `failed_peers` | plus `peer_failures`, one reason per peer |
| local execute, out of time | `500` "Service call failed" | `504` `not-responding`, names endpoint and budget |
| local execute, endpoint absent | `500` | `503` |

The classifier does not use the transport error alone. cpp-httplib reports both "the
read budget ran out" and "the connection broke while the answer was arriving" as
`Error::Read`. So the elapsed time is compared with the budget: a call that came back
inside its budget did not run out of it, and a peer that died in the middle of an
answer is reported as unreachable, not as slow.

### 3. The fault stream on an aggregating gateway

`GET /faults/stream` on an aggregating gateway answered `200 text/event-stream` and
then sent only keepalives. The gateway runs on its own ROS domain with its own
`fault_manager` that no producer reports to. So the route was open and valid and
silent, and a client reading it sees a healthy system. In many deployments this is the
only port an operator can reach.

`GET /faults` already fanned out. `SSEStreamProxy` was already written and unit
tested. The design document already described the relay. Nothing constructed it
outside its own test.

`PeerFaultRelay` opens each healthy peer's stream through that proxy and puts the
events into this gateway's own stream. Each relayed event carries `x-medkit.peer`.

Three rules shape it:

- It opens on the first attached client and closes with the last. A relayed stream
  holds one SSE client slot on every peer, and `sse.max_clients` is 2 by default.
- It reconciles from the streaming loop, so a peer that appears or goes away is picked
  up without a timer of its own.
- Its own request carries `X-Medkit-No-Fan-Out`, the header the collection routes
  already use. A stream request that carries it is served from the local graph only.
  Without this, a chain of aggregating gateways would pass one event around the loop.

Replay is limited on purpose and the limit is documented. Two peers number their
events independently, so a `Last-Event-ID` cannot address a position in a merged
stream. The aggregator gives every event an id of its own, and a reconnecting client
resumes from what the aggregator has buffered.

Relayed connections use up a peer's `sse.max_clients` without anyone seeing it, so
`GET /health` now reports `x-medkit-sse` with `connected_clients` and `max_clients`.
A `503` on `/faults/stream` was not explainable from outside before.

Two more defects fixed here, both in code this change touches:

- `X-Medkit-No-Fan-Out` was only checked where the relay is opened. The replay buffer
  is shared by all clients, so a relayed event still reached a client that had asked
  for the local graph only. The filter now runs where events are handed out.
- The replay buffer used the fault code alone as the supersede key. A fault code is
  unique only on the gateway that raised it, so a relayed fault could have replaced a
  local fault with the same code.

## Issue

- closes #528
- closes #611
- closes #612

## Type

- [x] Bug fix
- [x] New feature or tests
- [x] Breaking change
- [ ] Documentation only

The breaking part is small and it is the point of the change. A request that answered
`500` or `502` can now answer `504` or `503`, when that is what happened. No response
that existed before was removed. `failed_peers` keeps its wire shape, and the reasons
arrive next to it in `peer_failures`.

## Testing

All three defects are about behaviour between two processes, so the tests that decide
them run two or three real gateways and talk to them over HTTP. Three new files.

`test_aggregation_time_budgets` runs TWO aggregators in front of ONE peer. They differ
only in the key under test.

- A peer operation held open for 4 s returns the peer's own result through the
  aggregator whose forward budget covers it, and `504 not-responding` through the one
  whose budget does not.
- A lidar scan with 180000 readings arrives with the same number of readings as a read
  taken straight from the owning gateway, and times out on the tight aggregator.

The tight read runs first, on a cold topic. A gateway keeps a live subscription to a
topic it has served, so the first read pays for the subscription and later reads do
not. If the generous aggregator read first, the tight one would measure a warm cache
and return `200` under any budget.

`test_peer_failure_reasons` asks the same peer the same question from two aggregators
and requires two different answers: `timeout` from the one whose budget runs out while
the peer is still working, `error-status` from the one that waits for the peer's own
reply. A classifier that returned one constant would pass a test with one reason. It
then kills the peer and requires `unreachable`, not `timeout`.

`test_aggregator_fault_stream` raises a fault on a peer and requires it on the
aggregator's stream, with the peer named. The aggregator runs a `fault_manager` of its
own on purpose. Without one, "an event arrived" could only mean it came from a peer,
and the assertion about naming the peer could never fail. The cost of the relay is
measured on the peer's own SSE occupancy, because the claim is about a connection the
peer holds.

The three new budget keys are added to `test_startup_param_clamp_warnings`, which
reads the output of four gateways: each key at its floor, at its ceiling, at a
degenerate value, and above `INT_MAX`.

`demo_slow_calibration_service` is new. It is a service that takes a settable time to
answer. Every other service node in the test package answers at once, which cannot
tell a budget that is too small from one that is right.

Measured on Jazzy:

- gateway unit tests: 2813 tests, 0 errors, 0 failures
- integration tests, whole package: 3894 tests, 0 errors, 0 failures
- gateway build: 0 compiler warnings in our code
- the three new end-to-end files together: all passed, 227 s

One note on a test that is not part of this change.
`test_default_script_provider` reported "did not generate a result file" once in an
early batch run. It passed 20 out of 20 in two separate ten-run sweeps, and did not
happen again in four later full-suite runs. This branch does not touch that package.

## Checklist

### DIFF
--- a/src/ros2_medkit_gateway/config/gateway_params.yaml
+++ b/src/ros2_medkit_gateway/config/gateway_params.yaml
@@ -648,6 +648,19 @@ ros2_medkit_gateway:
       # Default: false (safe default - no token forwarding)
       forward_auth: false
 
+      # What this gateway presents as Authorization on connections it opens on
+      # its OWN behalf: the peer health check, the entity fetch, the fault-stream
+      # relay, and any forward whose caller sent no credential to pass on.
+      # Needed against peers that authenticate their reads - while it is empty
+      # such a peer answers 401, is recorded offline, and contributes neither
+      # entities nor events.
+      # Sent ONLY to peers named in peer_urls; a peer discovered over mDNS never
+      # receives it. Redacted from the configurations API, like auth.jwt_secret.
+      # Separate from forward_auth, which forwards an end user's token per
+      # request and has no effect on the relay.
+      # Default: "" (no credential)
+      peer_auth_header: ""
+
       # Security: Require TLS (https://) for all peer communication.
       # When true, peers with http:// URLs are rejected with an error log.
       # When false, http:// peers produce a warning about cleartext traffic.
--- a/src/ros2_medkit_gateway/config/gateway_params.yaml
+++ b/src/ros2_medkit_gateway/config/gateway_params.yaml
@@ -224,6 +224,15 @@ ros2_medkit_gateway:
       # Default: 3600 (1 hour)
       max_duration_sec: 3600
 
+      # Comment interval on an idle SSE stream, in seconds. Also how long the
+      # server takes to notice a client has gone: a closed socket is found on
+      # the next write. On an aggregating gateway that is how long a fault-stream
+      # relay keeps a client slot on a peer after nobody is watching.
+      # Shorten it behind a proxy that drops idle connections sooner than 30 s.
+      # Range 1-3600, clamped to the nearest end with a warning.
+      # Default: 30
+      keepalive_interval_sec: 30
+
     # Entity Cache Configuration
     entity_cache:
       # Pre-reserved capacity for the thread-safe entity cache.
@@ -592,10 +601,27 @@ ros2_medkit_gateway:
       # Default: false
       enabled: false
 
-      # HTTP timeout for peer requests (health checks, entity fetching, forwarding)
-      # in milliseconds
+      # Connect budget for every peer request, and the read budget for reading a
+      # peer's description of itself: health checks, entity fetching, and the
+      # collection fan-out a client waits on across every peer at once.
+      # Milliseconds. Range 100-600000, clamped to the nearest end with a warning.
       # Default: 2000
       timeout_ms: 2000
+
+      # Read budget for a forwarded request - an operation execution, or a read
+      # of a large resource - on the gateway that owns it. Sized above a peer
+      # gateway's own service_call_timeout_sec default of 10 s,
+      # because what this waits for is the peer doing that work.
+      # A value below timeout_ms is raised to it, with a warning.
+      # Milliseconds. Range 100-600000, clamped to the nearest end with a warning.
+      # Default: 15000
+      forward_timeout_ms: 15000
+
+      # Budget for pushing a request body to a peer. Raise for large uploads
+      # over a slow link.
+      # Milliseconds. Range 100-600000, clamped to the nearest end with a warning.
+      # Default: 5000
+      write_timeout_ms: 5000
 
       # Announce this gateway's presence via mDNS
       # Default: false (opt-in to avoid surprising network behavior)
--- a/src/ros2_medkit_gateway/src/core/aggregation/peer_client.cpp
+++ b/src/ros2_medkit_gateway/src/core/aggregation/peer_client.cpp
@@ -634,12 +634,18 @@ const char * peer_failure_reason(PeerFailureKind kind) {
   return "unreachable";
 }
 
-PeerClient::PeerClient(const std::string & url, const std::string & name, PeerTimeouts timeouts, bool forward_auth)
-  : url_(url), name_(name), timeouts_(timeouts), forward_auth_(forward_auth) {
+PeerClient::PeerClient(const std::string & url, const std::string & name, PeerTimeouts timeouts, bool forward_auth,
+                       std::string peer_auth_header)
+  : url_(url)
+  , name_(name)
+  , timeouts_(timeouts)
+  , forward_auth_(forward_auth)
+  , peer_auth_header_(std::move(peer_auth_header)) {
 }
 
-PeerClient::PeerClient(const std::string & url, const std::string & name, int timeout_ms, bool forward_auth)
-  : PeerClient(url, name, PeerTimeouts::uniform(timeout_ms), forward_auth) {
+PeerClient::PeerClient(const std::string & url, const std::string & name, int timeout_ms, bool forward_auth,
+                       std::string peer_auth_header)
+  : PeerClient(url, name, PeerTimeouts::uniform(timeout_ms), forward_auth, std::move(peer_auth_header)) {
 }
 
 const std::string & PeerClient::url() const {
@@ -689,7 +695,16 @@ void PeerClient::ensure_client() {
 void PeerClient::check_health() {
   std::lock_guard<std::mutex> lock(client_mutex_);
   ensure_client();
-  auto result = client_->Get(std::string(API_PREFIX) + "/health");
+  // The health check is this gateway asking on its own behalf, so it carries
+  // the gateway's own credential. Without it a peer that requires auth on
+  // reads answers 401, is recorded as unhealthy, and then appears in no
+  // fan-out and no relay - the peer is reachable and correctly configured, and
+  // the aggregator reports it as down.
+  httplib::Headers headers;
+  if (!peer_auth_header_.empty()) {
+    headers.emplace("Authorization", peer_auth_header_);
+  }
+  auto result = client_->Get(std::string(API_PREFIX) + "/health", headers);
   healthy_.store(result && result->status == 200);
 }
 
@@ -1010,6 +1025,11 @@ void PeerClient::forward_request(const httplib::Request & req, httplib::Response
   // Default is off to prevent token leakage to untrusted/mDNS-discovered peers.
   if (forward_auth_ && req.has_header("Authorization")) {
     headers.emplace("Authorization", req.get_header_value("Authorization"));
+  } else if (!peer_auth_header_.empty()) {
+    // The gateway's own credential, used when there is no client credential to
+    // pass on. A client's own token wins where forwarding is enabled, so
+    // forward_auth keeps naming the end user to the peer.
+    headers.emplace("Authorization", peer_auth_header_);
   }
   // Propagate fan-out suppression header to prevent recursive loops when
   // a forwarded request bounces back to the origin via bidirectional peering.
@@ -1133,6 +1153,8 @@ tl::expected<nlohmann::json, PeerError> PeerClient::forward_and_get_json(const s
   // Only forward auth header when forward_auth is enabled
   if (forward_auth_ && !auth_header.empty()) {
     headers.emplace("Authorization", auth_header);
+  } else if (!peer_auth_header_.empty()) {
+    headers.emplace("Authorization", peer_auth_header_);
   }
   for (const auto & [key, value] : extra_headers) {
     headers.emplace(key, value);
--- a/src/ros2_medkit_gateway/src/core/aggregation/peer_client.cpp
+++ b/src/ros2_medkit_gateway/src/core/aggregation/peer_client.cpp
@@ -16,6 +16,7 @@
 
 #include <algorithm>
 #include <cctype>
+#include <chrono>
 #include <iterator>
 #include <set>
 #include <string>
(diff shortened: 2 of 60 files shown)


src/ros2_medkit_gateway/src/core/aggregation/sse_stream_proxy.cpp
```diff
@@ -84,11 +85,23 @@ void SSEStreamProxy::reader_loop() {
...
     std::string buffer;
+    // Resume where the last connection stopped. A reconnect without this asks
+    // the peer to replay its whole buffer, and every replayed event is emitted
+    // again under a new aggregator id.
+    httplib::Headers attempt_headers = headers_;
+    if (!last_event_id_.empty()) {
+      attempt_headers.emplace("Last-Event-ID", last_event_id_);
+    }
+
     auto result = client.Get(
```
mfaferek93, line 183: The reader parks here in `client.Get` on the peer stream, but `close()` (line 47) only flips `should_stop_`, which the callbacks check only when bytes arrive, and `client` is a local in `reader_loop` (line 72) with no `stop()` handle. So on a connected-but-silent peer, `close()`/join blocks up to the 300s read timeout (~30s to the next keepalive). This is on the live path via `PeerFaultRelay::release()/shutdown()`: restarting the aggregator while a peer stays up between keepalives can overrun the shutdown budget and get SIGKILLed. Measured 118s in a probe. Suggest holding the client as a mutex-guarded member and calling `client.stop()` in `close()` after setting `should_stop_`.

src/ros2_medkit_gateway/src/core/aggregation/peer_client.cpp
```diff
@@ -511,8 +512,124 @@ SubResponse read_sub_response(const httplib::Result & result, const std::string
...
+      return PeerFailureKind::kTimeout;
+    case httplib::Error::Canceled:
+      return PeerFailureKind::kCanceled;
+    case httplib::Error::Read:
+      // The budget is always set on these clients, so a call that consumed it
+      // ran out of time. One that came back sooner broke for another reason -
+      // most often the peer process dying with the answer half sent - and
+      // reporting that as a timeout would tell an operator to wait for a peer
+      // that is gone.
+      return elapsed >= read_budget ? PeerFailureKind::kTimeout : PeerFailureKind::kUnreachable;
```
mfaferek93, line 589: `elapsed` here spans connect + read but is compared only against `read_budget`. With the default `connect_ms == metadata_read_ms == 2000`, a slow handshake then the peer dying mid-answer (read breaks at total ~2s) is classified `timeout` when the honest answer is `unreachable` - inverting the distinction this change adds. Suggest timing the read phase separately (or subtracting the connect deadline) so the comparison measures only the read.

src/ros2_medkit_gateway/src/core/aggregation/sse_stream_proxy.cpp
bburda: Your 118s probe matches what I measured. The join was bounded only by the 300s read timeout.

I could not use `client.stop()` for the fix. If it is called while the reader thread is inside `Get()`, it unlocks one of cpp-httplib's own mutexes from a thread that never locked it. TSan reports this as an unlock of an unlocked mutex, and glibc pthread debug mode treats it as fatal. This is the same reason `PeerClient` never stops its health-check client. I wrote your suggestion as it was, and the TSan job failed on it.

What I did instead: the reader saves a `dup()` of the connection socket in the socket-options callback, and `close()` calls `shutdown(fd, SHUT_RDWR)` on it. The read wakes up through the kernel, and no library state is touched. It has to be a dup. cpp-httplib closes its own descriptor inside `Get()`, and a freed number can be given to another thread in the process. Publishing the dup under the same mutex that `close()` reads also covers the case where `close()` runs before the socket exists.

`close_terminates_reader_thread` now returns in about 100 ms, and both integration tests are clean under TSan.

src/ros2_medkit_gateway/src/core/aggregation/peer_client.cpp
bburda: You are right about the inversion. cpp-httplib has no post-connect hook. Its socket-options callback runs before the connect, so there is no exact start of the read phase to measure from. The only way to get one is to patch the vendored header, and I would rather not carry that patch across upstream updates.

Instead the branch bounds the window. The connect budget is capped at 1000 ms, independent of `timeout_ms` (see `AggregationConfig::peer_timeouts`). A handshake can no longer use most of a read budget, and a handshake slower than the cap fails as `ConnectionTimeout`, which is named correctly. With the defaults you mention, the wrong-attribution window is 1 second instead of the full 2000 ms metadata budget. The comment at the check says what is left and what removing it would need.

src/ros2_medkit_gateway/src/http/handlers/sse_fault_handler.cpp
```diff
@@ -260,22 +388,52 @@ std::function<bool(httplib::DataSink &)> SSEFaultHandler::make_stream_loop(uint6
...
-          pending.emplace_back(queued.id, format_sse_event(queued));
+        if (queued.id <= last_event_id) {
+          continue;
+        }
+        // Suppression has to bite HERE, not only where the relay is opened.
+        // The buffer is shared by every client, so a relayed event put there
+        // for one client would otherwise be delivered to a client that asked
+        // for the local graph only - which is what an aggregating peer asks
+        // for, and is what would carry an event round a chain of them.
+        if (!relay_peers && !queued.peer.empty()) {
```
mfaferek93, line 430: **Minor** (accounting only; delivery is correct): this skip advances only the loop-local `last_event_id`, never the client's watermark in the queue. `has_pending_locked` is suppression-aware, so a tail made of relayed-only events never wakes this client, and its cursor catches up only at the next keepalive via `note_progress_locked`. A relayed-fault storm that overflows the 100-entry buffer inside one keepalive interval counts those evictions as drops owed to a client that never asked for them, inflating `dropped`/`slow_clients` for a healthy consumer.

Suggestion: call `note_progress_locked(cursor, queued.id)` in this branch (collect_pending already runs under `queue_mutex_`), or fold a per-client skip watermark into `delivered_watermark_locked`.

src/ros2_medkit_gateway/config/gateway_params.yaml
```diff
@@ -592,11 +601,28 @@ ros2_medkit_gateway:
...
       enabled: false
 
-      # HTTP timeout for peer requests (health checks, entity fetching, forwarding)
-      # in milliseconds
+      # Connect budget for every peer request, and the read budget for reading a
+      # peer's description of itself: health checks, entity fetching, and the
+      # collection fan-out a client waits on across every peer at once.
+      # Milliseconds. Range 100-600000, clamped to the nearest end with a warning.
       # Default: 2000
       timeout_ms: 2000
```
mfaferek93, line 609: **Minor**: `aggregation.peer_auth_header` is declared in `gateway_node.cpp` and documented in `docs/config/aggregation.rst`, but neither this reference yaml nor `gateway_params.secure.yaml`'s hardened aggregation block mentions it, while every sibling (`forward_auth`, `require_tls`, `peer_scheme`, `mdns_name`) has a commented entry. An operator working from the reference yaml won't find the one key that makes the relay work against authenticating peers. Suggest adding the commented entry to both files.

src/ros2_medkit_gateway/src/http/handlers/sse_fault_handler.cpp
bburda: You are right, and the comment on that line already claimed it moved the watermark while the code only moved the loop's own copy.

I traced the case you describe. `has_pending_locked` applies the same suppression, so a tail of relayed-only events never wakes that client, and its cursor only catches up at the next keepalive through `note_progress_locked`. Until then, eviction reads the shared cursor and books those entries as unread for a client that never asked for them, so `dropped` and `slow_clients` grow for a healthy consumer.

Fixed as you suggested: `collect_pending` now calls `note_progress_locked(cursor, queued.id)` in that branch. It already runs under `queue_mutex_`, so no extra locking. Unit suite and the SSE stream test are green.

src/ros2_medkit_gateway/config/gateway_params.yaml
bburda: Correct, and this was my miss. I declared the parameter and documented it in `docs/config/aggregation.rst`, but not in the reference configs, where every sibling key has a commented entry.

Added to both. In `gateway_params.yaml` the entry says what it is presented on (health check, entity fetch, relay, and a forward with no caller credential), that it goes only to peers named in `peer_urls` and never to an mDNS-discovered peer, and that it is redacted from the configurations API. In `gateway_params.secure.yaml` it sits in the hardened aggregation block with a note that it is preferable to `forward_auth` where peers authenticate, because it never carries an end user's token.

mfaferek93 (approved):

Issue #259, opened 2026-03-07
Security hardening: plugin safety and /health info leak


bburda:
## Summary

The plugin layer and /health endpoint need hardening before production use. A hanging or malicious plugin can DoS the gateway, and the /health endpoint leaks internal topology to unauthenticated callers.

---

## Proposed solution (optional)

Four related items, all plugin/security themed:

### 1. Plugin `discover()` timeout
`MergePipeline::execute()` calls each layer's `discover()` synchronously with no timeout. A hanging plugin blocks ALL discovery cycles permanently - the gateway serves stale data.

**File:** `src/ros2_medkit_gateway/src/discovery/merge_pipeline.cpp:407-415`

**Fix:** Run plugin `discover()` via `std::async` with `std::future::wait_for()`. Add configurable `discovery.plugin_timeout_ms` (default 5000ms). Disable plugin after N consecutive timeouts.

### 2. Per-layer entity count limit
`PluginLayer::discover()` has no cap on entity count. A plugin can return millions of entities, exhausting gateway memory.

**File:** `src/ros2_medkit_gateway/src/discovery/layers/plugin_layer.cpp:64-74`

**Fix:** Add `discovery.plugin_max_entities` config (default 10000). Truncate + warn if exceeded.

### 3. Plugin field content validation
Entity IDs are validated (format + length), but other fields (`name`, `description`, `tags`) accept arbitrary-length strings from plugins.

**Fix:** Add length limits: `name` max 256 chars, `description` max 4096 chars, `tags` max 100 entries with 256 chars each.

### 4. Gate /health discovery section behind auth
The `/health` endpoint is exempt from authentication but now exposes:
- Merge pipeline conflict details with entity IDs
- `linking->warnings` containing strings like `"App 'engine_ecu' cannot bind to '/nav/controller'"` - leaking ROS 2 graph topology

**File:** `src/ros2_medkit_gateway/src/http/handlers/health_handlers.cpp:42-57`

**Fix:** Return only `{"status": "healthy", "timestamp": ...}` for unauthenticated requests. Gate the `discovery` section behind auth, or strip entity IDs/FQNs.

---

## Additional context (optional)

Found during self-review of PR #258. All items are defense-in-depth - plugins already run in-process, but these changes limit the blast radius of bugs and prevent accidental info leaks.

PR #641, opened 2026-08-26
gateway: add a closed profile, a public-route list, and an environment switch that closes a container

bburda:
# Pull Request

## Summary

The gateway gets a closed **profile**, not closed defaults. `config/gateway_params.yaml` stays open exactly as it is today (auth off, `require_auth_for: "write"`, TLS off), so every existing launch, every existing configuration and the web UI keep working after an upgrade. `config/gateway_params.secure.yaml` carries the closed posture: auth on, `require_auth_for: "all"`, TLS on, rate limiting, an explicit CORS origin list. It is selected with `config_file:=`, and `gateway.launch.py` warns about a missing signing secret only when the profile in force needs one.

Independently of the profile, the environment closes the gateway, and the node reads it itself when it loads its parameters, so the rule is the same on every entry path: the container entrypoint, `ros2 run`, and any launch file.

- `MEDKIT_JWT_SECRET` turns auth on with `require_auth_for: "all"` and that secret, whatever a parameters file or a launch argument says.
- `MEDKIT_CLIENTS` replaces the client list: comma-separated `id:secret:role` entries, surrounding whitespace trimmed, a duplicate id dropped with the first entry winning and a WARN. An empty value means no clients and warns. A value whose every entry is refused stops the gateway. A comma inside a secret breaks the entry. `MEDKIT_CLIENTS` without `MEDKIT_JWT_SECRET` is ignored with a WARN.
- `MEDKIT_AUTH_DISABLED=1` opens any gateway, a hardened source install included. Being able to set it on a gateway's environment is being able to turn its authentication off.
- `ros2 param get` reports the posture the gateway enforces. While auth is on, every configured secret reads back as a sentinel naming where it came from (`<from MEDKIT_JWT_SECRET>`, `<set at start>`), and each client as `<id>:<sentinel>:<role>`. `auth.*` parameters are read at start; a `param set`, a `param load` entry or an atomic batch touching them is refused with a reason, and the other parameters in a `param load` still apply.
- `MEDKIT_JWT_SECRET` together with `auth.jwt_algorithm: RS256` stops the gateway. A refusal names the setting and where it came from, the environment only for the three values it can supply, never a value. A malformed file client entry is reported by position, not echoed. RS256 keys are read once at start, so a key rotation needs a restart.
- `auth.public_routes` is validated whether or not auth is on; a blank entry such as `[""]` means no routes.

Tokens and revocation:

- An access token is checked on its signature and the local client table. It survives a gateway restart and works on any gateway that shares the signing config, which is what `forward_auth` peers need. The role it grants comes from the receiving gateway's `auth.clients`: a client that is admin on the aggregator and viewer on a peer may only read on the peer, and a client absent from the peer's table is refused there.
- Refresh tokens are per process and do not survive a restart.
- `POST /auth/revoke` takes a refresh token and writes a revoked record keyed by its id, kept until the last access token minted on it expires; an access token presented there changes nothing (the answer stays 200). It works on a gateway that never issued the token, so revoking on a peer refuses the token on that peer. The record lives in memory, so a token revoked before a restart can return for at most `token_expiry_seconds` after it. Gateways that share a signing configuration must share `token_expiry_seconds` and `refresh_token_expiry_seconds`, or a revocation on the peer with the shorter value lapses while the issuer's tokens are live.
- Refresh records are swept every five minutes and on every refresh, which bounds the store for refresh-only workloads.

Requests:

- `auth.public_routes`, a list of exact `"METHOD /path"` entries that answer without a credential. It ships empty. When `GET /api/v1/health` is listed, an anonymous caller on it gets a reduced body: `status`, `timestamp`, an empty `warnings`, `warning_schema_version` and `x-medkit-reduced: true`; the entity-cache counts and linking warnings need a token. On a route that is not listed, `/health` keeps its sections, so an anonymous `GET /health` under `require_auth_for: "write"` answers as before.
- The rate limiter answers before any token work. A caller without a credential on a protected route gets one shape of 401 (`WWW-Authenticate` plus the error document) whether it is the first request or the one that exhausted the anonymous burst, and no `X-RateLimit-*` headers. A caller over the limit that presents any `Authorization` header, a valid one included, gets a bare 429 with no `Retry-After` and no `X-RateLimit-*`, before the token is parsed; on the closed profile a client therefore paces itself from `X-RateLimit-Reset` on its last accepted answer. A plain `OPTIONS` is metered like any other request; only a real preflight is exempt.
- `server.tls.min_version` is enforced, and `server.tls.ca_file` turns on mutual TLS: every client must then present a certificate signed by that CA.

The container image:

- The config is three layers, carried by the entrypoint and applied in this order: the image's own file (`server.host` 0.0.0.0, a 2 s refresh interval), the mount point `/etc/ros2_medkit/params.yaml`, then the command's arguments. Later wins, so a mounted file sets `server.host` and `refresh_interval_ms` again, and `docker run <image> --ros-args -p server.port:=9090` keeps `server.host` at 0.0.0.0. The Dockerfile `CMD` is empty.
- The image no longer allows the `localhost:3000` and `localhost:5173` CORS origins it used to. The compose example in docker.rst mounts a params file naming `http://localhost:3000`.
- The secure profile ships in the image at its documented path and binds 8443; TLS still needs the certificate files mounted, and the secret comes from the environment or from a second `--params-file` placed after the secure one.
- The publish workflow checks the image's posture on the built digest before naming it: open with no environment, closed with `MEDKIT_JWT_SECRET`, closed with `MEDKIT_JWT_SECRET` over a mounted file that says auth off. Every push to `main` also publishes `:main-<sha7>` next to `:latest`; release tags keep `:sha-<sha7>` for the multi-arch manifest list.

`MEDKIT_TLS_CERT_FILE` and `MEDKIT_TLS_KEY_FILE` are gone from the launch file. The environment contract is the three variables above.

`test_openapi_contract` waits for a discovered service before comparing operation items, which is what the humble run was racing.

---

## Issue

Refs #259 (item 4, the `/health` information leak).

---

## Type

- [ ] Bug fix
- [x] New feature or tests
- [x] Breaking change
- [ ] Documentation only

---

## Breaking changes

- Image users passing arguments: the config layers moved from the Dockerfile `CMD` into the entrypoint and the `CMD` is empty. Arguments no longer replace the image's values.
- The image drops its `localhost:3000` and `localhost:5173` CORS origins. A browser client needs `cors.allowed_origins` in a mounted file.
- Access tokens survive a restart; a revoked token can return for at most `token_expiry_seconds` after one.
- The role a token grants comes from the receiving gateway's client table, and a client absent from that table is refused, so peers under `forward_auth` need the client in their own `auth.clients`.
- An exhausted caller presenting a credential gets a bare 429 before the token is checked, and a plain `OPTIONS` is metered.
- `auth.public_routes` is validated with auth off; a malformed entry that used to pass unnoticed now stops the gateway.
- `MEDKIT_AUTH_DISABLED=1` opens any gateway.
- `MEDKIT_TLS_CERT_FILE` and `MEDKIT_TLS_KEY_FILE` no longer exist.

---

## Testing

- `test_open_default_profile` pins the compatibility promise: the packaged default file answers an anonymous `GET /areas` and a full anonymous `/health`; with the file mutated to auth on it fails on all three assertions.
- `test_secure_profile` and `test_closed_by_default` launch the secure profile. The second asks the running gateway for its route table and calls every route without a credential and with a fake token: a token whose signature has one byte changed, a correctly signed token naming an unknown client, and a refresh token presented as an access token.
- `test_env_auth_contract` drives the environment rule against the node: precedence over a file and a launch argument, the empty and the all-refused client list, the RS256 refusal, the parameter read-back and the refused `param set`. After shutdown it scans both gateways' output for the secret values.
- `test_forward_auth_across_gateways` runs two gateways under `forward_auth`: a token issued on one works on the other, is refused there after `POST /auth/revoke` on the peer, and grants the peer's role.
- `test_rate_limiting` and `test_auth_policy_contract` pin the single 401 shape, the bare 429, and that `X-RateLimit-*` reaches a credentialed caller and not an anonymous one. A unit test composes the middleware and the limiter the way the server does and asserts the verifier is not reached on the exhausted-header path.
- `test_tls_protocol_floor` drives `openssl s_client` against three real gateways: floors at 1.2 and 1.3, plus one requiring a client certificate.
- `scripts/smoke_image_auth_posture.sh` against a locally built jazzy image reads the gateway's startup line back from the container log: open, closed, closed over a mounted file, a mounted value applied, and `-p server.port` keeping `server.host`, eight cases green. The docker.rst closed-profile example was run against the same image.
- Gateway unit suite: 2948 tests, 0 failures. Integration package on jazzy: 1339 tests, 0 failures. Sphinx with warnings as errors clean.

---

## Checklist


src/ros2_medkit_gateway/src/core/auth/auth_manager.cpp
```diff
@@ -248,10 +282,27 @@ TokenValidationResult AuthManager::validate_token(const std::string & token, Tok
...
+  // credential somebody explicitly withdrew.
+  //
+  // The cost is deliberate and worth stating: a restart invalidates every
+  // access token, so clients re-authenticate after one. That is visible
+  // behaviour, and it is the trade this gateway makes elsewhere too - refusing
+  // is better than quietly allowing.
   if (claims.refresh_token_id.has_value()) {
     auto record = get_refresh_token(claims.refresh_token_id.value());
-    if (record.has_value() && record->revoked) {
+    if (!record.has_value()) {
```
mfaferek93, line 300: Every token this gateway mints carries `refresh_token_id` (lines 163, 229) and `refresh_tokens_` is per-process memory, so with `forward_auth: true` a peer that shares the JWT config but did not issue the token now answers 401 to every forwarded request: the "Secure Aggregation (TLS + Auth)" setup in docs/config/aggregation.rst stops working after this upgrade, and a `peer_auth_header` taken from a peer's `/auth/authorize` dies the moment that peer restarts. `ARestartInvalidatesAccessTokensRatherThanTrustingThem` asserts exactly this cross-instance rejection and no integration test runs forward_auth against an auth-on peer. Keep revocation as a denylist (mark the record revoked, hold it until the last access token's exp) instead of an allowlist that needs the issuer's memory.

src/ros2_medkit_gateway/launch/gateway.launch.py
```diff
@@ -120,6 +232,91 @@ def _launch_setup(context, *_args, **_kwargs):
...
+        # require_auth_for to "all" - "write" would leave every read open, and
+        # the reads are the disclosure.
+        auth_arg = LaunchConfiguration('auth_enabled').perform(context).strip()
+        jwt_secret = (LaunchConfiguration('jwt_secret').perform(context)
+                      or os.environ.get('MEDKIT_JWT_SECRET', ''))
+        clients = (LaunchConfiguration('auth_clients').perform(context)
+                   or os.environ.get('MEDKIT_CLIENTS', ''))
+        env_closes = (bool(os.environ.get('MEDKIT_JWT_SECRET', ''))
+                      and os.environ.get('MEDKIT_AUTH_DISABLED') != '1')
+        if auth_arg:
```
mfaferek93, line 289: `auth_enabled:=true` with `MEDKIT_JWT_SECRET` in the environment takes this branch, which writes only `auth.enabled`, so `auth.require_auth_for` stays at the file's `"write"` and every read answers anonymously while the docs promise the secret closes the gateway whatever the file says. Apply the `'all'` override whenever `env_closes` and auth ends up on, not only inside the `elif`. The same block also lets `auth_enabled:=true` beat `MEDKIT_AUTH_DISABLED=1`, while docker.rst says that flag wins over everything; pick one precedence and document it.

docker/entrypoint.sh
```diff
@@ -24,13 +24,54 @@ source "${COLCON_WS}/install/setup.bash"
...
 #   - empty, or starts with "-" (the default CMD "--ros-args --params-file ..."
 #     or an override like --ros-args -p server.port:=9090): run the gateway node
 #     directly, so `docker run <img>` and arg-only overrides keep working.
 #   - a full command (e.g. `ros2 launch ros2_medkit_gateway bringup.launch.py`
 #     or `bash`): exec it as-is, so the image can launch the whole bringup stack.
 if [ -z "$1" ] || [ "${1#-}" != "$1" ]; then
-  exec ros2 run ros2_medkit_gateway gateway_node "$@"
+  exec ros2 run ros2_medkit_gateway gateway_node "$@" "${AUTH_ARGS[@]}"
 fi
 exec "$@"
```
mfaferek93, line 77: With `MEDKIT_JWT_SECRET` set the entrypoint prints "authentication is ON" (lines 60-61) and then this exec runs the full command untouched: `ros2 run ros2_medkit_gateway gateway_node ...` never gets `AUTH_ARGS` and the node reads no `MEDKIT_*` variable, and `ros2 launch ros2_medkit_gateway gateway_https.launch.py` builds its own Node without looking at the environment, so both serve every route anonymously under a log line saying the opposite. Only bringup.launch.py is covered. Apply the rule in the gateway's own startup so every way of starting the process honours it, or at least prepend `AUTH_ARGS` for `ros2 run ... gateway_node` and make `gateway_https.launch.py` read the same variables.

src/ros2_medkit_gateway/src/http/rest_server.cpp
```diff
@@ -353,6 +395,18 @@ void RESTServer::setup_pre_routing_handler() {
...
     }
 
+    // 5. Now the limiter may speak, headers included. The caller reached this
+    // line with a credential this gateway accepts, or on a route that needs
+    // none, so the allowance and the reset time tell them something they are
+    // entitled to know.
+    if (rate_metered) {
+      RateLimiter::apply_headers(rl_result, res);
+    }
+    if (rate_limited) {
```
mfaferek93, line 405: With the 429 deferred to here, an over-limit request that carries any `Authorization: Bearer ...` still runs the whole of `process` first (JWT decode, signature verify, under RS256 also a `read_file_contents` of the public key per request, client lookup, RBAC) and only then gets refused, so the limiter no longer bounds the verification cost the comment at 352 says it exists to bound; `process` short-circuits only on a missing header (`auth_middleware.cpp:39`). Keep the anonymous 401 for a missing header, then answer 429 right after that check and before `validate_token`, and load the RS256 key once in the constructor.

src/ros2_medkit_gateway/src/http/handlers/health_handlers.cpp
```diff
@@ -50,13 +51,55 @@ ErrorInfo make_internal_error(const char * where, const std::exception & e) {
...
     dto::Health response;
     response.status = "healthy";
     response.timestamp = std::chrono::system_clock::now().time_since_epoch().count();
 
+    // Liveness and nothing else for an anonymous caller. Returned before any
+    // of the sections below are built, so a section added later is private by
+    // default rather than public until someone remembers to think about it.
+    // The flag is what keeps the empty `warnings` below from reading as
+    // "nothing is wrong here" to a monitor that never presented a credential.
+    if (is_anonymous(ctx_, req)) {
```
mfaferek93, line 98: `is_anonymous` keys on `auth.enabled` alone, so under `require_auth_for: "write"` (the server.rst example and every existing auth-on deployment) an anonymous `GET /health` now loses the discovery, entity-cache and linking sections too, while the PR text, server.rst, hardening.rst and the OpenAPI description on `x-medkit-reduced` all say this only happens on a route listed in `auth.public_routes`. Either gate the cut on the route actually being exempted by `PublicRouteExemptionPolicy` (expose `is_public(method, path)`), or keep it and state it as a behaviour change for auth-on `write` deployments.

Dockerfile
```diff
@@ -188,4 +191,14 @@ USER medkit
...
-CMD ["--ros-args", "--params-file", "/etc/ros2_medkit/params.yaml"]
+# The two values a container needs that a host install does not: bind every
+# interface, because the port is published rather than reached over loopback,
+# and refresh faster, because a container's graph turns over as sibling
+# containers come and go. Passed after the params file so they win over it.
+#
+# CORS names no origin here. A published image allowing development origins is
+# a setting nobody chose; a deployment that runs the web UI next to the gateway
+# sets cors.allowed_origins to its own origin.
+CMD ["--ros-args", "--params-file", "/etc/ros2_medkit/params.yaml", \
```
mfaferek93, line 202: The two `-p` overrides come after `--params-file`, so a mounted `/etc/ros2_medkit/params.yaml` can no longer set `server.host` or `refresh_interval_ms`: rcl takes the later value and says nothing, while docker.rst:104-125 still tells people to mount a file and shows one setting both keys. Someone on `network_mode: host` who mounts `host: "127.0.0.1"` to stay on loopback now gets 0.0.0.0. Put the two container values in a small image-side overlay loaded before the mounted file, or state in docker.rst that a mounted file cannot set them.

docs/tutorials/docker.rst
```diff
@@ -156,7 +187,15 @@ Example ``docker-compose.yml`` with the gateway and web UI:
        environment:
```
mfaferek93, line 187: This compose example still runs the web UI on :3000 against a gateway with no params mount, but the image now ships `cors.allowed_origins [""]` (the old docker params that allowed localhost:3000/5173 are deleted), so `CorsConfigBuilder` sets enabled=false and the browser at http://localhost:3000 blocks every call: the documented gateway+UI stack stops working as written. Give the example a mounted params file naming http://localhost:3000, and tick Breaking change in the PR: image users lose the localhost origins and every access token now dies on a gateway restart.

docker/entrypoint.sh
```diff
@@ -24,13 +24,54 @@ source "${COLCON_WS}/install/setup.bash"
...
+if [ "${MEDKIT_AUTH_DISABLED:-0}" = "1" ]; then
+  AUTH_ARGS+=(-p auth.enabled:=false)
+  echo "ros2_medkit: MEDKIT_AUTH_DISABLED=1 - starting WITHOUT authentication." >&2
+  echo "             Every route is readable by anyone who can reach this port." >&2
+elif [ -n "${MEDKIT_JWT_SECRET:-}" ]; then
+  AUTH_ARGS+=(-p auth.enabled:=true)
+  AUTH_ARGS+=(-p auth.require_auth_for:=all)
+  AUTH_ARGS+=(-p "auth.jwt_secret:=${MEDKIT_JWT_SECRET}")
+  if [ -n "${MEDKIT_CLIENTS:-}" ]; then
+    AUTH_ARGS+=(-p "auth.clients:=[${MEDKIT_CLIENTS}]")
```
mfaferek93, line 54: This splices `MEDKIT_CLIENTS` into a YAML flow sequence for rcl to parse, so a secret containing `[`, `]`, `{`, `}` or ` #` makes rcl reject the args and the node never starts, and a comma (nothing documents it as the separator) splits the entry into two "Invalid client format" fragments, leaving the container closed with no client and only a WARN. Split on commas in shell and emit each entry as a double-quoted YAML scalar with `\` and `"` escaped, document the separator, and have the smoke script's closed case actually `POST /auth/token` with the given credentials.

src/ros2_medkit_gateway/launch/gateway.launch.py
```diff
@@ -120,6 +232,91 @@ def _launch_setup(context, *_args, **_kwargs):
...
+            tls_enabled = parse_security_flag('tls_enabled', tls_arg)
+            param_overrides['server.tls.enabled'] = tls_enabled
+        else:
+            # Nothing said otherwise, so the config file decides - read it
+            # rather than assume a profile.
+            tls_enabled = config_says(
+                LaunchConfiguration('config_file').perform(context),
+                'server', 'tls', 'enabled')
+        cert_file = (LaunchConfiguration('cert_file').perform(context)
+                     or os.environ.get('MEDKIT_TLS_CERT_FILE', ''))
```
mfaferek93, line 252: `MEDKIT_TLS_CERT_FILE` and `MEDKIT_TLS_KEY_FILE` exist only on these two lines: docker.rst states the environment contract as exactly `MEDKIT_AUTH_DISABLED` / `MEDKIT_JWT_SECRET` / `MEDKIT_CLIENTS`, https.rst documents only the `cert_file:=`/`key_file:=` args, the entrypoint does not export them and no test covers them. They land in `param_overrides`, so they silently beat a config file's `server.tls.*`, and setting one of the pair gives a mismatched cert and key. Either document them in the precedence rule and the entrypoint, or drop the env fallback.

src/ros2_medkit_integration_tests/test/features/test_closed_by_default.test.py
```diff
@@ -0,0 +1,696 @@
...
+        """Status alone is not the contract; the body and the challenge are.
+
+        The refusal an exhausted anonymous caller gets is produced on a
+        different line from the one an unexhausted caller gets, so nothing
+        stops the two drifting apart. A client that reads the error document to
+        tell a missing credential from an expired one gets nothing to read from
+        a bare 401 - and a 401 that is visibly a different shape is itself the
+        disclosure that the limiter, not the credential, decided it.
+        """
+        fresh = requests.get(f'{RL_BASE_URL}/areas', timeout=15)
```
mfaferek93, line 655: `fresh` is already exhausted here: test_01 and test_02 have pushed 15 requests through this gateway's 5/min token bucket (refill is one request per 12 s, `rate_limiter.cpp:99`), so both samples come from the same state and the keys comparison can never catch the two refusal paths drifting apart. Take the unexhausted sample before the exhaustion loop, e.g. stash the anonymous response from test_01 on the class and compare against that.

src/ros2_medkit_gateway/src/core/auth/auth_manager.cpp
bburda: I chose the denylist (c1942922, auth_manager.cpp). An access token is checked on its signature and the local client table. So a forwarded token works on a peer that never issued it, and a token survives a restart. `POST /auth/revoke` takes a refresh token and writes a revoked record keyed by its id. It does this on whichever gateway gets the call, also for a token that gateway never issued and for a refresh token past its expiry. The record is held until the last access token minted from it can expire, and a foreign record is capped at this gateway's own refresh lifetime. `test_forward_auth_across_gateways` (8a0e58c9) runs two gateways under forward_auth: a token from one works on the other, is refused there after a revoke on the peer, and grants the peer's role. authentication.rst and aggregation.rst (e11ec25e) say what survives a restart and that peers must share both expiries.

src/ros2_medkit_gateway/launch/gateway.launch.py
bburda: Fixed in c1942922. gateway.launch.py no longer applies the environment rule. The node reads MEDKIT_JWT_SECRET, MEDKIT_CLIENTS and MEDKIT_AUTH_DISABLED itself (`resolve_auth_environment`, auth_environment.cpp). A non-empty MEDKIT_JWT_SECRET turns auth on with require_auth_for "all", whatever `auth_enabled:=` or the file says. MEDKIT_AUTH_DISABLED=1 wins over everything, `auth_enabled:=true` included. The launch file only prints a notice when the two disagree. docker.rst (e11ec25e) states this precedence.

docker/entrypoint.sh
bburda: Done in c1942922 and 90fab0ec. The rule now lives in the node, which reads the three variables when it loads its parameters. So the entrypoint's gateway, `ros2 run ros2_medkit_gateway gateway_node`, gateway_https.launch.py and every other launch file follow it. The entrypoint only exports the variables and prints nothing about whether authentication is on. The node logs at WARN on start whenever the environment overrides a setting. `ros2 param get auth.enabled` and `auth.require_auth_for` report what is enforced, the secret parameters read back as placeholders in every auth state, and setting `auth.*` at runtime is refused. `test_env_auth_contract` (8a0e58c9) drives this against the node.

src/ros2_medkit_gateway/src/http/rest_server.cpp
bburda: Changed in c1942922 (rest_server.cpp). The limiter still counts every request before authentication. A caller with no Authorization header on a protected route keeps the standard 401. A caller over the limit who sends any Authorization header on a protected route gets a bare 429 before the token is parsed or verified, with no X-RateLimit-* headers. Those headers reach only a caller who was accepted, or a route that needs no credential. The RS256 keys are read once in the AuthManager constructor. `TheVerifierIsNotReachedOnTheExhaustedHeaderPath` checks the verifier count on that path, and `test_04_an_exhausted_caller_with_a_header_gets_a_bare_429` (8a0e58c9) drives a running gateway.

src/ros2_medkit_gateway/src/http/handlers/health_handlers.cpp
bburda: Fixed in c1942922. The cut now needs auth on and the route listed in `auth.public_routes`, checked through `AuthManager::is_public_route`, the same policy the router applies (`serves_reduced_body` in health_handlers.cpp). Under require_auth_for "write" an anonymous `GET /health` keeps every section. `TestWriteModeHealthIsNotCutDown` (8a0e58c9) runs that mode against a gateway, and server.rst describes the reduced body only for a listed route.

Dockerfile
bburda: Changed in 90fab0ec. The image carries three params files, and the entrypoint passes them in this order: /etc/ros2_medkit/base.yaml (the packaged gateway_params.yaml), /etc/ros2_medkit/container.yaml (server.host 0.0.0.0 and a 2 s refresh), then the mount point /etc/ros2_medkit/params.yaml. The caller's arguments come after them. A mounted file can therefore set server.host and refresh_interval_ms again, and `docker run <image> --ros-args -p server.port:=9090` keeps server.host at 0.0.0.0. The Dockerfile CMD is empty. docker.rst (e11ec25e) lists the layers and their order, and scripts/smoke_image_auth_posture.sh checks a mounted value and the arg-only override on the built image.

docs/tutorials/docker.rst
bburda: Fixed in e11ec25e. The compose example in docker.rst mounts gateway-cors.yaml at /etc/ros2_medkit/params.yaml, and that file names http://localhost:3000 in cors.allowed_origins. The PR is marked as a breaking change, and its Breaking changes section lists the dropped localhost:3000 and localhost:5173 origins. The token part no longer applies: access tokens now survive a restart (see the auth_manager.cpp thread), and the PR text says a revoked token can come back for at most token_expiry_seconds after a restart.

docker/entrypoint.sh
bburda: The YAML splice is gone (c1942922, 90fab0ec). The node parses MEDKIT_CLIENTS itself (auth_environment.cpp). Entries are separated by commas and each is `id:secret:role`. Surrounding spaces and tabs are trimmed. A duplicate id keeps its first entry and logs a WARN. A value whose every entry is refused stops the gateway. docker.rst documents the comma separator and says a comma inside a secret breaks the entry. The smoke script's closed case now posts the MEDKIT_CLIENTS credential to /auth/authorize and reads /areas with the token it gets.

src/ros2_medkit_gateway/launch/gateway.launch.py
bburda: Removed in c1942922. gateway.launch.py no longer reads MEDKIT_TLS_CERT_FILE or MEDKIT_TLS_KEY_FILE, and nothing in the repository names them now. The environment contract is the three variables docker.rst lists. TLS files come from `cert_file:=` and `key_file:=` or from the config file.

src/ros2_medkit_integration_tests/test/features/test_closed_by_default.test.py
bburda: Fixed in 8a0e58c9. test_01 stashes its anonymous refusal on the class while the allowance is still there, and asserts X-RateLimit-Remaining is above zero at that moment. test_03 compares the exhausted refusal against that stash.

PR #643, opened 2026-08-27
fault_manager: make a time-based confirmation as visible as a reported one

### LINKED ISSUE #642: Debounce docs recommend a setting that silently disables faults for one-event-per-transition reporters
## Summary

The debounce documentation recommends a configuration that quietly does nothing for a whole class
of fault reporters, and the parameter that does work for them is documented but never connected to
the case.

The debounce counter only moves when an event arrives. `confirmation_threshold` and
`healing_threshold` therefore assume a reporter that keeps sending FAILED while the condition is
still there, and keeps sending PASSED after it goes away.

Many reporters do not work that way. They send one FAILED when a condition appears and one clear
when it goes away, and nothing in between. The README section "Advanced: Debounce Filtering"
recommends this to everyone:

```bash
-p confirmation_threshold:=-3 -p healing_enabled:=true -p healing_threshold:=3
```

For a reporter of the second kind that gives:

- The second FAILED never arrives, so the counter stops at -1 and the fault stays PREFAILED. It
  never confirms.
- `ListFaults` with an empty `statuses` array returns CONFIRMED only, so nobody sees the fault at
  all.
- Healing needs `healing_threshold - confirmation_threshold` consecutive PASSED events. Only one
  PASSED is ever sent, so a fault that did confirm stays CONFIRMED until someone calls
  `~/clear_fault`.

So the recommended setting turns a working fault path into a silent one, and nothing in the docs
warns about it.

`auto_confirm_after_sec` already solves this. It promotes a PREFAILED fault to CONFIRMED after a
timeout without touching the counter, which is exactly the filter such a reporter needs: hold the
first FAILED in PREFAILED, and confirm it only if it is still there when the window closes. The
docs list the parameter in the table but never say this is what it is for.

There is also no test on this path. No test covers `auto_confirm_after_sec`, and no test covers
`healing_threshold: 0`, which is the only value that lets a single PASSED event heal a fault.

## Proposed solution

Documentation, in the package README and in `docs/config/fault-manager.rst`:

- Say that the count-based settings need a reporter that repeats its events.
- Say what happens when they are used with a reporter that sends one event per transition.
- Point at `auto_confirm_after_sec` plus `healing_threshold: 0` as the pair for that case, and say
  the timeout should come from how often the reporter samples.

Test, in `ros2_medkit_fault_manager`:

An integration test driving the real node over the real services, using the event counts a
one-event-per-transition reporter actually sends. It should pin:

- one FAILED does not confirm
- an unconfirmed fault is not in the default fault list
- a condition that stays raises CONFIRMED after `auto_confirm_after_sec` with nobody acting
- a condition that clears inside the window never confirms
- one PASSED heals a confirmed fault with nobody acting

The fourth one matters most. Without it a config could pass by only delaying a false alarm instead
of filtering it.

## Additional context

No code change is needed. All four parameters already behave correctly, and both storage backends
agree on the arithmetic. This is a documentation gap plus a test gap.

bburda:
# Pull Request

## Summary

The debounce counter only moves when an event arrives, so `confirmation_threshold` and
`healing_threshold` are tunable only for a reporter that keeps sending events while a condition
holds. For a reporter that sends one FAILED when a condition appears and one clear when it goes
away, the second event never comes: confirmation stalls in PREFAILED, and healing needs
`healing_threshold - confirmation_threshold` consecutive PASSED events when only one is sent. The
docs recommended `-3` with `healing_threshold: 3` to everyone, with no mention of this.

Writing that down turned up four real defects on the same surface, all fixed here.

**A time-based confirmation was invisible.** The auto-confirm timer only wrote to the database and
the audit log. The report path also publishes `EVENT_CONFIRMED` and enqueues snapshot + rosbag
capture, so the same fault produced an event and a recording when confirmed by a report, and
neither when confirmed by the timer. Anything driven off the event stream saw no alarm. The capture
block is now a helper shared by both paths so they cannot drift apart again.

**`auto_confirm_after_sec` accepted NaN.** Every comparison against NaN is false, so the `< 0.0`
guard let it through and the `> 0.0` timer guard then rejected it, disabling time-based
confirmation with nothing logged at all. A negative value at least warned.

**The startup line reported the requested thresholds, not the effective ones.** A value the
sanitizer replaced was logged as if it had been accepted, so the log described a configuration that
was not running. Related: the configuration guide recommended `confirmation_threshold: 0`, which
the node rejects and replaces with `-1`.

**`bringup_params.yaml` never healed.** It paired `healing_enabled: true` with
`healing_threshold: 3` under a comment promising a fault heals when its action recovers. The action
bridge emits exactly one PASSED per recovery, and its own README already says healing needs
threshold 0.

---

## Issue

- closes #642

---

## Type

- [x] Bug fix
- [x] New feature or tests
- [ ] Breaking change
- [x] Documentation only

---

## Testing

Two integration tests, both driving the real node over the real services with SQLite storage and
the event counts a one-event-per-transition reporter actually sends.

`test_auto_confirm_visibility` subscribes to the event topic and asserts a timer-driven
confirmation publishes `EVENT_CONFIRMED`. Reverting the fix fails it with the defect in the message:

```
AssertionError: 'fault_confirmed' not found in []
  : a time-based confirmation published no event, so nothing downstream of it is told
```

`test_debounce_and_healing` is parametrized over `healing_threshold` so that 0 is shown to be what
makes healing reachable rather than asserted to be. The run at the default of 3 asserts the fault
latches instead of skipping, and a re-raise case pins that a healed fault confirms again when the
condition returns. Setting `healing_enabled: false` fails three cases in the `[0]` variant and none
in `[3]`, which is where the discriminating power should sit.

The NaN and startup-log fixes were checked by running the node at `.nan`, at `-1.0` and at
`confirmation_threshold: 0`, before and after.

Package suite: 761 tests, 0 errors, 0 failures, 37 skipped. Build clean, no warnings.

---

## Checklist

### DIFF
--- a/docs/config/fault-manager.rst
+++ b/docs/config/fault-manager.rst
@@ -75,6 +75,32 @@ The fault manager uses AUTOSAR DEM-style debounce filtering to prevent fault fla
    For immediate fault confirmation (no debounce), set ``confirmation_threshold: 0``.
    Faults with ``SEVERITY_CRITICAL`` always bypass debounce regardless of this setting.
 
+.. important::
+
+   The counter only moves when an event arrives, so ``confirmation_threshold`` and
+   ``healing_threshold`` work only for a reporter that keeps sending FAILED while the condition
+   is still there.
+
+   A reporter that sends one FAILED when a condition appears and one clear when it goes away
+   never sends the second event. ``confirmation_threshold: -3`` then leaves the fault in
+   PREFAILED forever, and the default fault list returns CONFIRMED only, so the fault is never
+   seen. ``healing_threshold: 3`` has the same problem: healing needs
+   ``healing_threshold - confirmation_threshold`` consecutive PASSED events, and only one is
+   sent, so the fault stays CONFIRMED until someone calls ``~/clear_fault``.
+
+   For such a reporter, filter by time instead of by count:
+
+   .. code-block:: yaml
+
+      confirmation_threshold: -2      # first FAILED stays PREFAILED
+      auto_confirm_after_sec: 3.0     # confirm it if it is still there after 3 s
+      healing_enabled: true
+      healing_threshold: 0            # heal on the single PASSED
+
+   Choose ``auto_confirm_after_sec`` from how often the reporter samples, so a condition has to
+   survive a few sampling cycles before it confirms. A glitch that clears in time never reaches
+   CONFIRMED, because the clear takes the fault out of PREFAILED before the timer fires.
+
 Near-Miss Retention
 ~~~~~~~~~~~~~~~~~~~
 
--- a/docs/config/fault-manager.rst
+++ b/docs/config/fault-manager.rst
@@ -65,7 +65,9 @@ The fault manager uses AUTOSAR DEM-style debounce filtering to prevent fault fla
      - When true, PASSED events can heal confirmed faults.
    * - ``healing_threshold``
      - ``3``
-     - Number of PASSED events to transition from CONFIRMED to HEALED.
+     - Counter value at which a fault heals, not a number of events. Healing costs
+       ``healing_threshold`` minus the counter the fault confirmed at, so with the
+       default ``-1`` it takes four PASSED events.
    * - ``auto_confirm_after_sec``
      - ``0.0``
      - Auto-confirm prefailed faults after this duration. Set to 0 to disable.
@@ -560,7 +562,7 @@ by default: with it off there is no table, no file and no write cost.
      - Turn the audit log on.
    * - ``audit_log.transitions``
      - ``"all"``
-     - Which transitions are recorded: ``all`` (occurred, confirmed, cleared) or
+     - Which transitions are recorded: ``all`` (occurred, confirmed, healed, cleared) or
        ``confirmed_only``. Any other value falls back to ``all`` with a warning.
    * - ``audit_log.retention_max_records``
      - ``0``
@@ -623,11 +625,13 @@ Complete Example
        storage_type: "sqlite"
        database_path: "/var/lib/ros2_medkit/faults.db"
 
-       # Debounce (require 3 FAILED events to confirm)
+       # Debounce for a reporter that repeats its events while a condition holds:
+       # three FAILED events confirm, and four PASSED events heal from there.
+       # For a reporter that sends one event per transition, use -1 with
+       # healing_threshold 0 instead - see the note under Debounce Settings.
        confirmation_threshold: -3
        healing_enabled: true
        healing_threshold: 3
-       auto_confirm_after_sec: 30.0
 
        # Per-entity debounce overrides
        entity_thresholds:
--- a/docs/config/fault-manager.rst
+++ b/docs/config/fault-manager.rst
@@ -72,34 +72,38 @@ The fault manager uses AUTOSAR DEM-style debounce filtering to prevent fault fla
 
 .. tip::
 
-   For immediate fault confirmation (no debounce), set ``confirmation_threshold: 0``.
+   For immediate fault confirmation (no debounce), set ``confirmation_threshold: -1``,
+   which is also the default. ``0`` is rejected: the threshold must be strictly
+   negative, and the node falls back to ``-1`` with a warning.
    Faults with ``SEVERITY_CRITICAL`` always bypass debounce regardless of this setting.
 
 .. important::
 
    The counter only moves when an event arrives, so ``confirmation_threshold`` and
-   ``healing_threshold`` work only for a reporter that keeps sending FAILED while the condition
-   is still there.
+   ``healing_threshold`` are tunable only for a reporter that keeps sending events while a
+   condition holds. Confirmation needs repeated FAILED; healing needs repeated PASSED.
 
    A reporter that sends one FAILED when a condition appears and one clear when it goes away
    never sends the second event. ``confirmation_threshold: -3`` then leaves the fault in
-   PREFAILED forever, and the default fault list returns CONFIRMED only, so the fault is never
-   seen. ``healing_threshold: 3`` has the same problem: healing needs
-   ``healing_threshold - confirmation_threshold`` consecutive PASSED events, and only one is
-   sent, so the fault stays CONFIRMED until someone calls ``~/clear_fault``.
+   PREFAILED, and the default fault list returns CONFIRMED only, so the fault is never seen.
+   Healing has the same shape: it needs ``healing_threshold - confirmation_threshold``
+   consecutive PASSED events counted from where the fault confirmed, and only one is sent.
 
-   For such a reporter, filter by time instead of by count:
+   For such a reporter, leave the confirmation threshold alone and make healing reachable:
 
    .. code-block:: yaml
 
-      confirmation_threshold: -2      # first FAILED stays PREFAILED
-      auto_confirm_after_sec: 3.0     # confirm it if it is still there after 3 s
-      healing_enabled: true
-      healing_threshold: 0            # heal on the single PASSED
+      fault_manager:
+        ros__parameters:
+          confirmation_threshold: -1    # the default; the single FAILED confirms
+          healing_enabled: true
+          healing_threshold: 0          # heal on the single PASSED
 
-   Choose ``auto_confirm_after_sec`` from how often the reporter samples, so a condition has to
-   survive a few sampling cycles before it confirms. A glitch that clears in time never reaches
-   CONFIRMED, because the clear takes the fault out of PREFAILED before the timer fires.
+   ``auto_confirm_after_sec`` promotes a fault that stayed PREFAILED for that long and looks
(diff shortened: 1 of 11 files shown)


bburda:
A downstream change sets these parameters in shipped configuration files and depends on this PR. It should not land before this one, or those files would assert a behaviour nothing proves here.

src/ros2_medkit_fault_manager/src/fault_manager_node.cpp
```diff
@@ -431,18 +437,24 @@ FaultManagerNode::FaultManagerNode(const rclcpp::NodeOptions & options) : Node("
         auto fault = storage_->get_fault(fault_code);
         if (fault) {
           audit_transition(kTransitionConfirmed, *fault, "auto_confirm_timer", confirmed_at_ns);
+          // A timer-driven confirmation is a confirmation: it has to reach the
+          // event stream and the black box exactly like one raised by a report,
+          // or subscribers see no alarm and no recording is ever made for it.
+          publish_fault_event(ros2_medkit_msgs::msg::FaultEvent::EVENT_CONFIRMED, *fault);
```
mfaferek93: The timer path publishes for muted symptoms. The report path gates every publish with `if (!should_mute)` (817, 824, 831); here `publish_fault_event(EVENT_CONFIRMED, *fault)` runs unconditionally. With a `mute_symptoms` rule and `confirmation_threshold: -2` the symptom sits muted in PREFAILED, the timer confirms it, and SSE / trigger subscriber / entity freeze-frame get a CONFIRMED alarm for a fault `handle_list_faults` hides.

Suggest skipping the publish when the engine has the code muted (an `is_muted(code)` on the engine, or `get_muted_faults()`), and keeping `capture_on_confirm` unconditional so it matches line 869.

src/ros2_medkit_fault_manager/src/fault_manager_node.cpp
```diff
@@ -126,8 +126,14 @@ FaultManagerNode::FaultManagerNode(const rclcpp::NodeOptions & options) : Node("
...
   // Time-based auto-confirmation parameter
   auto_confirm_after_sec_ = declare_parameter<double>("auto_confirm_after_sec", 0.0);
-  if (auto_confirm_after_sec_ < 0.0) {
-    RCLCPP_WARN(get_logger(), "auto_confirm_after_sec should be >= 0, got %.2f. Disabling.", auto_confirm_after_sec_);
+  // Positive test, then negated: every comparison against NaN is false, so a
+  // plain `< 0.0` accepts NaN and the timer is then never created either, which
+  // disables time-based confirmation with nothing logged. clang-tidy's
+  // readability-simplify-boolean-expr suggests the DeMorgan rewrite that puts
+  // that back - leave this form alone.
+  if (!(std::isfinite(auto_confirm_after_sec_) && auto_confirm_after_sec_ >= 0.0)) {
```
mfaferek93: Still lets through finite values that overflow downstream: `sqlite_fault_storage.cpp:1087` does `static_cast<int64_t>(auto_confirm_after_sec * 1e9)`, so anything above ~9.2e9 s is UB and `cutoff_ns` is garbage (memory backend stays in double, so only sqlite). Cheap to reject or clamp above `INT64_MAX / 1e9` in this same check, with the same warning.

docs/config/fault-manager.rst
```diff
@@ -72,9 +72,39 @@ The fault manager uses AUTOSAR DEM-style debounce filtering to prevent fault fla
 
 .. tip::
 
-   For immediate fault confirmation (no debounce), set ``confirmation_threshold: 0``.
+   For immediate fault confirmation (no debounce), set ``confirmation_threshold: -1``,
+   which is also the default. ``0`` is rejected: the threshold must be strictly
+   negative, and the node falls back to ``-1`` with a warning.
```
mfaferek93, line 79: The node still says the opposite: the warning at `fault_manager_node.cpp:111` reads `should be <= 0 (0 or -1 = immediate confirmation)`. Someone following that log sets `0` and gets the "Invalid debounce thresholds" warning at line 241 instead. Worth changing that string to `should be < 0 (-1 = immediate confirmation)` in the same PR.

docs/config/fault-manager.rst
```diff
@@ -72,9 +72,39 @@ The fault manager uses AUTOSAR DEM-style debounce filtering to prevent fault fla
...
+.. important::
+
+   The counter only moves when an event arrives, so ``confirmation_threshold`` and
+   ``healing_threshold`` are tunable only for a reporter that keeps sending events while a
+   condition holds. Confirmation needs repeated FAILED; healing needs repeated PASSED.
+
+   A reporter that sends one FAILED when a condition appears and one clear when it goes away
+   never sends the second event. ``confirmation_threshold: -3`` then leaves the fault in
+   PREFAILED, and the default fault list returns CONFIRMED only, so the fault is never seen.
+   Healing has the same shape: it needs ``healing_threshold - confirmation_threshold``
```
mfaferek93, line 91: The parameter table above (line 68) still says `healing_threshold` is "Number of PASSED events to transition from CONFIRMED to HEALED", which with the defaults `-1`/`3` is four events, as this box says. README.md:68 already has the counter wording; the table should match.

docs/config/fault-manager.rst
```diff
@@ -72,9 +72,39 @@ The fault manager uses AUTOSAR DEM-style debounce filtering to prevent fault fla
...
+   .. code-block:: yaml
+
+      fault_manager:
+        ros__parameters:
+          confirmation_threshold: -1    # the default; the single FAILED confirms
+          healing_enabled: true
+          healing_threshold: 0          # heal on the single PASSED
+
+   ``auto_confirm_after_sec`` promotes a fault that stayed PREFAILED for that long and looks
+   like it would allow a deeper threshold. It does not: HEALED is latched, leaving that latch
```
mfaferek93, line 105: The "Complete Example" at the bottom of this page (lines 627-630) pairs `confirmation_threshold: -3`, `healing_threshold: 3` and `auto_confirm_after_sec: 30.0`, which is exactly the combination this box warns about. Either drop `auto_confirm_after_sec` from the example or note it is for a repeating reporter.

src/ros2_medkit_fault_manager/src/fault_manager_node.cpp
bburda: Fixed, and the gate now has a test. The timer asks the correlation engine before it publishes, through a new `is_muted` on the engine. Capture stays ungated as you suggested, since `just_confirmed` on the report path is set regardless of muting.

`test_04_a_muted_symptom_is_confirmed_but_not_announced` reports `ESTOP_001` and then a symptom inside the rule window, and asserts three things: the root cause is announced, the symptom is not, and the symptom is still CONFIRMED in the store with `include_muted`. The third one is what makes the other two mean something, without it the case would also pass if the timer had simply never confirmed the symptom. Removing the gate fails it and prints the symptom's own `fault_confirmed` event.

src/ros2_medkit_fault_manager/src/fault_manager_node.cpp
bburda: Fixed. The check now also rejects anything above 9.0e9 seconds, so the product in the SQLite backend stays inside int64. Same warning, it just names the range now.

docs/config/fault-manager.rst
bburda: Changed in the same commit. The warning now reads `confirmation_threshold should be < 0 (-1 = immediate confirmation)`, so following the log no longer leads to a value the sanitizer rejects.

docs/config/fault-manager.rst
bburda: Changed. The row now says `healing_threshold` is a counter target, not a number of events, and gives the cost with the default: four PASSED events at `-1`.

docs/config/fault-manager.rst
bburda: Dropped `auto_confirm_after_sec` from that example and labelled the rest as being for a reporter that repeats its events, with a pointer to the note above for the other kind.

src/ros2_medkit_fault_manager/test/integration/test_debounce_and_healing.test.py
```diff
@@ -0,0 +1,249 @@
...
+
+
+@launch_testing.post_shutdown_test()
+class TestDebounceAndHealingShutdown(unittest.TestCase):
+    """Check the node exited cleanly."""
+
+    def test_exit_code(self, proc_info, fault_manager_node):
+        launch_testing.asserts.assertExitCodes(
+            proc_info, allowable_exit_codes=[0, -2, -15], process=fault_manager_node
+        )
```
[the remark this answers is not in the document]
bburda: Both were already changed before this comment landed: the exit codes are back to the default `[0]` and there is a cleanup test that asserts the directory is gone rather than ignoring errors.

mfaferek93 (approved):

PR #648, opened 2026-08-31
ci: make a CI failure legible - crash frames, budget overruns, shared memory, and a coverage job that fits its cap

bburda:
# Pull Request

## Summary

Everything here came out of one investigation: a lyrical job failed, and the failure was
unreadable. Four changes, each from a measurement.

**A node that dies leaves no trace.** `demo_brake_pressure_sensor` was killed by SIGSEGV
1.155 s into a run, having printed nothing at all - no output, and no core file, because a
container cannot set the host's `core_pattern`. The exit status alone cannot separate a
defect here from one in rclcpp, rmw or the DDS implementation. `run_demo_node()` now
installs a handler for SIGSEGV, SIGBUS and SIGABRT that writes the frames behind a
`MEDKIT-CRASH` marker and then lets the original signal terminate the process, so the exit
status a test asserts on is unchanged. Sanitizer builds keep their own handler: this one
stands down there, and the tests assert that case too rather than going quiet under it.

**A step timeout reads as a test failure.** A step killed by its cap is marked failed by the
runner, not cancelled, and the in-flight test leaves no result file; `colcon test-result`
then reports one erroring test named `<test>.xunit.missing_result` among thousands that
never ran. Over one month, 17 of 25 killed jobs died on a step cap, and every timeout in
this file has been raised at least once. A new step says so plainly and warns at 80% of the
cap. It detects the overrun by the clock, because a job cannot read its own step log.

**Shared memory was unmeasured and stranded.** Container jobs got Docker's default 64 MB of
`/dev/shm`. Fast DDS puts a 512 KB segment plus port files there per participant, measured
at ~0.65 MB, and a participant killed rather than shut down never gets them back: 15 live
participants take 9.6 MB, and 125 killed ones fill 63 MB of 64. Jobs now ask for 1 GB and
record what they used; `scripts/test.sh` reclaims stranded segments locally through
`fastdds shm clean`. The launch-domain probe child, which died from SIGINT with its
participant never destroyed, now shuts down on both SIGINT and SIGTERM, and its test
asserts an exit status of zero instead of accepting any non-crash.

**The coverage job was twelve seconds from failing.** 89m48s against a 90 minute cap on
2026-08-27, with four of its last five runs on main between 86 and 90 minutes, and a test
step that grew 166% in a month. The cause is one line: `build-and-test` and `jazzy-test`
both skip `ros2_medkit_graph_watchdog`, whose suite is 24 minutes, and `coverage` skips
only `ros2_medkit_opcua`. It is now sharded four ways by measured test time per package,
with a merge job carrying the completeness gate, the HTML report and the Codecov upload,
because a shard's tracefile is legitimately partial. `fail-fast` is off deliberately: with
it, one failing shard cancels the others, and a cancelled job cannot be told apart from one
killed by its cap.

---

## Issue

No issue filed - this came out of investigating a CI failure rather than from a report.
Happy to open one if the history is worth having separately.

---

## Type

- [ ] Bug fix
- [x] New feature or tests
- [ ] Breaking change
- [ ] Documentation only

---

## Testing

The sharded coverage job only runs on push to main, so it could not execute before landing
there. Two temporary commits widened that condition, the job ran on this branch, and the
condition is restored - the net diff against main carries no temporary condition. Results
from run 33426859082, every job green:

| Claim | Evidence |
|---|---|
| `--shm-size` is accepted and takes effect | `shm 1.0G 2.1M 1022M 1% /dev/shm`, 44 segments at the end of the suite - the first time this has been measured at all |
| The shard matrix expands correctly | Four green jobs: `coverage (integration/graph-watchdog/gateway/rest)`, including the folded-YAML catch-all shard |
| Every shard produces a tracefile | `merging 4 tracefiles` |
| Merging real gcov output works | Merged report 86.9%, 34539 of 39753 lines |
| The completeness gate holds on the merged report | `OK: all 15 checked C++ package(s) include ROS2MedkitCoverage and reached coverage.info` |
| The overrun guard runs and measures | `test step: 1573s elapsed of a 2700s cap (58%); results: 285 written, 0 empty, 0 missing_result` |
| Codecov still receives one report | Upload queued, `outcome=success` |

**Sharding loses no coverage.** The last unsharded run on main measured 86.9% at 34537 of
39753 lines; the sharded run measures 86.9% at 34539 of 39753. Identical denominator, so
the same instrumented line universe.

**Wall clock.** Worst shard `graph-watchdog` 3046s, then `integration` 2663s, `gateway`
2041s, `rest` 1462s. Against 86-90 minutes serial and a 90 minute cap, the margin goes from
twelve seconds to about 39 minutes. The cap stays at 90 here: those are cold-cache numbers,
and the figure to lower it by should come from a few runs of the margin line.

Verified outside CI:

- The crash reporter: 5/5 death tests on jazzy at `-O2`, 5/5 in a lyrical/resolute
  container, and 5/5 under `-fsanitize=address` where the marker correctly never appears.
- The domain-probe fix: the launch test passes, and reverting the child to a bare
  `rclpy.spin` makes it fail, so the assertion discriminates.
- The sweep: eight participants killed with SIGKILL stranded 5.24 MB across 66 files in a
  lyrical container; `fastdds shm clean` returned 5.23 MB of it.
- The overrun guard: exercised at 4%, 85% and 100% of a cap, and with no result files.
- `lcov -a`: checked on two tracefiles where a line uncovered in one and covered in the
  other merges to covered, with counts summing.

Two of these tests were wrong first, which is worth recording. The crash test asserted a
frame format that only matches one glibc release - resolute writes `") [0x"` where noble
writes `")[0x"` - and it was written with a `volatile` qualifier on the wrong side of the
pointer, so the compiler deleted the store at `-O2` and the process did not crash at all.
The completeness guard first compared registered tests against result files and reported
missing results on a run where everything passed, because `test_dds_domain_allocation` is
registered in every package and writes no xunit. All three were caught by running the
tests, not by reading them.

---

## Checklist

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -126,7 +126,7 @@ jobs:
             --ctest-args -LE linter \
             --event-handlers console_direct+
 
-      - name: Check the run completed, and record the margin
+      - name: Check the run completed, and record the environment
         if: always()
         run: |
           # A step killed by its cap is reported as a failure, not a cancellation,
@@ -134,8 +134,14 @@ jobs:
           # overrun reads as one flaky test out of thousands, and the packages
           # queued behind it leave no trace at all.
           ./scripts/ci_test_completeness.sh "${TEST_STEP_START}" 45
+          # A DDS participant killed rather than shut down leaves its segments
+          # behind for the rest of the job, and the installed middleware decides
+          # which transport those segments belong to. Neither is recoverable from
+          # a finished run, so both are recorded rather than assumed.
           df -h /dev/shm
           echo "shm segments: $(ls /dev/shm | wc -l)"
+          ls -la /dev/shm | head -20
+          dpkg -l | grep -E 'rmw-fastrtps|fastdds|cyclonedds' || true
 
       - name: Show test results
         if: always()
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -30,6 +30,10 @@ jobs:
     # run before that step's cap can - which loses the "which step ran long" answer. The build
     # side of this job is around 18 minutes.
     timeout-minutes: 90
+    env:
+      # One source for the test step's cap. The step's timeout-minutes and the
+      # guard that reports the margin both read this.
+      TEST_STEP_CAP_MINUTES: 45
     defaults:
       run:
         shell: bash
@@ -105,7 +109,9 @@ jobs:
         # arrived (lyrical 22m50s before them, jazzy 28m15s after). The jazzy job below caps the
         # whole job instead of this step, so only these two distros can be killed mid-package -
         # which reads as a test failure and hides every test the kill cut off.
-        timeout-minutes: 45
+        # The cap lives in the job's env, referenced here and by the guard below,
+        # so raising one raises the other and they cannot drift apart.
+        timeout-minutes: ${{ env.TEST_STEP_CAP_MINUTES }}
         env:
           # FastRTPS 2.6 on Humble has a known use-after-free in the
           # discovery-teardown path (EDP::unpairWriterProxy) that segfaults
@@ -133,7 +139,14 @@ jobs:
           # and colcon then names whichever test was in flight. Without this, the
           # overrun reads as one flaky test out of thousands, and the packages
           # queued behind it leave no trace at all.
-          ./scripts/ci_test_completeness.sh "${TEST_STEP_START}" 45
+          # The start is missing only when an earlier step failed before it ran.
+          # Say so and carry on to the census, rather than dying on a usage error
+          # and taking the environment evidence down with it.
+          if [ -n "${TEST_STEP_START:-}" ]; then
+            ./scripts/ci_test_completeness.sh "${TEST_STEP_START}" "${TEST_STEP_CAP_MINUTES}" || guard_rc=$?
+          else
+            echo "::notice::no test step start was recorded - the run failed before the tests"
+          fi
           # A DDS participant killed rather than shut down leaves its segments
           # behind for the rest of the job, and the installed middleware decides
           # which transport those segments belong to. Neither is recoverable from
@@ -142,6 +155,7 @@ jobs:
           echo "shm segments: $(ls /dev/shm | wc -l)"
           ls -la /dev/shm | head -20
           dpkg -l | grep -E 'rmw-fastrtps|fastdds|cyclonedds' || true
+          exit "${guard_rc:-0}"
 
       - name: Show test results
         if: always()
@@ -499,9 +513,13 @@ jobs:
         uses: actions/cache@v4
         with:
           path: /root/.cache/ccache
-          key: ccache-coverage-${{ matrix.shard }}-${{ github.sha }}
+          # One entry for the whole matrix, not one per shard: every shard builds
+          # the same workspace with the same flags, so per-shard entries would
+          # quadruple this cache against a 10 GB repository limit and evict the
+          # very entries that keep the builds warm. Concurrent shards racing to
+          # save the same key is benign - the first wins, the rest log it.
+          key: ccache-coverage-${{ github.sha }}
           restore-keys: |
-            ccache-coverage-${{ matrix.shard }}-
             ccache-coverage-
 
       - name: Install dependencies
@@ -597,7 +615,11 @@ jobs:
   # against four partial answers.
   coverage-merge:
     name: coverage (merge)
-    if: github.event_name == 'push' && github.ref == 'refs/heads/main'
+    # always(), so a failed shard does not take the merged report, the
+    # completeness gate and the Codecov upload down with it. Losing the
+    # diagnostic exactly when a shard broke is the wrong way round; the job
+    # reports what it merged and then fails on the shard's behalf.
+    if: always() && github.event_name == 'push' && github.ref == 'refs/heads/main'
     needs: coverage
     runs-on: ubuntu-latest
     container:
@@ -623,6 +645,9 @@ jobs:
         with:
           pattern: coverage-tracefile-*
           path: tracefiles
+          # A shard that failed uploaded nothing; merge what exists rather than
+          # failing here, so the report still describes the shards that ran.
+          merge-multiple: false
 
       - name: Merge the tracefiles
         run: |
@@ -658,7 +683,17 @@ jobs:
           name: coverage-report
           path: coverage_html/
 
+      # Last, so everything above is recorded first. The merged report is still
+      # worth having when a shard broke - it just does not describe the whole
+      # workspace, and saying so is this job's business.
+      - name: Fail if any shard did not succeed
+        if: needs.coverage.result != 'success'
+        run: |
+          echo "::error::a coverage shard did not succeed (matrix result: ${{ needs.coverage.result }}). The merged report above covers only the shards that finished, so its percentage is a floor, not the workspace figure."
+          exit 1
+
       - name: Upload coverage to Codecov
+        if: needs.coverage.result == 'success'
         uses: codecov/codecov-action@v7
         with:
           token: ${{ secrets.CODECOV_TOKEN }}
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -605,8 +605,9 @@ jobs:
       - name: Install tools
         run: |
           apt-get update
-          # gpg is required by the codecov action's dependency check.
-          apt-get install -y git lcov gpg
+          # gpg and curl are both required by the codecov action's dependency
+          # check, which fails the step rather than skipping the upload.
+          apt-get install -y git lcov gpg curl
 
       - name: Checkout repository
         uses: actions/checkout@v4
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -19,6 +19,13 @@ jobs:
             os_image: ubuntu:resolute
     container:
       image: ${{ matrix.os_image }}
+      # A container gets 64 MB of /dev/shm by default, which is Docker's number and
+      # not a considered one. Fast DDS puts a 512 KB segment plus port files there
+      # per participant - measured at ~0.65 MB - and a participant killed rather
+      # than shut down never gets its segments back. Measured: 15 live
+      # participants take 9.6 MB, and 125 killed ones fill 63 MB. 1g is tmpfs, so
+      # it costs only what is used, against 16 GB of runner memory.
+      options: --shm-size=1g
     # Must stay above the build plus the test step's own budget below, or the job cap kills the
     # run before that step's cap can - which loses the "which step ran long" answer. The build
     # side of this job is around 18 minutes.
@@ -88,6 +95,9 @@ jobs:
           ccache -s
           ./scripts/ccache_report.sh "${{ matrix.ros_distro }}"
 
+      - name: Record the test step start
+        run: echo "TEST_STEP_START=$(date +%s)" >> "$GITHUB_ENV"
+
       - name: Run unit and integration tests
         # The suite keeps outgrowing this budget as packages land, and every overrun so far has
         # been real work finishing rather than a hang: 15 minutes was raised to 25 when lyrical
@@ -115,6 +125,17 @@ jobs:
             --packages-skip ros2_medkit_opcua ros2_medkit_graph_watchdog \
             --ctest-args -LE linter \
             --event-handlers console_direct+
+
+      - name: Check the run completed, and record the margin
+        if: always()
+        run: |
+          # A step killed by its cap is reported as a failure, not a cancellation,
+          # and colcon then names whichever test was in flight. Without this, the
+          # overrun reads as one flaky test out of thousands, and the packages
+          # queued behind it leave no trace at all.
+          ./scripts/ci_test_completeness.sh "${TEST_STEP_START}" 45
+          df -h /dev/shm
+          echo "shm segments: $(ls /dev/shm | wc -l)"
 
       - name: Show test results
         if: always()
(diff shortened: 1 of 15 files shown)


.github/workflows/ci.yml
```diff
@@ -455,10 +564,24 @@ jobs:
...
             --ctest-args -LE linter \
             --event-handlers console_direct+
 
+          # colcon WARNS and exits 0 on a --packages-select name that does not
+          # exist, so a rename would leave this shard testing nothing while the
+          # merged report still looked complete - another shard's tests execute
+          # the same installed code and put it back in the report. Name-aware,
+          # because a shard selecting two packages still produces results when
+          # only one of the names rotted.
+          for pkg in ${{ matrix.select }}; do
```
mfaferek93: This loop walks every word of `matrix.select`, and for the `rest` shard that is `--packages-skip ros2_medkit_opcua ros2_medkit_integration_tests ...`. First non-flag word is `ros2_medkit_opcua`, which this job never builds, so the `find` is empty and the step exits 1 (the four skipped packages fail the same way: built, not tested here). This landed in the second-pass commit after the green sharded run in the description, and PR CI never runs this job, so the first push to main after merge fails `coverage (rest)` and the merge job fails with it.

Run the loop only when the flag is `--packages-select`, or add a separate `check:` matrix key with the packages to verify.

src/ros2_medkit_integration_tests/CMakeLists.txt
```diff
@@ -206,6 +206,15 @@ if(BUILD_TESTING)
   set_tests_properties(test_test_utils_constants PROPERTIES LABELS "unit")
   medkit_test_needs_no_domain(test_test_utils_constants)
 
+  # The crash reporter every demo binary installs. Death tests, because the
+  # behaviour under test only exists in a process that is being killed - a
+  # handler that is merely installed proves nothing about what a crash leaves
+  # behind. ROS-free: it creates no node, only signals.
+  find_package(ament_cmake_gmock REQUIRED)
+  medkit_add_gmock(test_crash_backtrace test/test_crash_backtrace.cpp)
```
mfaferek93, line 215: The sanitizer stand-down is never exercised by CI: sanitizers are opt-in per package via `include(ROS2MedkitSanitizers)` (seven packages do it), and this package includes only Compat/Linting/Warnings at lines 11-13. So `-DSANITIZER=asan,ubsan`/`tsan` in quality.yml is an unused variable here, `__SANITIZE_ADDRESS__` is never defined, the handler stays active in the ASan/TSan jobs and every `else` branch of the new test is dead. Either include the module (and accept instrumented demo nodes) or drop the "tests assert that case too" claim.

.github/workflows/ci.yml
```diff
@@ -496,12 +619,90 @@ jobs:
...
 
       # A package that never gets --coverage emits no .gcda and drops out of the
       # report entirely - out of the numerator and the denominator both, so the
-      # percentage silently stops describing the workspace. Assert the report
-      # still covers every package the source tree says holds production C++.
+      # percentage silently stops describing the workspace. Assert the merged
+      # report still covers every package the source tree says holds production
+      # C++. Sharding makes this stricter, not weaker: a shard that silently
+      # tested nothing now shows up as a package missing from the merge.
       - name: Verify every C++ package reached the coverage report
```
mfaferek93, line 715: No `if: always()` here. When a shard fails it uploads no tracefile, this gate exits 1 on the missing package, and both the HTML upload and "Fail if any shard did not succeed" below are skipped. The run then says "package X missing" instead of "shard Y failed" and the partial HTML is gone, which is the opposite of the comment at 644-647. `if: always()` on those two steps fixes it.

.github/workflows/ci.yml
```diff
@@ -270,6 +324,13 @@ jobs:
     runs-on: ubuntu-latest
     container:
       image: ubuntu:noble
+      # A container gets 64 MB of /dev/shm by default, which is Docker's number and
+      # not a considered one. Fast DDS puts a 512 KB segment plus port files there
+      # per participant - measured at ~0.65 MB - and a participant killed rather
+      # than shut down never gets its segments back. Measured: 15 live
+      # participants take 9.6 MB, and 125 killed ones fill 63 MB. 1g is tmpfs, so
+      # it costs only what is used, against 16 GB of runner memory.
+      options: --shm-size=1g
```
mfaferek93, line 322: This job gets the 1g shm but not the census, and neither jazzy-test nor graph-watchdog (45 min step cap on a 24 min suite) gets the record/guard pair build-and-test got. A kill there still shows up as one `missing_result` test. Worth a small composite action (record + guard + census) used in all three.

src/ros2_medkit_integration_tests/include/ros2_medkit_integration_tests/crash_backtrace.hpp
```diff
@@ -0,0 +1,179 @@
...
+  }
+
+  void ** frames = crash_frame_buffer();
+  const int depth = ::backtrace(frames, kMaxFrames);
+  // backtrace_symbols_fd writes through the raw fd and allocates nothing, which
+  // is what makes it usable here; backtrace_symbols would call malloc.
+  ::backtrace_symbols_fd(frames, depth, STDERR_FILENO);
+  write_literal(kCrashMarker);
+  write_literal(" end\n");
+}
```
mfaferek93, line 144: Re-raise is implicit here: return and let the faulting instruction re-execute. That does not hold for a SIGSEGV/SIGBUS delivered by `kill`/`raise`/`pthread_kill`: the process resumes and the still-armed `alarm(5)` kills it 5 s later with SIGALRM (status 14, not 11), contrary to README line 96. Ending the handler with `::raise(signum)` covers both cases and is async-signal-safe (default disposition is already back via SA_RESETHAND).

src/ros2_medkit_integration_tests/include/ros2_medkit_integration_tests/crash_backtrace.hpp
```diff
@@ -0,0 +1,179 @@
...
+  stack_t alt{};
+  alt.ss_sp = detail::crash_alt_stack();
+  alt.ss_size = detail::kAltStackBytes;
+  alt.ss_flags = 0;
+  ::sigaltstack(&alt, nullptr);
+
+  struct sigaction action {};
+  action.sa_handler = &detail::crash_handler;
+  // The handler writes into one static frame buffer, and SA_RESETHAND only
+  // resets the signal that fired. Without this, a second thread taking a
```
mfaferek93: `sa_mask` only applies to the thread running the handler, so this does not prevent what the comment says. Thread A in SIGSEGV, thread B takes SIGBUS: B runs `crash_handler` concurrently, both write the same static `frames[]` and interleave stderr. A `static std::atomic_flag` (`test_and_set` is lock-free) with a one-line marker for the second entrant would actually do it.

src/ros2_medkit_integration_tests/include/ros2_medkit_integration_tests/crash_backtrace.hpp
```diff
@@ -0,0 +1,179 @@
...
+  ::sigemptyset(&action.sa_mask);
+  ::sigaddset(&action.sa_mask, SIGSEGV);
+  ::sigaddset(&action.sa_mask, SIGBUS);
+  ::sigaddset(&action.sa_mask, SIGABRT);
+  // SA_RESETHAND restores the default disposition before the handler runs, so
+  // returning from it re-raises the signal and the process dies the way it
+  // would have without us.
+  // SA_ONSTACK puts the handler on the alternate stack above, which is what lets
+  // it run at all when the ordinary stack is the thing that overflowed.
+  action.sa_flags = SA_RESETHAND | SA_ONSTACK;
```
mfaferek93: `SA_RESETHAND | SA_ONSTACK` is `unsigned` going into `int sa_flags`: `-Wsign-conversion` "changes value from 2281701376 to -2013265920", once per including TU (34 times in a build). Bits are right; `static_cast<int>(SA_RESETHAND | SA_ONSTACK)` silences it.

src/ros2_medkit_integration_tests/test/test_crash_backtrace.cpp
```diff
@@ -0,0 +1,199 @@
...
+/// The object and offset pair is what addr2line turns back into a location, and
+/// glibc's spacing between the offset and the address differs by release
+/// (resolute writes ") [0x", noble writes ")[0x"), so the separator is loose. A
+/// frame carries a symbol name before the "+" whenever the binary exports its
+/// symbols, which CMake does by default, so that half is loose too.
+std::string own_binary_frame_pattern() {
+  char exe[PATH_MAX] = {};
+  const ssize_t len = ::readlink("/proc/self/exe", exe, sizeof(exe) - 1);
+  const std::string self = len > 0 ? std::string(exe, static_cast<std::size_t>(len)) : std::string();
+  return self + R"(\([^)]*\+0x[0-9a-fA-F]+\) ?\[0x[0-9a-fA-F]+\])";
```
mfaferek93: glibc prints the main executable as `argv[0]` (`dl-addr.c` uses `_dl_argv[0]` for the main map), not the realpath from `/proc/self/exe`. With a symlinked workspace path, or `./test_crash_backtrace` run by hand, the frame line and this pattern differ and the test fails on a correct report. Anchoring on the basename `test_crash_backtrace\(` is enough.

src/ros2_medkit_integration_tests/README.md
```diff
@@ -78,6 +78,30 @@ ros2 launch ros2_medkit_integration_tests demo_nodes.launch.py
 | `long_calibration` | `/powertrain/engine` | Action | Fibonacci-based long-running action |
 | `dual_calibration` | `/testrig/dual` | Services + Actions | `left/calibrate` and `right/calibrate`, `left/sweep` and `right/sweep` - one provider carrying each operation short name twice, so the ROS path is the only id that separates the copies |
 
+### Crash reports
+
+Every demo binary goes through `run_demo_node()` in
```
mfaferek93: Not every one: `demo_nodes/unresponsive_param_node.cpp:116` and `demo_nodes/managed_lifecycle_node.cpp:64` have their own `main` and never call `install_crash_backtrace()`, so those two still die silently. Calling it first thing in both mains makes this sentence (and the CMake comment at 209) true.

scripts/sweep_shm.sh
```diff
@@ -0,0 +1,77 @@
...
+  printf '%s' "${out}" | head -n 1 | cut -f1
+}
+
+if [[ "${FORCE}" -eq 0 ]]; then
+  # Match on the command line rather than the process name: colcon runs as
+  # python3, and a pattern that also matches this script would match itself.
+  # demo_ covers the launch fixtures the documented demo workflow starts, which
+  # the earlier pattern missed entirely - a developer running only demo nodes got
+  # a sweep in the middle of live work.
+  if pgrep -f '[g]ateway_node|[f]ault_manager_node|[d]emo_|[c]test|[l]aunch_test|[c]olcon test' >/dev/null 2>&1; then
```
mfaferek93: This pattern matches the caller: `bash ./scripts/test.sh test_demo_lifecycle` contains `demo_`, so the documented single-test presets always get "ROS processes are running" and never sweep (test.sh swallows the exit, so the only symptom is the misleading refusal). Filter `$PPID`/`$$` out of the pgrep result, or match executable names with `pgrep -x`.

.github/workflows/ci.yml
bburda: You are right, and this would have failed on main rather than here, because the coverage jobs do not run on pull requests. For the catch-all shard the first name after the flag is `ros2_medkit_opcua`, which that job never builds, so the find was always empty. The check now runs only when the flag is `--packages-select`.

src/ros2_medkit_integration_tests/CMakeLists.txt
bburda: Confirmed. This package includes only Compat, Linting and Warnings, so `__SANITIZE_ADDRESS__` is never defined here and every `else` branch was unreachable. I removed those branches and the claim about them. The stand-down stays in the header for the day this package opts in, and the comment now says it is not tested.

.github/workflows/ci.yml
bburda: Fixed. The gate, the HTML upload and the step that reports the failing shard all run with `if: always()` now, so a broken shard no longer hides its own cause behind "package X missing".

.github/workflows/ci.yml
bburda: Moved the record and guard pair into a composite action and wired it into all three jobs. `build-and-test` and `graph_watchdog` pass 45, `jazzy-test` passes 90. On `jazzy-test` the overrun branch cannot fire, because a job level kill stops every remaining step, but the 80 percent warning still arrives while there is time to act.

src/ros2_medkit_integration_tests/include/ros2_medkit_integration_tests/crash_backtrace.hpp
bburda: Measured, and you are right: `raise(SIGSEGV)` exited with 142, not 139, because the alarm killed the resumed process. The handler now ends with `::raise(signum)` and the same probe gives 139. Fault delivered signals still report their own status.

src/ros2_medkit_integration_tests/include/ros2_medkit_integration_tests/crash_backtrace.hpp
bburda: Correct, the comment claimed more than the code did. Replaced with an atomic flag: the first thread writes the report, the second writes one line saying it also faulted and gets out of the way. The mask stays for the same thread case only, and the comment now says so.

src/ros2_medkit_integration_tests/include/ros2_medkit_integration_tests/crash_backtrace.hpp
bburda: Added the cast. A build with `-Wall -Wextra -Wconversion -Wsign-conversion` is clean now.

src/ros2_medkit_integration_tests/test/test_crash_backtrace.cpp
bburda: Changed to the basename. I checked it at `-O2` and at `-O2 -rdynamic`, where frames carry symbol names, and both pass.

src/ros2_medkit_integration_tests/README.md
bburda: Both nodes call `install_crash_backtrace()` in their own `main` now, so the sentence is true. The README also names the two and says why they cannot use the helper.

scripts/sweep_shm.sh
bburda: Fixed. Our own process and our caller are excluded from the match now. Worth adding that on my machine the guard was also refusing for a real reason, gateway processes from another checkout, so I checked the filter logic separately.

mfaferek93 (approved):

Issue #649, opened 2026-09-01
PostgreSQL support


gstavrinos:
Hello Bartosz and Michal,

First of all, thank you for your contribution in the ROS community with this package. Impressive work!

We (https://github.com/aperion-robotics) are currently working on adding support for PostgreSQL as a fault storage database. The fork should be visible in the upcoming days. 

Unfortunately, sqlite could not satisfy our needs. Our main goal is to make the integration easier with the rest of our infrastructure, and simplify the local to online/remote database offloading. 

I am reaching out due to the rapid pace of development of this repository and the potential changes that occur due to newfound issues.

To be more specific, we had a complete PostgreSQL implementation about a week ago, and then noticed the large amount of changes that were introduced with the fix for multiple rosbags support per fault. Currently we are, once more, working on integrating PostgreSQL with the new database schema and filesystem handling. 

Our main concern is having the same "breaking" changes happen before our integration is complete. It is very important that all your changes are backwards compatible and we respect that. For a new integration though, all these new changes are non-applicable, and mainly just introduce extra work to check already implemented functions or create new ones. 

We are currently cleaning up functions (like the "old" way to check fault-rosbag referencing). Also new code should be implemented to generate the new schema right from the start and support the new functionalities.

At this point, it is very important for us to hear your thoughts on some matters.

Is PostgreSQL something that you have considered in the past? Do you have a roadmap regarding its integration? Do you see any drawbacks? Given the fact that we plan to open a PR soon, would it be something you would be interested in supporting in the future?

Additionally regarding our next steps, do you think that the codebase will be stable enough in the upcoming days in order for us to have the time to make a PR with the new implementation? (Without the need to re-visit our implemented code to integrate newer changes)

Sorry for the very long issue. Eager to hear your thoughts.

George

bburda:
Hi George! I'm really sorry for the churn on the schema. Multiple rosbags per fault was necessary because a fault usually happens more than once in a mission, in different places and under different conditions, and we want all of those occurrences under one fault entry with the matching rosbags. One bag per fault gives you a snapshot. With all of them you can replay the mission, see whether the problem is systematic, and see what to fix on the robot. We should have flagged the change before merging it.

[...]

PR #651, opened 2026-09-04
First steps towards PostgreSQL integration

gstavrinos:
# Pull Request

## Summary

As discussed in https://github.com/selfpatch/ros2_medkit/issues/649, this is an early implementation of the PostgreSQL integration. Keep in mind that currently the code does not compile because the testing suite is not included in this PR.

---

## Issue

- #649

---

## Type

- [ ] Bug fix
- [X] New feature or tests
- [ ] Breaking change
- [ ] Documentation only

---

## Testing

Tests are yet to be implemented, so the code remains in an early stage, completely untested even for basic functionality

---

## Checklist

- [X] Tests were added or updated if needed
- [X] Docs were updated if behavior or public API changed

## TODOs (based on your checklist, will tick the list as improvements come along)

- [X] Testing suite following the SQLite paradigm (Currently WIP, ETA next week)
- [X] Documentation update on how to use the new PostgreSQL fault storage
- [X] For now, no breaking changes have been added, and the goal is to not introduce any. 

As this is still a WIP, feel free to offer suggestions, recommendations or problems you might think of.

### DIFF
--- a/docs/config/fault-manager.rst
+++ b/docs/config/fault-manager.rst
@@ -18,8 +18,9 @@ Storage
 
    fault_manager:
      ros__parameters:
-       storage_type: "sqlite"              # Storage backend: "sqlite" or "memory"
+       storage_type: "sqlite"              # Storage backend: "sqlite" or "memory" or "postgres" 
        database_path: "/var/lib/ros2_medkit/faults.db"  # Path for sqlite storage
+       database_url: "postgresql://user:password@localhost:5432/ros2_medkit_faults_database"  # URL for postgres storage (alternatively:  "host=localhost port=5432 dbname=ros2_medkit_faults_database user=user password=password")
 
 .. list-table::
    :header-rows: 1
@@ -34,6 +35,9 @@ Storage
    * - ``database_path``
      - ``/var/lib/ros2_medkit/faults.db``
      - File path for SQLite database. Directory must exist and be writable.
+   * - ``database_url``
+     - ``postgresql://user:password@localhost:5432/ros2_medkit_faults_database``
+     - Connection URL for PostgreSQL database. The PostgreSQL server must be running with the appropriate user and password.
 
 Debounce Settings
 ~~~~~~~~~~~~~~~~~
@@ -549,7 +553,7 @@ by default: with it off there is no table, no file and no write cost.
    * - ``audit_log.database_path``
      - ``""``
      - Where the audit database lives. Empty puts it beside the fault database,
-       or in memory when the fault store is itself in memory or not SQLite.
+       or in memory when the fault store is itself in memory or unknown storage type.
 
 Correlation Configuration
 -------------------------
--- a/src/ros2_medkit_fault_manager/CMakeLists.txt
+++ b/src/ros2_medkit_fault_manager/CMakeLists.txt
@@ -32,6 +32,7 @@ find_package(ament_cmake REQUIRED)
 find_package(rclcpp REQUIRED)
 find_package(ros2_medkit_msgs REQUIRED)
 find_package(ros2_medkit_serialization REQUIRED)
+find_package(PostgreSQL REQUIRED)
 find_package(SQLite3 REQUIRED)
 find_package(nlohmann_json REQUIRED)
 # OpenSSL EVP SHA-256 for the tamper-evident audit log hash chain
@@ -49,6 +50,7 @@ add_library(fault_manager_lib STATIC
   src/capture_thread_pool.cpp
   src/fault_manager_node.cpp
   src/fault_storage.cpp
+  src/postgres_fault_storage.cpp
   src/sqlite_fault_storage.cpp
   src/fault_audit_log.cpp
   src/snapshot_capture.cpp
@@ -76,6 +78,7 @@ medkit_target_dependencies(fault_manager_lib PUBLIC
 target_link_libraries(fault_manager_lib PUBLIC
   SQLite::SQLite3
   nlohmann_json::nlohmann_json
+  pqxx
   yaml-cpp::yaml-cpp
   OpenSSL::Crypto
 )
@@ -135,6 +138,11 @@ if(BUILD_TESTING)
   medkit_add_gtest(test_sqlite_storage test/test_sqlite_storage.cpp)
   target_link_libraries(test_sqlite_storage fault_manager_lib)
   medkit_target_dependencies(test_sqlite_storage rclcpp ros2_medkit_msgs)
+  
+  # PostgreSQL storage tests
+  medkit_add_gtest(test_postgres_storage test/test_postgres_storage.cpp)
+  target_link_libraries(test_postgres_storage fault_manager_lib)
+  medkit_target_dependencies(test_postgres_storage rclcpp ros2_medkit_msgs)
 
   # Rosbag retention parity: every assertion runs against both storage backends.
   medkit_add_gtest(test_rosbag_storage_parity test/test_rosbag_storage_parity.cpp)
--- a/src/ros2_medkit_fault_manager/README.md
+++ b/src/ros2_medkit_fault_manager/README.md
@@ -61,8 +61,9 @@ ros2 service call /fault_manager/clear_fault ros2_medkit_msgs/srv/ClearFault \
 
 | Parameter | Type | Default | Description |
 |-----------|------|---------|-------------|
-| `storage_type` | string | `"sqlite"` | Storage backend: `"sqlite"` or `"memory"` |
+| `storage_type` | string | `"sqlite"` | Storage backend: `"sqlite"` or `"memory"` or `"postgres"` |
 | `database_path` | string | `"/var/lib/ros2_medkit/faults.db"` | Path to SQLite database file |
+| `database_url` | string | `"postgresql://user:password@localhost:5432/ros2_medkit_faults_database"` | Connection URL to the PostgreSQL database |
 | `confirmation_threshold` | int | `-1` | Counter value at which faults are confirmed |
 | `healing_enabled` | bool | `false` | Enable automatic healing via PASSED events |
 | `healing_threshold` | int | `3` | Counter value at which faults are healed |
@@ -133,6 +134,8 @@ format used by black-box capture (`snapshots.rosbag.format`, see Rosbag Capture 
 **SQLite (default)**: Faults are persisted to disk and survive node restarts. Uses WAL mode for optimal performance.
 
 **Memory**: Faults are stored in memory only. Useful for testing or when persistence is not required.
+
+**PostgreSQL**: Faults are persisted to disk and survive node restarts. Needs an external PostgreSQL server.
 
 ## Near-Miss Series
 
--- a/src/ros2_medkit_fault_manager/src/fault_manager_node.cpp
+++ b/src/ros2_medkit_fault_manager/src/fault_manager_node.cpp
@@ -26,6 +26,7 @@
 #include <sstream>
 
 #include "ros2_medkit_fault_manager/correlation/config_parser.hpp"
+#include "ros2_medkit_fault_manager/postgres_fault_storage.hpp"
 #include "ros2_medkit_fault_manager/sqlite_fault_storage.hpp"
 #include "ros2_medkit_fault_manager/time_utils.hpp"
 #include "ros2_medkit_msgs/msg/cluster_info.hpp"
@@ -118,6 +119,8 @@ FaultManagerNode::FaultManagerNode(const rclcpp::NodeOptions & options) : Node("
   // Declare and get parameters
   storage_type_ = declare_parameter<std::string>("storage_type", "sqlite");
   database_path_ = declare_parameter<std::string>("database_path", "/var/lib/ros2_medkit/faults.db");
+  database_url_ = declare_parameter<std::string>(
+      "database_url", "postgresql://user:password@localhost:5432/ros2_medkit_faults_database");
 
   auto confirmation_threshold_param = declare_parameter<int>("confirmation_threshold", -1);
   if (confirmation_threshold_param > 0) {
@@ -519,6 +522,11 @@ std::unique_ptr<FaultStorage> FaultManagerNode::create_storage() {
     return std::make_unique<SqliteFaultStorage>(database_path_);
   }
 
+  if (storage_type_ == "postgres") {
+    RCLCPP_INFO(get_logger(), "Using PostgreSQL fault storage: %s", database_url_.c_str());
+    return std::make_unique<PgFaultStorage>(database_url_);
+  }
+
   RCLCPP_ERROR(get_logger(), "Unknown storage_type '%s', falling back to in-memory", storage_type_.c_str());
   return std::make_unique<InMemoryFaultStorage>();
 }
@@ -558,7 +566,7 @@ std::unique_ptr<FaultAuditLog> FaultManagerNode::create_audit_log() {
   }
 
   if (audit_path.empty()) {
-    if (database_path_ == ":memory:" || storage_type_ != "sqlite") {
+    if (database_path_ == ":memory:" || (storage_type_ != "sqlite" && storage_type_ != "postgres")) {
       audit_path = ":memory:";
     } else {
       std::filesystem::path base(database_path_);
--- a/src/ros2_medkit_fault_manager/src/fault_manager_node.cpp
+++ b/src/ros2_medkit_fault_manager/src/fault_manager_node.cpp
@@ -523,7 +523,17 @@ std::unique_ptr<FaultStorage> FaultManagerNode::create_storage() {
   }
 
   if (storage_type_ == "postgres") {
-    RCLCPP_INFO(get_logger(), "Using PostgreSQL fault storage: %s", database_url_.c_str());
+    std::string redacted_database_url = database_url_;
+
+    // Handle URL style: user:password@host
+    std::regex url_regex(R"((://[^:/@]+:)([^@]+)(@))");
+    redacted_database_url = std::regex_replace(redacted_database_url, url_regex, "$1***$3");
+
+    // Handle parameter style: password=value
+    std::regex param_regex(R"(password=([^\s]+))");
+    redacted_database_url = std::regex_replace(redacted_database_url, param_regex, "password=***");
+    RCLCPP_INFO(get_logger(), "Using PostgreSQL fault storage (with redacted password): %s",
+                redacted_database_url.c_str());
     return std::make_unique<PgFaultStorage>(database_url_);
   }
 
--- a/src/ros2_medkit_fault_manager/src/postgres_fault_storage.cpp
+++ b/src/ros2_medkit_fault_manager/src/postgres_fault_storage.cpp
@@ -1,0 +1,1416 @@
+// Copyright 2026 gstavrinos
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+#include "ros2_medkit_fault_manager/postgres_fault_storage.hpp"
+
+#include "rcutils/logging_macros.h"
+#include "ros2_medkit_msgs/msg/fault.hpp"
+
+namespace ros2_medkit_fault_manager {
+
+PgFaultStorage::PgFaultStorage(const std::string & conn_info) : conn_info_(conn_info) {
+  try {
+    db_conn_ = std::make_unique<pqxx::connection>(conn_info);
+  } catch (const std::exception & e) {
+    throw std::runtime_error(std::string("Failed to connect to PostgreSQL: ") + e.what());
+  }
+  initialize_schema();
+}
+
+PgFaultStorage::~PgFaultStorage() {
+  if (!db_conn_->is_open()) {
+    return;
+  }
+  try {
+    db_conn_->close();
+  } catch (...) {
+  }
+}
+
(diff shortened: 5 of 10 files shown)


gstavrinos:
Oh, wow, Copilot came in aggressively! I will consider the LLM's remarks as I am implementing the tests.

bburda:
> Oh, wow, Copilot came in aggressively! I will consider the LLM's remarks as I am implementing the tests.

No worries, he is always like that. Use your own judgement, some of his comments are not worth fixing.
Also please ping me any time if you need guidance or get stuck, or once the PR is ready for review. 

Happy to take a look.

gstavrinos:
Hello, I am currently finishing up the unit tests.

I plan to update this PR with the new changes tomorrow. Apart from the integration and testing code, I also want to update the documentation. Can you please point me to the required documentation file(s) that need to be updated?

Thanks!

bburda:
> Hello, I am currently finishing up the unit tests.
> 
> I plan to update this PR with the new changes tomorrow. Apart from the integration and testing code, I also want to update the documentation. Can you please point me to the required documentation file(s) that need to be updated?
> 
> Thanks!

Hi @gstavrinos, two files to update: 
- The Storage section, and the same parameters again in the complete example at the bottom here: https://github.com/selfpatch/ros2_medkit/blob/main/docs/config/fault-manager.rst
- The Features list, the Parameters table and the Storage Backends section here: https://github.com/selfpatch/ros2_medkit/blob/main/src/ros2_medkit_fault_manager/README.md

That should be it, but if I find anything else, I'll let you know.

gstavrinos:
I have updated the source code, added the unit tests, a docker compose file and relevant documentation points.

Some remarks: 

* I could not find an easy and non-destructive way to also integrate PostgreSQL to the audit logs. They appear to be intertwined with sqlite.
* I have not tampered with the github actions of the repository. The provided docker container should also run while testing postgres fault storage.

I will be waiting for your feedback on the provided material and the next steps towards merging. 

Thanks again!

src/ros2_medkit_fault_manager/CMakeLists.txt
```diff
@@ -32,6 +32,7 @@ find_package(ament_cmake REQUIRED)
 find_package(rclcpp REQUIRED)
 find_package(ros2_medkit_msgs REQUIRED)
 find_package(ros2_medkit_serialization REQUIRED)
+find_package(PostgreSQL REQUIRED)
```
bburda, line 35: Could you put the PostgreSQL backend behind a CMake option that is OFF by default? Right now libpqxx is required for every build, and that breaks builds that never use PostgreSQL:

- Humble has libpqxx 6.4. The new file does not compile there: `connection::close()` is protected, a `std::vector<std::string>` cannot be a query parameter (`list_faults()`), and `<filesystem>` is not included.
- On Lyrical, 22 of 24 fault_manager tests fail. For example, `test_capture_thread_pool` passes all 12 tests and then aborts at exit with `free(): double free detected in tcache 2`. The cause is the libpqxx 7.10.0 package in Ubuntu 26.04. A program with only `#include <pqxx/pqxx>` and `int main() { return 0; }` aborts in the same way. If I link libpqxx and leave out the header, it exits normally.
- The runtime stage of the Dockerfile in this repository does not install libpqxx. `fault_manager_node` now needs `libpqxx-7.8.so`, so the Docker image cannot start the fault manager.

With the option OFF, please build nothing PostgreSQL-related: no `find_package`, no `postgres_fault_storage.cpp`, no `pqxx` link and no test target. In such a build, `create_storage()` should stop with a clear error when `storage_type` is `postgres`. The `#include` of `postgres_fault_storage.hpp` in `fault_manager_node.cpp` needs the same guard. Please also remove `<depend>libpqxx-dev</depend>` from package.xml, and write the option name and the `libpqxx-dev` package in the README, because rosdep does not install it without the `<depend>`.

src/ros2_medkit_fault_manager/src/postgres_fault_storage.cpp
```diff
@@ -0,0 +1,1419 @@
...
+  // Only a FAILED report can write two rows (fault row and near-miss row) and
+  // both have to land together: written separately, a failure on the second would
+  // leave the debounce counter already advanced, so the caller's retry would
+  // advance it a second time and the near miss it retried for would still be
+  // missing from the series.
+  //
+  // The SQLite backend has to opt into a transaction for this (BEGIN IMMEDIATE)
+  // In the PostgreSQL implementation the lock used is a transaction lock, so no extra handling is required apart from
+  // updating the tables properly
+  pqxx::work tx(*db_conn_);
```
bburda, line 322: I built this branch and ran the fault manager against PostgreSQL 16. I reported one fault, restarted the database container and reported a second fault. The node stopped on the second report:

```
terminate called after throwing an instance of 'pqxx::broken_connection'
  what():  Lost connection to the database server.
```

The object opens one `pqxx::connection` and never reconnects, and the service handlers do not catch storage exceptions. So every database restart or network drop stops the fault manager. If you want to fix this in the PR, two changes are needed:

1. In `PgFaultStorage`: when a transaction cannot start because the connection is broken, open a new connection and try the call once more. Do not retry on `pqxx::in_doubt_error`, because the commit may already be applied.
2. In the node: catch exceptions from `storage_` in the service and timer callbacks and return an error response, so the process keeps running while the database is down.

src/ros2_medkit_fault_manager/src/postgres_fault_storage.cpp
```diff
@@ -0,0 +1,1419 @@
...
+  // check inside the loop would find it still held by a sibling that a later
+  // iteration then evicts, leaking the directory.
+  const std::set<std::string> unique_paths(evicted.begin(), evicted.end());
+  for (const auto & path : unique_paths) {
+    if (path_referenced(path)) {
+      continue;
+    }
+    std::error_code ec;
+    std::filesystem::remove_all(path, ec);
+  }
```
bburda, line 1010: Could you make sure this loop cannot throw? The rows are already committed at this point. `path_referenced()` starts a new transaction, so it throws when the connection is lost. `RosbagCapture` treats any exception from `store_rosbag_files()` as "nothing was stored" and deletes the new bag, and the committed rows then point to a bag that no longer exists. With the change below, the worst case is an orphaned directory, which the other comments in this file already accept.

```suggestion
  for (const auto & path : unique_paths) {
    try {
      if (path_referenced(path)) {
        continue;
      }
    } catch (const std::exception &) {
      continue;  // The rows are already committed, so keep the directory.
    }
    std::error_code ec;
    std::filesystem::remove_all(path, ec);
  }
```

src/ros2_medkit_fault_manager/test/test_postgres_storage.cpp
```diff
@@ -0,0 +1,2278 @@
...
+    // Create a unique schema for each test using random_device for better entropy
+    std::random_device rd;
+    std::mt19937_64 gen(rd());
+    std::uniform_int_distribution<uint64_t> dist;
+    schema_ = "test_faults_" + std::to_string(dist(gen));
+
+    try {
+      db_conn_ = std::make_unique<pqxx::connection>(base_conn_info());
+    } catch (const std::exception & e) {
+      GTEST_SKIP() << "No PostgreSQL server reachable (set ROS2_MEDKIT_TEST_PG_CONN): " << e.what();
```
bburda, line 69: All 98 tests in this file are skipped in CI, because CI has no PostgreSQL server. The Jazzy test log shows `[  SKIPPED ]` for each of them. If you want to fix this in the PR, the `jazzy-test` job in `.github/workflows/ci.yml` needs:

- a `postgres` service at the minimum version you support
- `libpqxx-dev` installed and a build with the new option ON
- `ROS2_MEDKIT_TEST_PG_CONN` set to `postgresql://user:password@postgres:5432/<db>`

`jazzy-test` runs in a container, so the service name works as the host name. After that job is in place, please change `GTEST_SKIP()` to `FAIL()` here, so a missing server is reported as a failure. With the option OFF the test target is not built, so the other jobs are not affected.

src/ros2_medkit_fault_manager/src/postgres_fault_storage.cpp
```diff
@@ -0,0 +1,1419 @@
...
+
+  pqxx::work tx(*db_conn_);
+  // Apply the bound to what is already in the database. Without this, a database that grew under
+  // a larger bound (or none) stays over the new bound until each fault code happens to record
+  // another near miss - and a code that never does keeps its rows for good.
+  try {
+    auto res = tx.exec_params(
+        "DELETE FROM near_misses WHERE id IN ("
+        "SELECT id FROM (SELECT id, ROW_NUMBER() OVER "
+        "(PARTITION BY fault_code ORDER BY id DESC) AS rn FROM near_misses) "
```
bburda, line 886: Could you add an alias to this subquery? PostgreSQL accepts a subquery in FROM without an alias only since version 16. I ran the fault manager from this branch against `postgres:15` with default parameters. It stops at startup, because `near_miss.max_per_fault` defaults to 200 and the constructor runs this statement:

```
set_max_near_misses_per_fault PostgreSQL error: ERROR:  subquery in FROM must have an alias
```

With the alias below, the statement works on PostgreSQL 15. Ubuntu 22.04 still ships PostgreSQL 14. Please also write the minimum PostgreSQL version in the docs and run the tests against that version. The compose file uses `postgres:18`, so the tests could not catch this.

```suggestion
        "(PARTITION BY fault_code ORDER BY id DESC) AS rn FROM near_misses) AS ranked "
```

src/ros2_medkit_fault_manager/src/postgres_fault_storage.cpp
```diff
@@ -0,0 +1,1419 @@
...
+size_t PgFaultStorage::get_total_rosbag_storage_bytes() const {
+  std::lock_guard<std::mutex> lock(mutex_);
+  pqxx::work tx(*db_conn_);
+
+  // Sum per bag, not per fault: one recording can back a burst of correlated
+  // faults, and double-counting it would evict bags that still fit the quota.
+  try {
+    auto res = tx.exec(
+        "SELECT COALESCE(SUM(size_bytes),0) AS sz FROM (SELECT MAX(size_bytes) AS size_bytes FROM rosbag_files "
+        "GROUP BY file_path)");
```
bburda, line 1312: The same alias is missing here. `evict_bags_over_quota()` calls this after every stored recording, and `RosbagCapture` only logs a warning when it throws. So on PostgreSQL 15 the rosbag storage limit is never enforced. Could you add the alias here too?

```suggestion
        "GROUP BY file_path) AS per_bag");
```

src/ros2_medkit_fault_manager/src/fault_manager_node.cpp
```diff
@@ -118,6 +119,8 @@ FaultManagerNode::FaultManagerNode(const rclcpp::NodeOptions & options) : Node("
   // Declare and get parameters
   storage_type_ = declare_parameter<std::string>("storage_type", "sqlite");
   database_path_ = declare_parameter<std::string>("database_path", "/var/lib/ros2_medkit/faults.db");
+  database_url_ = declare_parameter<std::string>(
+      "database_url", "postgresql://user:password@localhost:5432/ros2_medkit_faults_database");
```
bburda, line 123: Could you make the default an empty string and keep the password out of this parameter? Any process on the ROS graph can read parameters. With the node from this branch running, `ros2 param get /fault_manager database_url` in another process printed `postgresql://user:password@...` in full. With an empty connection string, libpq takes host, user and password from its environment variables (`PGHOST`, `PGUSER`, `PGPASSWORD`) and from `~/.pgpass`. Then the password is never a parameter. Please also change the default of `database_url` to `""` in the README table and in `docs/config/fault-manager.rst` (the table and the YAML example). They still show the URL with the password.

```suggestion
  database_url_ = declare_parameter<std::string>("database_url", "");
```

bburda (commented):
Thanks for this contribution :)

There are also two bigger things: the PostgreSQL tests do not run in CI, and the fault manager stops on the next fault report after the database restarts. We can open separate issues for them, or you can fix them in this PR. Please also make sure that CI passes.

src/ros2_medkit_fault_manager/src/fault_manager_node.cpp
```diff
@@ -519,6 +522,21 @@ std::unique_ptr<FaultStorage> FaultManagerNode::create_storage() {
...
 
+  if (storage_type_ == "postgres") {
+    std::string redacted_database_url = database_url_;
+
+    // Handle URL style: user:password@host
+    std::regex url_regex(R"((://[^:/@]+:)([^@]+)(@))");
+    redacted_database_url = std::regex_replace(redacted_database_url, url_regex, "$1***$3");
+
+    // Handle parameter style: password=value
+    std::regex param_regex(R"(password=([^\s]+))");
```
bburda, line 533: These regexes miss valid connection strings. I ran both of them on two examples: `password = secret` (spaces around `=`) stays unchanged in the log, and `password='my secret'` becomes `password=*** secret'`. Could you remove the regexes and log `dbname()`, `hostname()` and `port()` of the connection after it is open?

src/ros2_medkit_fault_manager/src/postgres_fault_storage.cpp
```diff
@@ -0,0 +1,1419 @@
...
+}
+
+bool PgFaultStorage::report_fault_event_locked(const std::string & fault_code, uint8_t event_type, uint8_t severity,
+                                               const std::string & description, const std::string & source_id,
+                                               const rclcpp::Time & timestamp, const DebounceConfig & config,
+                                               pqxx::work & tx) {
+  int64_t timestamp_ns = timestamp.nanoseconds();
+  const bool is_failed = (event_type == EventType::EVENT_FAILED);
+
+  // NOTE: Do I need to SELECT here with FOR UPDATE to secure the selected row?
```
bburda, line 341: About this NOTE: with one fault manager you do not need `FOR UPDATE`, because `mutex_` already serializes all calls on the connection. It matters only when two fault managers write to the same database, and then the schema is the bigger problem. Faults are keyed only by `fault_code`, so the same code from two robots becomes one fault. When one robot applies the rosbag cap, it deletes rows of bags that are on the other robot's disk. Those bags become unreachable and are no longer counted in that robot's storage quota.

Do you plan to share one database between robots? For this PR, could you replace the NOTE with a comment and a sentence in the docs saying that one database belongs to exactly one fault manager? A separate schema per fault manager, selected with `options=-csearch_path=...` like in your tests, counts as its own database.

```suggestion
  // No row lock needed: mutex_ serializes every call on this connection, and one database belongs to exactly one
  // fault manager.
```

src/ros2_medkit_fault_manager/src/postgres_fault_storage.cpp
```diff
@@ -0,0 +1,1419 @@
...
+    // was introduced, we should not check for backwards compatibility
+
+    // Migration: releases that advanced last_occurred_ns on PASSED events left
+    // inflated rows behind, and a latched CONFIRMED fault that only ever heals
+    // would keep the wrong timestamp forever. last_failed_ns holds the true
+    // last occurrence; last_occurred_ns can only exceed it via that old bug.
+    tx.exec(
+        "UPDATE faults SET last_occurred_ns = last_failed_ns WHERE last_failed_ns > 0 AND last_occurred_ns > "
+        "last_failed_ns");
+
```
bburda, line 81: This UPDATE repairs rows from older SQLite releases. A database created by this backend cannot contain such rows. Could you remove it? The suggestion below removes the UPDATE and the NOTE above it. Several comments copied from the SQLite backend also describe SQLite history: the capture_id migration (lines 83-88), the rule about `UNIQUE` columns (140-145), the behaviour of older releases at N = 1 (1044-1047) and `json_each()` (1357-1360). Please rewrite them so they describe this backend, and move the NOTEs that compare it with SQLite to the PR description.

```suggestion
```

src/ros2_medkit_fault_manager/test/test_postgres_storage.cpp
```diff
@@ -0,0 +1,2278 @@
...
+  // Simulate the old bug: a PASSED event at t=9000s re-dated the row.
+  exec_raw("UPDATE faults SET last_occurred_ns = 9000000000000 WHERE fault_code = 'FAULT_MIG'");
+
+  storage_ = std::make_unique<PgFaultStorage>(conn_info_);
+
+  auto fault = storage_->get_fault("FAULT_MIG");
+  ASSERT_TRUE(fault.has_value());
+  EXPECT_EQ(rclcpp::Time(fault->last_occurred).nanoseconds(), failed_at.nanoseconds());
+}
+
```
bburda, line 303: Could you remove this test together with the migration in `initialize_schema()`? Without that UPDATE this test fails.

```suggestion
```

docs/config/fault-manager.rst
```diff
@@ -549,7 +553,7 @@ by default: with it off there is no table, no file and no write cost.
    * - ``audit_log.database_path``
      - ``""``
      - Where the audit database lives. Empty puts it beside the fault database,
-       or in memory when the fault store is itself in memory or not SQLite.
+       or in memory when the fault store is itself in memory or unknown storage type.
```
bburda, line 556: With `storage_type: postgres`, `create_audit_log()` writes the audit log to `fault_audit.db` next to `database_path`. By default that is `/var/lib/ros2_medkit/`, a path that PostgreSQL users do not use otherwise. Could you write that here, so users know the audit trail stays on the robot? The complete example at the end of this page also lists only `storage_type` and `database_path`. Please add `database_url` there.

```suggestion
       or in memory when the fault store is itself in memory or unknown storage type.
       With ``storage_type: postgres`` it is a local SQLite file ``fault_audit.db`` next to
       ``database_path``, so the audit trail stays on the robot.
```

src/ros2_medkit_fault_manager/README.md
```diff
@@ -134,6 +135,8 @@ format used by black-box capture (`snapshots.rosbag.format`, see Rosbag Capture
 
 **Memory**: Faults are stored in memory only. Useful for testing or when persistence is not required.
 
+**PostgreSQL**: Faults are persisted to disk and survive node restarts. Needs an external PostgreSQL server.
```
bburda, line 138: Could you update the Features list above too? It still says `Persistent storage: SQLite backend ensures faults survive node restarts`. Please also change this line so it says where the faults and the audit log are stored:

```suggestion
**PostgreSQL**: Faults are stored in an external PostgreSQL server and survive node restarts. The audit log stays a local SQLite file next to `database_path`.
```

src/ros2_medkit_fault_manager/CMakeLists.txt
```diff
@@ -135,6 +138,11 @@ if(BUILD_TESTING)
   medkit_add_gtest(test_sqlite_storage test/test_sqlite_storage.cpp)
   target_link_libraries(test_sqlite_storage fault_manager_lib)
   medkit_target_dependencies(test_sqlite_storage rclcpp ros2_medkit_msgs)
+
```
bburda, line 141: Could you remove the trailing whitespace on this line? `lint_cmake` in the format-lint job fails on it.

```suggestion

```

src/ros2_medkit_fault_manager/test/test_postgres_storage.cpp
```diff
@@ -0,0 +1,2278 @@
...
+  // delete_rosbag_file() has to survive a DELETE that throws - a lock timeout, a
+  // full disk, a read-only standby - without having already unlinked the bag. A
+  // surviving row whose directory is gone fails every later retrieval and keeps its
+  // bytes charged against the quota, which sums rows.
+  const auto bag_dir = temp_root_ / "failing_delete_bag";
+  std::filesystem::create_directories(bag_dir);
+  {
+    std::ofstream(bag_dir / "payload.mcap") << "data";
+  }
```
bburda, line 1369: Could you format this block as clang-format expects? The format-lint job fails on it. `pre-commit run --all-files` from the repository root catches this and the CMake whitespace before CI does.

```suggestion
  { std::ofstream(bag_dir / "payload.mcap") << "data"; }
```

gstavrinos:
Hey Bartosz,

Thank you for your input. I will look into your suggestions and return with fixes and/or questions.

I am a little bit surprised that the tests do not pass, since I run them on my (jazzy) setup. Seems like I will have to test against humble and lyrical too since Ubuntu uses different PostgreSQL versions.

Give me some time for all that and I will be back.

gstavrinos:
As the commit message implies, this is not finished. I think I need some guidance though, since the changes have started getting a bit disruptive.

In order to achieve the reconnection feature, many functions had to be changed to non-constant. 

At the same time, I presumably have guarded all disconnection exception in the fault manager node, but still a loose broken_connection is thrown somewhere. I will keep investigating tomorrow.

The PostgreSQL support has now been extended to 14-18. I have also provided the equivalent postgres14 container, so that tests can now run on both the oldest and latest versions supported.

Regarding Humble-Lyrical support, I have included a FetchContent declaration in the CMakeLists, but currently pqxx refuses to compile successfully. I will investigate further tomorrow.

Any input would be highly appreciated!

gstavrinos:
I have pushed another commit that allows for pqxx compilation through FetchContent. Reconnections should be working now. More testing might be required, since unit tests do not currently catch all potential issues.

Please advice on how we can proceed.

Have a nice weekend!

bburda:
@gstavrinos Ok, I will check it tomorrow and give you feedback. Have a nice weekend!

bburda:
Hi @gstavrinos 

I've created draft PR with potential solution to all those issues: https://github.com/selfpatch/ros2_medkit/pull/693
Please take a look, you might want to cherry pick commits so we can continue on this PR :)

gstavrinos:
I am continuing the discussion here to avoid splitting the thread. I have studied your fixes and additions.

I am angry (and ashamed) that I did not think of utilizing the `mutable` getaway to avoid the un-const spiral... Sorry about that!

I am also impressed with your attention to detail and thorough tests.

I don't really anything to add or change. Impressive work! I will now use this branch to have some further testing on our setup until it is merged.

Thanks for your help and effort!

bburda:
@gstavrinos If you can cherry pick my commit from https://github.com/selfpatch/ros2_medkit/pull/693 and verify that current solution works for you, I'm fine with merging it :)

gstavrinos:
I have cherry-picked your commits. Give me some time to test and I will be back. (Probably next week)

Issue #277, opened 2026-03-15
Extract GraphProviderPlugin to a loadable .so plugin


bburda:
## Summary

`GraphProviderPlugin` is currently hardcoded as a built-in plugin in `gateway_node.cpp:491`:
```cpp
plugin_mgr_->add_plugin(std::make_unique<GraphProviderPlugin>());
```

This couples the graph provider to the gateway binary. The plugin framework already supports dynamic `.so` loading via `dlopen` (see `test_gateway_plugin.cpp` for the pattern and `gateway_params.yaml` for YAML config). The graph provider should follow the same pattern so users can opt-in/out without recompiling.

---

## Proposed solution

1. **Create a separate build target** for the graph provider as a MODULE shared library (like `test_gateway_plugin.so`), e.g. `ros2_medkit_graph_provider_plugin.so`
2. **Add `extern "C"` exports**: `plugin_api_version()`, `create_plugin()`, `get_introspection_provider()`
3. **Remove the hardcoded `add_plugin`** from `gateway_node.cpp` and the `#include` of `graph_provider_plugin.hpp`
4. **Remove `graph_provider_plugin.cpp` from `gateway_lib`** static library sources in CMakeLists.txt
5. **Document loading** in `gateway_params.yaml` and `docs/tutorials/plugin-system.rst`:
   ```yaml
   plugins: ["graph_provider"]
   plugins.graph_provider.path: "/opt/ros2_medkit/lib/libros2_medkit_graph_provider_plugin.so"
   ```
6. **Keep `diagnostic_msgs` dependency** scoped to the plugin .so, not the gateway core

Follows the existing pattern established by the update plugin and test plugin (`test_gateway_plugin.cpp:117-131`).

---

## Additional context

Introduced in PR #270. Currently the graph provider includes `gateway_node.hpp` and uses `dynamic_cast<GatewayNode *>` to access entity cache and fault manager - these couplings should go through `PluginContext` instead when extracting to .so.

PR #657, opened 2026-09-06
opcua: make read-only a property of the build, not a setting

### LINKED ISSUE #655: Make read-only a property of the OPC UA plugin build, not a setting
## Summary

Read-only is a convention the OPC UA plugin follows, not a property of what ships. `OpcuaClient::write_value` is compiled into every build, `DataProvider::write_data` and the `x-plc-operations` route reach it, and config-less discovery marks a point writable from the server's `CurrentWrite` access bit (`infer_writable` defaults to true), so a walk with no node map exposes a write surface exactly where the controller allows one. Alarm acknowledge and confirm are method calls that change alarm state on the controller and go the same way.

A configuration value cannot answer the question a plant security review asks, which is whether this gateway is capable of changing a controller. A flag can be set back. Only the build can answer it, and only an inspection of the built object can prove the answer.

---

- A CMake option, default on, that builds the plugin read-only: the value-write path, the alarm acknowledge/confirm path and the `x-plc-operations` route are compiled out, `writable` is never inferred or honoured from a node map (with one startup warning naming the build property), and the provider methods refuse with a vendor error code naming the build property. A write-capable build stays available for development and is selected explicitly.
- A ctest that inspects the built shared object: no symbol of the write path, no OPC UA symbol in the export table, the plugin entry points present, and the read path present as a positive control so an empty or wrong object fails. Verified in both build types.
- An integration test that drives a gateway with the plugin against the in-tree test server in both variants: the read-only build refuses and never advertises, the write-capable build writes and the value reaches the server.
- CI builds and tests the write-capable variant once so that path cannot rot, and the OpenPLC integration job proves the read-only contract on the default image.

---

`src/ros2_medkit_plugins/ros2_medkit_opcua/` - `node_map.hpp`, `address_space_browser.cpp`, `opcua_client.cpp`, `opcua_plugin.cpp`, the package CMakeLists and `.github/workflows/opcua-plugin.yml`.

bburda:
# Pull Request

## Summary

Read-only was a convention the OPC UA plugin followed, not a property of what shipped. `OpcuaClient::write_value` was in every build, `DataProvider::write_data` and the `x-plc-operations` route reached it, config-less discovery marked a point writable from the server's `CurrentWrite` bit, and the alarm acknowledge and confirm operations changed condition state on the controller. A configuration value cannot answer the question a plant security review asks, which is whether this gateway can change a controller. Only the build can, and only an inspection of the built object can prove the answer.

**`MEDKIT_OPCUA_READ_ONLY`, a CMake option, default ON.** In the read-only build the value-write path, the alarm acknowledge/confirm path and the `x-plc-operations` route are compiled out. `writable` is never inferred from the server and never honoured from a node map; a map that asks for it gets one startup warning naming the build property. The provider methods still exist, because the gateway calls them through its provider interfaces, and they refuse with `x-medkit-plugin-error` and a message naming the property. `ConditionRefresh` stays: it asks the server to replay the conditions it already holds and changes nothing on the controller, and without it a restart loses the active fault set. A write-capable build is selected explicitly with `-DMEDKIT_OPCUA_READ_ONLY=OFF` and exists for development.

**The object carries no write path, and CI checks the object, not the flag.** `test_opcua_build_variant` runs `nm` on the built plugin and asserts, for the variant the build declared: the write markers absent (`OpcuaClient::write_value`, `handle_plc_operations`, `call_condition_method`, the open62541pp write templates and services), the read markers present as a positive control so an empty or wrong object fails, no `UA_` or `opcua::` symbol in the dynamic export table, the six plugin entry points exported, and none of open62541's `UA_*_write*` primitives anywhere in the object. The last two hold because the plugin now links with `--exclude-libs,ALL` and `--gc-sections`, with the section flags applied to open62541's object libraries: upstream sets them only under Release, and a default-type build previously kept twelve write primitives as local code. A companion self-check test links a module that exports `__UA_Client_writeAttribute` alongside the six entry points and requires the export rule to reject it.

**Both variants are exercised against a real server.** `test_opcua_read_only` drives a gateway with the plugin against the in-tree test server in whichever variant was built: the read-only build reports every point `writable: false` whatever the server and the map say, advertises no `x-plc-operations` and no acknowledge/confirm, refuses writes and condition operations with the vendor code, and keeps doing so after the server restarts and the tree is re-walked; the write-capable build writes, the value reaches the server, and an acknowledge acts on a live condition the fixture raised. The OpenPLC docker job became a two-leg matrix: the default image proves the read-only contract on the PLC (refusals, and both tags unchanged on re-read), the write-capable image proves writes. A Jazzy job builds and tests the write-capable variant so that path cannot rot.

---

## Issue

- closes #655

---

## Type

- [ ] Bug fix
- [x] New feature or tests
- [x] Breaking change
- [ ] Documentation only

Breaking for a deployment that wrote to a controller through this plugin: the default build refuses those writes. Rebuild with `-DMEDKIT_OPCUA_READ_ONLY=OFF` to keep them.

---

## Testing

- Read-only variant, Release: 434 tests, 0 failures. Write-capable variant: 438 tests, 0 failures. Lint on the package clean, zero compiler warnings in our sources, clang-tidy zero findings in changed hunks.
- The inspection test fails on the read-only object when told to expect write-capable, and on the write-capable object when told to expect read-only, on every marker. Measured in Release and in the default build type.
- The integration test forced to the wrong variant fails every variant-specific expectation.
- `nm -DC` on the read-only object: 115 dynamic exports, none from open62541 or open62541pp, the six entry points present.
- The OpenPLC script was run locally against a gateway backed by the in-tree fixture with the OpenPLC entity and tag names; every assertion added or moved passes in both legs, and the two tags the fixture does not have are what the CI job with the real PLC image covers.
- Sphinx: no new warnings.

---

## Checklist

### DIFF
--- a/.github/workflows/opcua-plugin.yml
+++ b/.github/workflows/opcua-plugin.yml
@@ -125,9 +125,21 @@ jobs:
         run: colcon test-result --verbose
 
   integration:
-    name: Integration (OpenPLC)
+    name: Integration (OpenPLC, ${{ matrix.variant }})
+    # Both write surfaces are exercised against the same real PLC. The
+    # read-only leg is the shipped image: it proves the refusal, its vendor
+    # code, and that the tag on the PLC did not move. The write-capable leg
+    # keeps proving that a write reaches OpenPLC and reads back.
     runs-on: ubuntu-latest
     timeout-minutes: 30
+    strategy:
+      fail-fast: false
+      matrix:
+        include:
+          - variant: read-only
+            read_only: 'ON'
+          - variant: write-capable
+            read_only: 'OFF'
     steps:
       - name: Checkout repository
         uses: actions/checkout@v4
@@ -141,6 +153,7 @@ jobs:
       - name: Build gateway + OPC-UA plugin image
         run: |
           docker build \
+            --build-arg MEDKIT_OPCUA_READ_ONLY=${{ matrix.read_only }} \
             -f src/ros2_medkit_plugins/ros2_medkit_opcua/docker/Dockerfile.gateway \
             -t gateway-opcua .
 
@@ -212,6 +225,8 @@ jobs:
           docker logs gateway 2>&1 | tail -10
 
       - name: Run integration tests
+        env:
+          MEDKIT_OPCUA_VARIANT: ${{ matrix.variant }}
         run: bash src/ros2_medkit_plugins/ros2_medkit_opcua/docker/scripts/run_integration_tests.sh
 
       - name: Dump gateway logs on failure
--- a/.github/workflows/opcua-plugin.yml
+++ b/.github/workflows/opcua-plugin.yml
@@ -157,6 +157,22 @@ jobs:
             -f src/ros2_medkit_plugins/ros2_medkit_opcua/docker/Dockerfile.gateway \
             -t gateway-opcua .
 
+      - name: Inspect the plugin object inside the built image
+        # The image is what ships, and Dockerfile.gateway builds it with
+        # BUILD_TESTING=OFF, so the inspection ctest does not exist inside the
+        # container. Pull the object out and run the same check on the runner:
+        # without this the image is only proven by its behaviour, and a
+        # build-arg or Dockerfile drift that shipped the wrong variant would be
+        # caught by the integration assertions rather than by the inspection.
+        run: |
+          cid=$(docker create gateway-opcua)
+          docker cp "$cid:$(docker run --rm --entrypoint sh gateway-opcua -c \
+            "find /root/ws/install -name libros2_medkit_opcua_plugin.so | head -1")" \
+            ./plugin-from-image.so
+          docker rm -v "$cid" >/dev/null
+          python3 src/ros2_medkit_plugins/ros2_medkit_opcua/test/inspect_build_variant.py \
+            ./plugin-from-image.so --expect ${{ matrix.variant }}
+
       - name: Start OpenPLC
         timeout-minutes: 3
         run: |
@@ -276,6 +292,18 @@ jobs:
         env:
           MEDKIT_OPCUA_VARIANT: ${{ matrix.variant }}
         run: bash src/ros2_medkit_plugins/ros2_medkit_opcua/docker/scripts/run_alarm_tests.sh
+
+      - name: Inspect the plugin object inside the built image
+        # Same reason as the OpenPLC job. This suite builds its own image, so
+        # the step runs after it; the tag is the one run_alarm_tests.sh builds.
+        run: |
+          cid=$(docker create gateway-opcua:alarm-test)
+          docker cp "$cid:$(docker run --rm --entrypoint sh gateway-opcua:alarm-test -c \
+            "find /root/ws/install -name libros2_medkit_opcua_plugin.so | head -1")" \
+            ./plugin-from-image.so
+          docker rm -v "$cid" >/dev/null
+          python3 src/ros2_medkit_plugins/ros2_medkit_opcua/test/inspect_build_variant.py \
+            ./plugin-from-image.so --expect ${{ matrix.variant }}
 
       - name: Dump container logs on failure
         if: failure()
--- a/.github/workflows/opcua-plugin.yml
+++ b/.github/workflows/opcua-plugin.yml
@@ -244,13 +244,24 @@ jobs:
           docker network rm plc-demo 2>/dev/null || true
 
   integration-alarms:
-    name: Integration (AlarmConditionType)
+    name: Integration (AlarmConditionType, ${{ matrix.variant }})
     # Issue #386: tests the native OPC-UA AlarmCondition subscription bridge
     # against the test_alarm_server fixture (open62541 with FULL ns0 + alarms
     # ON). Independent of the OpenPLC threshold-mode integration above; runs
     # in parallel.
+    #
+    # Both write surfaces, like the OpenPLC job: acknowledging a condition
+    # changes its state on the controller, so the read-only leg asserts the
+    # refusal and that the condition stays unacknowledged on the server, and the
+    # write-capable leg asserts the acknowledge reaches it. The script builds the
+    # gateway image itself and derives the build arg from the same variable, so
+    # the image and the expectations cannot drift apart.
     runs-on: ubuntu-latest
     timeout-minutes: 30
+    strategy:
+      fail-fast: false
+      matrix:
+        variant: [read-only, write-capable]
     steps:
       - name: Checkout repository
         uses: actions/checkout@v4
@@ -262,6 +273,8 @@ jobs:
           pip3 install --break-system-packages asyncua
 
       - name: Run alarm integration suite
+        env:
+          MEDKIT_OPCUA_VARIANT: ${{ matrix.variant }}
         run: bash src/ros2_medkit_plugins/ros2_medkit_opcua/docker/scripts/run_alarm_tests.sh
 
       - name: Dump container logs on failure
--- a/.github/workflows/opcua-plugin.yml
+++ b/.github/workflows/opcua-plugin.yml
@@ -342,3 +342,91 @@ jobs:
       - name: Show test results
         if: always()
         run: colcon test-result --verbose
+
+  write-capable-build:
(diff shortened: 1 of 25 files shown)


.github/workflows/opcua-plugin.yml
```diff
@@ -141,9 +153,26 @@ jobs:
       - name: Build gateway + OPC-UA plugin image
         run: |
           docker build \
+            --build-arg MEDKIT_OPCUA_READ_ONLY=${{ matrix.read_only }} \
             -f src/ros2_medkit_plugins/ros2_medkit_opcua/docker/Dockerfile.gateway \
             -t gateway-opcua .
 
+      - name: Inspect the plugin object inside the built image
+        # The image is what ships, and Dockerfile.gateway builds it with
```
mfaferek93, line 161: "The image is what ships, and Dockerfile.gateway builds it": the image that ships to DiagBox is `ghcr.io/selfpatch/ros2_medkit-jazzy` (platform `deploy/diagbox/.env`), built by the root `Dockerfile:100` with `colcon build --cmake-args -DBUILD_TESTING=OFF` and no `-DMEDKIT_OPCUA_READ_ONLY`, and `docker-publish-multiarch.yml` has no `inspect_build_variant.py` step. So the shipped plugin is read-only by option default only, and nothing would notice if the default flipped. Passing the flag explicitly in `Dockerfile` and running the docker-cp + inspect step in the publish workflow before push makes the property hold where it matters. Built both flavours on arm64 here: the read-only `.so` has 0 write-marker symbols and 0 exported open62541 symbols, the write-capable one 242, so the mechanism itself is right.

src/ros2_medkit_plugins/ros2_medkit_opcua/src/opcua_plugin.cpp
```diff
@@ -1678,6 +1716,16 @@ tl::expected<dto::DataValue, DataProviderErrorInfo> OpcuaPlugin::read_data(const
 tl::expected<dto::DataWriteResult, DataProviderErrorInfo> OpcuaPlugin::write_data(const std::string & entity_id,
                                                                                   const std::string & resource_name,
                                                                                   const nlohmann::json & value) {
+#if MEDKIT_OPCUA_READ_ONLY
```
mfaferek93, line 1723: A read-only build still has no positive marker from outside: `x-plc-status` carries `mode`, `poll_count` and friends but no build field, and the missing `x-plc-operations` capability is what a write-capable build with zero writable points also shows. #277 asked for one flag. `j["write_capable"] = !MEDKIT_OPCUA_READ_ONLY` in `handle_plc_status`, asserted in `test_opcua_read_only.test.py`, would give the box and the fleet UI something to check.

src/ros2_medkit_plugins/ros2_medkit_opcua/docker/scripts/run_integration_tests.sh
```diff
@@ -65,23 +79,77 @@ assert "tank_temperature has value" "$([ -n "$TEMP" ] && [ "$TEMP" != "null" ] &
...
+else
+    WRITE=$(curl -s -X POST "$API/apps/drain_valve/x-plc-operations/set_valve_position" \
+        -H "Content-Type: application/json" -d '{"value": 50.0}')
+    assert "Write valve position OK" "$(echo "$WRITE" | jq '.status == "ok"' 2>/dev/null)"
+fi
+
+# 8. Error Handling
+echo -e "\n${YELLOW}8. Error Handling${NC}"
 assert "404 unknown entity" "$(curl -s "$API/apps/nonexistent/x-plc-data" | jq 'has("error_code")' 2>/dev/null)"
 assert "404 unknown operation" "$(curl -s -X POST "$API/apps/tank_process/x-plc-operations/nonexistent" -H "Content-Type: application/json" -d '{"value":1}' | jq 'has("error_code")' 2>/dev/null)"
```
mfaferek93: In the read-only leg these two are vacuous: "404 unknown operation" and "400 invalid JSON" both POST to `x-plc-operations`, which is unregistered there, so both get the gateway's generic 404 and `has("error_code")` passes for the wrong reason, while the README table still counts them 3/3. Drive them through `PUT .../data/{name}` (bad JSON, unknown resource) in that leg, or skip them with a note.

src/ros2_medkit_plugins/ros2_medkit_opcua/docker/Dockerfile.gateway
```diff
@@ -7,8 +7,16 @@ FROM ros:jazzy-ros-base AS builder
...
+# rosdep update fetches the distribution index from raw.githubusercontent.com.
+# Both are network reads that a reset connection fails outright, and both run
+# before anything of ours compiles, so a blip fails the whole build for a reason
+# that has nothing to do with the change under test. Three attempts, five
+# seconds apart; each attempt's own error text goes to stderr, so the last
+# failure is what a reader sees.
 # hadolint ignore=DL3008
-RUN apt-get update && apt-get install -y --no-install-recommends \
+RUN retry() { n=1; while :; do "$@" && return 0; [ "$n" -ge 3 ] && return 1; echo "attempt $n of 3 failed: $*" >&2; n=$((n+1)); sleep 5; done; }; \
+    retry apt-get update && apt-get install -y --no-install-recommends \
```
mfaferek93: The retry commit is separate work, and it is half a fix: `apt-get update` gets the retry, `apt-get install`, the two `git clone`s in `openplc/Dockerfile` and the FetchContent clone of open62541pp inside `colcon build` (line 65) do not, and those are the fetches that actually hang. Either its own PR with the whole fetch stage retried, or leave it out of this one.

.github/workflows/opcua-plugin.yml
bburda: The image published today as ghcr.io/selfpatch/ros2_medkit-jazzy:latest is write-capable: 154 write markers and 1751 exported OPC UA symbols, because main has no MEDKIT_OPCUA_READ_ONLY yet. The root Dockerfile now builds the plugin in its own colcon pass with -DMEDKIT_OPCUA_READ_ONLY=ON and -DCMAKE_BUILD_TYPE=Release, and fails the layer when the object is missing. :latest is published by docker-publish.yml on pushes to main, and the multiarch workflow runs on tags, so both got the same step: push by digest, pull that digest, run inspect_build_variant.py on the plugin object, and only then tag. I built the root Dockerfile locally. The inspection passes on the produced image (1.5 MB Release object) and fails on the published one. In f7e7c12f.

src/ros2_medkit_plugins/ros2_medkit_opcua/src/opcua_plugin.cpp
bburda: Added write_capable to x-plc-status, derived from the same build macro that removes the write path. It is documented in the README and the design doc. The read-only integration test asserts false and the write-capable one asserts true, both against a real gateway and OPC UA server. In f7e7c12f.

src/ros2_medkit_plugins/ros2_medkit_opcua/docker/scripts/run_integration_tests.sh
bburda: Both cases now go through the endpoints the read-only build serves and assert status and error code. GET of an unknown point is the plugin's 404. PUT of an unknown point is the build's 501 naming MEDKIT_OPCUA_READ_ONLY, since the refusal precedes the lookup by design. PUT with a malformed body is the gateway's 400 before the plugin is reached, so that case moved to the block both variants share. Every error case in the script asserts status and error now, and the README counts are derived from the script (29 read-only, 26 write-capable). In f7e7c12f.

src/ros2_medkit_plugins/ros2_medkit_opcua/docker/Dockerfile.gateway
bburda: Dropped from this branch: it covered one fetch of four. The whole fetch stage (apt, both clones, the open62541pp fetch) is in a separate issue that proposes a prebuilt base image for this job, so a PR run stops fetching at all.

mfaferek93 (approved):

Issue #659, opened 2026-09-06
[BUG] gateway_node segfaults on shutdown when the OPC UA plugin is loaded


bburda:
# Bug report

### Steps to reproduce
1. Build with `colcon build --symlink-install --packages-up-to ros2_medkit_opcua ros2_medkit_gateway` (Jazzy dev image).
2. Start the gateway with the plugin loaded and no server listening:
   ```
   OPCUA_ENDPOINT_URL=opc.tcp://127.0.0.1:14840 OPCUA_NODE_MAP_PATH=<tank_demo_nodes.yaml> OPCUA_DISCOVERY_ENABLED=0 \
   gateway_node --ros-args --params-file docker/gateway_params.yaml \
     -p plugins.opcua.path:=<install>/ros2_medkit_opcua/lib/ros2_medkit_opcua/libros2_medkit_opcua_plugin.so
   ```
3. Wait 10 s, send SIGTERM, read the exit code.

Control: the same params with the `plugins` key removed.

### Expected behavior

A clean exit (0) on SIGTERM, with or without the plugin.

### Actual behavior

Exit 139 whenever the plugin is loaded, 3 of 3 runs on `main` at 8b1969f8 (0.7.0). Without the plugin the same teardown exits 0, 3 of 3. The crash lands in the node destructor (`node.reset()` in `main.cpp`) after the plugin has already logged `OPC-UA plugin shutdown complete`. Discovery was off and the endpoint pointed at a closed port, so the comms-lost path was active and no server was involved.

```
Thread 1 "gateway_node" received signal SIGSEGV, Segmentation fault.
#0  rclcpp::CallbackGroup::~CallbackGroup ()          from librclcpp.so
#2  rclcpp::node_interfaces::NodeBase::~NodeBase ()    from librclcpp.so
#5  rclcpp::Node::~Node ()                             from librclcpp.so
#6  std::_Sp_counted_base<...>::_M_release_last_use_cold ()
#7  main ()
```

Every `docker stop` of a gateway with the plugin ends in a core dump instead of a clean exit, which hides real shutdown problems behind a constant one and makes exit codes useless for supervisors.

### Environment
- ros2_medkit version: 0.7.0 (`main` at 8b1969f8)
- ROS 2 distro: Jazzy
- OS: Ubuntu 24.04 (the dev container)

### Additional information

[...]

PR #660, opened 2026-09-06
opcua: rescan while disconnected, heal PLC_COMMS_LOST on connect, and name a plugin freeze-frame's source

### LINKED ISSUE #658: opcua: rescan while disconnected, heal PLC_COMMS_LOST on connect, and name a plugin freeze-frame's source
## Summary

Running the OPC UA plugin config-less against an OpenPLC server (`OPCUA_DISCOVERY_ENABLED=1`,
`OPCUA_DISCOVERY_SUBNETS` set, no `OPCUA_ENDPOINT_URL`, no node map) shows four
behaviours that a single deployment hits in one boot.

1. Discovery is a one-shot scan at start-up. A gateway that powers on with its controller
   scans while the controller is still coming up, finds nothing, falls back to
   `opc.tcp://localhost:4840` and retries that endpoint for as long as it runs. Only a
   restart finds the controller. `OPCUA_DISCOVERY_INTERVAL_S` is parsed and only warned
   about. Follows #509 and #544, and the same shape of fix as the ADS re-probe.
2. `PLC_COMMS_LOST` raised before a restart stays CONFIRMED after the plugin reconnects.
   The clear is sent only from the poller's reconnect arm and only when this process was
   the one that raised it, so after a restart the first connect succeeds inside
   `set_context()` and nothing clears the persisted fault. Follows #496 and #564.
3. A freeze-frame on a plugin-backed entity carries an empty `topic` and an empty
   `message_type`, correctly (the values are not a ROS message), which leaves no field
   at all naming where the values came from. Two such frames from different bridges are
   indistinguishable, and so is a frame from a topic capture whose metadata went missing.
4. The gateway lists the nodes it runs inside its own process (`<gateway>_sub`,
   `<gateway>_fault_clients`, `<gateway>_lifecycle_state_reader`, and the gateway node
   itself) as apps. None of them starts with an underscore, so the hidden-node rule does
   not cover them, and the self-node predicate used for the start-up peer count is not
   applied to app discovery.

Two more came out of fixing the above and belong with them, because they reach an
operator on the same deployment.

5. The per-entity SOVD `DELETE /{entity}/faults/{code}` route lands on the plugin's
   `FaultProvider::clear_fault` for a plugin-owned entity, which sends
   `skip_correlation_auto_clear` off. An operator scoped to one entity can therefore
   cascade-clear correlated symptoms reported by apps in other entities, which is the
   opposite of what the `ClearFault` contract documents for scoped routes.
6. A start-up discovery sweep is not cancellable. It runs inside `set_context()`, before
   the executor exists, so a SIGTERM during a wide sweep waits the sweep out.

## Proposed solution

What should hold afterwards:

1. A gateway started before its server adopts that server without a restart, within two
   rescan intervals, and its SOVD component identity is re-derived from the device
   instead of staying pinned to the fallback endpoint.
2. An explicit `interval_s: 0` keeps discovery on with the start-up scan only, an unset
   `interval_s` takes the 30 s default, and a negative value is refused. The cadence is
   measured from the end of a sweep and the reconnect backoff does not stretch it.
3. `PLC_COMMS_LOST` is cleared on every successful connect, including a first connect in
   a process that never raised it.
4. A link-state clear and a scoped operator clear set `skip_correlation_auto_clear`. A
   clear the device itself reports still cascades.
5. A freeze-frame on a plugin entity carries `x-medkit.source` naming the capture path,
   and the field is absent (not empty) on a fault-manager frame taken from a real topic.
6. `/api/v1/apps` on a gateway with the plugin loaded lists none of the gateway's own
   in-process nodes, and a peer named `<node>_monitor` or `<node>2` is still listed.
7. A SIGTERM ends a start-up or rescan sweep within one probe.

How to prove it:

```bash
# Unit and lint
./scripts/test.sh
./scripts/test.sh lint

# Docker, from the plugin's docker directory
bash scripts/run_integration_tests.sh
bash scripts/run_alarm_tests.sh

# The start-up race, from the repo root
bash src/ros2_medkit_plugins/ros2_medkit_opcua/docker/scripts/run_discovery_race_test.sh
```

The race scenario is the falsifying one. On a build without the fix it stops at the
adoption step with `endpoint still 'opc.tcp://localhost:4840' (connected='False') after
50s`. With the fix it runs through to the config-less rename, and it also asserts the
container never restarted, since a restart would satisfy the endpoint check while
proving nothing. The existing docker suites all pin `OPCUA_ENDPOINT_URL` and so
short-circuit discovery, which is why none of them can see the start-up race.

## Additional context

Size: M. Plugin lifecycle and the fault path, plus two gateway-side changes that land on
plugin entities, with one breaking config change (`interval_s: 0`) announced in the
plugin changelog.

Where the work lives: `opcua_plugin.cpp` and `opcua_poller.cpp` (rescan on the reconnect
arm, endpoint adoption, identity re-derivation, the connect-time clear, the clear origin,
the pending-dispatch buffer), `network_discovery.cpp` (the cancel predicate and the
report levelling), `gateway_node.cpp` (one self-node predicate for the peer count and the
app filter), `entity_freeze_frame_capture.cpp` and `http/handlers/fault_handlers.cpp`
(the frame's capture path and `x-medkit.source`), the plugin README, `docs/api/rest.rst`,
`docs/tutorials/snapshots.rst`, the plugin `CHANGELOG.rst`, unit coverage in the plugin
and gateway suites, and a new docker scenario at
`src/ros2_medkit_plugins/ros2_medkit_opcua/docker/scripts/run_discovery_race_test.sh`.

Reproduced against an OpenPLC server on a controller reached by network discovery with
no node map. Related closed work: #496 (comms-lost raised), #564 (comms-loss fault
without a freeze frame), #478 (active-condition replay on reconnect), and the merged
discovery work in #509 and #544. Fault storage is untouched.

bburda:
## Summary

In config-less discovery mode (`OPCUA_DISCOVERY_ENABLED=1`, subnets set, no
`OPCUA_ENDPOINT_URL`, no node map), a gateway that boots before its controller never
finds it. The start-up scan runs while the controller is still coming up, finds nothing,
falls back to `opc.tcp://localhost:4840` and retries that endpoint until someone
restarts it. Three more things read wrong beside it: the gateway lists its three helper nodes
as apps, a plugin entity's freeze-frame names no origin, and a
`PLC_COMMS_LOST` raised before a restart stays CONFIRMED, because only the process that
raised it would clear it.

CI also showed the gateway listing a node after its process exited. The DDS RMWs can
re-create a removed participant from a late discovery message, and the graph then lists
the node, with no enclave and no endpoints, until the gateway stops. graph_watchdog reads
presence from the same list, so it can miss that node's death. The parameter_beacon and
topic_beacon configuration checks let NaN, infinite and out-of-range values through; they
are fixed here too.

## What changes

Discovery and reconnect

1. With no session up and no `endpoint_url` configured, the reconnect arm rescans on
   `discovery.interval_s`, adopts a server that appeared after start-up and resets the
   backoff. The cadence is measured from the end of a sweep, the backoff is capped at it
   so the documented interval is the real one, and an unchanged outcome reports at DEBUG.
   The plugin is bound to the ApplicationUri of the server a session reached: the start-up
   sweep and every rescan look for that server at any address and refuse every other, and a
   different server answering at the bound address is dropped at connect with the outage
   left standing. The binding is written to `discovery.binding_file` (default
   `/var/lib/ros2_medkit/opcua/binding`, env `OPCUA_DISCOVERY_BINDING_FILE`, empty
   disables) when it is established and read at start, so a restart keeps the identity;
   removing the file and restarting is the recommissioning gesture. The first adoption of a
   process with no file, a configured `endpoint_url` and a server that publishes no
   ApplicationUri are outside the binding. One file belongs to one plugin instance.
   The reconnect arm owns every reconnect: after each client iterate the session state is
   read, a session the library re-opened underneath is dropped and handed to the arm, which
   connects through its own path, runs the binding check and re-creates the alarm
   subscription. Every path that marks the client disconnected drops the session's
   subscription bookkeeping, so the re-created monitored item delivers again.
2. Both sweeps take a cancel predicate that reads the node's own context. A rescan that
   throws is caught on the poll thread, logged at WARN, and still stamps the cadence.
3. With no node map, the component identity is re-derived from the device once a session
   is up, before the session subscribes to events, so the provisional `opcua-<host>` name
   goes and the conditions the server replays land on the new entity. A name the device
   gave itself is never replaced by the fallback; an empty read is retried a bounded
   number of times per session (one read at connect, up to three more in the connect
   hook, then at most five on later polls). Conditions the poller already pinned under
   the old alarms entity are moved to the new one; the event path reads a copy of the
   alarm routing owned by the poller, refreshed from the poll thread.

Comms-lost clears and the pending buffer

4. On every successful connect the plugin asks the fault manager who reported a standing
   `PLC_COMMS_LOST` and clears it only when every reporting source is an id this bridge
   reports under: the current component id and the `opcua-<host>` stand-ins it assigned
   in this process. A row another bridge holds is left standing. The decision runs on the
   poll thread, only while the link is up, with one fault-store read outstanding at a
   time; a read the store does not answer within `fault_service_timeout_ms` (new key,
   default 5000) is dropped and taken again, and an answer that arrives after the link
   dropped is held until the next connect. A link-state clear the bounded buffer could
   not keep is owed and decided again once the sink is ready.
5. Link-state clears and the scoped operator clear set `skip_correlation_auto_clear`. A
   clear the device itself reports still cascades.
6. The buffer used while `fault_manager` is unreachable keeps the PLC's order: a clear is
   coalesced onto a pending clear for the same code only when nothing was reported for
   that code in between, so report, clear, report, clear flushes in that order. A device
   alarm's inactive edge ages out like a report.

Gateway entity listing

7. One predicate names the gateway's three helper nodes (`<node>_sub`,
   `<node>_fault_clients`, `<node>_lifecycle_state_reader`) for the peer count and the app
   filter, matched per exact suffix so a peer named `<node>_monitor` or `<node>2` stays
   visible. The gateway's own node stays an App: its parameters are that App's
   configurations. `_sub` follows the gateway's namespace; the other two are created in the
   root namespace, so a helper-named node there counts as plumbing whichever gateway created
   it. While `discovery.runtime.filter_internal_nodes` is true the same predicate keeps the
   helpers out of `/health`'s `linking.orphan_count` and out of the `unmanifested_nodes`
   warning; with the flag false they are served as Apps and counted there like any other
   undeclared node. A manifest App bound to a helper is dropped with one warning per
   distinct set. The peer count excludes `_lifecycle_state_reader`. Remote entities are
   untouched.

Freeze-frame source

8. An entity frame carries `x-medkit.source`, `plugin_data_provider` or
   `plugin_x_plc_data_route`. `topic` and `message_type` stay empty, and the field is
   omitted when the capture named no path.

Docs

9. Plugin README, `docs/api/rest.rst`, `docs/tutorials/snapshots.rst`,
   `docs/config/discovery-options.rst`.

Graph leftovers

10. The node-list reader (`ros2_common/graph_node_list.hpp`, header-only) leaves out an
    entry without an enclave for a node it saw running and that has departed, when the
    graph resolves no publisher or subscriber for it. A node it never saw running is
    listed, so nodes behind a DDS router and micro-ROS nodes read as they do on main. A
    departed name is remembered until it runs again, or until a read finds no entry of it
    more than 10 s after the first read that found none. At most 1024 departed names are
    remembered. Discovery logs one warning when it starts leaving a node out. Discovery and
    the startup peer count share one reader; parameter_beacon has its own.
11. A leftover also stays out of the Functions derived from namespaces and out of the
    bare-name collision check, so it cannot rename a live node that shares its name.

Discovery plugins

12. parameter_beacon keeps one parameter client per target across cycles, on its own
    executor. A request that times out is removed from its client, so a node that never
    answers leaves nothing pending. The list and get waits share one deadline,
    `param_timeout_sec`. A get answer with no values (rclpy does this for a typed parameter
    without a value) stores no hint and causes no backoff. The gateway's own node and its
    helper nodes are not polled. A graph read with no targets drops every client.
13. Duration keys of both beacons: NaN, -inf and values below the minimum become the
    minimum; values above 2147483647 s, +inf included, become 2147483647 s, because Fast
    DDS keeps the seconds of a wait in a signed 32-bit field. Each replacement is logged.
14. topic_beacon `max_messages_per_second` is limited to 1..10000.
15. `max_hints` is read as a 64-bit integer before it is narrowed: below 1 becomes 1,
    above 2147483647 becomes 2147483647, with a warning. The parameter parser reads an
    integer that does not fit in 64 bits as a double. A double, such as 1e12 or NaN, keeps
    the default of 10000.
16. Docs: `docs/config/server.rst` (departure section), `docs/config/discovery-options.rst`,
    the READMEs of the gateway, parameter_beacon, topic_beacon and graph_watchdog, and the
    two beacon design pages.

## Behaviour changes to know about

1. `interval_s` unset means the 30 s default, an explicit `0` means discovery on with no
   rescan. The old README line was `interval_s: 0`, so a config copied from it gains no
   rescanning on upgrade. Drop the key for the default. A negative value is refused.
   This one is breaking.
2. A per-entity `DELETE /{entity}/faults/{code}` served by this plugin now sets
   `skip_correlation_auto_clear`, as the `ClearFault` contract already promised.
3. Three `/api/v1/apps` entries disappear with the default `filter_internal_nodes: true`:
   the gateway's helper nodes. The gateway's own node stays.
4. A SIGTERM ends a start-up or rescan sweep within one probe.
5. A `PLC_COMMS_LOST` that two bridges raised into one fault manager is not cleared by
   either link coming back; an operator clears it. The fault manager has no per-source
   de-assert today.
6. New plugin config key `fault_service_timeout_ms` (default 5000, out of range refused
   with a warning). `comms_lost_debounce_ms` logs a warning when it clamps.
7. New plugin config key `discovery.binding_file`. The identity survives a restart; a
   replaced PLC at the bound address is refused until the file is removed. In the shipped
   image the directory exists but persists across a re-created container only when mounted.
8. A PLC swapped or rebooted under a config-less session with native alarms is now
   noticed: the outage is reported and the alarm subscription is re-created on the bound
   server when it returns. Before, the client library re-opened the session silently and
   device alarms stayed dead until a restart.
9. `GET /apps` stops listing a node the gateway saw running once only a leftover of it
   remains, and the gateway logs one warning for it.
10. topic_beacon `max_messages_per_second` above 10000 is lowered to 10000 with a warning.
    A config with a higher rate, or with `.inf` to remove the limit, now gets 10000 per
    second.
11. parameter_beacon no longer polls the gateway's own node and helper nodes.
12. Out-of-range beacon values are replaced as described in "What changes", items 13-15,
    with a warning. On main a
    NaN passed the topic_beacon checks: it then dropped every beacon or kept hints forever.

## Testing

1. Jazzy, clean tree: the opcua unit suite (405 tests) and the gateway unit suite pass with 0 failures, and the full integration package passes 1285 tests with 0 failures; the four
   contract tests that failed on every distro (`test_configuration_api`,
   `test_locking_disabled_contract`, `test_relay_peer_credential`,
   `test_unmanifested_policy`) pass, with three new feature tests beside them
   (`test_own_node_apps`, `test_own_node_apps_namespaced`, `test_own_node_undeclared`)
   that fail when the helper filter or the own-node rule is broken on purpose.
   `test_operation_handlers` pins the stop response on the contract (goal CANCELING or
   CANCELED), 10 of 10. The alarm-server test fixture runs every
   `UA_Server_*` call on the server's own thread, as `UA_MULTITHREADING 0` requires, and
   the identity tests tear their executor down on every exit path.
2. Plugin e2e against the fixture server with stub fault services: a fault raised under
   the stand-in heals after the device names itself; a row two bridges hold is left
   standing; no decision is taken while the link is down; a probe the store never answers
   is dropped and taken again; an answer to a timed-out probe is not read as the next
   one's; a link-state clear the buffer refused is sent last once the sink is ready; the
   connect hook runs before the event routing is copied; a rename moves the conditions
   already pinned. Unit tests pin the gate predicate, the decision predicates, the buffer
   order (a clear never merges past a report for the same code), the identity rule, the read
   budget and the burst cadence, the guarded rescan, the node-context cancel, and the
   linker skip in both flag positions. Each new test was made to fail first. The binding
   file has unit tests for its format (one line, BOM, control characters, CRLF, a second
   line, a symlinked path, an unwritable parent) and e2e tests against the fixture: a
   restarted plugin keeps the binding and refuses a foreign server at the bound address, a
   configured `endpoint_url` ignores the file, the start-up sweep selects the bound server
   and not the lowest address, a swap and an outage under a config-less session are caught,
   and a device alarm is delivered after the bound server reboots.
   `test_openapi_contract` sweeps every entity, requires a built item for every listed
   resource, checks the item each one names, and waits for calibration's operation through
   the shared class wait.
3. Docker: `run_integration_tests.sh` `Total: 18 / Passed: 18`, `run_alarm_tests.sh`
   passed, `run_discovery_race_test.sh` passed. That race scenario fails on the merge
   base at step 6 of 9, the adoption assertion (`endpoint still
   'opc.tcp://localhost:4840' (connected='False') after 50s`), and passes on the branch
   through step 9 of 9, the config-less rename, which now also asserts the stand-in
   component left `/components`.
4. Scoped clear live: `PLC_TANK_HIGH` still CONFIRMED after `PLC_COMMS_LOST` was cleared
   through the per-entity route. The control with the flag off cascades.
5. SIGTERM 14 s into a start-up sweep: container gone in 506 ms, against 98 s
   uncancelled.
6. TSan is verified by CI only: a sanitizer build of this package alone links against an
   uninstrumented rclcpp and is not usable on the development machine.
7. Graph leftovers are reproduced end to end. A fixture node runs in an rclcpp context of
   its own; its participant is removed, and its captured discovery message is published
   again afterwards, as the late message does. Every case first checks that the test's own
   graph lists the leftover. `test_graph_leftover_nodes`, `test_graph_leftover_nodes_scale`
   (past the 1024 cap) and `test_graph_leftover_nodes_startup` check discovery, the startup
   count and parameter_beacon. graph_watchdog's `ghost_departure` e2e checks that the
   node's death is still raised. Each rule element was mutated and turned a test red.
8. `test_param_beacon_out_of_range_config` and `test_topic_beacon_config_bounds` sweep
   NaN, +inf, -inf, below the minimum, the minimum, the maximum, above it and 3e9 on a real
   gateway, plus `max_hints` 0, -5, 1, 2147483647, 2147483648, 2^32 + 1, 1e12 and NaN. A
   beacon with NaN TTL and expiry is observed going stale and then removed.
9. Jazzy with Fast DDS on the final tree: gateway 4342 tests, integration 1285,
   parameter_beacon 44, topic_beacon 30, all with 0 failures. The changed tests also pass
   on Cyclone DDS.

## Not in this PR

1. A docker scenario that restarts the gateway with a standing `PLC_COMMS_LOST`. That
   case has plugin-level e2e coverage with stub fault services.
2. An empty `source_id` on `PLC_COMMS_LOST` in the `/api/v1/faults` list view while the
   fault manager logged a source. Unverified, filed on its own if it holds.
3. `gateway_node` exiting 139 on shutdown with the plugin loaded. It reproduces on `main`,
   see #659.
4. A per-source de-assert in the fault manager, which is what would let two bridges heal
   one shared `PLC_COMMS_LOST` row.
5. A node restarted while a leftover of its previous instance is still listed can show no
   services, and so no operations, because the graph answers per-node queries from
   whichever of the two participants sorts first. `docs/config/server.rst` states it.

## Issue

- closes #658

## Type

- [x] Bug fix
- [x] New feature or tests
- [x] Breaking change
- [ ] Documentation only

## Checklist


src/ros2_medkit_plugins/ros2_medkit_opcua/src/opcua_plugin.cpp
```diff
@@ -1276,42 +1312,111 @@ void OpcuaPlugin::send_report_fault(const std::string & entity_id, const std::st
...
+  // fire-and-forget, so a "Fault not found" answer for a code that was never
+  // raised costs nothing here and is the normal case on a healthy start.
+  //
+  // LinkState: this says the session came back, not that an operator resolved
+  // anything, so a correlation rule naming PLC_COMMS_LOST as a root cause must
+  // not cascade-clear the symptoms the outage produced. It is also the one clear
+  // the next reconnect re-derives, so the pending buffer may drop it before
+  // anything one-shot.
+  log_info(std::string("OPC-UA connection established, clearing any standing ") + kCommsLostFaultCode);
+  send_clear_fault(kCommsLostFaultCode, ClearOrigin::LinkState);
```
mfaferek93, line 1356: The fault store keys `PLC_COMMS_LOST` by code alone and the S7/ADS/Modbus bridges raise the same code into the same fault manager, so this clear on first connect (and the poller's clear on every reconnect at `opcua_poller.cpp:1216`, no longer gated on `comms_lost_raised_`) flips another bridge's CONFIRMED outage to CLEARED every time the OPC UA link comes up, gateway start included. Before sending, `GetFault` the code and only clear when `reporting_sources` names this component, so the restart case is still healed without touching a fault this plugin never raised.

src/ros2_medkit_plugins/ros2_medkit_opcua/src/opcua_plugin.cpp
```diff
@@ -1276,42 +1312,111 @@ void OpcuaPlugin::send_report_fault(const std::string & entity_id, const std::st
...
+  // At most one pending clear per fault code. A repeat moves to the BACK rather
+  // than overwriting in place, so an interleaved report-then-clear for the same
+  // code still flushes in the order the PLC produced it.
+  bool replaced = false;
+  if (is_clear) {
+    const auto same_code = std::find_if(buffer.begin(), buffer.end(), [&entry](const PendingFaultDispatch & pending) {
+      return pending.kind == PendingFaultDispatch::Kind::Clear && pending.fault_code == entry.fault_code;
+    });
+    if (same_code != buffer.end()) {
+      buffer.erase(same_code);
```
mfaferek93, line 1372: This erase runs before the capacity check and ignores a Report for the same code sitting after the older clear, so a buffered `Report(A), Clear(A), Report(A), Clear(A)` flushes as `Report(A), Report(A), Clear(A)`: `occurrence_count` stays at 1, one OCCURRED audit record is lost, and with a confirmation threshold below -1 the two FAILEDs stack the debounce counter and confirm a fault the real stream never confirmed. Only coalesce when the newest pending entry for this code is already a Clear (search from the back, stop at the first entry with this code), so `Clear, Clear` from repeated reconnects still collapses.

src/ros2_medkit_plugins/ros2_medkit_opcua/src/opcua_plugin.cpp
```diff
@@ -1396,6 +1501,56 @@ void OpcuaPlugin::run_auto_browse() {
            (result.depth_cap_hit ? " [depth cap reached on at least one branch]" : ""));
 }
 
+std::optional<ComponentIdentity> OpcuaPlugin::rederived_component_identity(const std::string & current_id,
+                                                                           const OpcuaClient::DeviceInfo & info,
+                                                                           const std::string & endpoint_url) {
+  const ComponentIdentity ci = derive_component_identity(info, endpoint_url);
+  if (ci.id.empty() || ci.id == current_id) {
```
mfaferek93, line 1508: `connect_generation` bumps on every fresh connect (`opcua_client.cpp:539`), so this re-derives on a plain reconnect to the same PLC, not only after an adoption, and it accepts any different id: one empty or partial `read_device_info` on the first read of the new session renames `siemens_ag_cpu_1505sp_f` to `opcua-192_168_1_10`, moves the component URL and the `_alarms` entity and orphans the faults standing under the old id, and since `component_identity_generation_` is stamped before the result is checked nothing retries until the next session. Only re-derive while the current id is still the endpoint-derived placeholder, never replace a nameplate-derived id with the host fallback.

src/ros2_medkit_plugins/ros2_medkit_opcua/src/opcua_plugin.cpp
```diff
@@ -1396,6 +1501,56 @@ void OpcuaPlugin::run_auto_browse() {
...
+  {
+    // Serialize against the REST read paths, which hold references into
+    // node_map_ (entity_defs) while they answer.
+    std::unique_lock<std::shared_mutex> lock(node_map_mutex_);
+    // The auto_alarms fallback entity is derived from the component id, so a
+    // default-derived one has to follow the rename. An operator-chosen entity_id
+    // does not match the derived form and is left alone.
+    auto & auto_alarms = node_map_.mutable_auto_alarms();
+    if (auto_alarms.entity_id == previous_id + "_alarms") {
+      auto_alarms.entity_id.clear();
```
mfaferek93, line 1543: This runs on the poll thread and clears `auto_alarms_.entity_id`, then reassigns it and rebuilds `entity_defs` in `finalize_auto_alarms_overlay`, all under `node_map_mutex_`, but the poller never takes that mutex: its event pump thread delivers `on_event` from `run_iterate` and reads `node_map_.auto_alarms().entity_id` and `find_by_node_id` lock-free (`opcua_poller.cpp:920, 975`). The ConditionRefresh burst lands on exactly this first session, so a `std::string` read races a clear-then-reassign and the fault report can carry a garbage entity_id. Give the poller its own copy of the alarm config taken at subscribe time and swap it from the poll thread.

src/ros2_medkit_plugins/ros2_medkit_opcua/src/opcua_plugin.cpp
```diff
@@ -1414,6 +1569,11 @@ void OpcuaPlugin::publish_values(const PollSnapshot & snap) {
   // Poll-thread hook: drain any fault reports buffered before fault_manager was
   // discovered, so a late sink still receives them.
   flush_pending_reports();
+  // Poll-thread hook: re-derive the config-less component identity from the
+  // device once a session is up, so an adopted PLC stops being served under the
+  // provisional endpoint-derived id. Runs BEFORE the re-walk below, which
+  // rebuilds entity_defs off the component id.
+  maybe_rederive_component_identity();
```
mfaferek93, line 1576: This runs too late for the alarms entity: in the reconnect arm the poller has already run `setup_event_subscriptions()` and ConditionRefresh before `poll_callback_` fires, so conditions replayed on the adopted session get pinned in `conditions_` (`apply_condition_state` keeps the first entity_id) and reported under `<old-id>_alarms`, an entity `finalize_auto_alarms_overlay` then drops from `entity_defs`, leaving those faults and all their later clears orphaned. Add a connected hook to `PollerConfig` (next to `rediscover_endpoint`) that the reconnect arm calls right after `connect()` and before `setup_event_subscriptions()`, and do the rename there.

src/ros2_medkit_plugins/ros2_medkit_opcua/src/opcua_plugin.cpp
```diff
@@ -1508,37 +1668,124 @@ void OpcuaPlugin::log_security_profile() const {
...
+  // thread would sweep back to back and only attempt a reconnect once a sweep.
+  // A sweep that threw still consumed that time, so the stamp is owed either
+  // way, or the next poll iteration would start another one immediately.
+  try {
+    const auto result = sweep();
+    *last_scan_end = now();
+    return result;
+  } catch (...) {
+    *last_scan_end = now();
+    throw;
```
mfaferek93, line 1712: This rethrow lands on the poll thread with nothing above it to catch it: `rescan_endpoint_for_reconnect`, the `rediscover_endpoint` lambda in `set_context` and `OpcuaPoller::poll_loop` have no try, so any exception out of the sweep (`std::system_error` from thread creation in `parallel_for` with `scan_concurrency=100`, `bad_alloc` on a /16 target list) is `std::terminate` for the whole gateway, not a skipped rescan. The "next poll iteration" rationale here and in `RescanStep.AThrowingSweepStillStampsTheCadence` assumes the thread survives. Catch in `rescan_endpoint_for_reconnect`, log at WARN and return nullopt.

src/ros2_medkit_plugins/ros2_medkit_opcua/src/opcua_plugin.cpp
```diff
@@ -1562,26 +1809,131 @@ void OpcuaPlugin::run_startup_discovery() {
...
+                                 discovery_reporter(&last_discovery_outcome_), [this]() {
+                                   return discovery_cancelled();
+                                 });
+      });
+  // The live client config, not client_config_: this runs on the poll thread
+  // and client_config_ is read by the refresh thread in introspect(). The
+  // client owns the endpoint once connect() has been called with it, and its
+  // accessors are mutex-guarded.
+  const std::string current = client_ ? client_->endpoint_url() : client_config_.endpoint_url;
+  if (!chosen || *chosen == current) {
```
mfaferek93, line 1931: This only checks that the rescan hit is a different URL, so once the adopted PLC drops and any other None/Anonymous data server is reachable on the subnet (a second PLC, Kepware, an engineering station), the next rescan hands that server to the poller, which connects, logs "connection restored", clears `PLC_COMMS_LOST` and in config-less mode renames the component after the new device. The real outage vanishes from diagnostics, which the merge base never did because it retried the adopted endpoint forever. Only rescan while the current endpoint has never carried a session, or accept a hit only when its ApplicationUri matches the adopted server.

src/ros2_medkit_plugins/ros2_medkit_opcua/src/opcua_plugin.cpp
```diff
@@ -1276,42 +1312,111 @@ void OpcuaPlugin::send_report_fault(const std::string & entity_id, const std::st
...
+    // Only a link-state clear is re-derivable: the next reconnect sends it
+    // again. A full buffer gives that up first, and refuses an incoming one
+    // rather than pushing out a dispatch nothing will re-send. Everything else -
+    // reports, a device alarm's inactive edge, an operator's scoped clear - is
+    // one-shot and ages out oldest-first.
+    const auto oldest_link_state = std::find_if(buffer.begin(), buffer.end(), [](const PendingFaultDispatch & pending) {
+      return pending.kind == PendingFaultDispatch::Kind::Clear && pending.link_state;
+    });
+    if (oldest_link_state != buffer.end()) {
+      buffer.erase(oldest_link_state);
```
mfaferek93, line 1388: This eviction assumes the next reconnect re-sends the clear, but when the clear was queued by a successful connect the link is up and no reconnect is coming: restart the gateway with a persisted `PLC_COMMS_LOST` while fault_manager is down, let 256 alarm edges arrive before it returns, and the flush leaves `PLC_COMMS_LOST` CONFIRMED against a healthy link until the next drop. The Refused branch below has the same hole. Set a flag when a link-state clear is evicted or refused and re-run `clear_comms_lost_on_connect()` from `flush_pending_reports` once the sink is ready.

src/ros2_medkit_plugins/ros2_medkit_opcua/src/opcua_plugin.cpp
```diff
@@ -1562,26 +1809,131 @@ void OpcuaPlugin::run_startup_discovery() {
...
+
+  client_config_.endpoint_url = *chosen;
+  log_info("OPC-UA discovery: auto-selected endpoint " + *chosen + " - handing to the connect + introspect path.");
+}
+
+bool OpcuaPlugin::discovery_cancelled() const {
+  // rclcpp::ok() only reads the default context's atomic shutdown flag, so it is
+  // safe to call from the set_context thread and the poll thread alike. The rule
+  // itself, and why both signals are needed, lives in discovery_cancelled_for.
+  return discovery_cancelled_for(shutdown_requested_.load(), rclcpp::ok());
```
mfaferek93, line 1887: `rclcpp::ok()` here reads the process-global default context, not the context the gateway node lives in, so any host that constructs `GatewayNode` with `NodeOptions().context(...)` gets `discovery_cancelled()` permanently true: the start-up sweep returns `{}` before a single probe and `rescan_endpoint_for_reconnect` bails at its top check, so the promised rescan never happens. Grab `node->get_node_base_interface()->get_context()` in `set_context` (as `graph_watchdog_plugin.cpp:379` does), keep it as a member and pass it to `rclcpp::ok(ctx)`.

src/ros2_medkit_gateway/src/discovery/manifest/runtime_linker.cpp
```diff
@@ -182,9 +183,21 @@ LinkingResult RuntimeLinker::link(const std::vector<App> & manifest_apps, const
...
+  // entities and the app filter removes them, so reporting them here would
+  // tell the operator to declare, in the manifest, nodes that can never
+  // become apps - and the 'error' policy words that report as an instruction
+  // ("Declare them in the manifest"). An empty self FQN, which is what a
+  // linker constructed without a node has, matches nothing and leaves this
+  // behaviour exactly as it was.
+  const std::string self_fqn = node_ != nullptr ? node_->get_fully_qualified_name() : std::string();
   for (const auto & rt_app : runtime_apps) {
     if (rt_app.bound_fqn.has_value() && matched_nodes.find(rt_app.bound_fqn.value()) == matched_nodes.end()) {
+      if (is_own_gateway_helper_node(rt_app.bound_fqn.value(), self_fqn)) {
```
mfaferek93, line 198: This skip runs regardless of `discovery.runtime.filter_internal_nodes`, but the app filter in `refresh_cache` is gated on that flag and discovery-options.rst says `false` re-exposes the three helpers. With the switch off the helpers are served as Apps yet never appear in `/health`'s `unmanifested_nodes` or `orphan_count`, while `_ros2cli_*` nodes appear in both. Either gate this on the same setting or state in discovery-options.rst that the health exclusion is unconditional.

src/ros2_medkit_plugins/ros2_medkit_opcua/docker/scripts/run_discovery_race_test.sh
```diff
@@ -0,0 +1,321 @@
...
+
+# The renamed component is the one actually polling the adopted PLC, so the id
+# the operator sees is not a second, stale entity next to a live opcua-localhost.
+endpoint="$(status_field_for "${renamed}" endpoint_url)"
+connected="$(status_field_for "${renamed}" connected)"
+[[ "${endpoint}" == "opc.tcp://${SERVER_IP}:${SERVER_PORT}" ]] \
+  || fail "renamed component reports endpoint '${endpoint}', expected the adopted server"
+[[ "${connected}" == "True" ]] \
+  || fail "renamed component reports connected='${connected}', expected a live session"
+[[ "${renamed}" != "${FALLBACK_COMPONENT_ID}" ]] \
```
mfaferek93, line 313: `renamed` can only be `DEVICE_COMPONENT_ID` here (set at 297, otherwise `fail` at 302 already exited), so this check cannot fail, and nothing asserts that `opcua-localhost` actually left `/components`, which is what the header and the comment at 305 say this pass is for. Assert `FALLBACK_COMPONENT_ID` is absent from `ids` after the rename. The `RestartCount` checks at 221 and 317 cannot fail either, the containers run with no `--restart` policy so it is always 0; compare `.State.StartedAt` before and after, or drop them.

src/ros2_medkit_plugins/ros2_medkit_opcua/src/opcua_plugin.cpp
bburda: Fixed in 276c374f. The connect-time clear and the poller's clear on every reconnect now both go to one decision on the poll thread (`drive_comms_lost_decision` in opcua_plugin.cpp), taken only while the link is up. It asks the fault manager with GetFault who reported the standing PLC_COMMS_LOST. It clears only when every reporting source is an id this bridge reports under: the current component id, the stand-in ids it assigned, and the NodeMap default. A row that another bridge also holds stays standing, because ClearFault clears the whole row and the fault manager has no per-source de-assert; the README says an operator clears it. One read is in flight at a time. An unanswered read is dropped after fault_service_timeout_ms and sent again, and an answer from before a link drop is discarded. `ASharedCommsLostRowIsLeftStanding` covers the two-bridge case against the test server.

src/ros2_medkit_plugins/ros2_medkit_opcua/src/opcua_plugin.cpp
bburda: Fixed in 276c374f. `enqueue_pending_dispatch` now searches from the back and stops at the first entry with the same code. It merges the incoming clear only when that entry is a clear. A report there means the code was raised again, so the clear goes behind it. Report, Clear, Report, Clear now flushes in that order, and repeated reconnect clears with nothing between them still collapse to one. `AClearNeverCoalescesPastAReportForTheSameCode` and `RepeatedReconnectClearsCollapseOnlyWhileNothingWasReportedBetween` pin both.

src/ros2_medkit_plugins/ros2_medkit_opcua/src/opcua_plugin.cpp
bburda: Fixed in 276c374f (`rederived_component_identity`, `maybe_rederive_component_identity`). A nameplate replaces a stand-in id, and nothing replaces a nameplate id. A read with no nameplate moves one stand-in to another only when the endpoint host changed, so an empty first read no longer turns siemens_ag_cpu_1505sp_f into opcua-192_168_1_10. The session is stamped only when the device names itself, or when its read budget is spent. An empty read is retried on later polls, up to kMaxIdentityPollReads (5) per session after the connect burst. `RederivedComponentIdentity.*` and `IdentityReads.*` cover the rules, and `AFaultRaisedUnderTheStandInHealsAfterTheDeviceNamesItself` runs them against the test server.

src/ros2_medkit_plugins/ros2_medkit_opcua/src/opcua_plugin.cpp
bburda: Fixed in 276c374f. The poller keeps its own copy of the alarm routing (event_alarms, auto_alarms and the node-id to entity map) behind a shared_ptr (`refresh_alarm_routing` in opcua_poller.cpp). It is taken in `setup_event_subscriptions` and refreshed from the poll thread when the plugin renames the component. `on_event` reads only that copy, so the event pump no longer reads node_map_. `TheConnectedHookRunsBeforeTheEventRoutingIsCopied` checks the order against the test server.

src/ros2_medkit_plugins/ros2_medkit_opcua/src/opcua_plugin.cpp
bburda: Fixed in 276c374f. PollerConfig has an `on_connected` hook. The reconnect arm calls it right after `connect()`, before the link-state edge and before `setup_subscriptions()` and `setup_event_subscriptions()`. `start()` calls it the same way for a session that is already up. The plugin re-derives the identity inside that hook, so conditions replayed by ConditionRefresh land on the new entity. Conditions pinned before a later rename are moved with `repin_auto_alarms_entity`. The hook also runs the binding check, and a session it rejects is dropped with the outage left standing.

src/ros2_medkit_plugins/ros2_medkit_opcua/src/opcua_plugin.cpp
bburda: Fixed in 276c374f. `rescan_endpoint_for_reconnect` runs the sweep through `rescan_guarded`. It catches std::exception and anything else, logs at WARN with a fixed format string, and returns nullopt, so the endpoint stands and the poll thread keeps running. `rescan_step` still stamps the cadence and rethrows, and its comment now names `rescan_guarded` as the caller on the poll thread. `RescanGuarded.AThrowingSweepEndsAsNoEndpoint` pins it.

src/ros2_medkit_plugins/ros2_medkit_opcua/src/opcua_plugin.cpp
bburda: Fixed in 276c374f and 84688539. The plugin binds to the ApplicationUri of the server a session reached. The start-up sweep and every rescan select only that server, at any address, so a foreign server that sorts lower does not win. A different server at the bound address is caught when its session comes up: the live ApplicationUri is read, the session is dropped, no link-state clear is sent and nothing is renamed. The binding is written to `discovery.binding_file` (default /var/lib/ros2_medkit/opcua/binding) and read at start, so a restart keeps it. Replacing a PLC is a recommissioning: remove the file and restart. A configured endpoint_url, a first adoption and a server that publishes no ApplicationUri have no binding, and the README lists these cases (d54e8c92). `TheBridgeStaysBoundToItsOwnServerAcrossAddressAndSwap` and `ARestartedPluginKeepsTheBindingAndRefusesAForeignServer` run this against two test servers.

src/ros2_medkit_plugins/ros2_medkit_opcua/src/opcua_plugin.cpp
bburda: Fixed in 276c374f. When the full buffer evicts a link-state clear, or refuses one, `send_or_buffer` marks the PLC_COMMS_LOST decision as owed. The poll thread takes it once the link is up and the fault manager answers, through the same source check as every other link-state clear. That covers the clear a successful connect queued, where no reconnect is coming. `AnOwedLinkStateClearIsSentAfterTheBufferDrains` holds the buffer full so the connect-time clear is refused, and checks that the clear goes out once the buffer drains.

src/ros2_medkit_plugins/ros2_medkit_opcua/src/opcua_plugin.cpp
bburda: Fixed in 276c374f. `set_context` keeps `node->get_node_base_interface()->get_context()`, and `discovery_cancelled()` passes it to `rclcpp::ok(context)`. A plugin that has no context yet reads the default one. `DiscoveryCancelledForContext.ReadsTheNodesOwnContextNotTheProcessDefault` builds a private context and checks both.

src/ros2_medkit_gateway/src/discovery/manifest/runtime_linker.cpp
bburda: Fixed in 02d72a4e. RuntimeLinker takes `discovery.runtime.filter_internal_nodes` and skips the helper nodes only while it is true. With it false, the helpers are served as Apps and counted like any other undeclared node, in orphan_count and in the unmanifested_nodes warning. discovery-options.rst states the coupling, and `RuntimeLinkerHelperNodes.TheOrphanSkipFollowsTheAppFilterSetting` links the same apps with the setting on and off.

src/ros2_medkit_plugins/ros2_medkit_opcua/docker/scripts/run_discovery_race_test.sh
bburda: Fixed in 77303bb4. The config-less pass now waits for one /components sample that lists the renamed component and no longer lists opcua-localhost, and fails when that sample never comes. The RestartCount checks are gone: the containers run with no restart policy, and a gateway that dies fails the endpoint and component checks.

PR #661, opened 2026-09-07
fault_manager, gateway: declare a planned stop, so a weekend is not a wave of faults

### LINKED ISSUE #656: Declare a planned stop, so a weekend is not a wave of faults
## Summary

A plant stops on purpose: a weekend, a changeover, a maintenance slot. Every controller the fault manager watches goes quiet at once, and there is no way to tell it that this was expected. With a debounce set and healing enabled, every one of those faults confirms and then heals, every transition is a real audit row, and the first Monday of a deployment starts with a fault list nobody can read. The lesson a maintenance team takes from that is to stop looking.

The fix is not suppression. A stop is a fact about the plant, and hiding the faults it produces would make the evidence describe a plant that did not exist. And the bridges report one `FAILED` per transition, so a fault dropped during a stop that outlives the stop would never be reported again while the alarm still stands on the controller.

---

- A planned-stop switch on the fault manager, turned on and off at runtime with a reason. It is a ROS service, so it is already an SOVD operation on the fault manager entity, and needs no new route and no change to any released message.
- While the switch is on, every fault whose cycle starts is recorded, debounced, captured and audited exactly as any other, and is muted the way a correlated symptom is muted: out of the default fault list, counted in `muted_count`, visible with `include_muted`, not published to the stream. The mute names the planned stop as its cause.
- When the switch goes off, faults that are still active are unmuted, take their place in the list and publish their confirmation, because the alarm still stands on the controller. Switching on and off is an audit event.
- Schedules, history and the button belong to the operator's interface, calling the SOVD API.

---

`src/ros2_medkit_fault_manager/` (the switch, the mute path, the audit event); the gateway already exposes the operation and the muted counts.

bburda:
# Pull Request

## Summary

A plant stops on purpose: a weekend, a changeover, a maintenance slot. Every controller the fault manager watches goes quiet at once, and there was no way to tell it that this was expected. With a debounce set and healing on, every one of those faults confirms and heals, every transition is an audit row, and the first Monday starts with a fault list nobody can read.

**A planned stop is a declared window, not a suppression.** An operator declares `{from, to, reason}` through `POST /api/v1/x-medkit-planned-stops`; the gateway forwards it to a new fault manager service, and the window is stored next to the faults, so a restart inside it keeps it. A fault whose cycle started inside a window (its `first_occurred`, which resets when a cleared fault is raised again) is reported as expected: `x-medkit.expected` and `x-medkit.planned_stop_id` on every fault item, `x-medkit.expected_count` on every fault list including aggregated ones, an `expected=true|false|all` query on all five fault-list routes, and the same flag on the fault event stream. Storage, debounce, healing, capture and the audit log treat the fault exactly as any other; no code path in the fault manager consults the windows to change a transition.

**No released interface changes.** `Fault.msg` and the existing services are untouched; the flag is derived in the gateway from the windows the fault manager serves. What is new is additive: `PlannedStop.msg`, `DeclarePlannedStop.srv`, `EndPlannedStop.srv`, `ListPlannedStops.srv`.

**The rules, each written where the code decides.** Time resolution is one millisecond end to end: the parser floors, storage keeps it, responses return it unchanged, and the covering test floors the fault instant the same way, because the fault item carries `first_occurred` as a double. A window that has not started is cancelled by `DELETE`, never ended early; a running one ends at now or at an instant within its running span; a finished one is immutable, judged against the fault manager's clock and never against an instant a caller sent. An instant at or before the epoch is refused everywhere; omitting `from` on the REST side means now, and the gateway always sends an explicit start. Retention, `planned_stop.max_windows` (default 100, clamped to [1, 10000]), bounds the windows that are no longer active; an active window is never dropped, and a declaration is never pruned by the retention it triggers. On the event stream an event carries no `expected` key until the gateway has read a window set, so an outage of the fault manager cannot read as "nothing is expected". Windows are per fault manager; an aggregating gateway reports what each peer derived and counts what it actually served.

---

## Issue

- closes #656

---

## Type

- [ ] Bug fix
- [x] New feature or tests
- [ ] Breaking change
- [x] Documentation only

The relaxed `required` on `FaultStreamXMedkit` is a schema change without a behaviour break: the object was previously present only when the entity resolved and is now present on every event with the flag; the entity fields are unchanged when they resolve.

---

## Testing

- Gateway end to end, real gateway with fault manager and demo node: declare over REST, report inside and outside, list with each filter value, per-entity list, `expected_count`, the stream flag on the specific event, a window declared after the fault, a window ended early, a second cycle after a clear, a cancel, the auth roles (viewer refused on POST and DELETE), and the no-fault-manager case (503 on POST, no `expected` key on the stream). Two-gateway aggregation: the filter reaches the peer and the count includes what the peer served.
- Fault manager over its services with SQLite: restart inside a window keeps it; an expected and an unexpected fault produce the same rows and transitions; retention at 1, at the default and past the cap; the end and cancel rules against the clock on both backends.
- Unit: the covering predicate at the millisecond boundaries and with overlapping windows, the ISO 8601 parser including the overflow and epoch bounds, the ns to `Time` conversion at negative and extreme values, the retention clamp sweep, and the SSE handler with and without window knowledge.
- Mutations run and recorded: dropping the derivation, freezing the end instant, removing the migration, forwarding the fan-out without the query, restoring the instant-based guards; each reddens the tests that pin it and nothing else.
- Suites: fault manager 579 unit and 113 integration, gateway and msgs unit sweep 3468, the six feature tests above, lint clean, clang-tidy zero findings in changed hunks, Sphinx with no new warnings, OpenAPI differential against main limited to the new routes, schemas, tag and fields.

---

## Checklist

### DIFF
--- a/docs/api/rest.rst
+++ b/docs/api/rest.rst
@@ -2793,7 +2793,15 @@ drops a window that is still running.
 What makes a fault expected
 ~~~~~~~~~~~~~~~~~~~~~~~~~~~
 
-A fault is expected when its ``first_occurred`` lies in ``[from, to]``, inclusive at both ends.
+A fault is expected when its ``first_occurred`` lies in ``[from, to]``, inclusive at both ends and
+compared at **millisecond** resolution: both sides are floored to the millisecond before the
+comparison, so a fault reported anywhere inside the millisecond a window opens or closes is inside
+the window. That is the resolution of the whole API - the parser rounds an ISO 8601 fraction down
+to it, a ``GET`` returns exactly what the ``POST`` sent - because an operator declares a stop at
+minute or second precision while ``first_occurred`` crosses the wire as a double resolving to about
+a quarter of a microsecond, and a boundary defined more finely than the data can express is one
+nobody can rely on.
+
 ``first_occurred`` is the start of the fault's **current cycle**, not its first report ever: it
 is reset when a FAILED event raises a CLEARED fault again. A code that failed during a
 changeover and failed again after it was acknowledged is therefore expected for the first cycle
@@ -2826,8 +2834,8 @@ window logic.
      - One window
    * - ``DELETE /api/v1/x-medkit-planned-stops/{planned_stop_id}``
      - operator
-     - End a window early. ``200`` with the ended window, not ``204``: the caller has to be told
-       where ``to`` landed
+     - End a window early, or cancel one that has not started. ``200`` with the window, not
+       ``204``: the caller has to be told where ``to`` landed, or that the window is gone
 
 Declaring a window
 ~~~~~~~~~~~~~~~~~~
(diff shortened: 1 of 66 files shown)


bburda:
Superseded. The same outcome ships as a planned-stop switch on the fault manager that reuses muting: recorded, audited, counted in muted_count, hidden from the default list, no new routes and no message change. A smaller pull request follows.

PR #662, opened 2026-09-07
ci: retry the network reads our own CI steps make

bburda:
# Pull Request

## Summary

Every CI job reads from the network before anything of ours is built, and `rosdep update` is the one of those reads this branch can wrap: it fetches the distribution index from raw.githubusercontent.com. A reset connection there fails the whole job with an error that has nothing to do with the change under test. Two jobs on two pull requests failed that way within an hour on the day this branch was opened, both on `rosdep update` with `Connection reset by peer`.

`rosdep update` now gets three attempts five seconds apart in every workflow step that calls it (`ci.yml`, `quality.yml`, `opcua-plugin.yml`). The ctcache install in the clang-tidy job gets the same loop: it is a `git+https` URL, so pip runs `git clone` in a subprocess to fetch it, and pip's own `--retries` only covers its HTTP session. A genuine failure still ends the job after the last attempt with its own error text; the loop only names the attempt.

The release-asset download in `.github/actions/ros-apt-source` gets a wider window. It already retried, but only for about 18 seconds, and a job on this branch died when GitHub answered 500 to all six attempts. That download is the one fetch in the action with nothing to fall back to: the version lookup above it can use the pinned value, but a 5xx on the .deb ends the step, and setup-ros then has no apt source, so the build stops at `colcon: command not found`. Measured against a server that answers 500 every time, the old flags make 6 attempts over 18 s and the new ones make 9 over 131 s.

One read stays unprotected, and it runs earlier than any of these. `ros-tooling/setup-ros@v0.7` removes `/etc/ros/rosdep/sources.list.d/20-default.list` and then runs `sudo rosdep init` once, with no retry around it (`src/setup-ros-ubuntu.ts:135-137` in the action). That call fetches `20-default.list` from raw.githubusercontent.com, the same host that fails here, and it happens before the step this branch touches, so a reset there still ends the job. Wrapping it would mean not using the action, which is a larger change than this one.

`apt-get update` is left alone on purpose. apt has defaulted `Acquire::Retries` to 3 since 2.3.2. Measured on the three images this repository builds in, a fetch against a dead endpoint makes the same four attempts with the flag and without it:

| image | apt | no flag | `-o Acquire::Retries=3` | `-o Acquire::Retries=0` |
|---|---|---|---|---|
| ubuntu:jammy | 2.4.14 | 4 attempts | 4 attempts | 1 attempt |
| ubuntu:noble | 2.8.3 | 4 attempts | 4 attempts | 1 attempt |
| ubuntu:resolute | 3.2.0 | 4 attempts | 4 attempts | 1 attempt |

---

## Issue

- none: a CI hardening with no behaviour change in the repository's code

---

## Type

- [ ] Bug fix
- [ ] New feature or tests
- [ ] Breaking change
- [ ] Documentation only

CI configuration only.

---

## Testing

- Every workflow file still loads as YAML.
- The diff against `main` is the 11 `rosdep update` call sites, the ctcache install, and the curl flags plus their comment in the apt-source action. No other line is touched.
- The loop was exercised the way the workflow runs it. Each step body was read out of the YAML and run under the shell GitHub Actions uses (`bash --noprofile --norc -eo pipefail`), with a stand-in `rosdep` and `pip3` that fail a set number of times. `set -e` does not stop the step on a failed attempt: with zero, one or two failures the step ends at 0 and the following `rosdep install` still runs; with three failures the step ends at 1 and the last error is visible.

---

## Checklist

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -41,7 +41,7 @@ jobs:
     steps:
       - name: Install Git
         run: |
-          apt-get update
+          apt-get -o Acquire::Retries=3 update
           apt-get install -y git
 
       - name: Checkout repository
@@ -68,13 +68,13 @@ jobs:
 
       - name: Install dependencies
         run: |
-          apt-get update
+          apt-get -o Acquire::Retries=3 update
           apt-get install -y ros-${{ matrix.ros_distro }}-test-msgs
           if [ "${{ matrix.ros_distro }}" = "humble" ]; then
             apt-get install -y ros-humble-rmw-cyclonedds-cpp
           fi
           source /opt/ros/${{ matrix.ros_distro }}/setup.bash
-          rosdep update
+          for attempt in 1 2 3; do rosdep update && break; [ "$attempt" = 3 ] && exit 1; echo "rosdep update attempt $attempt failed, retrying" >&2; sleep 5; done
           # Linters (clang-tidy / clang-format) only run in the Jazzy quality
           # job (quality.yml). Multiple package.xml files still list them as
           # <test_depend>, so we skip the rosdep keys here to keep build-and-
@@ -200,7 +200,7 @@ jobs:
     steps:
       - name: Install Git
         run: |
-          apt-get update
+          apt-get -o Acquire::Retries=3 update
           apt-get install -y git
 
       - name: Checkout repository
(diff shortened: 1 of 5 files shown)


.github/workflows/ci.yml
```diff
@@ -41,7 +41,7 @@ jobs:
     steps:
       - name: Install Git
         run: |
-          apt-get update
+          apt-get -o Acquire::Retries=3 update
```
mfaferek93: This half of the PR is a no-op: apt has defaulted `Acquire::Retries` to 3 since 2.3.2 (`acquire-item.cc`: `Retries(_config->FindI("Acquire::Retries", 3))`, changelog "Enable the Acquire::Retries option by default, set to 3", LP #1876035), and jammy/noble/ubuntu-latest all ship that. So the 25 `apt-get -o Acquire::Retries=3 update` lines behave exactly as before; only the 11 rosdep loops change anything. Either drop these hunks, or if you want more headroom write `Acquire::Retries "5";` once per job into `/etc/apt/apt.conf.d/80-retries`, which then also covers `apt-get install` and `rosdep install`. The one fetch that would actually benefit from the loop is the `pip3 install git+https://github.com/matus-chochlik/ctcache.git@...` clone in quality.yml:149, which pip's HTTP retry does not cover.

bburda: Both points check out. Measured on the three images this repo builds in - ubuntu:jammy (apt 2.4.14), noble (2.8.3), resolute (3.2.0) - a fetch against a dead endpoint makes 4 attempts with `-o Acquire::Retries=3` and 4 without it, so the 25 apt lines are gone and the branch is down to the 11 rosdep call sites. The ctcache install you pointed at got the same loop, since pip's `--retries` does not reach the `git clone` subprocess. One more fetch joined after a job here died on it: the ros-apt-source release download retried for 18 seconds and GitHub answered 500 to all six attempts, so its window is now 9 attempts over 131 seconds.

mfaferek93 (approved):

PR #664, opened 2026-09-08
fault_manager: a planned-stop switch, so a weekend is not a wave of faults

### LINKED ISSUE #656: Declare a planned stop, so a weekend is not a wave of faults

bburda:
# Pull Request

## Summary

A plant stops on purpose: a weekend, a changeover, a maintenance slot. Every controller the fault manager watches goes quiet at once, and there was no way to tell it that this was expected. With a debounce set and healing on, every one of those faults confirms and heals, every transition is an audit row, and the first Monday starts with a fault list nobody can read.

**A planned stop is a switch on the fault manager, and the operator's interface owns the calendar.** `~/set_planned_stop {active, reason, declared_by}` and `~/get_planned_stop` are ROS services, so they are already SOVD operations on the fault manager entity: `POST /apps/fault_manager/operations/set_planned_stop/executions`, behind the operations routes' existing role. No new route, no new DTO, no change to any released message or service type, no gateway source file touched; the served OpenAPI document is byte-identical before and after.

**Marked, not suppressed.** While the switch is on, a fault whose cycle starts (a new code, a reactivation from CLEARED, or a re-fail from HEALED) is recorded, debounced, captured and audited exactly as any other fault, and is muted the way a correlated symptom is muted: out of the default fault list, counted in `muted_count`, visible with `include_muted`, its confirmation and updates withheld from the event topic (a clear or heal is published as it is for any muted fault). The bridges report one FAILED per transition, so dropping such a fault would lose it for as long as the alarm stands on the controller; marking keeps it.

**Ownership is a persisted fact.** The stop owns the cycles that started inside it, as a flag on the fault row in the store; the correlation engine derives its mute map from that ownership plus any rule that overlays it, and when the rule's mute ends the stop's mute is back. A restart inside a stop keeps every owned fault muted. Switching off writes the declaration, announces one confirmation per CONFIRMED owned fault, then clears the flags, and a start-up that finds flags behind a withdrawn declaration finishes that release before it serves anything. Every switch transition is an audit row with the reason and the declarer (with `audit_log.enabled`), and `get_planned_stop` serves the last declaration after the withdrawal.

---

## Issue

- closes #656

---

## Type

- [ ] Bug fix
- [x] New feature or tests
- [ ] Breaking change
- [x] Documentation only

One observable change on a fault manager with no correlation rules: the engine and its cleanup timer now always exist, so `muted_count` and `cluster_count` are emitted as 0 and the timer costs about 190 ns per tick at the 5 s default, measured.

---

## Testing

- Gateway end to end (real gateway, fault manager and a reporting node over the real graph): the switch through the operations route, faults raised during the stop out of the default list and in `include_muted`, nothing on the stream for them, a fault confirmed before the stop untouched by it, one confirmation per released fault after switch-off with a settle window proving no second frame.
- Fault manager over its services with SQLite: the exact event set seen during a stop, a fault that clears during the stop, a re-fail from HEALED, a rule-muted symptom and a cluster-muted code composed with the stop, the scoped acknowledge, the auto-confirm timer under the stop, restart inside a stop, an interrupted release finished at the next start, a new stop declared right after such a start, a store that refuses the write, and 200 faults released at once.
- Unit: the engine's two mute sources and their hand-over, the storage flag and its migration on a database from before the column existed, the audit markers, the release ordering with a store that records call order.
- Mutations run and recorded: unmute-at-switch-off removed, announcement removed, persistence removed, cycle-start gate forced, PASSED gate removed, re-mute on rule expiry removed, flags cleared before the announcement; each reddens the test that pins it.
- Suites: fault manager unit and integration, msgs unit, the fault, operations, health, OpenAPI contract and error-coverage feature tests, lint; clang-tidy zero findings in changed hunks; Sphinx with no new warnings.

---

## Checklist

### DIFF
--- a/docs/config/fault-manager.rst
+++ b/docs/config/fault-manager.rst
@@ -608,7 +608,10 @@ Fault correlation identifies root causes and filters symptom faults.
      - Path to YAML file defining correlation rules.
    * - ``correlation.cleanup_interval_sec``
      - ``5.0``
-     - Interval for running correlation cleanup tasks.
+     - Interval for running correlation cleanup tasks. The correlation engine is
+       constructed on every fault manager, configured rules or not, because the
+       planned-stop switch needs none - so this timer runs everywhere. With no
+       rules loaded each tick walks two empty containers and returns.
 
 .. seealso::
 
@@ -644,18 +647,30 @@ Two configuration choices decide how much the switch can do:
        ends it.
    * - ``audit_log.enabled``
      - ``false``
-     - When on, each transition is recorded as ``planned_stop_started`` /
-       ``planned_stop_ended``, with the reason and the declarer. Recorded even
-       under ``audit_log.transitions: confirmed_only``.
+     - **Off by default, so a stock manager records no audit row for the switch
+       at all.** When on, each transition is recorded as ``planned_stop_started``
+       / ``planned_stop_ended`` under the ``__audit__`` fault code, with that
+       transition's own reason and declarer. Recorded even under
+       ``audit_log.transitions: confirmed_only``.
 
-While the stop stands, a fault whose cycle starts is reported, debounced,
+While the stop stands, a fault whose cycle *starts* is reported, debounced,
 confirmed, captured and audited unchanged, and is marked as muted rather than
-dropped: absent from the default fault list, counted in ``muted_count``, listed
-under ``muted_faults`` with ``rule_id: planned_stop`` when muted entries are
-requested, and never published on ``~/events``. Withdrawing the stop releases
-every fault it alone was muting and publishes one confirmation event for each of
-those that is CONFIRMED. Correlation rules are unaffected: a symptom a rule mutes
-stays muted when the stop is withdrawn.
+dropped: absent from the default fault list, counted in ``muted_count``, and
+listed under ``muted_faults`` with ``rule_id: planned_stop`` when muted entries
+are requested. A fault that was already up when the stop was declared is left
+alone - its confirmation has already been announced, and reporters re-send FAILED
+for as long as the condition holds.
+
+A marked fault is published exactly as a rule-muted symptom is: ``EVENT_CONFIRMED``
+and ``EVENT_UPDATED`` are withheld, while ``EVENT_CLEARED`` - the fault healing, or
+being acknowledged - is published as usual. Withdrawing the stop releases every
+fault it alone was muting and publishes one confirmation event for each of those
+that is CONFIRMED. Correlation rules are unaffected: a symptom a rule mutes stays
+muted when the stop is withdrawn, and the stop never takes a mute a rule owns.
+
+``~/get_planned_stop`` keeps serving the declaration after the withdrawal, with
+``ended_at`` stamped, so the reason stays readable once the plant is back up. A
+request the store cannot record answers ``success: false`` and changes nothing.
 
 Over HTTP the switch is an operation on the fault manager's App entity
 (``POST /apps/fault_manager/operations/set_planned_stop/executions``), which
--- a/docs/config/fault-manager.rst
+++ b/docs/config/fault-manager.rst
@@ -614,6 +614,60 @@ Fault correlation identifies root causes and filters symptom faults.
 
    :doc:`/tutorials/fault-correlation` for correlation rule syntax and examples.
 
+Planned Stop
+------------
+
+A planned stop is declared at runtime, not configured: an operator calls
+``~/set_planned_stop`` (or the matching SOVD operation on the ``fault_manager``
+entity) with a reason and a declarer, and calls it again with ``active: false``
+when the plant is back up. There is no parameter for it - a stop has a reason and
+an audit trail, and a parameter has neither.
+
+.. code-block:: bash
+
+   ros2 service call /fault_manager/set_planned_stop ros2_medkit_msgs/srv/SetPlannedStop \
+     "{active: true, reason: 'line 3 quarterly maintenance', declared_by: 'shift_lead'}"
+
+Two configuration choices decide how much the switch can do:
+
+.. list-table::
+   :header-rows: 1
+   :widths: 35 15 50
+
+   * - Parameter
+     - Default
+     - Effect on the planned stop
+   * - ``storage_type``
+     - ``sqlite``
+     - The declaration is stored beside the faults, so it survives a restart.
+       With ``memory`` it lives only for the process, and a restart inside a stop
+       ends it.
+   * - ``audit_log.enabled``
+     - ``false``
+     - When on, each transition is recorded as ``planned_stop_started`` /
+       ``planned_stop_ended``, with the reason and the declarer. Recorded even
+       under ``audit_log.transitions: confirmed_only``.
+
+While the stop stands, a fault whose cycle starts is reported, debounced,
+confirmed, captured and audited unchanged, and is marked as muted rather than
+dropped: absent from the default fault list, counted in ``muted_count``, listed
+under ``muted_faults`` with ``rule_id: planned_stop`` when muted entries are
+requested, and never published on ``~/events``. Withdrawing the stop releases
+every fault it alone was muting and publishes one confirmation event for each of
+those that is CONFIRMED. Correlation rules are unaffected: a symptom a rule mutes
+stays muted when the stop is withdrawn.
+
+Over HTTP the switch is an operation on the fault manager's App entity
+(``POST /apps/fault_manager/operations/set_planned_stop/executions``), which
+exists automatically in ``runtime_only`` discovery. Under ``hybrid`` or
+``manifest_only`` the fault manager must be declared in the manifest, or the
+entity is not there to address.
+
+.. seealso::
+
+   :doc:`/design/ros2_medkit_fault_manager/index` for how the switch composes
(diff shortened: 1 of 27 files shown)


src/ros2_medkit_fault_manager/src/fault_manager_node.cpp
```diff
@@ -490,6 +522,40 @@ FaultManagerNode::FaultManagerNode(const rclcpp::NodeOptions & options) : Node("
...
+  // the new stop is holding and drop its flag. A publication this early may reach
+  // no subscriber - the events topic is volatile and nothing has matched a
+  // publisher this young - and that is the same accepted loss as any other
+  // publication made at startup. The released faults are in the default fault list
+  // either way, which is where an operator looks after a restart.
+  RCLCPP_WARN(get_logger(),
+              "%zu fault(s) are still owned by a planned stop that was already withdrawn - a previous run did not "
+              "finish releasing them. Releasing them now.",
+              owned.size());
+  const size_t announced = announce_released(owned);
```
mfaferek93, line 551: The announce half of the interrupted release reaches nobody. The publisher is created at 258 and publishes milliseconds later, before any subscriber has matched. Checked in Docker with `ros2 topic echo --qos-reliability reliable /fault_manager/events` up 6 s before the node and a DB with `planned_stop.active=0`, `DURING_STOP.planned_stop_owned=1`: the node logged "released 1 fault(s), announced 1", the subscriber got nothing for `DURING_STOP`, and did get `fault_confirmed` for a control fault reported 8 s later. `test_planned_stop_interrupted_release.test.py:348-353` asserts only the log line. Keeping the flag clear here and deferring the publish to a one-shot timer that waits (bounded) for `get_subscription_count() > 0`, re-checking ownership per code at fire time, would make it real.

src/ros2_medkit_fault_manager/src/fault_manager_node.cpp
```diff
@@ -824,12 +921,28 @@ void FaultManagerNode::handle_report_fault(
...
-          correlation_engine_->process_fault(request->fault_code, correlation::severity_to_string(request->severity));
+          correlation_engine_->process_fault(request->fault_code, correlation::severity_to_string(request->severity),
+                                             std::chrono::steady_clock::now(), cycle_started);
 
       should_mute = correlation_result.should_mute;
 
+      // Ownership is persisted, not inferred: the switch-off, and a restart, both
+      // read it back from the store rather than comparing timestamps.
+      if (planned_stop_.active && cycle_started) {
+        storage_->set_planned_stop_owned(request->fault_code, true);
```
mfaferek93: This write lands after `report_fault` already committed (`sqlite_fault_storage.cpp:754`). A crash between the two leaves a confirmed fault with `planned_stop_owned=0`: restart recovery misses it, the stop's end does not release it, and its later reports announce updates during the stop. Persisting ownership inside the report transaction (pass `planned_stop_.active && cycle_started` into the storage call) removes the window.

src/ros2_medkit_fault_manager/src/correlation/correlation_engine.cpp
```diff
@@ -229,10 +257,94 @@ ProcessClearResult CorrelationEngine::process_clear(const std::string & fault_co
...
+
+  planned_stop_active_ = false;
+
+  std::vector<std::string> released;
+  released.reserve(planned_stop_owned_.size());
+  for (const auto & fault_code : planned_stop_owned_) {
+    auto it = muted_faults_.find(fault_code);
+    // A rule's overlay is not the stop's to lift: that fault stays muted, and
+    // stays unannounced, for as long as the rule holds it.
+    if (it != muted_faults_.end() && it->second.by_planned_stop) {
```
mfaferek93: Cluster muting only sets `result.should_mute` (644) and never touches the stop's entry, so a `show_as_single` non-representative fault that is also stop-owned still has `by_planned_stop == true` here, gets released, and `announce_released` publishes EVENT_CONFIRMED for it with no cluster check (fault_manager_node.cpp:1583-1593). Two confirmed symptoms of one cluster during a stop both get announced when the stop ends, which is the wave this PR exists to prevent. Skipping codes an active cluster still hides (or re-asking `process_fault`'s cluster verdict at announce time) keeps the rule in charge.

docs/config/fault-manager.rst
```diff
@@ -608,12 +608,91 @@ Fault correlation identifies root causes and filters symptom faults.
...
+marked as muted rather than dropped: absent from the default fault list, counted in
+``muted_count``, and listed under ``muted_faults`` with ``rule_id: planned_stop``
+when muted entries are requested. A fault that was already up when the stop was
+declared is left alone - its confirmation has already been announced, and reporters
+re-send FAILED for as long as the condition holds.
+
+Ownership is a flag on the fault row, so it survives a restart and is what the
+switch-off releases. The mute is derived from it: a correlation rule muting an owned
+fault overlays the stop rather than taking the fault from it, the withdrawal leaves
+a rule-held fault alone, and when the rule lets go - root cause acknowledged, window
```
mfaferek93: "when the rule lets go - window closed, cluster expired - the fault is muted by the stop again" (also README:105) is not something the engine does: `muted_faults_` entries are erased only in `process_clear` (151, 259, 264) and on the planned-stop paths (291, 313); `cleanup_expired()` erases none, so `reassert_planned_stop_mutes()` at 432 never finds work. A rule mute ends on acknowledge and nothing else; the docs should say that.

src/ros2_medkit_fault_manager/src/fault_manager_node.cpp
bburda: Fixed the way you suggested. The publish runs from a one-shot timer. It fires at the first matched subscriber on ~/events, or after planned_stop.interrupted_release_wait_sec (default 5 s), and delivery re-reads ownership per code from the store. Two things fell out of the deferral and are in the same commit. A fault that heals and fails again during the wait is announced once, because the new cycle drops the old flag in the same transaction. A stop declared during the wait takes the pending release over and announces it at its own switch-off. The interrupted-release test subscribes before the node starts and asserts the event; a second phase asserts the bound with nobody listening. I reproduced your Docker sequence with echo started 6 s before and 2 s after the node, and both receive the events now. In b446ee8c.

src/ros2_medkit_fault_manager/src/fault_manager_node.cpp
bburda: Fixed. The store is told whether a stop is in force and decides the cycle boundary itself, inside the transaction that writes the fault row. The separate call after the commit is gone, and its method with it, since nothing in production called it. On SQLite a test installs a trigger that fails the ownership statement, and the reopen shows the fault row rolled back with it. In 3af0e3c7.

src/ros2_medkit_fault_manager/src/correlation/correlation_engine.cpp
bburda: Fixed. At the switch-off, a fault an active show_as_single cluster still hides has its stop entry erased and is not announced. Nothing is written in its place: the cluster hides through the same verdict it gives every report, so a member that lived through a stop looks like one that did not. Two gaps came up on the way and are fixed too: a representative acknowledged after its window closed is promoted from the active cluster, and a fault that joins a cluster below min_count is not folded into it. Cluster membership is not persisted, so a stop that spans a restart announces everything the store owns; the docs say so. In 3af0e3c7 and ccfdf56c.

docs/config/fault-manager.rst
bburda: A rule mute ends on acknowledge only: cleanup_expired touches the pending structures and never muted_faults_. The docs now name that one trigger. The reassert path stays because the acknowledge path does reach it. In 3af0e3c7 and ccfdf56c.

mfaferek93 (approved):

bburda:
Waiting with merge for https://github.com/selfpatch/ros2_medkit/pull/651

PR #677, opened 2026-09-09
docs: say what clearing a fault deletes, and that a device de-assert is a clear

### LINKED ISSUE #671: docs: the snapshot deletion note does not say a rosbag recording is outside retain_on_clear, or that a device de-assert is a clear
## Summary

Two gaps in `docs/tutorials/snapshots.rst`. The parameters behave as their reference entries say. The note near the top does not.

**1. The note says `retain_on_clear` keeps snapshots. It keeps value snapshots only.**

The note: snapshots are deleted when a fault is cleared, set `snapshots.retain_on_clear: true` to keep them. The parameter table 70 lines later: the parameter keeps a fault's value snapshots. The code agrees with the table. `clear_fault` in `sqlite_fault_storage.cpp` guards only `DELETE FROM snapshots`. The rosbag recording is deleted by `RosbagCapture::on_fault_cleared`, gated on `snapshots.rosbag.auto_cleanup` and `snapshots.rosbag.max_bags_per_fault`. A reader who follows the note sets `retain_on_clear: true` and still loses the recording.

Measured on 0.7.0, one fault, one recording, alarm de-asserted with no operator action:

| config | after the clear |
|---|---|
| defaults | `Auto-cleanup: deleted bag file for fault 'PLC_JAM_INFEED'`, bag gone |
| `snapshots.retain_on_clear: true` | same log line, bag gone |
| `snapshots.rosbag.max_bags_per_fault: 3` | no cleanup line, recording kept |

**2. The note does not say a device de-assert is a clear.**

The note names a REST DELETE and the `~/clear_fault` service. The OPC UA plugin calls `~/clear_fault` itself when an alarm tag goes false. Its README says only "the fault is automatically cleared". So an intermittent PLC alarm deletes its own recording with nobody touching the API:

```
[INFO] [plugin.opcua]: Alarm cleared: PLC_JAM_INFEED on load_process
[INFO] [fault_manager]: Fault cleared: PLC_JAM_INFEED (auto-cleared 0 symptoms)
[INFO] [fault_manager]: Auto-cleanup: deleted bag file for fault 'PLC_JAM_INFEED'
```

That is the intermittent-fault case the tutorial names first as the reason to record.

The note also names `DELETE /api/v1/faults/{code}`, which the gateway does not serve. The per-fault clear is `DELETE /api/v1/{entity-path}/faults/{fault_code}`. The bare `DELETE /api/v1/faults` clears all.

## Proposed solution

In the note: say value snapshots, point recordings at `snapshots.rosbag.auto_cleanup` and `max_bags_per_fault` (configurable since #623), say that a plugin de-assert is a clear, and fix the route. Mirror the de-assert sentence in the OPC UA plugin README.

## Additional context

0.7.0 on Jazzy, Ubuntu 24.04, rosbag capture on, OPC UA plugin bridging an OpenPLC-style controller. After the de-assert the fault still answers 200 with `status: cleared` and keeps its `freeze_frame` (#491). The `rosbag` entry, the `rosbag_files` row and the bag directory are gone. An operator DELETE on the same fault produces the same log lines, both paths land on the one `~/clear_fault` handler. Point 2 also reproduces on the 0.6.0 image.

bburda:
## Summary

The snapshots tutorial note said `snapshots.retain_on_clear: true` keeps everything a fault captured, and named `DELETE /api/v1/faults/{code}` as the clear, a route the gateway does not serve.

The note now says a clear deletes the value snapshots and that `retain_on_clear` covers only those. The recording goes under `snapshots.rosbag.auto_cleanup`, unless that is off or `snapshots.rosbag.max_bags_per_fault` keeps a history. It names the per-fault and bulk DELETE routes, the `~/clear_fault` service and the correlation cascade as examples of a clear, and says a plugin calls that service when its device de-asserts an alarm.

Code behind the text: `retain_snapshots_on_clear_` in `sqlite_fault_storage.cpp` guards only `DELETE FROM snapshots`. `RosbagCapture::on_fault_cleared` deletes the recording under `auto_cleanup`.

The same claim in the fault manager, gateway and OPC UA READMEs and in the shipped `snapshots.yaml` is corrected.

## Issue

- closes #671

## Type

- [ ] Bug fix
- [ ] New feature or tests
- [ ] Breaking change
- [x] Documentation only

## Testing

Docs only. From `docs/`, both clean:

```
sphinx-build -b linkcheck . _build/linkcheck
sphinx-build -b html . _build/html -W --keep-going
```

`pre-commit` passes.

## Checklist

### DIFF
--- a/docs/tutorials/snapshots.rst
+++ b/docs/tutorials/snapshots.rst
@@ -32,18 +32,24 @@ configuration always overrides the zero-config fallback when present.
 .. note::
 
    By default, clearing a fault deletes its **value snapshots**, the per-topic
-   JSON captures this tutorial configures. A clear reaches the fault manager
-   either from the entity-scoped DELETE route, for example
-   ``DELETE /api/v1/apps/{app_id}/faults/{fault_code}`` (``components``,
-   ``areas`` and ``functions`` carry the same route, and there is no global
-   ``DELETE /api/v1/faults/{code}``), or from the ``~/clear_fault`` service. A
-   plugin whose device de-asserts an alarm calls that same service, so a device
-   going quiet clears the fault the way an operator does.
-   ``snapshots.retain_on_clear: true`` keeps the value snapshots across a clear,
-   and only those. The rosbag recording follows
-   ``snapshots.rosbag.auto_cleanup``, which deletes it on clear unless
-   ``snapshots.rosbag.max_bags_per_fault`` keeps a history, in which case that
-   cap governs the recording's retention instead.
+   JSON captures this tutorial configures. Every clear reaches storage the same
+   way, whatever asked for it: the per-fault
+   ``DELETE /api/v1/apps/{app_id}/faults/{fault_code}`` route (``components``,
+   ``areas`` and ``functions`` carry the same one, though there is no global
+   ``DELETE /api/v1/faults/{code}``), the bulk
+   ``DELETE /api/v1/apps/{app_id}/faults`` and ``DELETE /api/v1/faults``, the
+   ``~/clear_fault`` service, and the correlation cascade, which clears a root
+   cause's symptoms with no clear addressed to them. A plugin calls that same
+   service when its device de-asserts an alarm, so a device going quiet clears
+   the fault the way an operator does. Starting with healing disabled takes the
+   value snapshots of leftover HEALED rows as it reclassifies them, without a
+   clear at all. ``snapshots.retain_on_clear: true`` keeps the value snapshots
+   across all of it, and only those. The rosbag recording is not covered by that
+   setting. A clear deletes the recording under
+   ``snapshots.rosbag.auto_cleanup``, unless that is off or
+   ``snapshots.rosbag.max_bags_per_fault`` keeps a history. The cap then governs
+   retention, except at ``0`` (unlimited), where only
+   ``snapshots.rosbag.max_total_storage_mb`` bounds it.
 
 Quick Start
 -----------
--- a/docs/tutorials/snapshots.rst
+++ b/docs/tutorials/snapshots.rst
@@ -31,9 +31,19 @@ configuration always overrides the zero-config fallback when present.
 
 .. note::
 
-   By default, snapshots are deleted when a fault is cleared via the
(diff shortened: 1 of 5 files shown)


docs/tutorials/snapshots.rst
```diff
@@ -31,9 +31,25 @@ configuration always overrides the zero-config fallback when present.
...
-   Set ``snapshots.retain_on_clear: true`` to keep them across clears.
+   By default, clearing a fault deletes its **value snapshots**, the per-topic
+   JSON captures this tutorial configures. Every clear reaches storage the same
+   way, whatever asked for it: the per-fault
+   ``DELETE /api/v1/apps/{app_id}/faults/{fault_code}`` route (``components``,
+   ``areas`` and ``functions`` carry the same one, though there is no global
+   ``DELETE /api/v1/faults/{code}``), the bulk
+   ``DELETE /api/v1/apps/{app_id}/faults`` and ``DELETE /api/v1/faults``, the
+   ``~/clear_fault`` service, and the correlation cascade, which clears a root
+   cause's symptoms with no clear addressed to them. A plugin calls that same
```
mfaferek93: "A plugin calls that same service when its device de-asserts an alarm" is true for OPC UA (`opcua_plugin.cpp:1023 send_clear_fault`), EtherNet/IP and S7 (platform `eip_plugin.cpp:811`, `s7_plugin.cpp:195`), but not for every plugin: the ADS plugin (platform `ads_plugin.cpp:970`) and graph_watchdog (`fault_request.hpp:45`) de-assert with a PASSED report, which goes through debounce and healing and deletes nothing. Suggest naming OPC UA here and adding "a plugin that de-asserts with a PASSED event heals instead, and keeps the snapshots".

bburda: I named OPC UA and split it by alarm kind. A threshold alarm clears on de-assert. A native AlarmCondition stays CONFIRMED until acknowledge and, unless require_confirm_for_clear is off, confirm. The PASSED case, as the graph watchdog sends it, goes through debounce and healing, and nothing on that path deletes snapshots. The heal itself needs healing_enabled and the counter reaching healing_threshold, so with the default the fault stays CONFIRMED. Pushed in 81b1f28e.

mfaferek93 (approved):

Issue #672, opened 2026-09-09
first_occurred is reset when a fault re-occurs, so it does not name the first occurrence


bburda:
## Summary

When a FAILED event reactivates a CLEARED fault, the fault manager overwrites `first_occurred` with the new timestamp while `occurrence_count` keeps climbing. Both backends do it on purpose: `src/ros2_medkit_fault_manager/src/sqlite_fault_storage.cpp:729-735` and `src/fault_storage.cpp:172-175` at 8b1969f8 ("this is a new outage cycle, not a continuation of the one that just cleared"), pinned by `test_fault_manager.cpp:424` and `test_sqlite_storage.cpp:503`.

A jam alarm asserted at `18:54:39.059Z`, cleared on the controller, asserted again at `18:54:54.925Z`:

```json
"occurrence_count": 2,
"first_occurrence": "2026-09-08T18:54:54.925Z",
"last_occurrence": "2026-09-08T18:54:54.925Z"
```

`faults.db` holds one row per code, so the first time is gone from the fault store. The audit log keeps it, but it is off by default and no REST route reads it. A UI that labels the value "First occurrence" reads the API correctly and shows the wrong thing.

#618 documented this at the message layer. `Fault.msg` and `docs/api/messages.rst` say "timestamp when the current occurrence started". The name still says first, and `docs/api/rest.rst` shows `extended_data_records.first_occurrence` and `first_occurred` only as example values, with no definition.

## Proposed solution

1. Keep `first_occurred` at the first occurrence for the life of the code, let `last_occurred` and `occurrence_count` carry the history, and add `cycle_started` for the current outage's start. Additive.
2. Keep today's behaviour and rename the reset field to `cycle_started`. Breaking on `/api/v1/faults`, on the SOVD `extended_data_records` block and on `Fault.msg`.

Either way `docs/api/rest.rst` needs the definition `messages.rst` already has. A PR for that part follows. The naming choice stays open here.

## Additional context

Reproduces through the services alone: `report_fault` (FAILED), `clear_fault`, `report_fault` again, and `first_occurred` moves to the second report with `occurrence_count` 2. A FAILED re-report while the fault is still CONFIRMED leaves `first_occurred` and the count alone and only moves `last_occurred`, so the reset is tied to the CLEARED to FAILED edge. Seen on `main` at f873448d (0.6.0), unchanged at 0.7.0 (8b1969f8). ROS 2 Jazzy, Ubuntu 24.04.

PR #678, opened 2026-09-10
docs(gateway): define the fault occurrence timestamps in the REST reference

bburda:
## Summary

The fault detail in the REST reference showed `first_occurrence` and `last_occurrence` as example values with no definition.

The reference now defines both where the fault object's fields are described. `first_occurrence` is the start of the current occurrence. It is reset when a FAILED event reactivates a CLEARED fault, so it moves with `occurrence_count`. `last_occurrence` advances on FAILED events only. The second commit states the mapping to `Fault.msg`'s `first_occurred` and `last_occurred`.

These are the definitions #618 settled at the message layer. The naming question in #672 stays open.

## Issue

- related #672

## Type

- [ ] Bug fix
- [ ] New feature or tests
- [ ] Breaking change
- [x] Documentation only

## Testing

Docs only. From `docs/`, both clean:

```
sphinx-build -b linkcheck . _build/linkcheck
sphinx-build -b html . _build/html -W --keep-going
```

`pre-commit` passes.

## Checklist

### DIFF
--- a/docs/api/rest.rst
+++ b/docs/api/rest.rst
@@ -1500,7 +1500,10 @@ Query and manage faults.
 
    The ``environment_data.extended_data_records`` object carries the fault's
    occurrence timestamps, which are ``Fault.msg``'s ``first_occurred`` and
-   ``last_occurred`` under their REST names (see :doc:`messages`):
+   ``last_occurred`` (see :doc:`messages`). The list route serves the same
+   instants as ``first_occurred`` and ``last_occurred`` in epoch seconds; only
+   this detail route renames them and formats them as ISO-8601 UTC with
+   millisecond precision:
 
    - ``first_occurrence``: when the current occurrence started, reset when a
      FAILED event reactivates a CLEARED fault, so it moves with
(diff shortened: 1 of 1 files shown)


docs/api/rest.rst
```diff
@@ -1496,6 +1496,19 @@ Query and manage faults.
    - ``confirmedDTC``: Confirmed DTC indicator (``0`` or ``1``)
    - ``pendingDTC``: Pending DTC indicator (``0`` or ``1``)
 
+   **Extended Data Records:**
+
+   The ``environment_data.extended_data_records`` object carries the fault's
+   occurrence timestamps, which are ``Fault.msg``'s ``first_occurred`` and
```
mfaferek93, line 1502: "under their REST names" reads as if `first_occurrence`/`last_occurrence` were the names everywhere. The list route serves the same instants as `first_occurred`/`last_occurred` epoch-seconds floats (`fault_msg_conversions.cpp:36-37`, `dto/faults.hpp:52-53`); only the detail route renames them and formats ISO-8601 UTC with ms. Worth one sentence: "the list route serves them as `first_occurred`/`last_occurred` in epoch seconds; the detail route uses these names in ISO-8601 UTC". The reset rules themselves match `sqlite_fault_storage.cpp:724-754`.

bburda: I added the sentence. The list route serves them as first_occurred and last_occurred in epoch seconds. Only the detail route renames them and formats them as ISO-8601 UTC with millisecond precision. Pushed in 9b9a7920.

mfaferek93 (approved):

Issue #675, opened 2026-09-09
bulk-data rosbags: a recording split across several storage files downloads as one segment, chosen by directory order


bburda:
## Describe the bug

A recording that rosbag2 split past `snapshots.rosbag.max_bag_size_mb` (default 50 MB) is a directory with several storage files and one `metadata.yaml` naming all of them. `GET /api/v1/{entity}/bulk-data/rosbags/{recording_id}` serves exactly one of those files. `resolve_rosbag_file_path` in `src/ros2_medkit_gateway/src/http/handlers/bulkdata_handlers.cpp` returns the first `.db3` or `.mcap` a `std::filesystem::directory_iterator` yields and never reads `metadata.yaml`. On ext4 that order follows a hash of the file name, not the recording order.

Measured on main:

- The three-segment mcap recording from `RosbagCaptureIntegrationTest.ABoundaryRecordingSplitsAndReportsTheWholeBag` (`_0` 1641630 bytes, `_1` 1642300, `_2` 823197) resolves to `_0`. 2465497 of 4107127 recorded bytes never leave the gateway.
- A three-segment sqlite3 bag with the same naming resolves to `_1`, the middle segment.
- A two-segment bag (`split_0.db3` 16384, `split_1.db3` 53248) resolves to `split_1.db3`. Creating the files in the opposite order, or replacing `metadata.yaml` with garbage, changes nothing.

Nothing in the response says the recording has more. `Content-Disposition` carries `<recording_id>.<format>` and never names the segment. The fault manager stores the directory total (`RosbagCapture::calculate_bag_size`), so the descriptor `size` and the nested `size_bytes` describe the whole recording while the transfer is one piece of it. That mismatch is the only signal a client gets. The descriptor's `x-medkit` block carries no storage-file count.

## To reproduce

1. Set `snapshots.rosbag.max_bag_size_mb` low so a fault's recording splits, or use the integration test above.
2. `GET /api/v1/apps/{app}/bulk-data/rosbags` and note `size`.
3. `GET /api/v1/apps/{app}/bulk-data/rosbags/{id}` and compare `Content-Length` and the received bytes against the files `metadata.yaml` lists under `relative_file_paths`.

## Expected behavior

Either the whole recording is downloadable through the bulk-data route, or each storage file is addressable on its own and the descriptor says how many there are.

## Options

- Serve a split recording as one tar with its `metadata.yaml`, so `ros2 bag play` works on what arrives.
[...]

PR #679, opened 2026-09-10
fix: report a rosbag's served size, retire the removed snapshot routes

### LINKED ISSUE #674: bulk-data rosbags: the descriptor size counts bytes the download does not serve, and the README documents removed snapshot routes
## Summary

Three mismatches around rosbag snapshots over the REST API.

**1. The descriptor `size` is the bag directory, the download is the storage file alone.** `GET /api/v1/apps/load_process/bulk-data/rosbags` reports `"size": 35943` (32768 bytes of `.db3` plus 3175 bytes of `metadata.yaml`). Downloading the same item answers `Content-Length: 32768`. The nested `environment_data.snapshots[]` entry on `GET .../faults/{code}` carries the same `size_bytes: 35943`. `RosbagCapture::calculate_bag_size` in the fault manager sums every file under the bag directory, `BulkDataHandlers::download` in the gateway serves the inner storage file resolved by `resolve_rosbag_file_path`. A client sizing a transfer from the descriptor is off by the metadata file.

**2. Documented snapshot routes do not exist.** `src/ros2_medkit_gateway/README.md` lists `GET /api/v1/faults/{fault_code}/snapshots`, `.../snapshots/bag` and `GET /api/v1/components/{component_id}/faults/{fault_code}/snapshots` with examples, and the quick start in `docs/tutorials/snapshots.rst` ends with `curl .../api/v1/faults/MOTOR_OVERHEAT/snapshots`. All answer 404. The same tutorial's "Migration from Legacy Endpoints" table already lists them as removed. Snapshots are served inside `GET .../faults/{code}` and the bag through `.../bulk-data/rosbags/{id}`.

**3. `Access-Control-Expose-Headers` is documented without its condition.** `docs/api/rest.rst` lists it among the download's response headers. It is sent only when `cors.allowed_origins` matches the request's `Origin`, and then it names six headers, not one.

## Proposed solution

1. Report the bytes the route serves, so the descriptor `size`, the nested `size_bytes` and `Content-Length` agree.
2. Drop the removed routes from the gateway README and the tutorial quick start. The migration table stays.
3. Qualify the `Access-Control-Expose-Headers` line as CORS-only, or leave it to the OpenAPI document.

## Additional context

`ghcr.io/selfpatch/ros2_medkit-jazzy` at f873448d (0.6.0) and a 0.7.0 build carrying #660, same result. ROS 2 Jazzy, Ubuntu 24.04, OPC UA plugin on an OpenPLC-style controller, `snapshots.rosbag.format: sqlite3`. Headers taken with `curl -D -`, directory size with `du -sb` in the container. The served file is a valid rosbag2 bag, `ros2 bag info -s sqlite3` reads it standalone.

Already fixed in 0.7.0 and not part of this: `Accept-Ranges` on the download, the descriptor addressed by fault code, `creation_date` carrying the fault's timestamp. The missing `x-medkit.source` on a plugin freeze frame is #660.

bburda:
## Summary

A rosbag2 recording is a directory with one storage file plus `metadata.yaml`. The bulk-data download serves the storage file alone, but every size in the API was the fault manager's stored figure, the whole directory. Each listing overstated its own download by `metadata.yaml`.

Now:

- The descriptor `size` under `/bulk-data/rosbags` is measured on the file the bag's `metadata.yaml` names, when the download resolves that same file.
- The size the fault manager reports through `GetFault`, `GetSnapshots`, `GetRosbag` and `ListRosbags` is the storage file named by the bag's `metadata.yaml`. The stored `size_bytes` and the `max_total_storage_mb` quota still count the directory, which is what eviction frees.
- When no single file can be named (no metadata, unparsable, the named file gone, a name that points outside the bag directory, or a recording split past `max_bag_size_mb`) the reported size falls back to the stored total, never to zero. Both layers decide this from the same `relative_file_paths` field and both follow a name only to a direct child of the bag directory with a `.db3` or `.mcap` extension. In those cases the download hands over one file from the directory and the descriptor size can exceed the transfer, which is the client's signal; a directory with no storage file at all answers 500 on download. Which file the download serves for a split is unchanged, see #675.
- When `metadata.yaml` names exactly one present file inside the bag, the listing and the download resolve to that file. When the named file is absent, the download still hands over what the directory holds, and the listing keeps the stored figure for that row. A bare storage-file path is sized as that file.
- The listing never throws. An unreadable or vanished bag directory leaves its row with the stored figure and the rest of the listing intact.
- `rest.rst` states the rule once: descriptor `size`, nested `size_bytes` and `Content-Length` are the same number for a recording whose metadata names its single present storage file, and lists the cases where they part.

Stale leftovers removed: the gateway README and the tutorial quick start documented `GET /faults/{code}/snapshots` and `.../snapshots/bag`, both 404 since the endpoints were removed. The README playback block told the reader to `tar -xzf` the download. It now shows `ros2 bag info` and `ros2 bag play` on the file the download names, run on both storage formats. The postman collection drops its three requests at the removed routes. The `GetSnapshots` response no longer carries `rosbag.download_url`, which pointed at that removed route. Nothing in the repository reads it and no gateway route forwards that payload.

## Issue

- closes #674

## Type

- [x] Bug fix
- [ ] New feature or tests
- [x] Breaking change
- [ ] Documentation only

Breaking: the `GetSnapshots` service response no longer carries `rosbag.download_url`. Recorded in the fault manager changelog. The REST API shape is unchanged.

## Testing

Gateway, `test_bulkdata_handlers.cpp`: descriptor size equals the served file and not the directory total; a bag the process cannot see keeps the stored figure; a two-segment recording is listed at the stored total; a stray second file beside a single named one is sized by the named file, with the test naming whichever file directory order would not reach first; a named file that is gone is not sized by whatever is left beside it; a bare storage-file path; names that climb out of the bag, are absolute, sit in a subdirectory or lack a storage extension are not followed, and one such bag that also holds a real storage file still serves that file; a listing with a symlink-loop bag answers with every row; a mode-0000 bag is declined without a throw, with the test dropping `CAP_DAC_OVERRIDE` from its own thread when it runs as root so the permission is denied in CI too. Each test was made to fail first.

Fault manager, `test_rosbag_capture.cpp` and `test_fault_manager.cpp`: the helper on a well-formed bag, missing metadata, unparsable metadata, a named file that is gone, a name outside the bag (four shapes), a multi-file recording, real service calls through all four services, and a `GetSnapshots` payload with no `download_url`. Quota tests untouched.

Integration, `test_bulk_data_api.test.py`: descriptor size equals the downloaded body length.

Full gateway and fault manager package suites pass locally on jazzy. Lint and docs build clean.

## Checklist


src/ros2_medkit_gateway/src/http/handlers/bulkdata_handlers.cpp
```diff
@@ -121,6 +212,53 @@ std::string rosbag_recording_id(const std::string & file_path) {
...
+    const auto names = rosbag_relative_file_paths(bag_path);
+    if (!names || names->size() != 1) {
+      return std::nullopt;
+    }
+  }
+
+  // The same two steps `download()` performs, in the same order and through the
+  // same resolver, so the size a client is promised cannot drift from the size
+  // it is sent. Changing which file a recording resolves to changes both.
+  const std::string resolved = BulkDataHandlers::resolve_rosbag_file_path(bag_path);
```
mfaferek93, line 290: The gate above only checks that metadata names one file, not that it exists, so for a bag naming `recording_0.db3` with only a stray `recording_1.db3` on disk the resolver falls through to the directory scan and this row is sized by the stray, while the fault manager's `rosbag_served_bytes` returns the stored total for the same bag (its `ANamedFileThatIsNotOnDiskFallsBackToTheStoredTotal` builds exactly that directory). That breaks the one-recording-one-size rule rest.rst just added and the description's claim about the fallback. Return nullopt here unless `resolved == bag_path / names->front()`.

src/ros2_medkit_gateway/test/test_bulkdata_handlers.cpp
```diff
@@ -247,6 +253,352 @@ TEST_F(BulkDataHandlersTest, NoRowsYieldsNoDescriptors) {
...
+  write_file(strays / "recording_0.db3", std::string(4096, 'a'));
+  write_file(strays / "recording_1.db3", std::string(65536, 'b'));
+  write_metadata(strays, {"recording_0.db3"});
+
+  const uint64_t named_size = std::filesystem::file_size(strays / "recording_0.db3");
+  const uint64_t stray_size = std::filesystem::file_size(strays / "recording_1.db3");
+  ASSERT_NE(named_size, stray_size) << "the two files are the same size, so nothing is being told apart";
+
+  // The download resolves the named file, so the bytes on the wire are its bytes.
+  EXPECT_EQ(BulkDataHandlers::resolve_rosbag_file_path(strays.string()), (strays / "recording_0.db3").string());
```
mfaferek93: If the metadata branch in `resolve_rosbag_file_path` (lines 171-177) were dropped, this test only fails when `directory_iterator` happens to yield `recording_1.db3` before `recording_0.db3`, and that order is hash order on ext4 and APFS, so on a given runner the regression this test is named for can pass clean every time. Make it deterministic: walk the directory the way the fallback loop does and write metadata naming the file that walk would not return first, and replace the `4096u` literal on 461 with `named_size`.

src/ros2_medkit_gateway/test/test_bulkdata_handlers.cpp
```diff
@@ -247,6 +253,352 @@ TEST_F(BulkDataHandlersTest, NoRowsYieldsNoDescriptors) {
...
+  }
+  std::error_code ec;
+  std::filesystem::permissions(locked, std::filesystem::perms::none, ec);
+  ASSERT_FALSE(static_cast<bool>(ec)) << "could not drop the directory's permissions";
+
+  std::error_code probe_ec;
+  std::filesystem::directory_iterator probe(locked, probe_ec);
+  if (!probe_ec) {
+    std::filesystem::permissions(locked, std::filesystem::perms::owner_all, ec);
+    GTEST_SKIP() << "running as uid " << ::getuid() << ", which ignores mode bits, so a 0000 directory is readable. "
```
mfaferek93: The description says the listing is exercised with a symlink-loop bag and a mode-0000 bag, but this test never calls `fold_rosbag_rows_into_descriptors` and skips under uid 0, which is what every CI job container runs as. The 0000 directory is the only case that reaches the resolver's `directory_iterator(path, ec)` EACCES branch, so the branch the resolver comment calls the 500 bug has no CI coverage. Either fix the description, or run this one under a non-root uid so it counts.

src/ros2_medkit_gateway/src/http/handlers/bulkdata_handlers.cpp
bburda: Fixed in e5eed021. `rosbag_served_bytes` now measures only when the file the resolver returned is the file the metadata names, after `lexically_normal`. For your shape, a bag naming recording_0.db3 with only recording_1.db3 on disk, it declines, and the row keeps its stored figure on both sides. `ANamedStorageFileThatIsGoneIsNotMeasuredByWhateverIsLeftBesideIt` pins it. The metadata branch also followed absolute and `..` names. The gateway and the fault manager (5713977b, rosbag_capture.cpp) now follow a name only to a direct child of the bag with a .db3 or .mcap extension. rest.rst (cccf127d) states when the three sizes agree and the four shapes where they differ.

src/ros2_medkit_gateway/test/test_bulkdata_handlers.cpp
bburda: Fixed in e5eed021. `TheMetadataNamesTheStorageFileRatherThanDirectoryOrder` walks the directory the way the fallback loop does, then writes metadata naming the file that walk does not reach first. The loop alone can then never return the named file, whatever the directory order. The 4096u literal is gone: the assertions use `named_size`, and the listing is also checked against `stray_size`.

src/ros2_medkit_gateway/test/test_bulkdata_handlers.cpp
bburda: Fixed in e5eed021. `AModeZeroDirectoryIsAlsoDeclinedRatherThanThrown` no longer skips. When it runs as root, it drops CAP_DAC_OVERRIDE and CAP_DAC_READ_SEARCH from its own thread's effective set and puts them back afterwards. It first checks that the directory walk is denied. Then it runs the size helper, the resolver and `fold_rosbag_rows_into_descriptors` under that. So the CI jobs, which run as root, now reach the resolver's EACCES branch. The body stays on the test's own thread, so sanitizers and coverage still see it. The PR text describes the test as it is.

PR #682, opened 2026-09-10
ci: run the ROS jobs on ros-base images so nothing fetches rosdep's sources list

bburda:
# Pull Request

## Summary

Every container job in `ci.yml`, `quality.yml` and `opcua-plugin.yml` runs on a bare `ubuntu:<codename>` image and installs ROS with `ros-tooling/setup-ros@v0.7`. That action removes `/etc/ros/rosdep/sources.list.d/20-default.list` and then runs `sudo rosdep init` with no retry around it (`src/setup-ros-ubuntu.ts:135-137` in the action). `rosdep init` downloads that file from raw.githubusercontent.com, which is the host that has failed jobs here before, and it happens before the step that does have a retry loop. A `uses:` step cannot be wrapped in a loop, so the fetch cannot be protected where it is.

The 11 jobs now run on `ros:<distro>-ros-base`. Those images ship ROS, `rosdep` already initialised, the ROS apt source configured, and colcon, rosdep, git, gpg, curl and vcstool. With that, `setup-ros` has nothing left to do, and the unprotected `rosdep init` download of `20-default.list` is gone. What is not gone is the `rosdep update` fetch: Actions runs container jobs with `HOME=/github/home`, so the sources cache the image carries under `/root/.ros` is never read, and every job still downloads `base.yaml`, `python.yaml` and `index-v4.yaml` from raw.githubusercontent.com into `/github/home/.ros/rosdep/sources.cache`. That fetch sits inside the existing 3x retry loop, which is why the loop stays. The `ros-apt-source` composite action goes with it: it existed only so `setup-ros` would skip its own rate-limited lookup. The per-job `Install Git` steps go too, since all three images carry git.

Two follow-on changes are in here because the first one requires them. The `ros-base` images ship no apt package index, and the deleted steps were where the index came from, so each step that runs `apt-get install` now runs `apt-get update` first. And the `coverage` job's gpg install carried a comment about the Codecov action needing it; that upload happens in `coverage-merge`, which has its own gpg and is untouched, and nothing else in the `coverage` job uses gpg, so the install is gone.

`coverage-merge` keeps its plain `ubuntu:noble` image. It never used ROS.

---

## Issue

- none: CI configuration only, no change to the repository's code

---

## Type

- [x] Bug fix
- [ ] New feature or tests
- [ ] Breaking change
- [ ] Documentation only

---

## Testing

`ros-base` is narrower than the `ros-<distro>-desktop` metapackage that `setup-ros` installed. The difference is rviz2, demo_nodes, turtlesim, examples and rqt, and nothing in the workspace depends on any of them. A grep over the workspace finds no reference to them, and a second check goes further. The three step bodies of the `jazzy-test` job were read out of the YAML and run verbatim inside a real `ros:jazzy-ros-base` container, with no `apt-get update` added beyond the ones the workflow itself carries. Result: 18 packages built, `ros2_medkit_gateway` unit suite 2886 tests, 0 errors, 0 failures.

Mechanical checks on the branch: every workflow still parses as YAML; no `setup-ros@` reference remains; no `ros-apt-source` reference remains and the directory is gone; 16 image values across the 11 jobs name a `ros:*-ros-base` image; exactly one `image: ubuntu:noble` remains, in `coverage-merge`; and of the 24 step bodies that run `apt-get install`, every one also runs `apt-get update`.

Measured on all three images, so the apt change is not guesswork: `/var/lib/apt/lists` is empty and `apt-get install -y --simulate ccache` exits 100 on `ros:humble-ros-base`, `ros:jazzy-ros-base` and `ros:lyrical-ros-base`.

Two things only CI can show, and reviewers should read the first run with them in mind:

- humble and lyrical were not built locally. Only jazzy was.
- the first run compiles cold in every job. The images set `LANG=C.UTF-8` where `setup-ros` exported `en_US.UTF-8`, and ccache includes the locale in its hash, which was confirmed by compiling the same file twice under the two locales in `ros:jazzy-ros-base` with ccache 4.9.1: the second compile missed. The Actions cache archive still restores, but its entries do not match, so the first run pays a full build and later runs are warm again. Cache keys are deliberately unchanged, because renaming them would save only the archive download and not the cold build.

One weakness was found and deliberately left alone: ctcache's hash covers the preprocessed input, the dumped configuration and the arguments, but not the identity of the clang-tidy binary, so a changed analyser could return an old clean result. This change does not trigger it. `ubuntu:noble` and `ros:jazzy-ros-base` offer the same `clang-tidy` candidate (`1:18.0-59~exp2`) from the same apt suites, so the analyser is the same before and after.

---

## Checklist


.github/workflows/ci.yml
```diff
@@ -330,24 +308,13 @@ jobs:
...
 
-      - name: Pre-install ROS 2 apt source
-        uses: ./.github/actions/ros-apt-source
-
-      - name: Set up ROS 2 Jazzy
-        uses: ros-tooling/setup-ros@v0.7
-        with:
-          required-ros-distributions: jazzy
-
       - name: Install ccache
```
mfaferek93, line 314: One correction to the body, not the code: the image's rosdep sources cache is never reused in Actions. Container jobs run with `HOME=/github/home`, so the `rosdep update` step below still fetches `base.yaml`, `python.yaml` and `index-v4.yaml` from raw.githubusercontent.com on every job (jazzy-test log lines 411-438, `updated cache in /github/home/.ros/rosdep/sources.cache`). What is gone is `rosdep init` and the `20-default.list` download, and the remaining fetch sits inside the 3x retry loop, so the title holds. Worth rewording the body so nobody later removes the retry on the strength of it. Also: coverage and coverage-merge only run on main pushes, so that half of the change is unexercised until the first push; what they need is there (lcov installed, gcov in the image, gpg kept in the merge job).

bburda: I reworded the body. It now says that rosdep init and the 20-default.list download are gone. It also says that rosdep update still fetches base.yaml, python.yaml and index-v4.yaml on every job, because HOME=/github/home bypasses the cache the image carries, and that this is why the retry loop stays. Coverage and coverage-merge run on the first push to main; lcov, gcov in the image and gpg in coverage-merge are in place.

mfaferek93 (approved):

PR #687, opened 2026-09-12
test: keep the TSan job green on a future teardown report and a fan-out read

bburda:
# Pull Request

## Summary

The `sanitizer-tsan` job on `main` is red on two tests that share nothing but the job.

`test_cancel_outcomes` passes all 20 cases; the process exits non-zero because TSan reports one race. The last reference to a `GenericClient` future's shared state is dropped by the rclcpp executor thread, inside `librclcpp`, which is not instrumented. TSan sees the `operator delete` there but not the acquire on the `shared_ptr` refcount that orders it after our `future::get()`, so it pairs the free with the caller's earlier read of the result. Our use of the API is correct and the reference is held for the whole `get()`. `tsan_suppressions.txt` gets two lines for this class of report, both anchored on the read side, the wrapper's `get()`: `rclcpp::detail::FutureAndRequestId` on Jazzy and newer, and the gateway's `compat::GenericServiceClient` on Humble. The wrapper's destructor and `wait_for` are not matched, and no `std::future` the gateway builds itself is matched. The comment blocks around the pattern describe the synchronisation the way libstdc++ implements it. Reproduced under TSan with a standalone reproducer against the real suppressions file: 5 of 5 runs reported the race with no suppression. The `sanitizer-tsan` job of this pull request checks the pattern against the real gateway.

`test_peer_failure_reasons` failed on an assertion. Between the aggregator marking a peer online and publishing its contributor map for an entity, a read answers 200 with no fan-out at all, which is the documented purpose of the contributor map. The test's readiness gate waited for the first signal and read the failure reason from a single response, so it asserted a transient state; the window is about 0.9 s without instrumentation and wider under TSan. `_fan_out_failure` now polls until the answer carries the fan-out before reading the reason, retries a request that raised, and keeps every original assertion. With the readiness gate reduced to "the HTTP server answers", the old helper fails 5 of 5 runs with the exact CI message and the new one passes 5 of 5.

Seven feature tests applied `get_time_scale()` on top of `DISCOVERY_TIMEOUT`, which already carries the scale, and `test_aggregation_time_budgets` scaled `GENEROUS_FORWARD_MS` twice in two request timeouts. Every budget now scales once, so the longest single wait under TSan is 180 s. `test_relay_peer_credential` gets a 300 s ctest budget like the other multi-gateway tests; it starts three gateways and a fault manager and holds two sequential waits before its first case.

---

## Issue

No issue. Two independent failures of one CI job.

---

## Type

- [x] Bug fix
- [ ] New feature or tests
- [ ] Breaking change
- [ ] Documentation only

---

## Testing

- `test_cancel_outcomes` and `test_peer_failure_reasons` on jazzy, plain build: 10 of 10 and 10 of 10, plus the aggregation sweep (`ctest -R "peer|aggregat"`, 11 of 11).
- Red-green for both changes as described above; a mutation of the expected reason in `test_1` still fails the fixed test (`'timeout' != 'unreachable'`).
- The seven touched feature tests and `test_relay_peer_credential` pass on jazzy, plain build, with the new budgets in place; the integration package's linters pass.
- The `sanitizer-tsan` job of this pull request is the only place the suppression is verified against the real gateway under TSan.

---

## Checklist


src/ros2_medkit_integration_tests/test/features/test_peer_failure_reasons.test.py
```diff
@@ -80,7 +80,10 @@
 TIGHT_METADATA_MS = 800
 PATIENT_METADATA_MS = 20000
 
-TIMEOUT = DISCOVERY_TIMEOUT * get_time_scale()
+# DISCOVERY_TIMEOUT already carries the sanitizer time scale. Applying the scale
+# again squares it, and setUpClass runs three of these waits back to back, so the
+# class would ask for more wall clock than this test's ctest budget grants.
+TIMEOUT = DISCOVERY_TIMEOUT
```
mfaferek93, line 85: The squaring you removed here is still in three siblings: `test_aggregation_time_budgets.test.py:114`, `test_relay_peer_credential.test.py:101` and `test_aggregator_fault_stream.test.py:93` all do `DISCOVERY_TIMEOUT * get_time_scale()`. `test_relay_peer_credential` has no ctest override, so under TSan its `_wait_for_gateways` and `_wait_for_peer_status` deadlines are 540 s inside a 360 s ctest budget, and a stuck wait there gets killed with no test name in the output, the exact failure this comment says it avoids. Fix all four here or say why the other three stay out.

tsan_suppressions.txt
```diff
@@ -79,6 +79,18 @@ deadlock:rclcpp::Context*
...
+# reference to the state is dropped by whichever thread finishes with the
+# promise; when that is the rclcpp executor thread, ~_State_base and the
+# _Result::_M_destroy() it runs live inside librclcpp, which is not
+# instrumented. TSan sees the operator delete, because that is intercepted in
+# the allocator, but not the acquire on the shared_ptr refcount that orders it
+# after the caller's future::get(), so it pairs the free with the caller's read
+# of _Result::_M_error. Anchored on the libstdc++ frame that performs that read:
+# it is the only frame of the pair that is reliably symbolized, since librclcpp
+# is stripped and the freeing frame prints as <null> at librclcpp.so+0x...,
+# which no race: pattern can match.
```
mfaferek93: This last clause is false: TSan matches `race:` patterns against function, file and module name, the module is filled in even when symbolization fails (that is where the `(librclcpp.so+0x...)` suffix comes from), and matching is an unanchored substring, so `race:librclcpp.so` would catch that `<null>` frame. The real reason to anchor on `_M_get_result` is scope: a module pattern suppresses every report with any librclcpp frame on either stack. Reword to that, and note the pattern also covers the gateway's own `std::future<NodeQueryResult>` in the configuration fan-out, so it is wider than the one GenericClient report.

src/ros2_medkit_integration_tests/test/features/test_peer_failure_reasons.test.py
bburda: Fixed in 3c1ed9a0. Every feature test now uses DISCOVERY_TIMEOUT as it is, because it already carries the time scale. Seven more files squared it: the three you named, plus test_manifest_config_key, test_manifest_malformed_config, test_merge_provenance and test_plugin_entity_survives_ignore. test_aggregation_time_budgets also scaled GENEROUS_FORWARD_MS a second time in two request timeouts. Those now scale only the added margin. test_relay_peer_credential gets a 300 s ctest budget like the other multi-gateway tests (be714ca4, CMakeLists.txt).

tsan_suppressions.txt
bburda: A module pattern does match that frame, so the old clause was wrong. In c850953f the suppression anchors on the read side of the report: `rclcpp::detail::FutureAndRequestId::get()` on Jazzy and newer, and `compat::GenericServiceClient`'s `FutureAndRequestId::get()` on Humble. That frame is header code compiled into the gateway's own binaries, so it always has a symbol. The futures the gateway builds itself (run_sync, the aggregation and configuration fan-outs, the topic sampler) do not match it, and neither do the wrapper's destructor and wait_for. The comment in tsan_suppressions.txt now says why no module pattern is used: it would suppress every report with a librclcpp frame on either stack. It also names the setter chain that runs uninstrumented inside librclcpp.

mfaferek93 (approved):

